PR #122, opened 2015-10-19
Update logrus sentry path

abeMedia:
logrus sentry was moved into its own repo


lonelycode:
Can you pull this in the develop branch instead? Will need to reject this one :-/

PR #171, opened 2016-01-17
Add X-Forwarded-For support to IP Whitelist

cppd245:
Noticed that IP Whitelisting does not work when Tyk Gateway is behind a proxy/load balancer.  This commit adds support for X-Forwarded-For header.


lonelycode:
Fantastic, thanks! :-D

Issue #177, opened 2016-01-22
HandleAddOrUpdateApi create API spec to filesystem even use_db_app_configs is true


iwat:
"HandleAddOrUpdateApi" does not respect "use_db_app_configs". Calling POST /tyk/apis/ will create files to configuration directory even use_db_app_configs is true. And no API is registered to MongoDB.

lonelycode:
This is expected behavior, to write to DB use the advanced API, the gateway API does not write back to DB. its to ensure independence between the services.

iwat:
If that's so, what is the proper way to add new API definition to a cluster of Tyk(s)?

lonelycode:
Just use the dashboard API (aka Advanced API):

https://tyk.io/advanced-api-v0-9/api-definitions/#create-api-definition:cdac0ac5cf58124a5fc9102bcd4708e4

This will save to DB and signal cluster to hot reload

iwat:
Thank you very much, it seems to be what I was looking for.

By the way, could you please provide a URL on how can I have that "Authorization" header?

lonelycode:
Go to your dashboard -> Users -> Select a user (or create one for the app) -> The API Key is listed under the user profile

iwat:
For an unknown reason, an access key of generated user from setup.sh (Docker) was not working. Resetting or creating a new user can solve the problem.

lonelycode:
That can happen, sometimes the access key isn't saved properly and needs regenerating.

lonelycode:
Not sure why that happens, it could be simply the key was lost in Redis when it restarted.

iwat:
By the way, is it better if the gateway returns error on POST /tyk/apis/ when 'use_db_app_config' is true?

lonelycode:
That is a fantastic idea :-)

PR #178, opened 2016-01-23
Throw error on POST /tyk/apis/ if UseDBAppConfigs is enabled

iwat:
Refer to TykTechnologies/tyk#177

Calling POST to /tyk/apis/ when `UseDBAppConfigs` is enabled may
cause an unexpected behavior. It added a new API definition to
`config.AppPath` directory and didn't add to MongoDB, so it's not
available on the dashboard.

This commit changes the behavior of POST command by rejecting
request if `UseDBAppConfigs` setting is enabled and suggest user to
use an Advanced Management API instead.


lonelycode:
Lovely, thank you very much :-)

PR #197, opened 2016-02-16
RH: Closes #193 by adding new config option for hiding the X-Generator heaer

### LINKED ISSUE #193: Make X-Generator header optional
Hi All,

As seen in this line: https://github.com/TykTechnologies/tyk/blob/master/handler_error.go#L122

Any error response from Tyk will always have the header "tyk.io"

This is useful in development, so we can work out where the header is from, but not necessarily something we want to expose in production.

Would it be possible to have a configuration item which could turn this header off?

Cheers,

richard-hulm:
Made it so setting the config option to true 'hides' the header, so if the option is not set then behaviour continues as before

I've got redis running locally but can't get the tests to pass on master or develop or my branch, but as this is literally just hiding a header on a config change I assumed it would be OK

EDIT: I believe the only failing test locally is TestServiceDiscovery_ETCD_NOLIST - no real idea on where to start looking as to why that fails...? Seems to have passed on CI though

### DIFF
--- a/config.go
+++ b/config.go
@@ -121,6 +121,7 @@ type Config struct {
 	ControlAPIHostname   string `json:"control_api_hostname"`
 	EnableCustomDomains  bool   `json:"enable_custom_domains"`
 	EnableJSVM           bool   `json:"enable_jsvm"`
+	HideGeneratorHeader  bool   `json:"hide_generator_header"`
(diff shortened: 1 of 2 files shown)


lonelycode:
Lovely :-)

PR #210, opened 2016-03-03
Fix directory based middleware loading

philipbjorge:
My post middleware was being run as pre middleware.

I'm unable to build this on Windows (and therefore cannot run tests)
https://github.com/rcrowley/goagain/issues/24

How do you want to proceed with this?


lonelycode:
LGTM

taliesins:
@philipbjorge perhaps we could use https://github.com/tylerb/graceful as an alternative?

lonelycode:
As an alternative to what?

PR #219, opened 2016-03-19
Let user terminate the request in middleware

### LINKED ISSUE #207: Feature request: allow middleware to terminate request handling and return response
I have a backend service with an API something like:

```
/accounts/:id
/accounts/:id/preferences
```

In my Tyk middleware I'm routing `/account` to `/accounts/:id`, looking up the ID based on a token in a request header using my own web service.

If the request header doesn't contain a valid token, I'd like to stop processing the request and return an Unauthorized/Access Denied. At the moment I'm doing this by forwarding the request to an endpoint which returns the desired status code.

However ideally in the middleware I'd be given another option to instead stop all further processing by returning a response, e.g. `sampleMiddleware.ReturnResponse(response);`.

JosephEarl:
Fixes #207

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -18,7 +18,7 @@ import (
 
 // Lets the user override and return a response from middleware
 type ReturnOverrides struct {
-	ResponseCode int
+	ResponseCode  int
 	ResponseError string
 }
 
--- a/plugins.go
+++ b/plugins.go
@@ -13,7 +13,14 @@ import (
 	"net/http"
 	"net/url"
 	"time"
+	"errors"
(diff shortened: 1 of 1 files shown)


lonelycode:
Again, thanks! :-D

JosephEarl:
No problem Thanks for the great piece of software!

Will have a few more changes ready in the next few days:

* Updates to Docker docs on changes needed to move from the Quickstart to getting it run remotely
* Some updates to the advanced API docs (things I had dig around for a bit to automate our portal setup)

PR #221, opened 2016-03-19
Add Target Path to Service Discovery

### LINKED ISSUE #161: Add target path when using service discovery
Typically service registries will store and return the host address and port, but we need to proxy an api to a particular target path.  For example:

myapi.mytykserver.com/myapiresource/* > [discovered host]/api/*

There needs to be a way to append the literal string (example: "/api/") to the targets retrieved from service discovery.

JosephEarl:
Fixes #161. Not 100% sure I was adding the path in the correct place (or in all the places it is needed), so feedback welcome.

### DIFF
--- a/service_discovery.go
+++ b/service_discovery.go
@@ -21,6 +21,7 @@ type ServiceDiscovery struct {
 	dataPath            string
 	parentPath          string
 	portPath            string
+	targetPath          string
 }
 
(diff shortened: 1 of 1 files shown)


lonelycode:
LGTM, Service Discovery path creation is self contained in the handler.

Issue #242, opened 2016-04-18
swagger import fails with json unmarshal error


bboozzoo:
Swagger import has ill defined structure of `path` element.

The code https://github.com/TykTechnologies/tyk/blob/master/swagger.go#L61 defines a map of `map[string]PathMethodObject` assuming that every entry under `<path>` has a leaf node mappable to `PathMethodObject`. The assumption is not true. According to swagger spec of `pathItemObject` http://swagger.io/specification/#pathItemObject only `get`, `put`, `post`, `delete`, `options`, `head`, `patch`, have a `OperationObject` node (that roughly maps to `PathMethodObject` in the code).  Having `parameters`, or custom fields will trigger an unmarshaller error.

lonelycode:
Care to supply a pull request?

Thanks.

bboozzoo:
Why not, seems like an easy fix.

PR #243, opened 2016-04-18
swagger: fix import for v2.0 files

bboozzoo:
Addresses #242 

According to Swagger 2.0 spec,
pathItemObject (http://swagger.io/specification/#pathItemObject)
contains both known attributes (`get`, `put`, `post`, `patch`,
`options`, `delete`, `head`, `$ref`, `parameters`) and any number of
user defined fields.

Out of these set only `get`, `put`, `post`, `patch`, `options`, `delete`
and `head` are mappable to `PathMethodObject` struct. Having other
attributes in the API specification file triggers a JSON unmarshaller
error.

Since we are only interested in HTTP related pieces of the path
specification, add an intermediated structure to unmarshall only
interesting fields, while the rest is dropped silently.

### DIFF
--- a/swagger.go
+++ b/swagger.go
@@ -38,6 +38,16 @@ type PathMethodObject struct {
 	Responses   map[string]ResponseCodeObjectAST `json:"responses"`
 }
 
+type PathItemObject struct {
+	Get     PathMethodObject `json:"get"`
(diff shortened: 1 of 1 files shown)


lonelycode:
Lovely, thanks!

Issue #106, opened 2015-09-11
Transfer-Encoding: Chunked


AlexGiovanentti:
Is Transfer-Encoding: chunked supported?

We have streaming endpoints that require this feature.
We have successfully been able to proxy most of our endpoints with Tyk except the endpoints that depend on Transfer-Encoding: chunked.
With other API Gateway solutions we have evaluated and that feature nginx, setting proxy_buffering off; in nginx.conf has enable the feature.

Thank you for your time.

lonelycode:
We haven't tested for it, might be a bug.
Tyk isn't nginx based, so it's something we'll need to investigate.






On Thu, Sep 10, 2015 at 10:20 PM -0700, "AlexGiovanentti" <notifications@github.com> wrote:












Is Transfer-Encoding: chunked supported?



We have streaming endpoints that require this feature.

We have successfully been able to proxy most of our endpoints with Tyk except the endpoints that depend on Transfer-Encoding: chunked.

With other nginx proxy solutions setting proxy_buffering off; in nginx.conf has enable the feature.



Thank you for your time.

lonelycode:
Have you tried setting close_connections in the Tyk.conf file to false? 
This setting forces the proxy to close the upstream connection as soon as it gets a response.







On Thu, Sep 10, 2015 at 10:20 PM -0700, "AlexGiovanentti" <notifications@github.com> wrote:












Is Transfer-Encoding: chunked supported?



We have streaming endpoints that require this feature.

We have successfully been able to proxy most of our endpoints with Tyk except the endpoints that depend on Transfer-Encoding: chunked.

With other nginx proxy solutions setting proxy_buffering off; in nginx.conf has enable the feature.



Thank you for your time.

lonelycode:
**Confirmed solution**: To enable processing of Transfer-Encoding: chunked, set the `close_connections` property in the `tyk.conf` file to `false` (or remove the line altogether).

Tyk aggressively closes TCP connections by default, to stop this from happening you can unset this option and Tyk will respect HTTP 1.1 convention to keep open TCP channels per client.

AlexGiovanentti:
Would your proposed solution pass-through the chunks of data as they are received or wait till the end of the stream?

Case Study:
We have an endpoint that uses Transfer-Encoding: chunked; to continuously send chunks of data with updates on any changes to the price of a stock. The stream does not have a deterministic end. The expected api gateway behavior would be to pass through the chunks of data as they are received from the upstream server.

[...]

PR #262, opened 2016-06-14
Use FlushInterval as time.Millisecond

matiasinsaurralde:
Hi, I think this is related to #106.

I've been trying to use Tyk for chunked responses but I find it too slow, so I started to play with the ```FlushInterval``` parameter and I found that's limited to ```time.Second```, I applied the current PR as a patch to my development server and now I'm able to specify a flush interval that's less than 1 second, and the experience of sending the request through Tyk is almost like what I'm getting when sending a request directly to my endpoint.

**Note:** My endpoint does some async stuff and sends chunks as fast as possible, so I get a chunk after 500 ms, another one after 2000 ms and so on. So during my tests, setting ```FlushInterval``` to 500 or 200 ms, improves the experience (I get the chunks ASAP), this is a very specific thing, but I do find useful to use milliseconds instead of seconds, it's more flexible!

Best regards!

### DIFF
--- a/tyk_reverse_proxy_clone.go
+++ b/tyk_reverse_proxy_clone.go
@@ -159,7 +159,7 @@ func TykNewSingleHostReverseProxy(target *url.URL, spec *APISpec) *ReverseProxy 
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


lonelycode:
Nice one :-)

PR #268, opened 2016-06-24
Upgrade to go-jwt 3

sje397:
Implement the fix recommended in the migration guide for breaking changes in go-jwt version 3.

### DIFF
--- a/middleware_jwt.go
+++ b/middleware_jwt.go
@@ -8,14 +8,15 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/Sirupsen/logrus"
-	"github.com/dgrijalva/jwt-go"
-	"github.com/gorilla/context"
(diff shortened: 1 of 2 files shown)


lonelycode:
We've already done all this in our develop branch :-(

sje397:
No problem - it was small...and tests fail. But atm if you just checkout and build master, it fails.

lonelycode:
Yeah, we recently did a small patch and found the same - the OpenID middleware is also broken, so our fork needed updating.

Develop will be merged to master very soon so it should all be stable then. 

For hot fixes have been temporarily vendoring older versions

lonelycode:
Thank you though! Really do appreciate it, nothing worse than upstream library changes breaking your code! M.

sje397:
No worries at all. Thanks for building Tyk! :)

Issue #239, opened 2016-04-13
Feature Request: Set Listen IP


yourpalal:
Currently, you can set the listen port, it would be nice to be able to set the listen IP as well. This way, when Tyk is behind nginx, it can be set to listen on 127.0.0.1 for improved/redundant security along with a firewall.

lonelycode:
NIce idea.

PR #277, opened 2016-07-15
Add Listen Address

richard-hulm:
Addresses #239 

Allows you to set a port that the Tyk process will bind to

Will ignore requests made otherwise

I've left blank as the default, which is the same as current behaviour.

See https://golang.org/pkg/net/#Listen


richard-hulm:
Didn't do it into develop again... sorry again!

PR #281, opened 2016-07-26
 fix use -h | --help option process not exit bug

stevensu1977:
fix use -h | --help option process not exit bug

### DIFF
--- a/main.go
+++ b/main.go
@@ -1051,7 +1051,7 @@ func init() {
 		--as-version=<version>       The version number to use when inserting
 	`
 
-	arguments, err := docopt.Parse(usage, nil, true, VERSION, false, false)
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -1051,7 +1051,7 @@ func init() {
 		--as-version=<version>       The version number to use when inserting
 	`
 
-	arguments, err := docopt.Parse(usage, nil, true, VERSION, false, false)
+	arguments, err := docopt.Parse(usage, nil, true, VERSION, false)//fix use -h|--help process not exit
```
buger: Probably you need to remove this comment from source.

mvdan:
Took over the patch, keeping the original author. See the new PR.

PR #285, opened 2016-08-12
OAuth - Multiple Redirect URIs

nazarhussain:
Added a configuration `oauth_redirect_uri_separator` when set will allow to use multiple redirect URIs for OAuth Client. 

If set blank or empty will allow only one single URL as redirect URI of OAuth Client.

### DIFF
--- a/config.go
+++ b/config.go
@@ -84,6 +84,7 @@ type Config struct {
 	}
 	OauthRefreshExpire int64 `json:"oauth_refresh_token_expire"`
 	OauthTokenExpire   int32 `json:"oauth_token_expire"`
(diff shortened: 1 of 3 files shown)


lonelycode:
This is really cool - I didn't know the oauth server could handle separated urls, where does it check the fields?

lonelycode:
Just realised its part of Osin... Derp ;-)

lonelycode:
This looks good, but could you make the PR against develop? That will be the next cut.

nazarhussain:
Initially I refactored tyk own OAuthClient structure to support multiple URIs. Later I noticed that feature in osin and I used it.

nazarhussain:
Coming this PR in favour to https://github.com/TykTechnologies/tyk/pull/286

PR #286, opened 2016-08-14
OAuth - Multiple Redirect URIs

nazarhussain:
Added a configuration oauth_redirect_uri_separator when set will allow to use multiple redirect URIs for OAuth Client.

If set blank or empty will allow only one single URL as redirect URI of OAuth Client.

### DIFF
--- a/config.go
+++ b/config.go
@@ -84,6 +84,7 @@ type Config struct {
 	}
 	OauthRefreshExpire int64 `json:"oauth_refresh_token_expire"`
 	OauthTokenExpire   int32 `json:"oauth_token_expire"`
+	OauthRedirectUriSeparator   string `json:"oauth_redirect_uri_separator"`
 	SlaveOptions       struct {
 		UseRPC                          bool   `json:"use_rpc"`
 		ConnectionString                string `json:"connection_string"`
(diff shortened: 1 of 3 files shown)


lonelycode:
That's ok, will be on our end, Dev is still a little unstable.

lonelycode:
Ah, no, it's actually a bug with the graylog hook... have raised an issue

lonelycode:
Submitted PR: https://github.com/gemnasium/logrus-graylog-hook/pull/13

nazarhussain:
Related to this feature we also have to make changes in Tyk Dashboard. So for multiple urls we show textbox instead of text input field for oAuth clients. 

Also updating multiple URLs is a frequent use case, so we need to gave a feature there to update existing oAuth clients as well. How an I contribute in it?

lonelycode:
@nazarhussain Dashboard is closed source I'm afraid, we'll push these changes in upstream, [it's in the board now](https://trello.com/c/Pru7NC5N/376-ensure-oauth-multi-client-url-is-supported-in-dashboard). 

AFAICT this only works if a `oauth_redirect_uri_separator` has been specified, we can't necessarily assume this is the case unless we make it a default in the config we ship, which we will need to do for Pro installs.

There's some other roadmap OAuth changes to allow self-registering of OAuth data in the portal, so it will probably be bundled with that.

nazarhussain:
Even if `oauth_redirect_uri_separator` is not set, we still can develop the feature for updating the oath client and also change text field to textbox. Since even right now long urls does not fit into that text field.

nazarhussain:
@lonelycode Thanks for merging it. What's the plan for next release that includes this feature.

lonelycode:
The plan is our trello board, and we'll release when it's ready :-)

PR #328, opened 2017-01-05
Travis and code cleanups

mvdan:


mvdan:
This is a very weird setup. From what I can see, develop is @lonelycode's WIP branch. It even has "FOR TESTING ONLY" commits. Master is not the stable branch, and this policy isn't documented anywhere, so I don't fully understand what the logic here is.

lonelycode:
Master always reflects the last published on-prem release (v2.3.1 atm), all development by team members happens in feature branches, then merged to `develop` either via PR or directly. 

We keep `master` as a mirror of latest packaged release to ensure that if someone uses one of our packaged binaries from packagecloud/docker and needs to modify it, they can be sure that what is in master will behave the same way as the version branch they have. 

We could achieve the same result from version tags (I think we may have in the past), but as far as the project goes, master is stable at the cutoff point of the latest release.

There's no policy documented because we don't have contributor docs yet ;-)

mvdan:
I still don't understand the logic, but I'll get back to that later :)

PR #333, opened 2017-01-11
Fix -h/--help not exiting the program

mvdan:
Rebase and cleanup of #281. Thanks @stevensu1977!

### DIFF
--- a/main.go
+++ b/main.go
@@ -973,7 +973,7 @@ func getCmdArguments() map[string]interface{} {
 		--log-instrumentation        Output instrumentation data to stdout
 	`
 
-	arguments, err := docopt.Parse(usage, nil, true, VERSION, false, false)
+	arguments, err := docopt.Parse(usage, nil, true, VERSION, false)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


buger:
This is what last argument stands for:

> By default, docopt calls `os.Exit(0)` if it handled a built-in option such as
`-h` or `--version`. If the user errored with a wrong command or options,
docopt exits with a return code of 1. To stop docopt from calling `os.Exit()`
and to handle your own return codes, pass an optional last parameter of `false`
for `exit`.
https://github.com/TykTechnologies/tyk/blob/60003580cd3e50c4809ddae9ec0f06898da2a1db/vendor/github.com/docopt/docopt.go/docopt.go#L36

lonelycode:
@buger Are you happy for this to be merged?

buger:
So basically right now with last `false` option it expected us to manually call os.Exit in case of `-h` flag. By removing this option, we allow docparse to exit program by himself. 

If my assumption is right, then 👍

mvdan:
Yeah, that's my understanding too.

PR #337, opened 2017-01-11
Explicitly support Go 1.6

mvdan:
CI: remove `go get -d` lines.

We vendor all deps, so this is unnecessary.


.travis.yml
```diff
@@ -1,16 +1,15 @@
...
+  - 1.6.x
   - 1.7.x
 
 services:
   - redis-server
   - mongodb
 
 install:
-  - go get -d -v ./...
-  - go get -d -v -tags coprocess ./...
```
buger: What is the reason of removing this lines?

mvdan: See discussion on slack. They do nothing, as they would only download but everything is already on disk since all deps are vendored.

mvdan: (also explained in the commit, let me know if you'd like my commits to be more verbose)

buger:
👍

Issue #334, opened 2017-01-11
Too many files in the root dir


mvdan:
There are too many files in the root directory. This means that visiting github.com/TykTechnologies/tyk is a bad experience, since:

* One has to scroll way down to see the README content
* Important files such as `.travis.yml`, `Makefile` or `main.go` are hard to find

Having so many go files in a single directory is also pretty bad as far as development goes, as it's hard to keep track and they won't fit in a single screen when listing them.

I suggest that we either merge some files or that we move some to separate packages. Ideally, if the root package is to be the executable, it should be fairly small. Luckily enough, since we're not exporting any identifiers from this package, moving things out of it is doable and won't break anything.

matiasinsaurralde:
I think that organizing everything in separate packages is very interesting, if we do it at some point we should also consider the idea of having a "pluggable" Tyk gateway.

mvdan:
I do agree that we need more packages. Some projects use way too many packages, though. I suggest we split up the current one into 2 or 3, and then see if we need to split further.

If we start exposing parts of Tyk as a library, then we have the issue of backwards compat. So we should be really sure about what to expose. Until then, we are free to keep things in the main package or in `internal/...`.

mvdan:
Looking at the files, two easy candidates come to mind:

* `middleware*.go`
* `coprocess_*.go`

lonelycode:
I can see this as a positive move, but we don't want to lose momentum because of refactor - small incremental updates to this issue should help

mvdan:
Yeah, this is low priority. At the same time I'd like to see the low-hanging fruit happen sooner than later, as this disorganization slows things down a bit.

abeMedia:
Breaking tyk up into libraries also brings it a little closer to being able to wrap it within a binary as mentioned in #308 
Ideally I think the main package should be very small, using internal library for most operations. This not only makes it easier to understand the code but also makes it more pluggable.

mvdan:
Agreed. One step at a time, though. You could say this issue is a dependency of yours.

mvdan:
FWIW, down from 148 to 138, both including directories. We haven't started splitting the main package into multiple ones.

mvdan:
@matiasinsaurralde thoughts on moving all of the coprocess stuff to `tyk/coprocess`? I guess what's already in that package can stay there.

matiasinsaurralde:
@mvdan Yes, we should find a place for the `coprocess` files in the root as well.

[...]

PR #338, opened 2017-01-12
Remove Makefile and Dockerfile

mvdan:
These were contributed by a user a long time ago. We don't use them to
ship our releases as docker images though, so they have been
unmaintained since. Remove them to make it clearer that they are not our
official docker release workflow.

Updates #334.


buger:
👍

PR #342, opened 2017-01-13
Fix test panics if redis isn't running

mvdan:
Two are in test code, one is in actual code. Running `go test` when redis isn't running now fails instead of panicking and stopping without running all the tests.


lonelycode:
LGTM

PR #344, opened 2017-01-13
Make golint even happier

mvdan:


mvdan:
Rebased after conflicts with my other PRs.

buger:
If there is no custom changes except linter ones, i'm ok 👍

mvdan:
The only change that affects the behaviour is the changes in error strings. Errors in Go should not be capitalized or have punctuation.

PR #345, opened 2017-01-13
Some more code simplifications

mvdan:
Also fix a potential bug with string handling.


buger:
`createConnectionStringFromDashboardObject` function is a really good candidate for unit test. 
Not too complex and without side effects.

mvdan:
@buger you mean it should stick around?

Ping @lonelycode in case any of these were left unused on purpose.

buger:
I mean you should write unit test if you think there is potential bug in it.

mvdan:
Ah, I assumed you meant a function I was removing.

As far as I can tell none of this redis stuff has independent unit tests (no `redis_*_test.go` files) and I'd rather not add yet another file to the root dir. Not sure it's the thing to do right now.

lonelycode:
> Ping @lonelycode in case any of these were left unused on purpose.

They may have been part of old tests that are no longer part of the codebase. 

`createConnectionStringFromDashboardObject` looks like a simple change, no bugfix, just variable renaming so not sure if it needs a unit test just yet ;-)

PR #347, opened 2017-01-13
ci: move script to separate file, enforce gofmt

mvdan:
Don't enforce it with -s for now.

Also run tests on all of the non-vendor packages, since we were only
running tests for the root package until now.


lonelycode:
LGTM

PR #348, opened 2017-01-13
README: use single CI badge, remove gitter

mvdan:
Two CI badges are a bit redundant. Since stable is just an older version
of master, and we will now enforce green PRs, its green badge means very
little.

Also, gitter is a 404 so nuke it.


lonelycode:
👍

PR #359, opened 2017-01-16
Remove more unused code

mvdan:
This time it's not test code.


mvdan:
@lonelycode I can understand the purpose of `middleware_sample.go`, but it should definitely not be here. If you'd rather have it moved elsewhere than deleted, let me know.

lonelycode:
Can we move it into a middleware.md then? Would like a reference version slmewhere in the repo.

matiasinsaurralde:
`coprocess_python` uses `getBundlePaths`, it's part of `coprocess_bundle`. I put it there because it might get called from other language integrations.

lonelycode:
Coprocessor build and tests should be covered by ci no?

matiasinsaurralde:
Yes, the test fail for me / locally (just ran it), with the build tags of course!

lonelycode:
Why didn't Travis pick it up then?

matiasinsaurralde:
`getBundlePaths` is used but not present in the tests, it's a small helper function that was added when we switched from the standard middleware path to bundle paths (in the path, rich plugins were loaded from the standard middleware path).
I suggest adding a test for it, and not removing that part of the code, as it's clearly not unused code.

mvdan:
> Why didn't Travis pick it up then?

I must have broken it. `go install -tags coprocess` fails here too. Will fix.

> I suggest adding a test for it, and not removing that part of the code, as it's clearly not unused code.

Instead I would suggest that Travis should build all the tag combinations. If I understand correctly, then Travis would throw a compile error since `getBundlePaths` would then be missing.

In any case, it should be moved to that file since it's not needed in the regular build with no tags.

mvdan:
> Can we move it into a middleware.md then? Would like a reference version slmewhere in the repo.

I suggest we make it an example then. Examples are like tests, also run as tests, but they're meant as useful source code for users. Also shown in the godoc.

matiasinsaurralde:
Sounds interesting, the CP builds are tested in a very basic way (and I think this is ok for now, we test the core functionality, and object passing, etc.). An extensive test will require having Python libs, Lua libs, etc. in the test environment, and then test the Python modules, a lot of stuff!

mvdan:
Oh, okay. I'll open an issue for that then.

lonelycode:
> I suggest we make it an example then. Examples are like tests, also run as tests, but they're meant as useful source code for users. Also shown in the godoc.

Great idea

> Instead I would suggest that Travis should build all the tag combinations. If I understand correctly, then Travis would throw a compile error since getBundlePaths would then be missing.

Yep, but then all the deps need to be in place as it can get messy :-)

For our CD we have these installed:

```
apt-get update && apt-get install -y ca-certificates
apt-get install -y wget
apt-get install -y build-essential
apt-get install -y libluajit-5.1-2
apt-get install -y libluajit-5.1-dev
apt-get install -y luarocks
apt-get install -y python3-setuptools
apt-get install -y python3-dev
apt-get install -y libpython3.4
apt-get install -y python3-pip
apt-get install -y ruby-dev gcc make rubygems
apt-get install -y gnupg

luarocks install lua-cjson
wget https://github.com/google/protobuf/releases/download/v3.1.0/protobuf-python-3.1.0.tar.gz
tar -xvzf protobuf-python-3.1.0.tar.gz
cd protobuf-3.1.0/ &&  ./configure -prefix=/usr && make && make install
cd protobuf-3.1.0/python && python3 setup.py build --cpp_implementation && python3 setup.py install --cpp_implementation
pip3 install grpcio
```

matiasinsaurralde:
> Sounds interesting, the CP builds are tested in a very basic way (and I think this is ok for now, we test the core functionality, and object passing, etc.). An extensive test will require having Python libs, Lua libs, etc. in the test environment, and then test the Python modules, a lot of stuff!

@mvdan See this, for example: https://github.com/TykTechnologies/tyk/blob/master/coprocess/python/dispatcher.py

mvdan:
Awesome, opened #359.

Will address issues and fix Travis in a separate PR.

mvdan:
FWIW, Travis has always been broken with regards to tags. It wouldn't have picked up build issues with tag combinations before I started changing things either. See #360.

mvdan:
Done. Verified that coprocess still compiles via `go install -tags coprocess,python`.

lonelycode:
@matiasinsaurralde Want to confirm all ok? Then I'll merge.

matiasinsaurralde:
@lonelycode Sure!

matiasinsaurralde:
@lonelycode It's ok!

PR #367, opened 2017-01-17
CI: use gofmt with -s, enforce goimports

mvdan:


lonelycode:
@buger You happy with this approach?

buger:
yes 👍

Issue #368, opened 2017-01-17
tykcommon.HostList contains sync.RWMutex and is copied often


mvdan:
This makes go vet very angry, as locks (`sync.RWMutex` in this case) should never be copied implicitly.

Filing a bug instead of fixing it because it's in tykcommon, a few nested levels deep, and not sure what the correct thing to do is. Possible fixes that come to mind:

* Remove the mutex from `HostList` somehow
* Remove `HostList` from one of the parent structs, such as `APIDefinition`
* Avoid copying any of these structs in this codebase

#346 depends on this.

lonelycode:
`HostList` is used extensively in the service discovery system, it *must* be thread safe and the list can change at any time.

I'd say keep the mutex but find where it's being copied, and make that just a pointer reference to be swapped around. 

This code is known to be a bit buggy, if the mutex is being copied all over the shop it might actually explain why some users report random nil pointer dereference panics during SD switchovers.

mvdan:
@lonelycode so which should be a pointer, `APIDefinition` or the `HostList` inside `APIDefinition`?

> if the mutex is being copied all over the shop it might actually explain why some users report random nil pointer dereference panics

Unless the mutex isn't actually used in between copies, or the copies don't happen in practice, this is bound to produce some sort of bug. Can't hurt to fix anyway.

lonelycode:
API Defintions are huge, so they should always be pointers (and I think they are IIRC, went through a refactor a ways back to make sure it wasn;t copied too often).

Hostlist is dynamic and shared across proxies requests (it must be up to date for any single configuration), so it too should be a pointer

lonelycode:
> Unless the mutex isn't actually used in between copies, or the copies don't happen in practice, this is bound to produce some sort of bug. Can't hurt to fix anyway.

Yes I think we should fix it.

mvdan:
> API Defintions are huge, so they should always be pointers (and I think they are IIRC, went through a refactor a ways back to make sure it wasn;t copied too often).
> 
> Hostlist is dynamic and shared across proxies requests (it must be up to date for any single configuration), so it too should be a pointer

Ah, thanks for the pointers. PR incoming soon.

mvdan:
Now fixed, see commits above.

PR #371, opened 2017-01-17
Always use pointers for HostList

mvdan:
As discussed in #368. Together with #369, this will make go vet have zero warnings.

### DIFF
--- a/api.go
+++ b/api.go
@@ -618,8 +618,7 @@ func HandleGetAPIList() ([]byte, int) {
 	var responseMessage []byte
 	var err error
 
-	var thisAPIIDList []tykcommon.APIDefinition
-	thisAPIIDList = make([]tykcommon.APIDefinition, len(*ApiSpecRegister))
+	thisAPIIDList := make([]*tykcommon.APIDefinition, len(*ApiSpecRegister))
 
 	c := 0
 	for _, apiSpec := range *ApiSpecRegister {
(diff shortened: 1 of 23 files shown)


mvdan:
BTW, I'm aware that I'm changing the vendored tykcommon. This makes my life easier to apply all the changes in a single PR and be able to test them and get it reviewed at once. IMO that's better than a first PR to tykcommon.

If this is merged, then I'd do a PR to tykcommon to sync the change.

buger:
I'm not sure if it is a good idea, but for example if we make "tykcommon" git submodule, instead of folder, it may work. What do you think?

mvdan:
I would instead suggest just making it part of this repo. I don't see an advantage in a separate one.

lonelycode:
@mvdan 
> I would instead suggest just making it part of this repo. I don't see an advantage in a separate one.

API Definition objects are big, and they change a lot, instead of having to manually copypasta the object into Tyk Pump and Tyk Dashboard (and MDCB, and Host Manager and Tyk CLI), and hope they are all in sync we have it in a third party repo, it makes life a lot easier.

@buger 
> I'm not sure if it is a good idea, but for example if we make "tykcommon" git submodule, instead of folder, it may work. What do you think?

Not a fan of submodules, I'd rather just not vendor it. The vendoring thing came about because our pipeline CD was having connectivity problems with GH and I was tired of the go get taking an age.

I'd ultimately prefer changes to be made to the original repo instead of the vendor version - we use govendor to handle it and so far it's done an ok job.

With regards to this PR, it looks and smells ok to me, a quick smoke test implies it's safe, needs more rigorous test, but that can be handled at integration.

Any objections? Otherwise I will merge it.

buger:
👍

mvdan:
The same could be done with this repo - all others could depend on this one. I still don't see why a separate git repo helps - if anything, it makes changes like this more complicated.

lonelycode:
> The same could be done with this repo - all others could depend on this one.

Fair point, and one to consider later down the line. Modifying the dependencies across a whole host of projects isn't particularly palatable at the moment :-)

PR #374, opened 2017-01-18
Cleanup some more code

mvdan:


buger:
👍

tyk_reverse_proxy_clone.go
```diff
@@ -238,7 +236,7 @@ func TykNewSingleHostReverseProxy(target *url.URL, spec *APISpec) *ReverseProxy
 
 		// No override, and no load balancing? Use the existing target
 		targetToUse := target
-		if switchTargets {
```
lonelycode: There isn't a test to cover this functionality, it looks ok, but we can't guarantee it works (Use case is: a path that is being rewritten has a hostname in it, and so needs re-parsing so the proxy target changes)

mvdan: Agreed, will undo. This does change functionality somewhat, as empty strings will be treated differently.

mvdan: @lonelycode done, PTAL

mvdan:
It seems I can't mark the requested change as solved. I assume only you can do that.

Issue #340, opened 2017-01-12
URL parameters that contain http:// as a field group (even url encoded) are misinterpreted when the proxy URL is reconstructed.


lonelycode:
The issue appears to be related to a "uri" as a path parameter which results in a path parameter of `http://somedomain.com/123123` to be converted into `http:/somedomain.com/12312` (leading double slash trimmed to one slash).

mvdan:
This stinks of `path.Join`.

lonelycode:
Matias already found a fix, it's got to do with the `DefaultServeMux`

lonelycode:
This issue is related to #410 and need revising so re-opening.

PR #377, opened 2017-01-19
Expose mux.Router SkipClean functionality

matiasinsaurralde:
This is related to #340, when `HttpServerOptionsConfig.OverrideDefaults` is enabled, Tyk will check `HttpServerOptionsConfig.SkipURLCleaning`.

Sample request:

```
curl http://localhost:8080/quickstart//headers
```

**SkipClean = true, no redirections**

**SkipClean = false, you get a 301 redirect to `/quickstart/headers`**

This behavior is described [here](http://www.gorillatoolkit.org/pkg/mux#Router.SkipClean).

### DIFF
--- a/config.go
+++ b/config.go
@@ -111,6 +111,7 @@ type HttpServerOptionsConfig struct {
 	ServerName       string     `json:"server_name"`
 	MinVersion       uint16     `json:"min_version"`
 	FlushInterval    int        `json:"flush_interval"`
+	SkipURLCleaning  bool       `json:"skip_url_cleaning"`
 }
 
 type AuthOverrideConf struct {
--- a/main.go
+++ b/main.go
@@ -1320,6 +1320,8 @@ func listen(l net.Listener, err error) {
 
 		// Use a custom server so we can control keepalives
 		if config.HttpServerOptions.OverrideDefaults {
+			defaultRouter.SkipClean(config.HttpServerOptions.SkipURLCleaning)
+
 			log.WithFields(logrus.Fields{
 				"prefix": "main",
 			}).Info("Custom gateway started")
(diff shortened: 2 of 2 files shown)


buger:
Are you sure that this option should be configurable?

matiasinsaurralde:
I think it's a very specific option (and perhaps that's why it's disabled by default in mux).

mvdan:
I also strongly believe this should not be exposed. Why should this ever be off?

mvdan:
Also, I feel like the three commits other than 2e0a131d49b32d95f895c98790502df56cda2592 should be squashed together, as they just change and move around the same lines of code.

lonelycode:
The only reason I prefer this exposed is because it changes default behaviour, and makes the muxer stricter, so it will impact any clients our customers are using that have sloppy URL schemes, which is quite likely.

lonelycode:
@matiasinsaurralde Can you squash the commits as @mvdan suggested?

matiasinsaurralde:
@lonelycode Sure!

matiasinsaurralde:
👍

mvdan:
It makes accepted urls more liberal, not stricter. An example of what could break?

lonelycode:
> It makes accepted urls more liberal, not stricter. An example of what could break?

It does, that's not what I'm getting from the docs:

> When true, if the route path is "/path//to", it will *remain with the double slash*. This is helpful if you have a route like: /fetch/http://xkcd.com/534/

So if we default it to "on" then anyone who has a client that might have a reconfigured URL builder (e.g. that doesn't join strings correctly with slashes, or accidentally adds a `/` in the middle, such as: `strings.join(["http://mydomain.com/", "myfolder/"], "/")` (this happens more often than you think). Basically, by making this assumption on behalf of users, and *changing it* we break code they may be using to access their own services, because suddenly the muxer became strict.

So we need to retain default behaviour (set it to false), and if a user has an API method that requires a double-slash as part of the path, they can explicitly make that work.

mvdan:
Right - I was confused by this because of the double negative (`Skip` set to `false`). I thought this was about turning off path cleaning by default, when it's about keeping it on by default and allowing to turn it off.

SGTM

buger:
👍

matiasinsaurralde:
Thanks 🍾

Issue #376, opened 2017-01-18
OPTIONS should not be cached


gwshaw:
**BUG**
The HTTP standard defines that OPTIONS is not cachable, 
https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html, section 9.2, says:
"Responses to this method are not cacheable."

**What is the current behavior?**
The Gateway Dashboard and related documentation describe that OPTIONS requests are cached. From the Gateway:
"Cache all safe requests (GET, HEAD, OPTIONS) - overrides per-path cache settings in versions, applies across versions"

**What is the expected behavior?**
Not caching OPTIONS.

PR #378, opened 2017-01-19
Remove OPTIONS from the safe method check.

matiasinsaurralde:
This is described in #376.


buger:
👍

mvdan:
If the commit message includes "Fixes #n", the issue will be automatically closed when the commit is merged. Unless I misunderstand and this doesn't actually fix it?

matiasinsaurralde:
> If the commit message includes "Fixes #n", the issue will be automatically closed when the commit is merged. Unless I misunderstand and this doesn't actually fix it?

Updated

buger:
I just realized that "OPTIONS" was not removed from dashboard UI. @matiasinsaurralde can you create an Issue in analytics repo and explain @lghiur how to find it?

PR #379, opened 2017-01-19
Travis cleanup

mvdan:
Now the script is more generic, so it will be easier to port it to other repos that have subtle differences like not using build tags.

### DIFF
--- a/utils/ci-test.sh
+++ b/utils/ci-test.sh
@@ -1,31 +1,42 @@
 #!/bin/bash
 
+# SKIP_LINT=true - only run the tests
+
 set -e
+
+MATRIX=(
+	""
+	"-tags coprocess"
+)
 
 # print a command and execute it
 show() {
-	echo "$@"
+	echo "$@" >&2
 	eval "$@"
 }
 
 fatal() {
-	echo "$@"
+	echo "$@" >&2
 	exit 1
 }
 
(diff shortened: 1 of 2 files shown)


utils/ci-test.sh
```diff
@@ -1,31 +1,42 @@
...
 
+MATRIX=(
+	""
+	"-tags coprocess"
+)
+
 # print a command and execute it
 show() {
-	echo "$@"
+	echo "$@" >&2
```
buger: Are you sure that "show" function should echo to stderr?

mvdan: Why not? The tools we run use stdout and stderr, so it doesn't make much of a difference. This was for consistency with fatal, mainly.

buger: Well stderr for errors, stdout for warns and info messages. Just about semantics. 
In this concrete case, there is high chance that Travis renders "stderr" and "stdout" with different color, so there may be some confusion.

mvdan: It depends - I usually only use stdout for useful output, so not log messages even if they are not errors :) Otherwise, pipes would be of little use.

Travis doesn't treat the text differently either, see https://travis-ci.org/TykTechnologies/tyk/jobs/193369280#L304

I don't see a clear winner, so I went with this for consistency.

buger: If so I'm ok :)

PR #381, opened 2017-01-19
Allow "/tyk/apis" without trailing slash

### LINKED ISSUE #372: Creating API using `/tyk/apis` should not require trailing slash
When I try to send API definition to "/tyk/apis" endpoint it returns 404 HTTP error.
Right now you have to add trailing slash "/tyk/apis/" to make it work.

buger:
Should fix #372 

This is the first URL user see when initializing Tyk, so it’s better to avoid any confusions. 

gorilla.Mux require you to make two handlers mappings if a route can be used with or without nested routes.

This is somehow related to #377, but our /tyk/* router already cleans the paths by default.

It is quite easy to test, before this change, any request to "/tyk/apis" (GET or POST) will return 404.

### DIFF
--- a/api.go
+++ b/api.go
@@ -778,7 +778,12 @@ func HandleDeleteAPI(APIID string) ([]byte, int) {
 }
 
 func apiHandler(w http.ResponseWriter, r *http.Request) {
-	APIID := r.URL.Path[len("/tyk/apis/"):]
+	var APIID string
+
+	if r.URL.Path != "/tyk/apis" {
+		APIID = r.URL.Path[len("/tyk/apis/"):]
+	}
+
 	var responseMessage []byte
 	var code int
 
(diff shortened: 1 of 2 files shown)


mvdan:
Add the "Fixes #n" to the commit?

api.go
```diff
@@ -778,7 +778,14 @@ func HandleDeleteAPI(APIID string) ([]byte, int) {
 }
 
 func apiHandler(w http.ResponseWriter, r *http.Request) {
-	APIID := r.URL.Path[len("/tyk/apis/"):]
+	var APIID string
+
+	if r.URL.Path == "/tyk/apis" {
+		APIID = ""
```
mvdan: This is line is redundant, negate the if and remove this.

buger: Right!

buger:
> Add the "Fixes #n" to the commit?
Fixed the comment, rebased, and added "fixes" to commit message.

mvdan:
LGTM

Issue #341, opened 2017-01-12
Runtime error isn't handled if a user attempts to generate an OAuth Client for a non-existant API


lonelycode:
When a user attempts to POST an OAuth Client to a new API before it has had a chance to propagate through the system, a runtime error  ("runtime error: for a invalid memory address or nil pointer dereference") is generated. This error should be caught by Tyk and provide the user with a more useful error message.

STR:

1) I set Authentication mode: OAuth 2.0 for my API Def
2) Go to OAuth Clients in the Dashboard and have error: Could not retrieve apps list
3) In the Tyk Dashboard log:

```
Dec 23 16:30:07 db1 tyk-analytics: time="Dec 23 16:30:07" level=error msg="Failed to delete key"
Dec 23 16:30:07 db1 tyk-analytics: time="Dec 23 16:30:07" level=error msg="Get http://gateway1.dev.xxx:8080/tyk/oauth/clients/de77e9f39ce64bc65dfdbdc46837e7d9: EOF"
Dec 23 16:30:07 db1 tyk-analytics: time="Dec 23 16:30:07" level=error msg="App request failed"
4) In the Tyk Gateway Log:
Dec 23 16:30:07 gateway1 tyk: 2016/12/23 16:30:07 http: panic serving 192.168.38.24:51508: runtime error: invalid memory address or nil pointer dereference
Dec 23 16:30:07 gateway1 tyk: goroutine 89648 [running]:
Dec 23 16:30:07 gateway1 tyk: net/http.(*conn).serve.func1(0xc420a8a480)
Dec 23 16:30:07 gateway1 tyk: /usr/local/go/src/net/http/server.go:1491 +0x12a
Dec 23 16:30:07 gateway1 tyk: panic(0xcb24a0, 0xc420010080)
Dec 23 16:30:07 gateway1 tyk: /usr/local/go/src/runtime/panic.go:458 +0x243
Dec 23 16:30:07 gateway1 tyk: main.getOauthClients(0xc420debbd7, 0x20, 0xd9cdf5, 0x3, 0xc420ddb901, 0x1)
Dec 23 16:30:07 gateway1 tyk: /src/github.com/TykTechnologies/tyk/api.go:1851 +0x78
Dec 23 16:30:07 gateway1 tyk: main.oAuthClientHandler(0x12541e0, 0xc42095f520, 0xc4205b4b40)
[...]

PR #382, opened 2017-01-20
Perform an additional check before retrieving OAuth client list, fixe…

matiasinsaurralde:
Related to #341.

### DIFF
--- a/api.go
+++ b/api.go
@@ -1840,6 +1840,20 @@ func getOauthClients(APIID string) ([]byte, int) {
 		return responseMessage, 400
 	}
 
+	if thisAPISpec.OAuthManager == nil {
+		log.WithFields(logrus.Fields{
+			"prefix": "api",
+			"apiID":  APIID,
+			"status": "fail",
+			"err":    "API not found",
+		}).Error("Failed to retrieve OAuth client list.")
+
+		notAvailable := APIStatusMessage{"error", "OAuth client list isn't available or hasn't been propagated yet."}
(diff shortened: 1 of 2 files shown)


api.go
```diff
@@ -1840,6 +1840,20 @@ func getOauthClients(APIID string) ([]byte, int) {
 		return responseMessage, 400
 	}
 
+	if thisAPISpec.OAuthManager == nil {
```
lonelycode: Wouldn;t it also be possible for the API Spec to be nil too?

buger:
Is there is way to add test for this change?

api.go
matiasinsaurralde: That check is right before this one!

lonelycode: Oh!~ Hurr durr....

api_test.go
```diff
@@ -538,3 +538,22 @@ func TestAPIAuthOk(t *testing.T) {
 		t.Error("Access to API should have been blocked, but response code was: ", recorder.Code)
 	}
 }
+func TestGetOAuthClients(t *testing.T) {
+	var testAPIID = "1"
+	var responseCode int
+
+	var tempSpecRegister = make(map[string]*APISpec)
+	ApiSpecRegister = &tempSpecRegister
```
buger: Since ApiSpecRegister is a global object 😞 you need to clean up for yourself and the end of the test. In your case it probably will be ApiSpecRegister = nil

matiasinsaurralde: Updated

buger:
Thank you!

PR #383, opened 2017-01-20
Create separate contributing guidelines

buger:
I tried to make compilation of our existing docs and took parts from top OSS projects :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,79 @@
+# Contributing to Tyk
+
+**First**: if you're unsure or afraid of anything, just ask or submit the issue or pull request anyways. You won't be yelled at for giving your best effort. The worst that can happen is that you'll be politely asked to change something. We appreciate any sort of contributions, and don't want a wall of rules to get in the way of that.
+
+However, for those individuals who want a bit more guidance on the best way to contribute to the project, read on. This document will cover what we're looking for. By addressing all the points we're looking for, it raises the chances we can quickly merge or address your contributions.
+
+## Filing issues
+
+If you have a question about Tyk or have a problem using it, please
+start with the GitHub search and our [community forum](https://community.tyk.io). If that doesn't answer your questions, or if you think you found a bug, please [file an
+issue](https://github.com/kubernetes/kubernetes/issues/new).
+
+## How to become a contributor and submit your own code
+
+### Contributor License Agreements
+
+We'd love to accept your patches! Before we can take them, we have to jump a couple of legal hurdles.
+
+The Tyk CLA [must be signed](https://github.com/TykTechnologies/tyk/blob/master/CLA.md) by all contributors. You will be automatically asked to sign CLA once PR will be created.
+
+Once you are CLA'ed, we'll be able to accept your pull requests. For any issues that you face during this process, please create an GitHub issue explaining the problem and we will help get it sorted out.
+
+***NOTE***: Only original source code from you and other people that have
+signed the CLA can be accepted into the repository. This policy does not
+apply to [vendor](vendor/).
+
+### Finding Things That Need Help
+
+If you're new to the project and want to help, but don't know where to start,
+we have a semi-curated list of issues that have should not need deep knowledge
+of the system.  [Have a look and see if anything sounds
+interesting](https://github.com/TykTechnologies/tyk/issues?q=is%3Aopen+is%3Aissue+label%3Ahelp-wanted).
+
+Alternatively, read some of the [many docs on the system](https://tyk.io/tyk-documentation/get-started/), and pick a component that seems
+interesting.  Start with `main()` and read
+until you find something you want to fix.  The best way to learn is to hack!
+There's always code that can be clarified and variables or functions that can
+be renamed or commented.
+
+### `master` is unstable
+
+We will do our best to keep master in good shape, with tests passing at all times. But in order to move fast, we will make API changes that your application might not be compatible with. We will do our best to communicate these changes and version appropriately so you can lock into a specific version if need be. For stable releases check our tags and `stable` branch, which contains our latest stable release.
+
+
+### Contributing A Patch
+
+If you're working on an existing issue, such as one of the `help-wanted` ones
+above, simply respond to the issue and express interest in working on it.  This
+helps other people know that the issue is active, and hopefully prevents
+duplicated efforts.
+
+If you want to work on a new idea of relatively small scope:
+
+1. Submit an issue describing your proposed change to the repo in question.
+1. The repo owners will respond to your issue promptly.
+1. Clone the repo, develop, and test your changes.
+1. Submit a pull request.
+1. If your proposed change is accepted, and you haven't already done so, sign a
+   Contributor License Agreement (see details above).
+
+If you want to work on a bigger idea, we **strongly** recommend that you start with
+some bugs or smaller features. It is always better to discuss your idea with our team first, before implementing it.
+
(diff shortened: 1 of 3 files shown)


lonelycode:
LGTM

CONTRIBUTING.md
```diff
@@ -0,0 +1,80 @@
...
+If you have a question about Tyk or have a problem using it, please
+start with the GitHub search and our [community forum](https://community.tyk.io). If that doesn't answer your questions, or if you think you found a bug, please [file an
+issue](https://github.com/kubernetes/kubernetes/issues/new).
+
+## How to become a contributor and submit your own code
+
+### Contributor License Agreements
+
+We'd love to accept your patches! Before we can take them, we have to jump a
+couple of legal hurdles.
```
mvdan: A couple? What's the second?

CONTRIBUTING.md
```diff
@@ -0,0 +1,80 @@
...
+of the system.  [Have a look and see if anything sounds
+interesting](https://github.com/TykTechnologies/tyk/issues?q=is%3Aopen+is%3Aissue+label%3Ahelp-wanted).
+
+Alternatively, read some of the [many docs on the system](https://tyk.io/tyk-documentation/get-started/), and pick a component that seems
+interesting.  Start with `main()` and read
+until you find something you want to fix.  The best way to learn is to hack!
+There's always code that can be clarified and variables or functions that can
+be renamed or commented.
+
+### `master` is unsafe
```
mvdan: `unstable`?

CONTRIBUTING.md
```diff
@@ -0,0 +1,80 @@
...
+If you want to work on a new idea of relatively small scope:
+
+1. Submit an issue describing your proposed change to the repo in question.
+1. The repo owners will respond to your issue promptly.
+1. Clone the repo, develop, and test your changes.
+1. Submit a pull request.
+1. If your proposed change is accepted, and you haven't already done so, sign a
+   Contributor License Agreement (see details above).
+
+If you want to work on a bigger idea, we STRONGLY recommend that you start with
```
mvdan: Replace caps with bold text

CONTRIBUTING.md
```diff
@@ -0,0 +1,80 @@
...
+some bugs or smaller features. It is always better to discuss your idea with our team first, before implementing it.
+
+### Downloading the project
+You need to clone Tyk from GitHub to your GOPATH folder, or alternatively you can run `go get -d github.com/TykTechnologies/tyk` which automatically downloads project to the right path.
```
mvdan: I would add `-u` here.

CONTRIBUTING.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Building the project
+You need to have working Go enironment: see [golang.org](https://golang.org/doc/code.html) for more info on how Go works with code.
+
+To build and test Tyk use built-in `go` commands: `go build` and `go test -v`. If you want to just test a subset of the project, you can pass the `-run` argument with name of the test.
```
mvdan: If this is for people not used to Go, recommend `install` instead of `build`. Otherwise any tiny change will take dozens of seconds to compile.

CONTRIBUTING.md
```diff
@@ -0,0 +1,80 @@
...
+
+Currently in order for tests to pass, a **Redis host is required**. We know, this is terrible and should be handled with an interface, and it is, however in the current version there is a hard requirement for the application to have its default memory setup to use redis as part of a deployment, this is to make it easier to install the application for the end-user. Future versions will work around this, or we may drop the memory requirement. Simplest way to run Redis is to use official Docker image [https://hub.docker.com/_/redis/](https://hub.docker.com/_/redis/)
+
+### Adding dependencies
+
+If your patch depends on new packages, ensure that they will be put in `/vendor` folder.
```
mvdan: With what tool? What commands to run?

CONTRIBUTING.md
buger: Well, this section for contributors, and named "Building the project" the project. And it still generate binary, if you want just to test Tyk.

`install` is a tricky command and rarely anyone use it. By default, $GOPATH/bin does not add to the $PATH, so it will only confuse newcomers. And frankly, I did not saw anyone using it in documentation.

CONTRIBUTING.md
buger: Well, I'm not sure if @lonelycode used anything like that. I usually just use `git clone` with specified path, it automatically creates directories.

CONTRIBUTING.md
buger: I'm not sure why, but right now with "-u" it generate lot of warnings like this:

```
package github.com/TykTechnologies/tyk/vendor/gopkg.in/vmihailenco/msgpack.v2: github.com/TykTechnologies/tyk is a custom import path for https://github.com/TykTechnologies/tyk, but /Users/buger/Documents/work/go/src/github.com/TykTechnologies/tyk is checked out from https://github.com/TykTechnologies/tyk.git
```

CONTRIBUTING.md
mvdan: This isn't about binaries. This is about incremental compilation. This is why I rarely ever use `build`. The tricky command is `build`, not `install` :)

CONTRIBUTING.md
mvdan: If we're to add docs about vendoring, we should at least mention what tool we use. Otherwise I would remove this section.

CONTRIBUTING.md
mvdan: That's because you did the clone yourself and used the `.git` at the end.

CONTRIBUTING.md
buger: It still moves the binary :) 
But trick caching during running install indeed nice, make sense try it with our dev environment project.

Well, this is trick, anyway. Those who know prefer about install, ignore our guide anyway.

CONTRIBUTING.md
buger: Added vendoring example

CONTRIBUTING.md
mvdan: > But trick caching during running install indeed nice, make sense try it with our dev environment project.

I don't understand this sentence at all, sorry.

I still think we should not encourage the use of `go build`. Newbies will end up rebuilding all the dependencies from scratch every single time, wondering why build times are so ridiculously slow.

mvdan:
Why was this PR red right before merging?

CONTRIBUTING.md
mvdan: I'm not particularly happy with the example. I'm fairly sure we don't use git clone to add dependencies. This method does not support updating existing deps and will break with nested dependencies.

CONTRIBUTING.md
buger: > trick **with ** caching

I specifically checked top Golang projects I know, now one use it. It means something. `go build` is a default by Go community, and everyone knows how to use it. So, leave it as it is.

buger:
There are NO code changes here, red was caused by Github issues.

CONTRIBUTING.md
mvdan: I've only seen `go build` being the default when people don't fully understand that it doesn't do incremental builds.

https://twitter.com/davecheney/status/700099491655909376
https://twitter.com/davecheney/status/697171533215576064

I still don't see why using `go install` by default is wrong. If any users are bothered by the cached object files or the binaries in `$GOPATH/bin`, they can just use build instead of install. Fast builds are definitely more important.

PR #386, opened 2017-01-24
Make tests less chatty, other fixes

mvdan:
I noticed the rest while looking for `var log`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 sudo: false
 language: go
 
+env:
+  global:
+    - TYK_LOGLEVEL=info
+
 matrix:
   include:
     - go: 1.6.x
--- a/api.go
+++ b/api.go
@@ -61,7 +61,7 @@ func GetSpecForApi(APIID string) *APISpec {
 		return nil
 	}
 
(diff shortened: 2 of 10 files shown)


mvdan:
Travis, wake up.

buger:
Yeah, I was thinking about those maps as well.

@lonelycode is there is any reason why you used pointer to the map? I mean map is already pointer by design. There is lot of places in Tyk when we use double pointers.

mvdan:
There's a lot of `*[]*Foo` that I wanted to tackle next. Those can be useful sometimes, though. Map pointers are certainly wrong.

lonelycode:
> There's a lot of *[]*Foo that I wanted to tackle next. Those can be useful sometimes, though. Map pointers are certainly wrong.

Careful about these :-/

mvdan:
Yep, which is why I left them for another time.

lonelycode:
>  is there is any reason why you used pointer to the map? I mean map is already pointer by design. There is lot of places in Tyk when we use double pointers.

@buger Ah, my golang-fu failing me... if they are already pointers and treated as such then yeah, we can kill the double pointers to maps.

PR #387, opened 2017-01-24
Fix api_loader race

mvdan:


api_loader.go
```diff
@@ -573,17 +571,13 @@ func loadApps(APISpecs *[]*APISpec, Muxer *mux.Router) {
...
 	// Create a new handler for each API spec
 	loadList := make([]*ChainObject, len(*APISpecs))
 	generateListenPathMap(APISpecs)
 	for i, referenceSpec := range *APISpecs {
-		var subrouter *mux.Router
-
-		go func(chChan chan *ChainObject, referenceSpec *APISpec, i int, subrouter *mux.Router) {
-			defer wg.Done()
+		go func(referenceSpec *APISpec, i int) {
+			var subrouter *mux.Router
```
buger: Since you started refactoring it, probably make sense to simplify it even more and just `subrouter := Muxer` and remove `else` statement below.

mvdan: Done. This made for a manual rebase, though.

buger:
👍

PR #388, opened 2017-01-24
Simplify code a little more

mvdan:


instrumentation_statsd_sink.go
```diff
@@ -308,7 +308,7 @@ func (s *StatsDSink) writeStatsDMetric(b []byte) {
 
 	lenUdpBuf := s.udpBuf.Len()
 
-	if (lenb + lenUdpBuf) > maxUdpBytes {
+	if lenb+lenUdpBuf > maxUdpBytes {
```
mvdan: > having spaces around arithmetic operations is kinda required

gofmt reigns here, so I don't think this matters.

> And in this case brackets are ok as well

Since spaces are not used, operator precedence is obvious from the style. I don't think parentheses here improve readability at all.

buger: Well, if this was a gofmt suggestion, then for sure will be easier not to fight with it. 👍

PR #391, opened 2017-01-24
Remove the test dependency on redis

### LINKED ISSUE #361: Remove the test dependency on redis
Seems like the easiest way would be to use https://github.com/alicebob/miniredis, proposed by @buger.

If that doesn't work for us, perhaps we could mock it ourselves at a higher level.

mvdan:
Use miniredis instead.

main.go reads the config and sets up the storage, so there is no easy
way to replace the port when running the tests. The smallest change that
I could think of is not resetting it when running tests.

Fixes #361.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,8 @@ matrix:
       env: SKIP_LINT=true
     - go: 1.7.x
 
-services:
-  - redis-server
-
 install:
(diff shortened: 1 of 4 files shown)


buger:
Good work! Like how small change is 👍

lonelycode:
Lovely!

PR #392, opened 2017-01-25
Expose pprof http interface

buger:
If you pass `--httpprof` argument, you will be able to get access to a standard [Go profiler](https://golang.org/pkg/net/http/pprof/) in runtime `go tool pprof ./tyk http://localhost:8181/debug/pprof/heap`.

It is very helpful when you debug memory leaks or performance issues https://blog.golang.org/profiling-go-programs

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,6 +16,8 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


lonelycode:
LGTM

PR #394, opened 2017-01-25
Fix data race in the host checker

mvdan:
Fix data race in the host checker

StartPoller() will start its own goroutine and StopPoller doesn't loop,
so both goroutines here are redundant. In fact they introduce a race:

        Write at 0x00c4202564e9 by goroutine 70:
          github.com/TykTechnologies/tyk.(*HostUptimeChecker).ResetList()
              /home/mvdan/go/src/github.com/TykTechnologies/tyk/host_checker.go:257 +0x40
          github.com/TykTechnologies/tyk.(*HostCheckerManager).UpdateTrackingList()
              /home/mvdan/go/src/github.com/TykTechnologies/tyk/host_checker_manager.go:346 +0x674
          [...]

        Previous write at 0x00c4202564e8 by goroutine 47:
          github.com/TykTechnologies/tyk.(*HostCheckerManager).StartPoller()
              /home/mvdan/go/src/github.com/TykTechnologies/tyk/host_checker_manager.go:169 +0x9cb

This is because we might use the poller before it's finished starting
up.

### DIFF
--- a/host_checker_manager.go
+++ b/host_checker_manager.go
@@ -106,14 +106,14 @@ func (hc *HostCheckerManager) CheckActivePollerLoop() {
 					"prefix": "host-check-mgr",
 				}).Info("Starting Poller")
 				hc.pollerStarted = true
-				go hc.StartPoller()
+				hc.StartPoller()
 			}
 		} else {
 			log.WithFields(logrus.Fields{
 				"prefix": "host-check-mgr",
 			}).Debug("New master found, no tests running")
 			if hc.pollerStarted {
-				go hc.StopPoller()
+				hc.StopPoller()
 				hc.pollerStarted = false
 			}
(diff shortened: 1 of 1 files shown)


buger:
I briefly checked if StartPoller or EndPoller could be blocking, and seems like this change may work. 
However pls re-check all the dependent code, it is potentially dangerous change.

mvdan:
This is covered by `go test`, which is how I found the race. The goroutine handling is entirely in the covered code, so it's IMO safe.

If you guys want to wait until the bugfix is released to merge this just in case, that's fine by me.

lonelycode:
Lets wait until the bugfix release, then we can get schwifty.

mvdan:
reping since bugfix is ready

PR #396, opened 2017-01-25
Remove unused code, fix interface and use pointer receivers

mvdan:


mvdan:
Also, I realise this makes the `New()` lines more complex. But they're already complex anyway. I plan to refactor that into a proper constructor func soon.

ilijabojanovic:
@buger should we add this in some milestone?

lonelycode:
@ilijabojanovic This isn't a feature, so milestone here does not matter.

PR #402, opened 2017-01-26
Test all build tag combinations on CI

mvdan:


mvdan:
Travis, wake the fuck up.

mvdan:
@matiasinsaurralde as we discussed, this only tests their builds.

PTAL @buger @lonelycode

mvdan:
Also note that this only includes the build dependencies, not the runtime ones. If we ever need to run any python or lua as part of the coprocess tests, we'll need more packages.

buger:
I think it should be a good start for now 👍

Issue #362, opened 2017-01-16
Run tests in parallel


mvdan:
I expect many of the tests could be run in parallel. This would speed up the whole test suite, which is currently a bit slow.

This would also help us find race bugs and other kinds of issues that won't arise for as long as the tests run sequentially. This won't work well with all the log messages (#330), so we first need to get rid of those.

buger:
I think this is related to #361. If Redis directly used in tests, there is high chance that one of them re-use same keys.

mvdan:
That's a good point. There probably will be more issues since tests have always been run sequentially - temporary files, used ports, global variables...

mvdan:
The tests pass after 40s on my laptop, with none of my CPU cores ever going past 5% use. 40s is way too long for me to quickly iterate. I think this is pretty high priority, as I'd rather not waste my time.

lonelycode:
The tests should be ok to run in any order, however they have always run sequentially. Since they ised to run on an instance with a standing redis DB, "duplication" of data shouldn;t be an issue. You may get other problems with port bindings though.

Has anyone actually tried running in parallel?

mvdan:
Nope, but it's on my list lf things to do soon. Would self-assign but can't anymore :)

lonelycode:
As you wish.... ;-)

mvdan:
As it turns out, quite a lot of tests modify global state (like the global `config`). Ideally this wouldn't be true, but it's unrealistic to refactor everything to remove global state just so we can run every test in parallel.

A good part of the tests don't modify global state though, and would likely still work in parallel. Keeping this issue open to track those.

mvdan:
After removing lots of network use from the tests, the total test time is down from ~4s to ~2s. I think that's acceptable for now, and it's no longer high priority to make lots of tests parallel.

Not a fan of umbrella issues with non-specific goals, so going to close this.

PR #403, opened 2017-01-26
[WIP] Mark some tests as parallel

mvdan:
Some of the others are flakey if run in parallel. Prioritized trivial
and slow ones here.

'go test' time on my machine is down from ~5s to ~1.5s.

Updates #362.

### DIFF
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -134,6 +134,7 @@ func createDefinitionFromString(defStr string) *APISpec {
 }
 
 func TestExpiredRequest(t *testing.T) {
+	t.Parallel()
 	uri := "/v1/bananaphone"
 	method := "GET"
 
@@ -158,6 +159,7 @@ func TestExpiredRequest(t *testing.T) {
 }
 
 func TestNotVersioned(t *testing.T) {
+	t.Parallel()
 	uri := "v1/allowed/whitelist/literal"
 	method := "GET"
 
@@ -185,6 +187,7 @@ func TestNotVersioned(t *testing.T) {
 }
 
 func TestMissingVersion(t *testing.T) {
+	t.Parallel()
 	uri := "/v1/bananaphone"
 	method := "GET"
 
@@ -208,6 +211,7 @@ func TestMissingVersion(t *testing.T) {
 }
 
 func TestWrongVersion(t *testing.T) {
+	t.Parallel()
 	uri := "/v1/bananaphone"
 	method := "GET"
 
@@ -232,6 +236,7 @@ func TestWrongVersion(t *testing.T) {
 }
 
 func TestBlacklistLinks(t *testing.T) {
+	t.Parallel()
 	uri := "v1/disallowed/blacklist/literal"
 	method := "GET"
 	param := make(url.Values)
@@ -274,6 +279,7 @@ func TestBlacklistLinks(t *testing.T) {
 }
 
 func TestWhiteLIstLinks(t *testing.T) {
+	t.Parallel()
 	uri := "v1/allowed/whitelist/literal"
 	method := "GET"
 	param := make(url.Values)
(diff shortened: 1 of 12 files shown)


buger:
A lot of this tests using RedisClusterStorageManager with same key prefixes. 
Even if we switched to miniredis, it is still a global storage for all tests.

So, maybe for tests that marked as parallel, we will use random prefixes? Assuming that removing global Redis object is not an option right now, smth like this may work:

```
# For purpose used fixed vocabulary, because reading from random may break redis
const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"

func randomizePrefix(prefix string) string {
    b := make([]byte, 10)
    for i := range b {
        b[i] = letterBytes[rand.Intn(len(letterBytes))]
    }
    return prefix + "-" + string(b)
}

...

RedisStorageManager{KeyPrefix: randomizePrefix("apikey")}
```

mvdan:
@buger I was thinking the same, but apparently this doesn't happen. If it did, I assume many of the tests would fail. I made sure via `-count 100`, i.e. firing all the tests 100 times. If any keys were stomped, a good part of the runs would have failed then.

Your solution doesn't convince me either - there is still space for flakiness. I would use something like the test name as the prefix, which is known to be unique. I can do that change, if you wish.

I'd first like input from @lonelycode as to whether the keys are repeated or not.

buger:
Maybe in our case it make sense not to add parallel tests at all (considering that test time already great), and instead in TestMain ensure to Cleanup redis after each test call. This should be bulletproof way to avoid any redis related issues.

mvdan:
> Maybe in our case it make sense not to add parallel tests at all (considering that test time already great)

I disagree. Four seconds is better than before, but it's still pretty bad. Remember that we have a 30% test coverage. Once that reaches decent numbers like 80%, the 4s could easily duplicate or triplicate. Parallel tests don't increase linearly in time consumed.

> and instead in TestMain ensure to Cleanup redis after each test call. This should be bulletproof way to avoid any redis related issues.

This is not how TestMain works, unless I'm missing something. In any case, it would be a hack and a workaround for our messy use of redis in the tests (if it is messy).

buger:
Well, we have to find some solution to global Redis object somehow, otherwise chances of getting crazy bugs and flaky tests too high. Especially when we start adding more tests.

mvdan:
@buger sure, but giving up parallel tests altogether isn't a good option IMHO. A few of our tests already take long (half a dozen are around `0.2s`) while not using almost any CPU, so it's definitely the way to go.

This PR is not high priority, we can discuss keys with Martin once he's back.

buger:
>  I would use something like the test name as the prefix, which is known to be unique. I can do that change, if you wish.

This solution you proposed looks fine to me, and should solve our problem.

mvdan:
> I'd first like input from @lonelycode as to whether the keys are repeated or not.

Just in case there's already something in place for this that we've both missed :)

buger:
If you look at the used keys, they are like "zxczxc" or "56423". So I'm sure he have no clue either :)

Even if current keys does not conflict, we can't rely on it in new tests.

lonelycode:
As far as I can remember the keys are not reused between tests (or they shouldn't be), in some other tests I think we started adding random numbers.

mvdan:
@lonelycode thanks, what I'll do is make key prefixes unique.

Unfortunately this will come with 1.8: https://tip.golang.org/pkg/testing/#T.Name

Will pick this up in a couple of days once it's released.

lonelycode:
sgtm

mvdan:
Unfortunately, 1.8 has been delayed by two weeks. So delaying this PR further.

buger:
Even when 1.8 get's released how you gonna keep compatibility between 1.7 and 1.8, assuming that we not going to ship tyk with 1.8 until there will be at least few patch releases?

mvdan:
@buger I can implement the helper func (the one that sets up the prefixes) twice, with test names for 1.8+ and with a random string pre-1.8. This way we can assure no failures going forward. We'll have to live with the potential collisions until we drop support for 1.7, but should be fine as long as the strings are long enough.

Then again, I can implement this right now, no need to wait for 1.8.

buger:
👍

mvdan:
I need to send other incremental patches first, so will redo this PR from scratch.

PR #404, opened 2017-01-26
Deduplicate some code

mvdan:


rpc_storage_handler.go
```diff
@@ -596,31 +596,8 @@ func (r *RPCStorageHandler) SetRollingWindow(keyName string, per int64, val stri
 
 }
 
-// SetScrollingWindow is used in the rate limiter to handle rate limits fairly.
 func (r *RPCStorageHandler) SetRollingWindowPipeline(keyName string, per int64, val string) (int, []interface{}) {
```
buger: So, if this function do not have any differences from `SetRollingWindow`, maybe make sense to remove it a all, and replace all references by `SetRollingWindow` ?

mvdan: I tried this first, but it doesn't work at all. I forgot to mention why in the commit, will update.

mvdan: @buger done, see the new commit message

buger: Got it, make sense

Issue #410, opened 2017-01-29
Hot reloads do not register new APIs or changed API Definition settings


lonelycode:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
When adding a new Middleware setting or a new API, and triggering a hot-reload on the gateway (e.g. by updating the API or addintg the API in the dashboard), the gateway hot reloads, but the settings do not take effect - e.g. adding aheader injection.

**What is the expected behavior?**
That on hot reload the muxer is properly replaced and the new API definition settings take effect.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. Start the gateway
2. Create an API with the dashboard
3. Wait for the reload
4. Use the APi, it will 404

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
Master

lonelycode:
Looks like this is caused by #340, removing the relevant commit (`2f5b2247f195296568fcdc259613701b652e1c46`) makes reload behaviour revert back to the desired state.. Suggest re-opening #340 as it will need to be rolled back.

PR #412, opened 2017-01-29
mux.Router SkipClean fix, restoring listen() logic.

matiasinsaurralde:
Related to #340 and #410.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1381,11 +1381,10 @@ func listen(l net.Listener, err error) {
 			log.WithFields(logrus.Fields{
 				"prefix": "main",
 			}).Printf("Gateway started (%v)", VERSION)
-			if RPC_EmergencyMode {
-				go http.Serve(l, nil)
-			} else {
-				go http.Serve(l, mainRouter)
+			if !RPC_EmergencyMode {
+				http.Handle("/", mainRouter)
 			}
+			go http.Serve(l, nil)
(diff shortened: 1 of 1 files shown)


mvdan:
What makes this fix correct in comparison to the old fix?

buger:
Yes, pls describe how it fix the bug. And what was the reason behind the bug. 

Also, I'm not sure why this PR even exist, we already have this functionality in our master https://github.com/TykTechnologies/tyk/blob/master/main.go#L1362

To do release we for sure can merge it like this, if it not waits. But we need to add test case which replicate this bug, and verify your fix.

lonelycode:
IIRC when custom http server overrides we're added they did have issues with hot reloads registering properly in some cases, so this might be a deeper issue with how we listen / reload.

matiasinsaurralde:
I thought the previous commits were removed, so this is the actual change we need, I've been able to reproduce this error, and have just tried the steps described by @lonelycode, can you check?

buger:
Is there is a way to add simple test for hot-reload?

mvdan:
+1, tests for this would be very useful since it seems particularly flakey.

buger:
I can confirm that this fix is working 👍 

I'm going to merge it both in master and 2.3.2 and will create a separate issue for adding tests.

Issue #411, opened 2017-01-29
JWT Claims are not included in context vars in subsequent requests


lonelycode:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
When accessing an API with a JWT, and context variables enabled, the context variables are only set the first time the JWT is received.

**What is the expected behavior?**
The JWT claims should always be processed and exposed.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. Create a JWT enabled API
2. Add a header injector that includes one of the claims using the `$tyk_context.jwt_claims_*` namespace
3. Check the request headers after making 2 requests

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
v2.3.x

mvdan:
Is this a regression?

ammaskartik:
Dont mind attempting the fix, looking at the middleware code, looks like the setContextVars are only set if the session doesnt exist, which makes sense on the issue. Should be a simple fix, good for my first one =)

ammaskartik:
Also, we are kinda blocked by this issue, and would rather not do a workaround.

buger:
@ammaskartik cool! Will wait for your PR then.

ammaskartik:
submitted new pull request: https://github.com/TykTechnologies/tyk/pull/413

ammaskartik:
Looks like @lonelycode merged my pull request and your team is pushing for release v2.3.2, this is great and very much appreciated.

ammaskartik:
@lonelycode whats your policy on closing issues, etc... Only after released?

buger:
@ammaskartik  2.3.2 is a patch release, and have own branch now. You pull request needs to be manually merged into 2.3.2 branch (we will do it, should happen soon), that's why issue is not yet closed :)

Hope it makes sense! Thank you for the contribution!

ammaskartik:
@buger that makes perfect sense. 

Thanks.

PR #413, opened 2017-01-30
Fix for #411: JWT Claims are not included in context vars in subsequent requests

ammaskartik:
Simple enough fix, just added the setContextVars outside of first time session check.


lonelycode:
Bumping this into 2.3.2 as it's pretty serious (but small)

lonelycode:
Also, LGTM

PR #415, opened 2017-01-30
Move handler_proxy.go code elsewhere

mvdan:


api_loader.go
```diff
@@ -555,6 +555,15 @@ func processSpec(referenceSpec *APISpec,
 	return &chainDef
 }
 
+type DummyProxyHandler struct {
```
buger: Why did you decide to separate DummyProxyHandler from ProxyHandler definition? Aren't they related?

mvdan: I was driven by use, not by names. The two bits of code aren't related at all, and the file `handler_proxy` seemed unnecessary.

Moreover, this code hasn't been touched in a long while, so I doubt this structure will be useful in the future.

PR #426, opened 2017-01-31
Disable Travis email notifications

buger:
Since some (when I say some, I mean myself) prefer using branches, lot of people may get spammed by Travis notifications. We do not really need them for Tyk, because we using PR to check Travis builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 sudo: false
 language: go
 
+notifications:
+  on_success: never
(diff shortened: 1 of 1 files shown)


mvdan:
Note that this also disables email notifications if master or any other important branch breaks. I'd rather keep those on.

buger:
Modified to send notifications only on failure

mvdan:
sgtm, squash though?

Issue #330, opened 2017-01-11
Tests are too chatty


mvdan:
By default, the tests should not write anything to stderr unless anything went wrong. This is far from the truth right now.

If this is needed for CI or anything else, it should be behind a custom flag like `-test.log` or https://golang.org/pkg/testing/#Verbose.

```
 $ go test
THIS IS THE TEST SETUP INIT
[Jan 11 11:08:23] ERROR Couldn't load configuration file
[Jan 11 11:08:23] ERROR open /etc/tyk/tyk.conf: no such file or directory
[Jan 11 11:08:23]  INFO Writing a default file to ./tyk.conf
[Jan 11 11:08:23] ERROR Failed to process environment variables: specification must be a struct pointer
[Jan 11 11:08:23]  INFO Loading default configuration...
[Jan 11 11:08:23]  INFO Connection dropped, connecting..
[Jan 11 11:08:23]  INFO main: PIDFile location set to: /var/run/tyk-gateway.pid
[Jan 11 11:08:23] ERROR main: Failed to write PIDFile: error opening pidfile /var/run/tyk-gateway.pid: open /var/run/tyk-gateway.pid642434050: permission denied
[Jan 11 11:08:23] ERROR Failed ot get host pid: open /var/run/tyk-gateway.pid: no such file or directory
[Jan 11 11:08:23] ERROR Instrumentation is enabled, but no connectionstring set for statsd
[Jan 11 11:08:23]  WARN Legacy path detected! Upgrade to extended.
[Jan 11 11:08:23]  WARN Legacy path detected! Upgrade to extended.
[Jan 11 11:08:23]  WARN Legacy path detected! Upgrade to extended.
[Jan 11 11:08:23]  WARN Legacy path detected! Upgrade to extended.
[Jan 11 11:08:23]  WARN Legacy path detected! Upgrade to extended.
[Jan 11 11:08:23]  WARN Legacy path detected! Upgrade to extended.
[Jan 11 11:08:23]  WARN Legacy path detected! Upgrade to extended.
[Jan 11 11:08:23]  INFO main: Initialising Tyk REST API Endpoints
[Jan 11 11:08:23]  INFO main: Loading API configurations.
[Jan 11 11:08:23]  INFO main: Tracking hostname api_name=Tyk Test API ONE domain=(no host)
[Jan 11 11:08:23]  INFO main: Loading API api_name=Tyk Test API ONE
[...]

PR #427, opened 2017-01-31
Remove all prints from test code

mvdan:
They look like debugging prints used while writing the tests, so just
remove them.

Updates #330.


lonelycode:
LGTM

PR #432, opened 2017-01-31
Better disk file cleanup for tests

mvdan:
These were fixed in a somewhat sloppy way. This is much cleaner.

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -27,12 +27,13 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		panic(err)
 	}
+	WriteDefaultConf(&config)
 	config.Storage.Port, _ = strconv.Atoi(s.Port())
+	config.EnableAnalytics = true
 	initialiseSystem(map[string]interface{}{})
 	exitCode := m.Run()
 
 	s.Close()
-	os.Remove("tyk.conf")
 	os.Remove(filepath.Join("apps", "1.json"))
 	os.Exit(exitCode)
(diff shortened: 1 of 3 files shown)


gateway_test.go
```diff
@@ -27,13 +26,15 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		panic(err)
 	}
+	WriteDefaultConf(&config)
+	config.AppPath, err = ioutil.TempDir("", "tyk-test-")
 	config.Storage.Port, _ = strconv.Atoi(s.Port())
+	config.EnableAnalytics = true
```
buger: Previously this option was enabled only for 2 tests, did you enabled it for all for purpose?

mvdan: The former method was flawed. They enabled it, but didn't disable it. Any tests after those two would have it enabled, while the rest would have it disabled. Having it consistently on is much better.

On top of that, such setting of global variables will break even worse when we use parallel tests.

The easy fix is to not fiddle with globals in tests unless we have to. And if we have to, that test can't be parallel and must clean after itself.

In any case, no tests seem to break. We can as Martin to be sure if you'd like.

buger: Nope, it fine. Just noticed that behaviour changed.

mvdan: Yeah, my bad for not mentioning it in the commit message. Let me update it.

mvdan: done.

PR #433, opened 2017-01-31
README: clarify what targets we actively support

mvdan:
Also clarify that Go 1.6 or later is required to build, not to run.

cc @lonelycode

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,8 @@
 Tyk is a lightweight, open source API Gateway and Management Platform enables you to control who accesses your API, when they access it and how they access it. Tyk will
 also record detailed analytics on how your users are interacting with your API and when things go wrong.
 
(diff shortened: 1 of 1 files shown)


buger:
Make sense add "officially" supported. Because it actually runs in Mac OS, Windows, and probably bunch of another platforms we not officiall support.

mvdan:
I thought about adding `officially`, but I think it's redundant. What kind of support is not official? What you're describing is more of a "happens to work on", since we don't even test it on those.

buger:
Support as "word" in this case a bit confusing, because is it about customer support, yes, it is redundant, but most users I believe read as a technological limitation and that it will not work on rest of the platforms. I would still add "official" here, just removes all the confusion.

mvdan:
fair enough, done (github hasn't reloaded yet)

PR #436, opened 2017-01-31
Remove the deprecated RedisStorageManager

mvdan:
It was replaced by RedisClusterStorageManager more than a year ago, but
some tests weren't updated. Update them and remove the code for good.

cc @lonelycode


lonelycode:
I guess we can always add it back in again if we need to step back.

mvdan:
Yeah, it's in the history.

PR #439, opened 2017-02-01
Remove hour-long sleep from analytics.go

mvdan:
This func is always called from a new goroutine. Having the goroutine
sleep for an hour for no reason just wastes resources.

@lonelycode please confirm this sleep has no purpose.


lonelycode:
LGTM

PR #440, opened 2017-02-01
Apply all misspell suggestions

mvdan:
See https://github.com/client9/misspell. Only two were wrong, concerning
plurals.


lonelycode:
LGTM

PR #443, opened 2017-02-01
Make ineffassign slightly happier

mvdan:
A couple of errors assigned to but never used.

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -28,6 +28,9 @@ func TestMain(m *testing.M) {
 	}
 	WriteDefaultConf(&config)
 	config.AppPath, err = ioutil.TempDir("", "tyk-test-")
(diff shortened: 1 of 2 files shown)


gateway_test.go
```diff
@@ -28,6 +28,9 @@ func TestMain(m *testing.M) {
 	}
 	WriteDefaultConf(&config)
 	config.AppPath, err = ioutil.TempDir("", "tyk-test-")
+	if err != nil {
+		panic(err)
```
buger: Forgot to remove debug?

mvdan: What do you mean? This is the correct way to handle errors in `TestMain`.

buger: Ah, I see, did not noticed the context. 👍

PR #444, opened 2017-02-01
test: deduplicate spec init code

mvdan:
Also add *testing.T to the whole flow in preparation for using t.Name()
to use unique key prefixes.

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -51,13 +51,10 @@ var apiTestDef = `
 
 `
 
-func makeSampleAPI() *APISpec {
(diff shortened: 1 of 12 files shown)


mvdan:
@buger the idea is that once this is merged, making key prefixes truly unique will be a tiny patch as we only set them up in three lines.

mvdan:
I broke the coprocess tests again. Thank god CI checks for it now.

mvdan:
@buger PTAL

PR #446, opened 2017-02-01
test: don't set up global analytics three times

mvdan:
It's already set in initialiseSystem, and the two other times we set it
up in tests it's with an equal store. Moreover, this is a global so it
ends in data races when 'go test -race' is used.

Since it's a global we can't use a per-test unique KeyPrefix, so just
give up and set it up in TestMain.

Also fix the stutter in the test name.

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -776,16 +776,7 @@ func TestQuota(t *testing.T) {
 	}
 }
 
-func TestWithAnalyticsTestWithAnalytics(t *testing.T) {
-	AnalyticsStore := RedisClusterStorageManager{KeyPrefix: "analytics-"}
-	log.Info("Setting up analytics DB connection")
-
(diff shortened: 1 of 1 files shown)


mvdan:
@lonelycode is it possible to do away with the global, or somehow be able to set a unique `KeyPrefix` for each test as described above?

lonelycode:
The store uses a global driver, and the logic there probably needs more elegance - it's basically a bad singleton. 

But the storage handlers enforce the prefix on all operations on the store and basically ensure a namespace.

The key prefixes are specific to the functionality, so should be consistent in the tests.

mvdan:
@lonelycode makes sense, thanks - do you LGTM this?

I'm not going to change the global right now.

lonelycode:
Yeah this looks good :-)

PR #449, opened 2017-02-01
Remove all uses of *[]*Foo

mvdan:
This only makes sense if the variable is a parameter and it is modified,
e.g. appended to. None of these fall under that case.

Note that there are still some `*[]Foo` around, but I'm leaving those for another day.

@lonelycode PTAL

### DIFF
--- a/api_definition_manager.go
+++ b/api_definition_manager.go
@@ -239,7 +239,7 @@ func (a *APIDefinitionLoader) readBody(response *http.Response) ([]byte, error) 
 }
 
 // LoadDefinitionsFromDashboardService will connect and download ApiDefintions from a Tyk Dashboard instance.
-func (a *APIDefinitionLoader) LoadDefinitionsFromDashboardService(endpoint string, secret string) *[]*APISpec {
+func (a *APIDefinitionLoader) LoadDefinitionsFromDashboardService(endpoint string, secret string) []*APISpec {
 	var APISpecs = []*APISpec{}
 
 	// Get the definitions
@@ -261,20 +261,20 @@ func (a *APIDefinitionLoader) LoadDefinitionsFromDashboardService(endpoint strin
 	response, err := c.Do(newRequest)
 	if err != nil {
 		log.Error("Request failed: ", err)
-		return &APISpecs
+		return APISpecs
 	}
 
 	retBody, err := a.readBody(response)
 	if err != nil {
 		log.Error("Failed to read body: ", err)
-		return &APISpecs
+		return APISpecs
 	}
 
 	if response.StatusCode == 403 {
 		log.Error("Login failure, Response was: ", string(retBody))
 		reloadScheduled = false
 		ReLogin()
-		return &APISpecs
+		return APISpecs
 	}
 
 	// Extract tagged APIs#
@@ -293,14 +293,13 @@ func (a *APIDefinitionLoader) LoadDefinitionsFromDashboardService(endpoint strin
 	if err := json.Unmarshal(retBody, &list); err != nil {
 		log.Error("Failed to decode body: ", err, "Response was: ", string(retBody))
 		log.Info("--> Retrying in 5s")
-		return &APISpecs
-		// return &APISpecs
+		return APISpecs
 	}
 
 	rawList := make(map[string]interface{})
 	if err := json.Unmarshal(retBody, &rawList); err != nil {
 		log.Error("Failed to decode body (raw): ", err)
-		return &APISpecs
+		return APISpecs
 	}
 
 	// Extract tagged entries only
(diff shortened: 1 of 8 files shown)


buger:
I wonder why it was even used originally. Arrays are already pointers, same as maps.

mvdan:
> Arrays are already pointers, same as maps.

Careful with the wording there - you mean slices, not arrays :)

Also, grouping slices with maps isn't exactly correct. Maps are a single pointer, while slices are a three-field struct (arr, len, cap) where the actual backing array is a pointer. Using pointers for slices is still useful to update any of the three fields, as explained above.

buger:
Well, for sure I meant slices.

But when they used as arguments, they send as pointers (yes to struct), not by value. So this is still technically a pointer. And you can access real pointer to "slice" anytime, by dereferencing it anyway, without explicitly sending(or returning it) to function.

Anyway, I did not saw usage of such functionality anywhere, that's why I'm asking, maybe we missing something.

buger:
Actually, I may be wrong about dereferencing it, so it will be possible to update capacity. Still, we not used it anyway :)

lonelycode:
The reason it's this way is to stop copying API Definitions around. Using this was just lighter and faster. We would need to confirm whether this change has a detrimental effect on performance and load times of large API Definitions sets for a gateway (2,000 and up). Until we can be sure of that am reticent to make a change.

mvdan:
Slices don't actually contain the objects directly, so I doubt this would affect performance in any way. Rather the opposite - you're using pointers to 24-byte structs, which is likely to degrade performane a tiny bit.

But my main concern here is making the code overly obtuse for no reason.

This isn't blocking anything else of mine though, so feel free to run your tests first.

lonelycode:
> so I doubt this would affect performance in any way. Rather the opposite - you're using pointers to 24-byte structs, which is likely to degrade performane a tiny bit.

Assumption is the mother of all f*ckups, prove that performance isn't affected and I'm a happy man.

mvdan:
Well, I would if we had benchmarks :) I don't know how you run yours, so I can't prove much.

buger:
Well, while I agree that performance should be ok, try very basic synthetic benchmark of inserting few thousand APISpecs and accessing few fields in contexts of *[] and [], it should be enough.

mvdan:
I've opened https://github.com/TykTechnologies/tyk/issues/453 about our gateway benchmark discussion on Slack.

If I write an APISpec benchmark, would it be useful to keep in the repo?

buger:
This synthetic test unlikely will be really useful, so probably you can just show results and throw it away. But in the context of #453 , it can be useful to add some benchmarks which test specific parts, but it will be separate task anyway.

lonelycode:
> Well, while I agree that performance should be ok, try very basic synthetic benchmark of inserting few thousand APISpecs, iterating over them, and accessing few fields in contexts of *[] and [], it should be enough.

The synthetic test should be:

- Insert a few thousand APIs
- Hot reload them a few times
- Check speed and memory consumption

buger:
@mvdan any updates on this?

mvdan:
Not yet, was distracted with other stuff.

mvdan:
I've added a somewhat stupid benchmark:

```
name               old time/op    new time/op    delta
ApiInsertReload-4    14.9ms ± 1%    14.8ms ± 0%  -0.48%  (p=0.015 n=6+6)

name               old alloc/op   new alloc/op   delta
ApiInsertReload-4    4.37MB ± 0%    4.37MB ± 0%    ~     (p=0.485 n=6+6)

name               old allocs/op  new allocs/op  delta
ApiInsertReload-4     47.2k ± 0%     47.2k ± 0%    ~     (p=0.608 n=6+6)
```

I say it's stupid because it doesn't use http to create the apis, and it doesn't reload them.

The way the tests are written is a bit confusing:

* do I need to set up `http.DefaultServeMux`?
* do I need to use `httptest`?
* don't I need a secret to do these things, like shown in https://tyk.io/tyk-documentation/get-started/with-tyk-on-premise/tutorials/tyk-premise-community-edition/create-an-api/?

Also, @lonelycode, when you said "Insert a few thousand APIs", do you mean one by one or all at once?

mvdan:
Actually, ignore the numbers above as the gateway is erroring under the covers and not actually adding the endpoints.

mvdan:
Actually, these are the real numbers for the silly benchmark - I hadn't noticed the gateway was erroring and not actually doing much.

```
name               old time/op    new time/op    delta
ApiInsertReload-4    77.7ms ± 0%    76.9ms ± 1%  -1.07%  (p=0.030 n=5+6)

name               old alloc/op   new alloc/op   delta
ApiInsertReload-4    33.7MB ± 0%    33.7MB ± 0%    ~     (p=0.937 n=6+6)

name               old allocs/op  new allocs/op  delta
ApiInsertReload-4      400k ± 0%      400k ± 0%    ~     (p=0.619 n=6+6)
```

mvdan:
Okay - I'll open an issue about making the benchmark better.

lonelycode:
👍

PR #455, opened 2017-02-02
Fix call to envconfig when setting conf defaults

mvdan:
envconfig.Process requires a pointer to a struct (i.e. *Config), but
we're passing it a pointer to a pointer to a struct (**Config). This
results in envconfig apparently never being useful:

	ERROR Failed to process environment variables: specification must be a struct pointer

@buger @lonelycode this seems like a bad regression, perhaps it could go into a bugfix.

### DIFF
--- a/config_test.go
+++ b/config_test.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"os"
+	"testing"
+)
+
+func TestWriteDefaultConf(t *testing.T) {
+	conf := &Config{}
+	WriteDefaultConf(conf)
+	if conf.ListenPort != 8080 {
+		t.Error("Expected ListenPort to be set to its default")
+	}
+	*conf = Config{}
+	os.Setenv("TYK_GW_LISTENPORT", "9090")
+	WriteDefaultConf(conf)
+	if conf.ListenPort != 9090 {
+		t.Error("Expected ListenPort to be set to 9090")
+	}
+}
--- a/config_utils.go
+++ b/config_utils.go
@@ -38,7 +38,7 @@ func WriteDefaultConf(conf *Config) {
 	conf.HideGeneratorHeader = false
 	conf.OauthRedirectUriSeparator = ""
 	newConfig, err := json.MarshalIndent(conf, "", "    ")
-	overrideErr := envconfig.Process(ENV_PREVIX, &conf)
+	overrideErr := envconfig.Process(ENV_PREVIX, conf)
 	if overrideErr != nil {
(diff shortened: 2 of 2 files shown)


mvdan:
Also I can attach a test, but I wanted input on whether my intuition regarding this fix was correct.

buger:
"Benefits" of using interface{} as arguments...

Adding test here definitely make sense, at least overriding few arguments using ENV's. 

From what I understand it affects only tests and when tyk run without specifying configuration file, right?

mvdan:
It affects the tests and when tyk is run with no config file, yes. In both cases env vars are discarded when they shouldn't be.

mvdan:
@buger PTAL, confirmed test was failing without the fix.

buger:
I started filling bug for this issue (to move it through our pipeline), and noticed interesting behavior. Tyk binary actually works well, and `TYK_GW_LISTEN_PORT=8080 TYK_GW_STORAGE_HOST=tyk-redis ./tyk` works as expected and overrides config. 

Can it be that only tests are affected? And why tyk works?

mvdan:
As said, this only affects the real binary when the config is missing. Have you tried that?

buger:
So, tyk works because `WriteDefaultConfig` only writes it to disk, and environment changes will be missing there (still bug), but tyk loads this config using `loadConfig` which evaluate environment variables without issues. So in the end, only generated config have bug, but tyk still override its changes from env.

mvdan:
True, but it can still break if you do:

```
rm -f tyk.conf
TYK_GW_LISTENPORT=9090 tyk
tyk
```

Only the second `tyk` run will have the proper setting.

buger:
Yes it still the bug, but not as major as I thought. Including it to 2.4 milestone unless @lonelycode thinks different.

buger:
https://github.com/TykTechnologies/tyk/issues/457

mvdan:
Let me fix the commit message then, to reference the issue.

buger:
Checked, fix is working 👍

mvdan:
Merged before green?

buger:
You only edited commit message, and it was green before. Don't be so pedantic 😛

mvdan:
Travis gets borked if we merge before it can pull from git, so this merged PR will stay red. So it makes things unnecessarily confusing.

Also I'd rather be safe than sorry :)

buger:
Btw, in future try to create Issue first, so we can follow our workflow.

mvdan:
Yeah, that was my bad

PR #456, opened 2017-02-02
Make 'go test' hide all output by default

mvdan:
As discussed with @buger. This can still be overriden at any time, like:

    TYK_LOGLEVEL=info go test -run SomeTest


buger:
Make sense update our contributing guidelines on how our logging works, since this is non-standard behavior.

mvdan:
Done.

PR #462, opened 2017-02-02
Merge latest release 2.3 to stable

buger:


buger:
@lonelycode since buddy works now does not use "stable" branch now, it is safe to merge.

Issue #431, opened 2017-01-31
[Test request] Hot-reload functionality


buger:
When we update API in our Dashboard, Tyk Gateway receives "reload" signal and should update its API spec cache. Recently we found a bug preventing it: https://github.com/TykTechnologies/tyk/pull/412

We should cover it with tests. 

In general, I want to cover reload without Dashboard as well: "/tyk/reload" and "/tyk/reload/group" endpoints.

- [ ] PRO: handle "reload" signal add new API
- [ ] PRO: handle "reload" signal update existing API
- [ ] Hybrid: handle RPC reload signal add new API
- [ ] Hybrid: handle RPC reload signal update existing API
- [x] Community: handle "reload" signal add new API using files
- [x] Community: handle "reload" signal update existing API using files
- [ ] Community: handle "reload" signal add new API using `/tyk/apis`
- [ ] Community: handle "reload" signal update existing API  using `/tyk/apis`

buger:
RPC tests now possible, check https://github.com/TykTechnologies/tyk/pull/516 as reference.

buger:
Will close it for now

Issue #363, opened 2017-01-16
CI: also run tests with -race


mvdan:
I would first like to see other issues solved, like #330, #361 and #362.

Also, if nobody has ever run the tests with `-race`, I expect it will uncover quite a few new issues. Best to first fix them all, and only add the line to CI once it passes consistently.

mvdan:
Some updates:

* I've fixed two races, see fedf73f808de6a8ca905cba33113e9e49ec477b6 and 0936e1a4b644480bd729a544e3cc12974c054272
* The three issues mentioned above are almost completely fixed
* A race remains in `HostCheckerManager` that I have failed to fix so far
* A couple of races remain in `RedisClusterStorageManager` that I haven't looked at yet

mvdan:
I'll revisit this for 2.4 once we have swapped the redis cluster library. It's likely that the redis races are caused by it.

PR #465, opened 2017-02-02
Add tests for reset handlers.

matiasinsaurralde:
A first approach to #431.

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/TykTechnologies/tyk/apidef"
+	"github.com/garyburd/redigo/redis"
 	"github.com/gorilla/mux"
 )
 
@@ -48,8 +49,8 @@ const apiTestDef = `{
 	}
 }`
 
-func makeSampleAPI(t *testing.T) *APISpec {
-	spec := createSpecTest(t, apiTestDef)
+func makeSampleAPI(t *testing.T, def string) *APISpec {
+	spec := createSpecTest(t, def)
 	loadApps([]*APISpec{spec}, discardMuxer)
 	return spec
 }
@@ -72,7 +73,7 @@ func TestHealthCheckEndpoint(t *testing.T) {
 	recorder := httptest.NewRecorder()
 	param := make(url.Values)
 
-	makeSampleAPI(t)
+	makeSampleAPI(t, apiTestDef)
 
 	req, err := http.NewRequest(method, uri+param.Encode(), nil)
 
@@ -126,7 +127,7 @@ func TestApiHandler(t *testing.T) {
 		recorder := httptest.NewRecorder()
 		param := make(url.Values)
 
-		makeSampleAPI(t)
+		makeSampleAPI(t, apiTestDef)
 
 		req, err := http.NewRequest(method, uri+param.Encode(), strings.NewReader(string(body)))
 
@@ -163,7 +164,7 @@ func TestApiHandlerGetSingle(t *testing.T) {
 	recorder := httptest.NewRecorder()
 	param := make(url.Values)
 
-	makeSampleAPI(t)
+	makeSampleAPI(t, apiTestDef)
 
 	req, err := http.NewRequest(method, uri+param.Encode(), strings.NewReader(string(body)))
 
@@ -252,7 +253,7 @@ func TestKeyHandlerNewKey(t *testing.T) {
 	recorder := httptest.NewRecorder()
 	param := make(url.Values)
 
-	makeSampleAPI(t)
+	makeSampleAPI(t, apiTestDef)
 	param.Set("api_id", "1")
 	req, err := http.NewRequest(method, uri+param.Encode(), strings.NewReader(string(body)))
 
@@ -285,7 +286,7 @@ func TestKeyHandlerUpdateKey(t *testing.T) {
 
 	recorder := httptest.NewRecorder()
 	param := make(url.Values)
-	makeSampleAPI(t)
+	makeSampleAPI(t, apiTestDef)
 	param.Set("api_id", "1")
 	req, err := http.NewRequest(method, uri+param.Encode(), strings.NewReader(string(body)))
 
@@ -311,7 +312,7 @@ func TestKeyHandlerUpdateKey(t *testing.T) {
 }
 
 func TestKeyHandlerGetKey(t *testing.T) {
-	makeSampleAPI(t)
+	makeSampleAPI(t, apiTestDef)
 	createKey()
 
 	uri := "/tyk/keys/1234"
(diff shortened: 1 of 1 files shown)


api_test.go
```diff
@@ -564,3 +565,75 @@ func TestGetOAuthClients(t *testing.T) {
...
+		store.StartPubSubHandler(pubSubChannel, func(message redis.Message) {
+			notif := Notification{}
+			if err := json.Unmarshal(message.Data, &notif); err != nil {
+				t.Fatal("Unmarshalling message body failed, malformed: ", err)
+			}
+			if notif.Command == NoticeGroupReload {
+				signalChan <- true
+			}
+		})
+	}(RedisPubSubChannel, cacheStore, signalChan)
```
mvdan: Why are you passing these as arguments?

api_test.go
```diff
@@ -564,3 +565,75 @@ func TestGetOAuthClients(t *testing.T) {
...
+	cacheStore.Connect()
+
+	go func(pubSubChannel string, store RedisClusterStorageManager, signalChan chan bool) {
+		store.StartPubSubHandler(pubSubChannel, func(message redis.Message) {
+			notif := Notification{}
+			if err := json.Unmarshal(message.Data, &notif); err != nil {
+				t.Fatal("Unmarshalling message body failed, malformed: ", err)
+			}
+			if notif.Command == NoticeGroupReload {
+				signalChan <- true
```
mvdan: If the condition isn't met, or this line doesn't execute for any other reason, the main goroutine will block forever. The test will use up all of the timeout. Whatever your purpose here is, there has to be a better way to do it.

matiasinsaurralde: Yes, just saw it on CI. We need to make sure that `NoticeGroupReload` is published after the signal is triggered.

mvdan: Since you're always listening, you always have to send (if you want to use a channel).

matiasinsaurralde: If for any reason this message doesn't arrive, it's ok to hit the timeout and report it as an error (like Travis is reporting now).

mvdan: I'd say otherwise - this wastes everyone's time. The test should fail as soon as possible, instead of hitting the timeout on purpose (which by default is 10m).

matiasinsaurralde: Any suggestion?

buger: Yeah, but probably it make sense to control this situation with own timeout:
```
select {
case signalChan <- true:
case <-time.After(20 * time.Millisecond):
   t.Fatal("Timeout reached")
}
```

mvdan: Having such a tiny timeout will make the test racy and it will fail under `-race`, so please don't do that.

> Any suggestion?

Yes, always send on the channel. Either true or false.

buger: Well, when we are talking about concurrency code there is no true or false, there is a lot of maybe. For sure we trying to control it, but bugs will happen anyway. So why not to have controlled failure.

mvdan: There definitely is a true or false here. This is the wrong way to write tests. This is a version of testing `time.Sleep` with a slightly larger timeout. Of course it will work 99.99% of the times when you run the test, but that's besides the point. The test takes longer than needed and it's flakey.

And as said above, any test using a timeout to work around introduced races will make https://github.com/TykTechnologies/tyk/issues/363 impossible.

buger: For sure it is better to avoid timeouts, but in this case, you have to install triggers into the tyk internals, to check if requests finished/failed and etc, for example checking if `NoticeGroupReload` is pushed. 

If there is sane way to get this event without rewriting tyk, I'm for sure prefer this approach. Just sometimes it can be really tricky.

mvdan: The way Matias wrote it now should be fine, as long as the internals aren't broken and the event never fires. As long as there is no sleep/after that will break `-race`, I'm happy.

matiasinsaurralde:
I'm getting:
```
[Feb  2 17:52:15]  INFO api: Group reload accepted. status=ok
[Feb  2 17:52:15] ERROR Error trying to set value:
[Feb  2 17:52:15] ERROR ERR unknown command 'publish'
[Feb  2 17:52:15] ERROR Could not send notification
```
Don't we have Redis on the testing environment?
I know about `miniredis` but seems it doesn't support `pubsub`?

mvdan:
Nope, search for it on https://github.com/alicebob/miniredis.

We should probably open an issue about it and consider undoing the miniredis switch, unfortunately.

matiasinsaurralde:
@mvdan Just saw this: https://github.com/alicebob/miniredis/issues/1

mvdan:
@matiasinsaurralde rebase on master, pub/sub should work now.

mvdan:
@matiasinsaurralde this needs a rebase now, apologies :)

mvdan:
@matiasinsaurralde looks like 1.8rc3 hit the 10m timeout again

mvdan:
reping @matiasinsaurralde please rebase and fix the timeout issue

matiasinsaurralde:
It was rebased but wasn't able to reproduce it last Friday, seems like some
of the pubsub messages just get lost.

matiasinsaurralde:
@mvdan Passed 🎆

api_test.go
```diff
@@ -552,6 +553,86 @@ func TestGetOAuthClients(t *testing.T) {
...
+	if len(ApiSpecRegister) == 0 {
+		t.Fatal("Hot reload (group) was triggered but no APIs were found.")
+	}
+
+	for {
+		recvSignal := <-signalChan
+		if recvSignal {
+			break
+		}
+	}
```
mvdan: Why the `for`? Should we receive more than once?

matiasinsaurralde: We may receive other messages/notifications (this is what I've observed in my local environment, by using `redis-cli monitor`), so we can't expect that the first message will be the one we're looking for.

mvdan: Could you add a comment, perhaps a TODO if you don't understand why multiple messages are sometimes received?

matiasinsaurralde: I've added a comment, this seems to be the first test to involve pubsub, so when writing tests in the future, this could make it easier to run multiple tests that produce/consume notifications.

buger:
@matiasinsaurralde so what was the issue? I can't figure it out because it is already rebased. How did you make it pass after failures?

matiasinsaurralde:
@mvdan In the second rebase, it just started to work, so I think it's related to @mvdan changes, on `loadApps`, etc.

buger:
@matiasinsaurralde should we merge it now, and you will create separate tests for rest of tests? If so, pls mark checkboxes of completed tests here #431.

matiasinsaurralde:
@buger That makes sense!

Issue #390, opened 2017-01-24
Merge tykcommon repo with this one


mvdan:
There is no advantage to keeping them as separate git repos. I see multiple disadvantages, such as:

* Separate issue trackers
* Changes affecting code in common like API definitions need two PRs
* Unnecessary vendoring

#389 should happen first.

mvdan:
@lonelycode this is my proposal for the new package paths:

* `tykcommon.GetLogger()` -> `tyk/log.Get()`
* `tykcommon` API stuff -> `tyk/api/v0.*` (package name `api`)
* `tykcommon.coprocess.*` -> `tyk/coprocess.*

If it sounds OK with you, I'll handle the move.

buger:
Just to clarify, this will be a new separate package in tyk repo, smth like `tyk/common`?

mvdan:
No, when I say `tyk/log.Get` the package will be `tyk/log`. Same for the rest - they all hang from the root package here without a `common` in between. That's IMO as meaningless as `util`.

lonelycode:
Wouldn't the entire `tyk` codebase be compiled into the binary for whatever is using it then? If I import a package from another large project, I thought the entire project doesn't get compiled in, just the package (it's already a big binary) - but I could be wrong?

Regarding the API Definition objects, I like the idea of versioning them, opens us up to change in the future, though when we were discussing versioning I was talking about the path interfaces exposed by the dashboard and the gateway, not the objects, so maybe `tyk/apidef/v0.*` would be better for the objects as it is different from `api`, which might end up being the actual service?

mvdan:
I missed this comment, apologies.

> Wouldn't the entire tyk codebase be compiled into the binary for whatever is using it then? If I import a package from another large project, I thought the entire project doesn't get compiled in, just the package (it's already a big binary) - but I could be wrong?

If you import a Go package, only it and its dependencies will get linked. What I'm changing here is git repos - all the `tykcommon` stuff will keep on being separate Go packages. The new packages being subpackages of the big `tyk` one doesn't change anything. I don't undesrta

> Regarding the API Definition objects, I like the idea of versioning them, opens us up to change in the future, though when we were discussing versioning I was talking about the path interfaces exposed by the dashboard and the gateway, not the objects, so maybe tyk/apidef/v0.* would be better for the objects as it is different from api, which might end up being the actual service?

In general an `api` package means API objects. It's a name that you'll have to type very often, so IMO the shorter the better.

[...]

PR #467, opened 2017-02-02
Don't overwrite log.Level with InfoLevel

### LINKED ISSUE #466: TYK_LOGLEVEL broken on master
I broke TYK_LOGLEVEL in 6c629371.

Before that commit:

```
$ TYK_LOGLEVEL=error tyk
[logs only errors]
```

After:

```
$ TYK_LOGLEVEL=error tyk
[logs at info level]
```

Fix incoming.

mvdan:
When we run GetLogger(), that func already reads TYK_LOGLEVEL to set the
proper value, defaulting to InfoLevel. If we overwrite it here, we are
ignoring TYK_LOGLEVEL. For example:

	$ TYK_LOGLEVEL=error go test
	[prints all logs, including INFO]

I introduced this regression in 6c629371.

Fixes #466.

@buger I did open an issue this time. I'd do a regression test, but this isn't easy now that our logger level logic is split among multiple git repos. We can open an issue about it and I'll add the test once #390 is done.

### DIFF
--- a/main.go
+++ b/main.go
@@ -900,9 +900,6 @@ func initialiseSystem(arguments map[string]interface{}) {
 		log.WithFields(logrus.Fields{
 			"prefix": "main",
 		}).Debug("Enabling debug-level output")
-	} else {
-		// default is info
-		log.Level = logrus.InfoLevel
 	}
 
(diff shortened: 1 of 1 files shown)


buger:
Works for me 👍

Issue #469, opened 2017-02-02
Reconsider the redis test dependency


mvdan:
See https://github.com/TykTechnologies/tyk/pull/465#issuecomment-277032009.

Miniredis doesn't support pub/sub but we need it. It's a limitation deep in its stack though, so unlikely to be in our power to fix it in the short term.

Giving up on pub/sub isn't an option either, at least not until we do away with Redis. But that won't happen for at least a few months yet.

What options do we have, besides undoing my commit that adds miniredis and depending on redis again?

cc @TykTechnologies/engineering

matiasinsaurralde:
We could continue using Redis and prepare a fix to support pubsub on miniredis, even if it's limited.

mvdan:
The discussion at https://github.com/alicebob/miniredis/issues/1 tells me that this wouldn't be easy. If anyone wants to have a go at it they're welcome to, but I don't think we should hold up new tests like your PR until we have it working.

buger:
What about trying https://github.com/rafaeljusto/redigomock ?

lonelycode:
Most CI tools support redis as a service - it's only for local dev where it's annoying. Instead of miniredis we should just `FLUSHAL` during teardown.

mvdan:
I want to avoid mocking libraries as that would be quite a lot of work to integrate. The advantage of miniredis is that it was a drop-in replacement.

> it's only for local dev where it's annoying

Yep, this is about making our lives easier when running the tests locally. @lonelycode you said Redis was going to be replaced, but you also said something like it would still stick around. Is it always going to be needed for the tests?

> Instead of miniredis we should just FLUSHAL during teardown.

I assume this is to clean up. That doesn't seem to have affected us at all while we used the real redis, though.

buger:
> That doesn't seem to have affected us at all while we used the real redis, though.
This does not matter for Travis, because on each run it have clean state, but matter for local testing.

As an alternative, we may try make `RedisClusterStorageManager` interface, and for the tests write own implementation (which in most cases will just proxy original calls), which mock pubsub internally using channels.

mvdan:
> This does not matter for Travis, because on each run it have clean state, but matter for local testing.

That's what I meant - I fire up Redis in the morning and keep it up. So it should have been affecting me.

> As an alternative, we may try make RedisClusterStorageManager interface, and for the tests write own implementation (which in most cases will just proxy original calls), which mock pubsub internally using channels.

[...]

PR #470, opened 2017-02-03
test: depend on a real redis server again

mvdan:
Remove our use of miniredis. It's lacking pub/sub, which we need for
some upcoming tests.

Don't delete the vendored package yet, as we might use it again after
some tweaks like pub/sub support on top. If that happens, better not
pollute the history twice.

Updates #469.

cc @buger @lonelycode @matiasinsaurralde


buger:
Looks good 👍

PR #473, opened 2017-02-03
Merge tykcommon into tyk

mvdan:
See https://github.com/TykTechnologies/tyk/issues/390.

@lonelycode @buger PTAL

### DIFF
--- a/api.go
+++ b/api.go
@@ -15,7 +15,7 @@ import (
 	"time"
 
 	"github.com/TykTechnologies/logrus"
-	"github.com/TykTechnologies/tykcommon"
+	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/gorilla/context"
 	"github.com/nu7hatch/gouuid"
 	"golang.org/x/crypto/bcrypt"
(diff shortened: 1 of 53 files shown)


mvdan:
Also, I went with the simplest forms of package names, `/log` and `/apidef`. We can hide `log` inside `/internal` and version via `/apidef/v0` later.

buger:
It looks sane to me, but since this is mostly not about tech but about vision, I would like to get @lonelycode approval here. 

Also, I assume rest of repositories which use `tykcommon` should be updated as well?

mvdan:
> It looks sane to me, but since this is mostly not about tech but about vision, I would like to get @lonelycode approval here.

I got his approval on the idea in #390, so this should only be about execution.

> Also, I assume rest of repositories which use tykcommon should be updated as well?

We should do it the sooner the better. But since we're not taking down the other repos yet, we can give it a couple of weeks.

buger:
Ok, let's merge it. But please create Issue to update all tykcommon references (you can find them using https://github.com/search?q=tykcommon&type=Code&utf8=%E2%9C%93)

lonelycode:
@buger Seconded, please raise an issue in each relevant repository, assign a milestone and get it into the v2.4 (or related project) milestone / board backlog.

mvdan:
@lonelycode @buger you can find them referenced in #390. I just opened them, haven't assigned/milestoned them.

PR #474, opened 2017-02-03
Add MaxMind GeoIP test DB to use in the tests

### LINKED ISSUE #352: GeoIP Failure (not recorded): ipAddress passed to Lookup cannot be nil
In some cases when GeoIP is enabled in the gateway, if the IP is nil or not extracted properly, the gateway will throw this useless error.

mvdan:
Obtained from: https://github.com/maxmind/MaxMind-DB/blob/689bbf927360cee2285fd472fc1eb6671f335979/test-data/MaxMind-DB-test-ipv4-24.mmdb

The database is tiny and contains just a few records. That doesn't
matter too much, as we simply want to be able to initialise and use
analytics.GeoIPDB in the tests.

initialiseSystem doesn't return any error, so we have to check for nil
ourselves to be sure that the database was loaded correctly.

In preparation to fix #352 with a test.

Updates #352.


mvdan:
> Coverage increased (+0.1%) to 32.525%

Woo, this thing works!

PR #483, opened 2017-02-07
Drop Go 1.6 build support, make 1.8 the main one

### LINKED ISSUE #405: Support Go 1.8, drop 1.6
1.8 will be out next week.

mvdan:
See https://golang.org/doc/go1.7 for all the things that we can use now.

This will be especially useful since we'll be able to use sub-tests and
sub-benchmarks. Right now we use lots of copy-pasted code in our tests.

Fixes #405.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,9 @@ addons:
 
 matrix:
   include:
-    - go: 1.6.x
     - go: 1.7.x
-      env: LATEST_GO=true # run linters and report coverage
     - go: 1.8rc3
+      env: LATEST_GO=true # run linters and report coverage
 
 services:
   - redis-server
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 Tyk is a lightweight, open source API Gateway and Management Platform enables you to control who accesses your API, when they access it and how they access it. Tyk will
 also record detailed analytics on how your users are interacting with your API and when things go wrong.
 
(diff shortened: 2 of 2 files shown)


mvdan:
I realise 1.8 isn't out yet, but it will be out within a week. This also doesn't force us to use 1.8 to build our release binaries as we will still support 1.7.

buger:
I'm generally ok, but we need to compile list of places which needs to be updated. Like docker images, build scripts, servers and etc.

mvdan:
@buger you mean to update to 1.7 or to 1.8? I don't know what we use right now.

buger:
I meant 1.6, just need to ensure that we do not use it anywhere. 

For example, dev-env already use 1.7, buddy-works builds Tyk with 1.7 as well. Anything else what comes to the mind?

mvdan:
Perhaps some of the docker images? Otherwise not that I know of.

Once 1.8 is released, perhaps we should use that instead of 1.7. After all, we support two versions and the newer one is almost always going to be better for us, as long as we stay within stable Go releases.

buger:
Well, latest Go version had 5 patch releases, so it was quite buggy. We can switch to 1.8 in dev environment, but BuddyWorks should stay on 1.7 until there will be at least few path releases fro 1.8.

mvdan:
I don't mind what policy we take too much, just want to easily know what Go version our binaries and cloud are running. Right now it's not documented or listed anywhere, that I know.

PR #484, opened 2017-02-07
analytics: don't throw errors on invalid IPs

### LINKED ISSUE #352: GeoIP Failure (not recorded): ipAddress passed to Lookup cannot be nil

mvdan:
Empty IPs remain silent as before. Invalid IPs are logged better - now
they are:

	GeoIP Failure (not recorded): invalid IP address "foobar"

insteaf of:

	GeoIP Failure (not recorded): ipAddress passed to Lookup cannot be nil

Add tests too.

Also, these are now warnings instead of errors, since they are often
just noise.

Fixes #352.

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"net"
 	"regexp"
 	"time"
@@ -65,22 +66,17 @@ func (a *AnalyticsRecord) GetGeo(ipStr string) {
 	if !config.AnalyticsConfig.EnableGeoIP {
 		return
 	}
-
 	// Not great, tightly coupled
 	if analytics.GeoIPDB == nil {
 		return
 	}
 
-	// Sometimes it is empty, we can't look up mpty IP addresses
-	if ipStr == "" {
+	record, err := geoIPLookup(ipStr)
+	if err != nil {
+		log.Error("GeoIP Failure (not recorded): ", err)
 		return
 	}
-
-	ip := net.ParseIP(ipStr)
-
-	var record GeoData // Or any appropriate struct
-	if err := analytics.GeoIPDB.Lookup(ip, &record); err != nil {
-		log.Error("GeoIP Failure (not recorded): ", err)
+	if record == nil {
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


mvdan:
@buger @lonelycode PTAL

Another option that comes to mind, instead of demoting from error to warning, is to keep it as an error but see what errors we get now that we'll see what string is being passed. Then we can decide what to do.

The problem is that we'd need to ship the patch in 2.4 to get the extra info, and then actually get rid of all the useless errors in a following release. Since I assume this is not low priority, I've gone ahead and demoted to warning directly.

buger:
I think part of this issue is to understand why we ever get such malformed IP? 

To get IP we use `GetIPFromRequest` from `util_http_helpers.go`. And I may assume that some of the customers have issues in `X-Forwarded-For` header, or we parsing it wrong. Can you check this part too?

And if this is an issue with X-Forwarded-For header format, it may be interesting to include this information to the Sentry.

buger:
Just an idea, maybe this is issue with IPv6 address format?

buger:
For example in case of IPv6 address format ("2001:0db8:85a3:0000:0000:8a2e:0370:7334"), doing strings.Split(remoteIPString, ":") to find the port not the best thing. And instead we should rely on net.IP package for parsing IP addresses.

mvdan:
Yeah, will have to look into it.

mvdan:
@buger PTAL at the two new commits.

buger:
So, it actually fails now at "panic: runtime error: slice bounds out of range"

mvdan:
Ah, was running just my test. Forgot about the rest.

mvdan:
@buger done

analytics.go
```diff
@@ -65,22 +66,17 @@ func (a *AnalyticsRecord) GetGeo(ipStr string) {
...
 	// Not great, tightly coupled
 	if analytics.GeoIPDB == nil {
 		return
 	}
 
-	// Sometimes it is empty, we can't look up mpty IP addresses
-	if ipStr == "" {
+	record, err := geoIPLookup(ipStr)
+	if err != nil {
+		log.Warn("GeoIP Failure (not recorded): ", err)
```
buger: Just out curiosity, can we temporary make it Error, and add "ipStr" to logging info, to see if number errors is changed, and if our assumption about Ipv6 is right. Maybe there are more `ipStr` issues we do not aware about?

Overall looks good.

mvdan: Sure thing.

mvdan: done.

analytics.go
```diff
@@ -65,22 +66,17 @@ func (a *AnalyticsRecord) GetGeo(ipStr string) {
...
 	// Not great, tightly coupled
 	if analytics.GeoIPDB == nil {
 		return
 	}
 
-	// Sometimes it is empty, we can't look up mpty IP addresses
-	if ipStr == "" {
+	record, err := geoIPLookup(ipStr)
+	if err != nil {
+		log.Error("GeoIP Failure (not recorded): ", err)
```
buger: pls add ipStr to error message too

mvdan: It's added in `geoIPLookup` already.

PR #485, opened 2017-02-07
Fix RPC memory leaks

buger:
It is not completely clear why RCP purger creates so many connections.
But seems like using global RPCFuncClientSingleton fix the issue.

I checked it and confirmed that number of opened connections to Sink stays the same. 

@lonelycode  I rebased your changes and already 2.3 compatible. I think it makes sense include into 2.3.2.

While this change definitely fix "too many opened files" issue, I still see that `heap` increase from time to time. @mvdan do you have experience in debugging heap files? I attach both heap and goroutine report.
[goroutine.txt](https://github.com/TykTechnologies/tyk/files/758177/goroutine.txt)
[heap.txt](https://github.com/TykTechnologies/tyk/files/758178/heap.txt)

### DIFF
--- a/main.go
+++ b/main.go
@@ -138,7 +138,7 @@ func setupGlobals() {
 
 		if config.AnalyticsConfig.Type == "rpc" {
 			log.Debug("Using RPC cache purge")
-			purger := RPCPurger{Store: &AnalyticsStore, Address: config.SlaveOptions.ConnectionString}
+			purger := RPCPurger{Store: &AnalyticsStore}
 			purger.Connect()
 			analytics.Clean = &purger
 			go analytics.Clean.PurgeLoop(10 * time.Second)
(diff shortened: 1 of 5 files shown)


mvdan:
@buger how were these reports generated?

buger:
Related links:
https://golang.org/pkg/runtime/pprof/
https://golang.org/pkg/net/http/pprof/ (<- what I used do get this files)
https://blog.golang.org/profiling-go-programs

mvdan:
I've used cpu and mem (heap) profiles plenty of times before, but not goroutine block ones. Will have a look later.

buger:
Worth noticing that this profiles were generated **after** load testing, so theoretically this numbers should have returned to normal.

mvdan:
This could use some rebasing/squashing :)

PR #490, opened 2017-02-08
Vendor updates

mvdan:


mvdan:
@lonelycode @buger we're also lagging behind on lots of other packages by quite a lot. This raises two questions:

* What should our policy be regarding updating `vendor/`? I think now is a good time, when we have a few months to test things before the next minor release.
* Why do we have so many `TykTechnologies/*` and `lonelycode/*` forks? Some have modifications, like `rediscluster`, but most don't. We should minimize forks to simplify staying up to date with upstream.

buger:
> What should our policy be regarding updating vendor/? I think now is a good time, when we have a few months to test things before the next minor release.

Ideally, we should update only to latest bugfix release, and consider major package updates only if there are features we REALLY want. Unlikely anyone of us would like to fix bugs of vendored packages. But unfortunately is it very rare for Go packages to use semver o similar, which makes hard updating them.

buger:
I will merge this PR, but @lonelycode feel free to answer here.

mvdan:
> Ideally, we should update only to latest bugfix release, and consider major package updates only if there are features we REALLY want. Unlikely anyone of us would like to fix bugs of vendored packages. 

This doesn't apply to `golang.org/x/...` - those are rolling and maintained by the Go team.

> But unfortunately is it very rare for Go packages to use semver or similar, which makes hard updating them.

It's actually fairly common, and should become more so since it will be used by the upcoming official dep tool.

lonelycode:
My policy is: if it ain't broke don't fix it, so keeping them as is and pinned is just fine.

If a vendor is updated, then it should be done one at a time with a full regression test.

We can get rid of the forks if there are no mods, the modified versions we should keep.

Some of the forks are because the original author at some point f*cked us with a breaking change (a la logrus and osin).

PR #493, opened 2017-02-09
test: remove a bunch of unnecessary mux work

mvdan:
These set up muxers and load endpoints into them, but then just throw
them away. They are set up as http.DefaultServeMux, but that's never
used.

@lonelycode @buger if either of you know why this was written like this, please let me know.

Also weird that removing the `loadApps(..., discardMuxer)` lines makes the tests fail and the gateway panic.

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -50,17 +50,7 @@ const apiTestDef = `{
 
 func makeSampleAPI(t *testing.T) *APISpec {
 	spec := createSpecTest(t, apiTestDef)
-
-	specs := []*APISpec{spec}
-	newMuxes := mux.NewRouter()
-	loadAPIEndpoints(newMuxes)
-	loadApps(specs, newMuxes)
-
-	newHttmMuxer := http.NewServeMux()
-
-	newHttmMuxer.Handle("/", newMuxes)
-
-	http.DefaultServeMux = newHttmMuxer
+	loadApps([]*APISpec{spec}, discardMuxer)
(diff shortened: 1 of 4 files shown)


buger:
Will this actually change the logic a bit, because before it was using clean state (router) each time, and now use global object, which can be potentially polluted during the tests. Are you sure you want introduce a new global object, or I'm missing something?

Also seems it does not call `loadAPIEndpoints` at all, why?

mvdan:
> Will this actually change the logic a bit, because before it was using clean state (router) each time, and now use global object, which can be potentially polluted during the tests. Are you sure you want introduce a new global object, or I'm missing something?

Before, a local router was used that was discarded. Now, a global one is used and is discarded. I did it this way because it's more obvious and cheaper. Either way, the router is pointless.

> Also seems it does not call loadAPIEndpoints at all, why?

Both `loadAPIEndpoints` and `loadApps` seem unnecessary, which is why I'm asking for input as to what the logic behind all this was.

lonelycode:
`loadAPIEndpoints` loads up endpoints for the gateway API, and so the routes are not needed for API Definiiton / proxied API tests.

`loadApps` loads up the API definition into the muxers, if they were in the tests it was to try and adhere to the logic of the application and the order it does things.

If the tests still pass then I can't see a problem with it.

mvdan:
@buger rebased, PTAL

PR #507, opened 2017-02-13
Refactor url rewrite tests into sub-tests

mvdan:


buger:
Nice

mvdan:
Yep :) This will take forever and it's tedious work, though. But definitely worth it for our own sanity.

PR #508, opened 2017-02-13
Organize filenames better

mvdan:


mvdan:
After this, the only `foo_test.go` files without `foo.go` are:

```
 $ for f in *_test.go; do test -f ${f//_test/} || echo $f; done
extended_method_versioning_test.go
gateway_test.go
middleware_example_test.go
multiauth_test.go
```

Only the first stands out, but I don't know where to put it. @lonelycode should we move it anywhere, or leave it be?

lonelycode:
`extended_method_versioning_test.go` can just be `acl_lists_test.go` because those tests essentially check white list, black list and ignore list. Because we used to have a different structure that handled these, they are part of the "extended" set, but we can actually just get rid of the old ones, they are truly legacy and need decommissioning.

analytics_test.go
```diff
@@ -21,3 +21,57 @@ func TestGeoIPLookup(t *testing.T) {
...
+
+	recordUUID1.NormalisePath()
+	recordUUID2.NormalisePath()
+	recordUUID3.NormalisePath()
+	recordUUID4.NormalisePath()
+	recordID1.NormalisePath()
+	recordCust.NormalisePath()
+
+	if recordUUID1.Path != "/{uuid}/search" {
+		t.Error("Path not altered, is:")
```
buger: Why you move group all to multiline t.Error call to single one?

mvdan: Sorry, what? If you're asking me why I'm using multiple `Error` calls, this is what the code did. I'm not touching any of the code, just moving it.

This particular anti-pattern is everywhere in both prod and test code, so I'd like to tackle it separately.

mvdan:
> Because we used to have a different structure that handled these, they are part of the "extended" set, but we can actually just get rid of the old ones, they are truly legacy and need decommissioning.

Get rid of what, exactly?

analytics_test.go
buger: I understand that it was there before you, just thought it make sense fix it right away here. If you prefer separate PR it is ok.

mvdan: Yeah, since there are hundreds of these. Fixing a few won't do much.

lonelycode:
> Get rid of what, exactly?

[These](https://github.com/TykTechnologies/tyk/blob/master/apidef/api_definitions.go#L160) 

The same whitelist/blacklist functionality, but it wasn;t quite as feature rich, had to change the data structure to make it work. But that was a long time ago and is no longer supported by the dashboard.

PR #511, opened 2017-02-14
Ensure that RPC loader returns an object, not nil. 

### LINKED ISSUE #510: Ensure that RPC loader returns an object, not nil
Bug

If the RPC API Definition loader has a disconnect event, it can return a nil object which fails up the chain and causes a runtime panic

The RPC loader should return an empty object

Not easily reproducible, it requires longitudinal monitoring of a live process against MDCB.

Hybrid / RPC mode of all gateways

lonelycode:
Fixes #510

### DIFF
--- a/api_definition_manager.go
+++ b/api_definition_manager.go
@@ -540,13 +540,13 @@ func (a *APIDefinitionLoader) processRPCDefinitions(apiCollection string) *[]*AP
 
 	if jErr1 != nil {
 		log.Error("Failed decode: ", jErr1)
-		return nil
+		return &APISpecs
 	}
 
 	jErr2 := json.Unmarshal([]byte(apiCollection), &StringDefs)
 	if jErr2 != nil {
 		log.Error("Failed decode: ", jErr2)
-		return nil
+		return &APISpecs
 	}
 
 	for i, thisAppConfig := range APIDefinitions {
--- a/api_definition_manager.go
+++ b/api_definition_manager.go
@@ -540,13 +540,13 @@ func (a *APIDefinitionLoader) processRPCDefinitions(apiCollection string) *[]*AP
 
 	if jErr1 != nil {
 		log.Error("Failed decode: ", jErr1)
-		return &APISpecs
+		return nil
 	}
 
 	jErr2 := json.Unmarshal([]byte(apiCollection), &StringDefs)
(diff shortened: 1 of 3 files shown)


mvdan:
Could the commit message or the code mention why this is necessary?

Also, I assume this also affects master? Or it doesn't because we now return `[]*ApiSpec` instead of `*[]*ApiSpec`?

mvdan:
We generally don't use PRs in milestones, we use the issues they fix/refer, so removing.

buger:
So basically, when it experience error it fallback to last cached value, which looks sane. 👍

buger:
@matiasinsaurralde updated comment

mvdan:
> So basically, when it experience error it fallback to last cached value, which looks sane.

Where do you get that from? This just returns an empty list.

buger:
Oops, did not expanded context till that point and was confused a bit by variable name, we have similar global name... :grimacing:

buger:
Actually, I do not think that this fix may work. I will try to add basic test.

But basically, returning `nil` is expected and handled everywhere, like here:
```go
specs := getAPISpecs()
if specs != nil {
    loadApps(specs, defaultRouter)
    getPolicies()
}
```

mvdan:
@buger this is why I'm asking for more details as to why this fixes it.

In any case, we will be doubly safe in 2.4 since it's no longer a pointer in master.

buger:
I wrote a test and found the reason for a `panic`. It fails on this, and following lines:

```
	log.WithFields(logrus.Fields{
		"prefix": "main",
	}).Printf("Detected %v APIs", len(*APISpecs))
```
https://github.com/TykTechnologies/tyk/blob/f82315b8d49a5e4902837af7a443427164a1b49e/main.go#L244:1-L245:1

Basically all `getAPISpecs()` calls handle if it returns `nil` but `getAPISpecs()` itself internally does not handle such situation. Will push a commit soon.

buger:
Added proper fix and tests. 

Since *[] was removed in master, this bug should not affect it, but I still want to merge tests with few changes.

mvdan:
Having the tests in master sounds like a good idea.

mvdan:
Your changes look fine, but I would squash the two commits. Your commit is partly just reverting all of martin's changes.

PR #514, opened 2017-02-14
Make golint happier with consistent receiver names

mvdan:
While at it, also simplify many IsEnabledForSpec funcs.


buger:
Should we included golint to CI as well?

mvdan:
I don't think so. Golint still has hundreds of warnings, and a lot of them are not critical - missing docs, names that could be improved, etc.

PR #515, opened 2017-02-14
Test cleanup, flush redis, remove randSeq

mvdan:


buger:
Since you are flushing Redis, is there is a chance to tell the "Tests" use different database number, so it will not override any local data?

mvdan:
You mean flush and use a specific database, instead of flushing everything?

buger:
I'm not sure if it is currently possible, without adding new configuration option which specifies Redis database. In the case of `redisgo` it will look like: "c.Do("SELECT", db);", where `db` is int. By default Redis use `0` database.

mvdan:
We're not using a database that I could see. We could do either of:

* Use a database name for both production and the tests
* Keep the default database for production, but use a specific one for the tests

Or keep things as they are. The only inconvenient now is that it might stomp data when running the tests.

cc @lonelycode

buger:
Actually looks like it is possible, and if in `TestMain` you will set `config.Storage.Database` and `config.CacheStorage.Database` to 1, and will use `c.Do("SELECT", db)` when flushing, tests will use separate Redis database.

mvdan:
SGTM, will update the PR.

mvdan:
@buger done and rebased, PTAL

PR #518, opened 2017-02-15
test: use a local version of httpbin

mvdan:
We no longer depend on an internet connection nor on these services being up for a lot of our tests.

Test time down from ~5s to ~1.5s.

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -24,9 +24,65 @@ func init() {
 	runningTests = true
 }
 
-var discardMuxer = mux.NewRouter()
+var (
+	// to register to, but never used
+	discardMuxer = mux.NewRouter()
+	// we need a static port so that the urls can be used in static
+	// test data and the requests aren't randomized for checksums
+	// port 16500 should be obscure and unused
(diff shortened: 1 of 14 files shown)


mvdan:
There are a few more external requests, but they're not straightforward to change. Will do them in separate PRs.

gateway_test.go
```diff
@@ -24,9 +24,66 @@ func init() {
...
 }
 
+// to register to, but never used
 var discardMuxer = mux.NewRouter()
 
+const (
+	// we need a static port so that the urls can be used in static
+	// test data and the requests aren't randomized for checksums
+	// port 16500 should be obscure and unused
+	testHttpListen = "127.0.0.1:16500"
```
buger: If you get rid of fixed port, you can use ":0", so OS will assign random free port, and getting it back using `server.Addr`. The same way you can get rid of global http server.

It can be a separate PR later if you prefer. Or maybe we can leave it as it is.

mvdan: Read the comment :)

I agree that that would be better, but it would need refactoring so that the tests won't fail. Right now some depend on the url being static.

Issue #520, opened 2017-02-15
Tyk attempts to open thousands connections to Redis machine on failure


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
If for some reason there are problems with connections to Redis (for example master failed), Redis attempts to make thousands of connections to Redis server.

**What is the expected behavior?**
There should not be a significant increase of connection attempts; it should gracefully retry requests to Redis using some interval. 

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
We have no clear way to replicate the issue, but worth trying stopping Redis server after the connection was initialized, and making few requests.

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
All

This behavior can be related to the following code https://github.com/TykTechnologies/tyk/blob/master/redis_cluster_handler.go#L136
If key is not found it tries to connect in loop.

mvdan:
I looked at this last week for a somewhat related issue - a race. Calls to `Connect` are not controlled at all, so the code is racy.

I did not post a fix because the obvious fix of adding a mutex of some sort could be tricky - it could lead to `GetKey` blocking all the time or for really long amounts of time.

A change that would fix this bug but not the race would be simpler - we would have to limit the retries to a limited amount (three, for example) and have it wait between attempts. Note that it would still be really racy, but it can't be worse than what it is now. In fact the chances of races would probably be lower since there would be fewer reconnection attempts.

@buger @lonelycode thoughts on this fix to backport to 2.3? The race fix likely doesn't need backporting.

buger:
Race itself not, but actually fixing related endless connection loop should be included into 2.3.

mvdan:
Okay, will have a PR soon.

mvdan:
I've tried to reproduce this with a test using a tcp server, but no luck. Basically, the singletons are only ever set to non-nil pointers, so the logic that fires the reconnects (`GetRelevantClusterReference(r.IsCache) == nil`) will never be true as soon as we've set up the cluster for the first time.

We set either of the singleton pointers to this:

```
cluster := rediscluster.NewRedisCluster(seed_redii, poolConf, false)
return &cluster
```

`rediscluster` never errors when creating the cluster, as it does the actual connections in the background.

So I don't think our reconnect code is the cause for all those connections. It could be `rediscluster` internally trying to open more connections, but then it would be a bug in its reconnect code, not ours.

@lonelycode thoughts?

[...]

PR #522, opened 2017-02-15
Redis reconnect cleanup

mvdan:
In preparation for #520.

### DIFF
--- a/redis_cluster_handler.go
+++ b/redis_cluster_handler.go
@@ -490,7 +490,7 @@ func (r *RedisClusterStorageManager) GetAndDeleteSet(keyName string) []interface
 	redVal, err := redis.Values(GetRelevantClusterReference(r.IsCache).DoTransaction([]rediscluster.ClusterTransaction{lrange, delCmd}))
 	if err != nil {
 		log.Error("Multi command failed: ", err)
-		r.Connect()
+		return []interface{}{}
 	}
 
 	log.Debug("Analytics returned: ", redVal)
--- a/redis_cluster_handler.go
+++ b/redis_cluster_handler.go
@@ -128,13 +128,25 @@ func (r *RedisClusterStorageManager) cleanKey(keyName string) string {
 	return setKeyName
 }
 
+func (r *RedisClusterStorageManager) ensureConnection() {
+	if GetRelevantClusterReference(r.IsCache) != nil {
+		// already connected
+		return
+	}
+	log.Info("Connection dropped, reconnecting...")
+	for {
+		r.Connect()
+		if GetRelevantClusterReference(r.IsCache) != nil {
+			// reconnection worked
+			return
+		}
+		log.Info("Reconnecting again...")
+	}
+}
+
 // GetKey will retrieve a key from the database
 func (r *RedisClusterStorageManager) GetKey(keyName string) (string, error) {
-	if GetRelevantClusterReference(r.IsCache) == nil {
-		log.Info("Connection dropped, connecting..")
-		r.Connect()
-		return r.GetKey(keyName)
-	}
+	r.ensureConnection()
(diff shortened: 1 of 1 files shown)


mvdan:
Also note that this removes tail calls, which can be nothing but trouble.

mvdan:
Also a nice coverage bump as this repeated code isn't being tested (I can't even make it run in a test).

redis_cluster_handler.go
```diff
@@ -128,13 +128,25 @@ func (r *RedisClusterStorageManager) cleanKey(keyName string) string {
...
-func (r *RedisClusterStorageManager) GetKey(keyName string) (string, error) {
-	if GetRelevantClusterReference(r.IsCache) == nil {
-		log.Info("Connection dropped, connecting..")
+func (r *RedisClusterStorageManager) ensureConnection() {
+	if GetRelevantClusterReference(r.IsCache) != nil {
+		// already connected
+		return
+	}
+	log.Info("Connection dropped, reconnecting...")
+	for {
```
buger: Why is it inside `for` loop?
`r.Connect` can never return `false` and `GetRelevantClusterReference` seems like never can be `nil` if `Connect` was called.

mvdan: These two things are what I was talking about on slack and in the original issue.

This PR is not about changing or fixing the logic, it's about deduplicating it. I simply rewrote the recursive call into a for loop. Which makes the broken logic more obvious too.

buger:
I see, so just to summarize:

All the storage handlers share the same Redis pool and use singletons. 
Once any of `RedisClusterStorageManager` instances makes the first `Connect()` call, it initializes singletons (and actually never returns `false`).
There are no places in code where singletons can be set to `nil`, which means all the reconnect logic does nothing and can be safely removed. To be precise, reconnecting still handled but inside `redigocluster` and `redigo`.

If we talking about #520 bug itself, I do not see the way how our code can cause them, so it is probably inside `redigocluster`. <ake sense to check if we need to update our fork.

@mvdan anything missing?

buger:
Worth noticing that I tried simulating Redis failure, like stopping it after it was initialized, single and multiple nodes: `GetKey` always returned proper error, without looping.

mvdan:
> All the storage handlers share the same Redis pool and use singletons.

I haven't looked at this deeply, but you pointed me towards the reconnect logic - and that only touches the singletons.

> There are no places in code where singletons can be set to nil, which means all the reconnect logic does nothing and can be safely removed. To be precise, reconnecting still handled but inside redigocluster and redigo.

Yes.

> If we talking about #520 bug itself, I do not see the way how our code can cause them, so it is probably inside redigocluster. Make sense to check if we need to update our fork.

Could be somewhere else in our redis-handling code, but likely in redigocluster, yes. Especially since we're behind on upstream changes and @lonelycode made changes that noone else is using besides tyk. So it's likely that it has obscure bugs like this one.

> Worth noticing that I tried simulating Redis failure, like stopping it after it was initialized, single and multiple nodes: GetKey always returned proper error, without looping.

I also failed to reproduce the looping. It could be that the looping is due to a race condition, though. Have you tried running your tests with `-race`? Although it's likely that it will panic due to a data race before it's able to reproduce anything.

mvdan:
@buger is there anything else I should do or comment on?

mvdan:
reping @buger this is for 2.3.3

buger:
If we are talking in the context of 2.3.3, it seems like you did not change any functionality, just refactoring. I do not see how it can fix Redis connection count, or I'm missing something?

mvdan:
Note that we haven't found the cause of the reconnect stuff, so there can't really be a fix for it.

@lonelycode mentioned getting rid of the tail calls for 2.3.3, which is what this PR is doing. It's a good first step until we can reproduce the bug.

PR #527, opened 2017-02-16
vendor: update google.golang.org/grpc to 1.0.5

mvdan:
We were using a slightly older version. These do have stable versions,
so pin the latest.


buger:
What was the previous version in terms of semver?

mvdan:
None; we were using HEAD as of mid november, I assume that's when @matiasinsaurralde's coprocess stuff went in.

The bump in time is of around 15 days.

buger:
Yes, but what we have commit id, so it is possible to get version which was vendored. I wanted to check the changelog.

mvdan:
Sure, you can do that. I'm not sure what your question is.

buger:
Ok, let me rephrase it. How do you ensure that updating this package is safe?

mvdan:
I can't - this is why I'm doing this PR now and not a week before the 2.4 release.

In any case, using a stable version is better than using a random HEAD.

buger:
Here is changelog https://github.com/grpc/grpc-go/compare/b13ef794997191ec801db90a42c50034df91850a...708a7f9f3283aa2d4f6132d287d78683babe55c8

buger:
Seems like mostly fixes and documentation updates. Our version is between 1.0.4 and 1.0.5.

PR #529, opened 2017-02-16
Map usage cleanup

mvdan:


api_definition.go
```diff
@@ -205,8 +205,8 @@ func (a *APIDefinitionLoader) MakeSpec(appConfig *apidef.APIDefinition) *APISpec
 		}
 	}
 
-	newAppSpec.RxPaths = make(map[string][]URLSpec)
-	newAppSpec.WhiteListEnabled = make(map[string]bool)
+	newAppSpec.RxPaths = make(map[string][]URLSpec, len(appConfig.VersionData.Versions))
```
buger: Why you specify map length here? From what I understand it may be used only for performance reasons, but I doubt it is the case here.

mvdan: Yeah. You may call it premature optimization, but it can't hurt. The default value is 0.

buger: > You may call it premature optimization
I do :)

It is non-default behavior and usually not used for maps. That's why I was wondering, and someone else who will work on this code can too.

If you really want to keep them here, I'm ok, but it just attracts attention.

mvdan: It's common practice to do it if the amount of inserts on a slice or map is known at compile-time. You can argue that if it's not a hot path it's not worth the time, but in this case the time is already invested :)

Issue #385, opened 2017-01-23
Get duplicated code under control


mvdan:
Can be detected via https://github.com/mibk/dupl. Using 200 tokens already yields half a dozen results, so the situation is far from good.

This issue is to track progress.

mvdan:
Ignoring test files is a good start. For example:

```
 $ dupl -t 100 $(ls *.go | grep -v _test)
found 2 clones:
  storage_handlers.go:709,730
  storage_handlers.go:732,753
found 2 clones:
  command_mode.go:37,103
  swagger.go:142,208
found 2 clones:
  middleware_url_rewrite.go:75,98
  middleware_url_rewrite.go:119,142
found 3 clones:
  main.go:461,484
  main.go:515,538
  main.go:543,566
found 2 clones:
  oauth_manager.go:641,661
  oauth_manager.go:743,763
found 2 clones:
  rpc_storage_handler.go:571,595
  rpc_storage_handler.go:598,622
found 2 clones:
  middleware_rate_limiting.go:40,60
  middleware_rate_limiting.go:62,82
found 2 clones:
  middleware_jwt.go:427,449
  middleware_openid.go:232,254

Found total 8 clone groups.
```

mvdan:
The sample above is down from 8 to 5. One of them is the redis code that we will likely refactor. The rest aren't particularly bad, as the code isn't easily deduplicated.

I'm going to close this, since we no longer have large bits of very similar code. Smaller refactors can be done over time.

PR #530, opened 2017-02-16
Deduplicate js middleware code from main.go

mvdan:
The four blocks of code are nearly identical, except the folder name,
the resulting slice/var and requiresSession. Use a table instead.

Updates #385.


buger:
cc @matiasinsaurralde

matiasinsaurralde:
Makes sense!

PR #534, opened 2017-02-17
Set version to v2.4-HEAD

mvdan:
master will become 2.4 at some point, so it should be 2.4* not 2.3*.
Don't call it just 2.4.0 as it's not yet finished. Don't call it alpha
or beta either as we haven't released anything.

Ideally, this would include the commit hash of the build. But this is
currently not possible to do with a simple `go get`, so just call it
HEAD.

This removes the confusion of Tyk reporting "version 2.3.2" on master,
which isn't true.

### DIFF
--- a/version.go
+++ b/version.go
@@ -1,3 +1,3 @@
 package main
 
-var VERSION = "v2.3.2"
+var VERSION = "v2.3.99"


buger:
I think currently it will break our deploy regexps which parse version number, so deployment to staging will be broken. As you mentions HEAD not really the right name, without having commit sha. 

Setting it to 2.4.0 probably should be good at this stage.

mvdan:
Ah, right, forgot about the regexes.

> Setting it to 2.4.0 probably should be good at this stage.

I don't like it for the same reason that I don't like 2.3.2 - once 2.4.0 is actually released, there will be no way to discern people running master older than 2.4.0 from people actually running 2.4.0.

Do we have to use numbers in the regexes? What about `2.4.HEAD`?

buger:
Right now only numbers. You can try to modify scripts here https://github.com/TykTechnologies/tyk/tree/master/utils and ensure that it will work with your scheme.

mvdan:
@buger are those the only ones? What about regexes in other repos, could they break with this version here?

This could be useful in our other open-source repos like pump, where people can just build master.

buger:
I think other repositories should not be affected, but we have lot of dependencies, can't be 100% sure. Also not sure about external stuff like docker hub and package cloud.

mvdan:
Okay, let's wait for @lonelycode then.

mvdan:
Another, less intrusive option would be to name it `2.3.99`. But that's not as descriptive and warning-y as `2.4.HEAD`.

lonelycode:
Please don't modify the scripts :-) there are counterpoints in the CD system so one impacts the other. (And I recently had to re-fix them)

mvdan:
What about `2.3.99` then? I can't think of a better number.

lonelycode:
That's a beautiful number :-)

mvdan:
Done.

PR #535, opened 2017-02-17
WIP: policy: use api_name and api_id tags

### LINKED ISSUE #192: Wrong field apiid in /api/portal/policies/ result
When getting all available policies using GET /api/portal/policies/ (without id), the response contains an `apiid` (in access_rights.'id').

When getting the policy details using GET /api/portal/policies/'id', the response contains an `api_id` (in access_rights.'id').

(Notice the difference in the underscore)

I would expect that both responses contain the same field(name)s

mvdan:
These are what the rest of the API objects use. Be consistent.

Fixes #192.

Note that I don't know if this change will break anything. The only place I can find after a quick search is https://github.com/TykTechnologies/tyk-analytics/blob/4521ce0d6c81a7d835f68ab50edbabc94df76631/ApiAdminMethods.go#L204.

There's also https://github.com/TykTechnologies/tyk-analytics/blob/7440dfdb619d70589f609e18ad9ec52d178bf358/ModelKeys.go#L35, but that's bson so should be unaffected.

This is the best search query I could come up with, since GitHub doesn't allow searching for special characters like `"` or `:`:

https://github.com/search?utf8=%E2%9C%93&q=org%3ATykTechnologies+language%3Ago+string+json+apiid&type=Code&ref=searchresults

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -33,8 +33,8 @@ type Policy struct {
 }
 
 type DBAccessDefinition struct {
-	APIName     string       `json:"apiname"`
-	APIID       string       `json:"apiid"`
+	APIName     string       `json:"api_name"`
+	APIID       string       `json:"api_id"`
 	Versions    []string     `json:"versions"`
 	AllowedURLs []AccessSpec `bson:"allowed_urls"  json:"allowed_urls"` // mapped string MUST be a valid regex
 }


buger:
Well, chances that our customers use this fields are too high. This is a breaking change, and can be introduced only in major versions, do not think 2.4 is major enough. 

Usually changes like this made by adding both fields to JSON payload, with deprecation notice of old key names.

mvdan:
Good point, that still fixes the issue and won't break anything.

We should have a list of things to fix once we break the API, perhaps in 3.0 or once we define apidef v1. Perhaps TODOs, like `TODO(apidef): ...`?

lonelycode:
As @buger said, this can only be merged in a major release, well aware of the issue - it causes *no end* of issues with the api import and export functionality as well as inter-connectivity between the dashboard and the gateway. This would have really wide ranging (subtle) problems and need to be approached holistically.

buger:
So, what are we going to do with this PR?

lonelycode:
Too many external side effects - will very seriously break user installs and other functionality. This needs Togo against a major version.

mvdan:
I want to update the patch so that it keeps the two keys in place, to not break backwards compat while fixing the issue. Is that not an option?

mvdan:
Yeah, that's not possible in an easy way.

Since this is not high priority, I'm dropping it and adding it to the v3 to-do list.

PR #538, opened 2017-02-17
api: disallow PUT /tyk/apis/

### LINKED ISSUE #157: [PUT] tyk/apis/ creates `.json`
I don't think it should create a `.json` file if you PUT without providing an api-id to `tyk/apis/{api-id}`. If I look in `apps/` I see just a `.json` file (no filename, just file-extension).

mvdan:
PUT without an argument doesn't make sense. Even worse, it can lead to
confusing outcomes like the creation of apps/.json.

Add tests too.

Fixes #157.

### DIFF
--- a/api.go
+++ b/api.go
@@ -781,7 +781,6 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 	var responseMessage []byte
 	var code int
 
-	log.Debug(r.Method)
 	switch r.Method {
 	case "GET":
 		if APIID != "" {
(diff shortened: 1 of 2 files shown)


mvdan:
Rolled back the POST changes as we have a test for it, and even though it's a bit silly some users might depend on it.

buger:
Can you give link to POST changes? Just curious what it does.

mvdan:
GitHub loses old commits in a PR, so I don't think I can retrieve them easily.

In any case, it just forbade `POST /foo` just like `PUT /`. Both make little sense, as `POST` is for something new and `PUT` is for something that already exists. But the former is more widely used and we had a test for `/foo`, so I preferred to be careful and keep it that way in case any users depend on this edge case being supported.

ilijabojanovic:
Verified @buger approved for merge

PR #544, opened 2017-02-20
Do all hot reloads with a queue and a worker

### LINKED ISSUE #506: Handle API reloads in a single goroutine
See https://github.com/TykTechnologies/tyk/issues/496#issuecomment-279342688.

Our current handling of hot reloads (API reloads) is racy because it concerns multiple goroutines and a global bool that isn't protected by any sort of lock.

I propose that a single goroutine is responsible for it, receiving jobs (e.g. reload requests) via a channel.

mvdan:
The former method was racey and could lead to multiple reloads happening
simultaneously.

Fixes #506.

### DIFF
--- a/api.go
+++ b/api.go
@@ -566,7 +566,7 @@ func handleURLReload() ([]byte, int) {
 	var responseMessage []byte
 	var err error
 
-	ReloadURLStructure()
+	ReloadURLStructure(nil)
 
 	code := 200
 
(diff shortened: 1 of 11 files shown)


mvdan:
Okay, fuck, forgot about the coprocess tests.

mvdan:
@buger some tests were timing out, so I've set the reload interval to 0 (instead of 10s) in the tests.

mvdan:
Hanging again. Investigating...

mvdan:
@buger green and ready, PTAL

PR #545, opened 2017-02-20
A bunch of test fixes to fight redis being flakey

mvdan:
@buger PTAL, this is badly needed for my tests

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -588,7 +588,7 @@ func TestGroupResetHandler(t *testing.T) {
 	cacheStore.Connect()
 
 	go func() {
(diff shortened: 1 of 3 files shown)


api_test.go
```diff
@@ -599,6 +599,11 @@ func TestGroupResetHandler(t *testing.T) {
 				signalChan <- false
 			}
 		})
+		if err != nil {
+			t.Log(err)
+			t.Fail()
```
buger: Why not `t.Fail(err)`?

mvdan: You mean `Fatal`, which can't be run from a separate goroutine. I need to log the error, mark the test as failed, and make the main goroutine stop, which is what this does.

PR #550, opened 2017-02-21
Remove useless code from redis pub/sub loop

mvdan:
Calling Connect() multiple times is useless, as it doesn't actually
reconnect. It is handled under the hood by rediscluster.

We also call StartPubSubHandler twice per loop iteration. It's not
particularly harmful, but it's worrying that we run it here without
checking for an error.

### DIFF
--- a/redis_signals.go
+++ b/redis_signals.go
@@ -31,9 +31,6 @@ func StartPubSubLoop() {
 			log.WithFields(logrus.Fields{
 				"prefix": "pub-sub",
 			}).Warning("Reconnecting")
-
-			CacheStore.Connect()
-			CacheStore.StartPubSubHandler(RedisPubSubChannel, HandleRedisEvent)
 		}
 
 	}


mvdan:
@lonelycode @buger I was unable to reproduce an escalating number of redis connections. I wrote a test doing one of each:

* Set up a DRL and send notifications every 50ms, verifying that they get sent and received
* Set up a health check and poll it every 50ms, verifying the returned error

Then, I would monitor the amount of connections with `redis-cli client list` under the following scenarios:

* redis works normally
* redis is restarted a bunch of times
* redis is stopped for a long time and brought back up

Under none of those cases did the number of tyk redis connections go above 2-3.

I say we backport this to 2.3 as a potential fix (we can't do much else unless we're able to reproduce or get a crash log) and also say that all of this will get much better soon as we'll swap out the redis driver for something much saner and more stable (for 2.4, maybe?).

lonelycode:
Fantastic news - the del was he most likely culprit for the issue we saw, so this is definitely a fix worth including in 2.3

PR #555, opened 2017-02-22
Remove useless code from redis pub/sub loop (backport)

mvdan:
Calling Connect() multiple times is useless, as it doesn't actually
reconnect. It is handled under the hood by rediscluster.

We also call StartPubSubHandler twice per loop iteration. It's not
particularly harmful, but it's worrying that we run it here without
checking for an error.

### DIFF
--- a/redis_signals.go
+++ b/redis_signals.go
@@ -29,9 +29,6 @@ func StartPubSubLoop() {
 			log.WithFields(logrus.Fields{
 				"prefix": "pub-sub",
 			}).Warning("Reconnecting")
-
-			CacheStore.Connect()
-			CacheStore.StartPubSubHandler(RedisPubSubChannel, HandleRedisMsg)
 		}
 
 	}


buger:
Linking original PR https://github.com/TykTechnologies/tyk/pull/550

mvdan:
Ah, yes, forgot to link.

buger:
Actually when backporting, and you have no new functionality, you can skip description and just link to original PR.

buger:
Looks like that Github webhook hanged, because Travis build actually passed. Merging.

mvdan:
> Looks like that Github webhook hanged

What do you mean?

> Actually when backporting, and you have no new functionality, you can skip description and just link to original PR.

If you mean the PR body, not the commit message - sure, I'll do that from now on.

buger:
> What do you mean? 
Github UI showed that tests is still running, but Travis was green
> If you mean the PR body
right

PR #557, opened 2017-02-23
test: make api method tests keep /1 in place

mvdan:
Lots of tests create the api /1, and some depend on it - like the reload
tests, since a reload won't happen if there are no apis. Make this test
leave things as it found them, by not having the DELETE be the last
command.


mvdan:
@buger green again :)

PR #558, opened 2017-02-23
Don't check for nil and empty slices

mvdan:
If a slice is nil, its len() is already 0. Checking both is redundant.

This might have been useful back when these were *[]Foo, but we got rid
of them.

### DIFF
--- a/main.go
+++ b/main.go
@@ -643,14 +643,6 @@ func doReload() {
 
 	// load the specs
 	specs := getAPISpecs()
-
-	if specs == nil {
-		log.WithFields(logrus.Fields{
-			"prefix": "main",
-		}).Warning("No API Definitions found (nil), not reloading")
-		return
-	}
-
 	if len(specs) == 0 {
 		log.WithFields(logrus.Fields{
 			"prefix": "main",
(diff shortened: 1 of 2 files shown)


buger:
what about https://github.com/mvdan/tyk/blob/6d884c4ed4ebf48148bd675e3e2dc76950dcd6b1/main.go#L1306
https://github.com/mvdan/tyk/blob/6d884c4ed4ebf48148bd675e3e2dc76950dcd6b1/main.go#L1377
?

mvdan:
Those check that the slice is nil. This is unrelated.

You could argue that they should check if the slice is empty, but I have't looked at when `getApiSpecs` returns an empty and nil slice versus an empty non-nil slice.

(to clarify, empty and nil is `var a []foo` or `a := []foo{}`, empty and non-nil is `a := make([]foo, 0)`)

buger:
Good point. It can send nil in case of error, and non-nil but zero length in case of expected empty api specs. We should cover it with specs...

mvdan:
@buger really, our code should not make distinctions. For example, back in the day the compiler wasn't very clever and `[]foo{}` was non-nil - now it is.

In the future, it might get even cleverer and avoid more allocations. If our funcs have to differentiate between a nil response and an empty response, it should be with `([]foo, error)` as return values, not playing with nil vs non-nil.

On top of being more robust, it's also more readable and understandable.

buger:
Yep, agree about returning error.

PR #561, opened 2017-02-23
Rework hot reload worker to use a ticker, mock it in the tests

mvdan:
@buger as promised


mvdan:
I somehow fucked 1.7, need to poke around a bit.

mvdan:
@buger not sure what the 1.7 failure was, but I've rebased and fired it multiple times and can't reproduce the flake.

PR #564, opened 2017-02-24
Better support for cpu and mem profiles

mvdan:
* Use different file handles, so both can be used at the same time
* Check for create errors


buger:
^ have conflicts

mvdan:
Approved, rebased and green - self-merging.

PR #568, opened 2017-02-27
Fix TestQuota with -count N

mvdan:


gateway_test.go
```diff
@@ -710,34 +703,27 @@ func TestQuota(t *testing.T) {
...
 		t.Error("Initial request failed with non-200 code: \n", recorder.Code, " Header:", recorder.HeaderMap)
-
 	}
 
 	secondRecorder := httptest.NewRecorder()
 	chain.ServeHTTP(secondRecorder, req)
 	thirdRecorder := httptest.NewRecorder()
 	chain.ServeHTTP(thirdRecorder, req)
 
-	if thirdRecorder.Code == 200 {
```
buger: Why this test was removed?

mvdan: It's checked against `403` just below. This is redundant.

PR #579, opened 2017-03-02
Remove go:generate msgp lines

mvdan:
These are only useful if we actually run them and commit the files.
Nobody did, as we have no *_gen.go files.

Remove them. If they are to be re-added, the generated files should be
too.

### DIFF
--- a/coprocess.go
+++ b/coprocess.go
@@ -1,8 +1,5 @@
 // +build coprocess
 
-//go:generate msgp
-//msgp:ignore CoProcessor CoProcessMiddleware CoProcessMiddlewareConfig TykMiddleware
-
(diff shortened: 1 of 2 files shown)


mvdan:
Worth noting that this will likely conflict with https://github.com/TykTechnologies/tyk/pull/580, so better merge that first and I'll rebase this.

lonelycode:
@matiasinsaurralde Can you let us know the significance of these lines? I imagine they are redundant but want to be sure.

matiasinsaurralde:
You can remove these lines, they were used by the `msgpack` binding generator, but we're using Protocol Buffers so it's ok.

PR #580, opened 2017-03-02
Lowercase lots of things

mvdan:
Capitalised in Go usually means exported, and none of these are nor should be.


mvdan:
For the curious, lowercasing funcs in the package doesn't seem to affect the output binary. This is likely because this is a `main` package, so the Go compiler knows exported names can't really be used elsewhere and treats them like unexported ones (with regards to inlining, etc).

PR #581, opened 2017-03-02
Fix two flaky tests

mvdan:
`TestCoProcessReload` and `TestHotReloadSingle` - the latter is what CI had run into twice over the last week.

### DIFF
--- a/redis_signals.go
+++ b/redis_signals.go
@@ -72,8 +72,12 @@ func handleRedisEvent(v interface{}) {
 		OnServerStatusReceivedHandler(notif.Payload)
 	case NoticeGatewayLENotification:
 		OnLESSLStatusReceivedHandler(notif.Payload)
-	default:
+	case NoticeGroupReload:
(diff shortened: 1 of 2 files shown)


mvdan:
@lonelycode rebased

redis_signals.go
```diff
@@ -72,8 +72,12 @@ func handleRedisEvent(v interface{}) {
 		OnServerStatusReceivedHandler(notif.Payload)
 	case NoticeGatewayLENotification:
 		OnLESSLStatusReceivedHandler(notif.Payload)
-	default:
+	case NoticeGroupReload:
```
lonelycode: I think this is perfectly OK, so I will merge this, but there's a potential case where a reload is triggered by a policy update signal which we might need to keep an eye on in regression testing with the dashboard.

mvdan: Oh, okay. Better have that explicitly listed here, in any case.

PR #584, opened 2017-03-02
Unindent more code, remove dead var

mvdan:


handler_error.go
```diff
@@ -227,11 +226,9 @@ func (e *ErrorHandler) HandleError(w http.ResponseWriter, r *http.Request, errMs
 		if prior, ok := r.Header["X-Forwarded-For"]; ok {
 			clientIP = strings.Join(prior, ", ") + ", " + clientIP
 		}
-		ip = clientIP
 	} else {
 		log.WithFields(logrus.Fields{
 			"prefix":      "gateway",
-			"user_ip":     ip,
```
lonelycode: We need the IP in there, this might be a dead var, but we can;t losr it from the log message - some users require IP traces for security auditing.

mvdan: Even if this ip is always empty string?

I can have it be `"user_ip": "",` if that is still useful.

lonelycode: Maybe make it `remoteAddr`? Then it may or may not be an empty string.

mvdan: Same key `user_ip`?

lonelycode: Yes please

mvdan: Done.

mvdan:
How do I mark the requested changes as done? Is it only you that can do that?

PR #586, opened 2017-03-03
Simplify and unindent even more code

mvdan:


lonelycode:
All good - though we might want to look at what debug logs we want to include, lots of people run with debug on.

PR #590, opened 2017-03-06
Shorten EventMessage fields

mvdan:
There is no reason for the fields to repeat Event. This simplifies its
usage.

Also simplify event meta selector in the tests.


lonelycode:
LGTM

PR #591, opened 2017-03-07
Remove *string parameters, simplify type switches

mvdan:


buger:
Nice trick with type switch!

PR #595, opened 2017-03-07
Use GetIPFromRequest helper func everywhere

mvdan:
We had semi-working copies all over the place. They were incomplete and
buggy like the old GetIPFromRequest, like how they missed ipv6 support.

Use the common helper func, which is also unit tested.

### DIFF
--- a/handler_error.go
+++ b/handler_error.go
@@ -212,6 +212,7 @@ func (e *ErrorHandler) HandleError(w http.ResponseWriter, r *http.Request, errMs
 		obfuscated = "****" + keyName[len(keyName)-4:]
 	}
 
+	// TODO: what is this bit of code for?
 	clientIP, _, err := net.SplitHostPort(r.RemoteAddr)
 	if err == nil {
 		// If we aren't the first proxy retain prior
--- a/api.go
+++ b/api.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"io/ioutil"
-	"net"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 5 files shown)


mvdan:
@lonelycode note that this touches the log call I modified recently to use `r.RemoteAddr`. I don't see how that log call could ever fire or why it's useful, as it's from an error to get an IP address that we never use.

handler_error.go
```diff
@@ -212,32 +209,6 @@ func (e *ErrorHandler) HandleError(w http.ResponseWriter, r *http.Request, errMs
 		w.Header().Add("Connection", "close")
 	}
 
-	var obfuscated string
```
buger: It seems like this part was completely removed, without being replaced by `GetIPFromRequest`. Even if `clientIP` not used anywhere, it still miss error logging, in case if `RemoteAddr` has issues.

lonelycode:
> note that this touches the log call I modified recently to use r.RemoteAddr. I don't see how that log call could ever fire 

When could it fire: When `SplitHostPort()` fails, which admittedly is unlikely, but it would be abnormal to fail, and in the interest of exposing abnormal behaviour, I'd like an error message there to cover the scenario. (See all the error scenarios in the source: https://golang.org/src/net/ipsock.go?s=3572:3638#L104)

> or why it's useful, as it's from an error to get an IP address that we never use.

The IP address *is* used, it is used to modify the `X-Forward-For` header, which informs the upstream server  of the proxy path that this request has gone through, if someone were messing with the chain, it might be useful to inform the user...

mvdan:
> When SplitHostPort() fails, which admittedly is unlikely, but it would be abnormal to fail, and in the interest of exposing abnormal behaviour, I'd like an error message there to cover the scenario.

Let me rephrase this - what we pass is `r.RemoteAddr`, which `net/http` says will be a valid addr+port in the very same format. Unless we have messed with `RemoteAddr` (hopefully not), this would only fire if Go's standard library is buggy.

> The IP address is used, it is used to modify the X-Forward-For header

Sorry, where? We only get the header - we don't set it nor modify it. Really, all we were doing there is the old version of `GetIPFromRequest`.

handler_error.go
mvdan: See my comment above and the discussion below.

lonelycode:
Sweet Christmas. Just leave it alone please.

mvdan:
As you wish.

mvdan:
@buger rebased, PTAL

Issue #531, opened 2017-02-16
panic: http: multiple registrations for /


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
Tyk starts failing on start, restarting it does not help.
```
[Feb 16 17:39:01] INFO main: Processed and listening on: /isAlive{rest:.*}
[Feb 16 17:39:01] INFO main: Processed and listening on: /v2.0/{rest:.*}
[Feb 16 17:39:01] INFO main: Processed and listening on: /doc/{rest:.*}
[Feb 16 17:39:01] DEBUG Checker host list
[Feb 16 17:39:01] INFO host-check-mgr: Loading uptime tests...
[Feb 16 17:39:01] DEBUG host-check-mgr: --- Setting tracking list up
[Feb 16 17:39:01] DEBUG Checker host Done
[Feb 16 17:39:01] INFO main: Initialised API Definitions
[Feb 16 17:39:01] INFO main: API reload complete
[Feb 16 17:39:01] DEBUG main: Downloading API Configurations from Dashboard Service
[Feb 16 17:39:01] INFO main: Detected 0 APIs
[Feb 16 17:39:01] INFO main: Loading API configurations.
[Feb 16 17:39:01] DEBUG Storage Engine already initialised...
[Feb 16 17:39:01] DEBUG Received DRL data: {datafront-1 <redacted> 4 0 }
[Feb 16 17:39:01] DEBUG Active Nodes: : 6 [Token Bucket Value]: 600 [Current Load p/s]: 1 [Current Load]: 0.000000
[Feb 16 17:39:02] DEBUG Calling: http://<redacted>:5000/register/ping
[Feb 16 17:39:02] DEBUG Sending Heartbeat as: <redacted>
[Feb 16 17:39:02] DEBUG Sending notification{NoticeGatewayDRLNotification {"HostName":"<redacted>","ID":"<redacted>","LoadPerSec":1,"Percentage":0,"TagHash":""} }
[Feb 16 17:39:02] DEBUG Received DRL data: {datafront-2 <redacted> 1 0 }
[Feb 16 17:39:02] DEBUG Active Nodes: : 6 [Token Bucket Value]: 600 [Current Load p/s]: 1 [Current Load]: 0.000000
[Feb 16 17:39:02] DEBUG Hearbeat Finished: Nonce Set: <redacted>
[Feb 16 17:39:02] DEBUG Checker host list
[...]

PR #596, opened 2017-03-07
Add router tests and fix related issues

buger:
Integration router tests run all combinations of using `goagain`, `config.HttpServerOptions.OverrideDefaults`, and control API on a separate port.

Following fixes are made:

* Fixed `/tyk/api` endpoint without slash #372
* Fixed control API when `config.HttpServerOptions.OverrideDefaults` is `true` (both when with
Goagain and without) #612 
* Fixed: do not bind control API if RPC_EmergencyMode turned on #612
* Fix Control API on a separate port in forked process #612
* Potential #531 fix by not using `http.Handle("/", …)` in favor of `http.Serve`

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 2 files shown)


buger:
@matiasinsaurralde since it touch Control API functionality, can you check this changes?

mvdan:
You can't add `start()` to `TestMain` like that, as it also calls `reloadLoop`. If you do, you must move it out of `start`.

mvdan:
Also, could you split up the fixes in separate commits? Otherwise, if any of the independent changes needs reverting, it would be much harder. It's also easier to see what code changes are from what fix.

mvdan:
Oh, forgot an important point - having fixes as individual commits will also enable backporting into 2.3. At least a couple of these we want to backport, I think.

mvdan:
I have to run now, but I'll review this tomorrow morning.

main.go
```diff
@@ -1370,6 +1370,10 @@ func listen(l, controlListener net.Listener, err error) {
 			if specs != nil {
 				loadApps(specs, defaultRouter)
 				getPolicies()
+
+				if config.ControlAPIPort > 0 {
+					loadAPIEndpoints(controlRouter)
+				}
```
mvdan: > If it bind on separate port, it not works

"If it binds on a separate port, it doesn't work."

main.go
```diff
@@ -1380,6 +1390,16 @@ func listen(l, controlListener net.Listener, err error) {
...
+
+			if controlListener != nil {
+				cs := &http.Server{
+					ReadTimeout:  time.Duration(readTimeout) * time.Second,
+					WriteTimeout: time.Duration(writeTimeout) * time.Second,
+					Handler:      controlRouter,
+				}
+				go cs.Serve(controlListener)
+			}
+
```
mvdan: These two blocks are identical copies. Dedup in a func?

gateway_test.go
```diff
@@ -804,3 +805,127 @@ func TestWithAnalyticsErrorResponse(t *testing.T) {
...
+// Admin api located on separate port
+func TestControlListener(t *testing.T) {
+	tests := []tykHttpTest{
+		{"GET", "/", 404, false, false},
+		{"GET", "/tyk/apis", 404, false, false},
+
+		// Querying control API
+		{"GET", "/", 404, false, true},
+		{"GET", "/tyk/apis", 403, false, true},
+		{"GET", "/tyk/apis/", 200, true, true},
```
mvdan: My only nit with this is that you have to mentally keep track of what each bool means. Perhaps using keyed vaues would help. If you think that's overkill for just the two bools, leaving it as-is sounds fine too.

main.go
buger: Yeah, we can do that in future, this block does have differences, but minor ones. Let's make it as separate PR, maybe you will touch this part someday :)

gateway_test.go
buger: Yeah, I do not like this set of booleans as well. Will think about using keyed values.

PR #598, opened 2017-03-08
test: wait for TestResetHandler to finish reload

### LINKED ISSUE #588: TestHotReloadSingle is still flaky
We're still getting the "reload wasn't queued" flake. I thought I had fixed it in https://github.com/TykTechnologies/tyk/pull/581, but apparently I didn't fix it completely.

mvdan:
First off, the tick was correct but it was added to the wrong test. The
group reset test doesn't actually fire a hot reload, as we use our own
pub/sub handler that doesn't fire one.

It's the single-node reset (reload) test that does fire a hot reload.

Moreover, we also need to wait for the reload to be done. Otherwise, it
could be that the reload worker doesn't have the chance to pick up the
job from the queue and other tests that need the queue to be empty would
fail:

	$ go test -run '^Test(ResetHandler|HotReloadSingle)$' -v
	=== RUN   TestResetHandler
	--- PASS: TestResetHandler (0.01s)
	=== RUN   TestHotReloadSingle
	--- FAIL: TestHotReloadSingle (0.00s)
		api_test.go:692: reload wasn't queued

Given that I was able to reproduce the issue locally and the logs are
identic to those of the Travis failures, I'm confident to say that this
particular flake is now fixed.

Fixes #588.

### DIFF
--- a/api.go
+++ b/api.go
@@ -545,11 +545,11 @@ func handleDeleteHashedKey(keyName, apiID string) ([]byte, int) {
 	return responseMessage, 200
 }
 
-func handleURLReload() ([]byte, int) {
+func handleURLReload(fn func()) ([]byte, int) {
 	var responseMessage []byte
 	var err error
 
-	reloadURLStructure(nil)
+	reloadURLStructure(fn)
 
 	statusObj := APIErrorMessage{"ok", ""}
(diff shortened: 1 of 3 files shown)


mvdan:
@buger PTAL, this is somewhat urgent

PR #600, opened 2017-03-08
Simplify API code even more

mvdan:
Removed the last few "if r.Method == ..." ifs covering entire methods.
4/5 levels of indentation are much less common now.


mvdan:
@buger rebased

PR #609, opened 2017-03-09
Replace nu7hatch/gouuid with satori/go.uuid

mvdan:


mvdan:
Also note that I haven't changed any of the code regarding `.String()` and `strings.Replace` to keep this PR smaller. Both libs use UUID v4 and both libs have the same `String()` implementation.

buger:
You also removed this transformation `strings.Replace(u5.String(), "-", "", -1)`. Does it expected? Such change can have unexpected side effects .

mvdan:
This is because `String()` is just hex with some dashes in between. It's easier to just do hex of the whole thing.

https://github.com/nu7hatch/gouuid/blob/master/uuid.go#L171
https://github.com/satori/go.uuid/blob/master/uuid.go#L197

buger:
It changes the length of encoded string. Better to call @lonelycode here to clarify why it was added.

mvdan:
I can do this simplification in a separate PR if that helps, but I don't see what you mean about the length.

mvdan:
Actually, this was my intention all along:

> Also note that I haven't changed any of the code regarding .String() and strings.Replace to keep this PR smaller.

I guess I got mixed up while posting it.

mvdan:
@buger undid strings.Replace bit

buger:
I assume you should add same change (`strings.Replace`) to `b64UUID` function too.

mvdan:
Hmm, only one of the three base64 calls actually used that. This is confusing.

I guess we could use it everywhere, but that will sure change behaviour. It's also odd that we first convert to hex and then to base64. I would assume the hex step is redundant.

buger:
I see, well, let's wait for @lonelycode here. I also do not like this `string.Replace` stuff, so if we can confirm it is legacy, it should be safe to remove.

lonelycode:
Keep the strings.Replace stuff please - there's a wider refactor where we base64 encode every UUID but for now we use them as is pls.

buger:
bump

mvdan:
Nothing to do - this is on hold until the security releases are out.

mvdan:
I've removed all of the refactoring, so now it's only swapping the libs. @buger PTAL

PR #610, opened 2017-03-09
Un-fork logrus, vendor stable versions of its pkgs

mvdan:
The reason why we forked all of these libs is because of the Sirupsen vs
sirupsen (case sensitive) issue. But we are immune to those issues if we
vendor all the libs and we're careful when updating them.

Moreover, we were running HEAD versions of all these libs. Use the latest
stable version of all of them as of March 2017.


mvdan:
I realise this is a bit of a mindfuck at the moment - it was tricky to do too - but I think this is important and should simplify things going forward. Keeping our own forks is not sustainable and adds quite a lot of complexity.

And, as said above, keeping our own forks doesn't really help us in any way.

lonelycode:
So, the question is, at any point were any of the forks modified or touched? I can't remember messing with any, but might be worth confirming first.

buger:
bump

mvdan:
Nothing to do - this is on hold until the security releases are out.

mvdan:
Confirming that the only changes we ever did to all of those libs was just replace `foo/*` with `TykTechnologies/*` in the imports.

Rebased on master. @buger PTAL

Issue #613, opened 2017-03-10
Make it possible to exclude a node from the DRL


lonelycode:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
If a node is reserved as a management node, then it can affect the way the rate limiter calculates bucket sizes making rate limits lower than they are set at.

**What is the expected behavior?**
The rate limits in a set of gateways with a mangement node (reserved, non-active node), should maintain the rate limit appropriately.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Set up three nodes, two behind a load balancer, then send traffic to the LB against a rate limited token, the rate limit will be lower than what is set.

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
v2.3.x

buger:
Is the there is a way to know if it is a management node? Or it require adding a new flag?

lonelycode:
It will require a flag to saybitnis one

lonelycode:
Actually, the flag should just stop it from broadcasting its details on the pub sub channel. It should be very straightforward.

buger:
@lonelycode should we also exclude it from the quota calculations?

buger:
Also, it does not look like a proper fix. Maybe it makes sense to add an option which disables any rate limiting and quotas for the whole gateway?

lonelycode:
So quotas are not broadcast and do not affect the server table that gets built up by the nodes. Only the DRL does that.

So the fix is: 

- Add a flag ManagementNode:true
- if set the DRL does not activate at all

This stops the broadcast, the other rate limiters do not make use of broadcast so that's ok and the quota system uses a direct redis counter so this should not need modifying.

A user could just disable the rate limiter by enabling one of the others, but that isn't intuitive.

buger:
We should not forget to update documentation

PR #614, opened 2017-03-10
redis: remove another duplicate pub/sub call

mvdan:
In b9530aba, we removed a duplicate StartPubSubHandler call that wasn't
checking for errors. This is another case. Like in the previous commit,
the extra Connect() call is unnecessary as RedisClusterStorageManager
only connects once - it's the underlying rediscluster that manages
reconnects.

Updates #520.

@buger this would be a good candidate to backport into 2.3, as it could help with #520 and we backported the other commit that did the same but somewhere else.


mvdan:
Also, I found this by chance while looking into https://github.com/TykTechnologies/tyk/issues/613#issuecomment-285602284 for 2.3.4. But since this is also a 2.3.4 thing, I figured why not.

PR #619, opened 2017-03-13
Add ManagementNode config option to disable DRL

### LINKED ISSUE #613: Make it possible to exclude a node from the DRL

mvdan:
Fixes #613.

### DIFF
--- a/config.go
+++ b/config.go
@@ -194,6 +194,7 @@ type Config struct {
 	EnableNonTransactionalRateLimiter bool                   `json:"enable_non_transactional_rate_limiter"`
 	EnableSentinelRateLImiter         bool                   `json:"enable_sentinel_rate_limiter"`
 	EnableRedisRollingLimiter         bool                   `json:"enable_redis_rolling_limiter"`
+	ManagementNode                    bool                   `json:"management_node"`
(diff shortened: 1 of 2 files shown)


mvdan:
Note: I had written a "test" for this, like:

```
func TestDRLManagementNode(t *testing.T) {
        config.ManagementNode = true
        defer func() {
                config.ManagementNode = false
        }()
        DRLManager.Ready = false
        startDRL()
        if DRLManager.Ready {
                t.Fatal("Expected DRL to not be set up")
        }
}
```

But on one hand, this test is of little use (all it tests is a conditional, really) and it can break other tests that depend on the DRL. And if we make the test re-enable the DRL after it's done, it will leak a goroutine since `startRateLimitNotifications` will never stop.

So I opted to not include the test.

PR #621, opened 2017-03-14
Fix API reloads and add related tests

### LINKED ISSUE #606: Gateway returns 404 error for all api requests
ENV: local docker dev-env
Branch: Current gateway master, current dashboard master
Version: On-premise 

Reproduction steps:
1. Log on dashboard
2. Go to API's
3. Create "test-api" api definiton 
4. Set keyless for Authentication mode
5. Save
6. Open api url(in my case http://localhost:8181/test-api/)

Actual:
Gateway returns 404 error for all requests

Expected:
Gateway should proxy request to https://httpbin.org/

Note:
This is visible only on latest gateway master, on release 2.3.3 this issue is not visible, also only on-prem version is affected. Community version is ok

buger:
It broke because `doReload` functionality was using `DefaultServeMux` for routing, but our initial `listen` logic used custom muxer, which is created automatically if you pass non nil second argument to `http.Serve`.

Change which broke this functionality (#596) was intended to fix `panic` caused by multiple registrations of “/“ route handle #531.

We did not noticed this bug because we had no tests covering API reloads, and this PR adds them as well.

This PR fixes API reloads, and still ensure that no multiple route registrations will be made. It is achieved by creating new `newServeMux` and updating `http.DefaultServeMux` with it.

Note that our routing and reload functionality have a lot of duplication (in 3 places), and should be unified one day. Goal of this PR is to fix the bug, and not to make refactoring.

Should fix #606

### DIFF
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"testing"
 
@@ -464,5 +465,32 @@ func TestGetAPISpecsRPCSuccess(t *testing.T) {
 	if len(specs) != 1 {
 		t.Error("Should return array with one spec", specs)
 	}
+}
 
+func TestGetAPISpecsDashboardSuccess(t *testing.T) {
+	// Mock Dashboard
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path == "/system/apis" {
+			w.Write([]byte(`{"Status": "OK", "Nonce": "1", "Message": [{"api_definition": {}}]}`))
+		} else {
+			t.Fatal("Unknown dashboard API request", r)
+		}
+	}))
+	defer ts.Close()
+
+	ApiSpecRegister = make(map[string]*APISpec)
+
+	config.UseDBAppConfigs = true
+	config.DBAppConfOptions.ConnectionString = ts.URL
+
+	defer func() {
+		config.UseDBAppConfigs = false
+		config.DBAppConfOptions.ConnectionString = ""
+	}()
+
+	doReload()
+
+	if len(ApiSpecRegister) != 1 {
+		t.Error("Should return array with one spec", ApiSpecRegister)
+	}
 }
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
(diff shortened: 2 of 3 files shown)


buger:
@ilijabojanovic can you check it?

api_definition_test.go
```diff
@@ -464,5 +465,30 @@ func TestGetAPISpecsRPCSuccess(t *testing.T) {
...
+	config.DBAppConfOptions.ConnectionString = ts.URL
+
+	doReload()
+
+	if len(ApiSpecRegister) != 1 {
+		t.Error("Should return array with one spec", ApiSpecRegister)
+	}
 
+	config.UseDBAppConfigs = false
+	config.DBAppConfOptions.ConnectionString = ""
```
mvdan: set these back in a `defer` right after you change them - this way, it won't break if the test exits early (even if it doesn't yet - it will be future-proof)

gateway_test.go
```diff
@@ -843,6 +846,9 @@ func testHttp(t *testing.T, tests []tykHttpTest, separateControlPort bool) {
 			config.HttpServerOptions.OverrideDefaults = false
 		}
 
+		// Ensure that no local API's installed
+		os.RemoveAll(config.AppPath)
+		config.AppPath, _ = ioutil.TempDir("", "tyk-test-")
```
mvdan: `AppPath` is already set to a temp dir in `TestMain`. Why are these two lines necessary?

gateway_test.go
```diff
@@ -900,13 +916,40 @@ func testHttp(t *testing.T, tests []tykHttpTest, separateControlPort bool) {
...
+                "expires": "3000-01-02 15:04"
+            }
+        }
+	},
+    "proxy": {
+        "listen_path": "/",
+        "target_url": "http://127.0.0.1:16500",
+        "strip_listen_path": true
+    },
+    "active": true
```
mvdan: Please indent with tabs like the rest of raw JSONs

gateway_test.go
buger: This tests need to ensure that NO apis exists at start of the tests. So it removes all the previous API's before each test run. `os.RevemoAll` also removes the test directory, so I have to re-create it.

mvdan: I'd instead empty the directory. Or, if you want to nuke it entirely, at least don't ignore errors.

api_definition_test.go
```diff
@@ -464,5 +465,32 @@ func TestGetAPISpecsRPCSuccess(t *testing.T) {
...
+		}
+	}))
+	defer ts.Close()
+
+	ApiSpecRegister = make(map[string]*APISpec)
 
+	config.UseDBAppConfigs = true
+	config.DBAppConfOptions.ConnectionString = ts.URL
+
+	defer func(){
```
mvdan: `gofmt`?

mvdan:
LGTM, please squash

mvdan:
The failure looks like a coveralls 504, so not our problem.

ilijabojanovic:
@buger approved for merge and also https://github.com/TykTechnologies/tyk/issues/606 is fixed

PR #631, opened 2017-03-21
Fix dashboard API reloads

### LINKED ISSUE #606: Gateway returns 404 error for all api requests

buger:
Not so long ago during refactoring, we removed handling of unknow redis
messages, which by default was triggering API reloads.
https://github.com/TykTechnologies/tyk/commit/8f6223e2328b579d1e95063191
97499874f2a0b9

It actually broke dashboard API reloads, because they were handling in
this `default` clause.

Another refactoring also removed some unused constants, which was
indeed unused, but I need to roll them back, because they are required
to whitelist redis messages which could trigger API reloads.

Should fix #606, again...

Worth noticing that previous PR was fixing API reloads as well, and it was a valid fixes, but not for API dashboard reloads. Previous PR included tests covering Dashboard API reloads, but it did not used redis messaging, only mocked http endpoint. This PR fix tests as well.

### DIFF
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -4,7 +4,9 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/garyburd/redigo/redis"
 	"github.com/lonelycode/gorpc"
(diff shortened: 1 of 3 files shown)


mvdan:
On one hand apologies for breaking master. On the other, I'm so glad this was spotted and we no longer depend on default cases :)

buger:
@mvdan tests are failing in a strange way, and I have a feeling this is similar to why you removed this default case. Error similar to what you experienced previously, do you have ideas?

mvdan:
This is likely some test firing one of these events that used to fall into the default warning but now trigger a reload, throwing off the manual reload ticker.

Let me poke around a bit once I'm done with the dashboard PR and get back to you.

ilijabojanovic:
@buger 404 issue is fixed, approved for merge

mvdan:
@buger should be working now, PTAL. Ilija's OK should still be fine as there is no logic change in the non-test code.

PR #635, opened 2017-03-24
Fix Uptime tests reverse logic

### LINKED ISSUE #632: Upstream targets not taken out of pool when upstream tests fail
bug
Uptime Tests feature doesn't work well, with this option enabled gateway sends traffic on host which is down
Gateway should exclude down host from routing 

Current Master and Stable 2.3.3

More details:
https://community.tyk.io/t/upstream-targets-not-taken-out-of-pool-when-upstream-tests-fail/1481

buger:
Should fix #632

When limit set to 1, it did not mark host as failed and continue sending traffic to it, until next host check. Worth noticing that now it also use `>=` logic instead of `>`, so it will trigger failed host check earlier than before, but it looks more logical: if `trigger_limit` set to 1, we want to send notification on the first failure. 

Also, fixed Uptime tests reverse logic: `IsHostDown` function currently has reverse logic, and returns true for active hosts. Fun thing, Tyk still works fine because `GetNextTarget` have reverse
logic as well, and they fix each other :)

Writing tests for this functionality was tricky because functionality spread up across the app, but thankfully we can set the callback, similar to how webhooks work, and it worked quite well. 

Added test covering part of uptime tests functionality.

### DIFF
--- a/host_checker.go
+++ b/host_checker.go
@@ -102,25 +102,21 @@ func (h *HostUptimeChecker) HostReporter() {
 			go h.pingCallback(okHost)
 
 		case failedHost := <-h.errorChan:
+			newVal := 1
+			if count, found := h.sampleCache.Get(failedHost.ID); found {
+				newVal = count.(int) + 1
+			}
 
-			cachedHostCount, found := h.sampleCache.Get(failedHost.ID)
-			if !found {
-				go h.sampleCache.Set(failedHost.ID, 1, cache.DefaultExpiration)
+			h.sampleCache.Set(failedHost.ID, newVal, cache.DefaultExpiration)
 
-			} else {
-				newVal := cachedHostCount.(int)
-				newVal++
-				go h.sampleCache.Set(failedHost.ID, newVal, cache.DefaultExpiration)
-
-				if newVal > h.sampleTriggerLimit {
-					log.Debug("[HOST CHECKER] [HOST WARNING]: ", failedHost.CheckURL)
-					// Reset the count
-					go h.sampleCache.Set(failedHost.ID, 1, cache.DefaultExpiration)
-					// track it
-					h.unHealthyList[failedHost.ID] = true
-					// Call the custom callback hook
-					go h.failureCallback(failedHost)
-				}
+			if newVal >= h.sampleTriggerLimit {
+				log.Debug("[HOST CHECKER] [HOST WARNING]: ", failedHost.CheckURL)
+				// Reset the count
+				h.sampleCache.Set(failedHost.ID, 1, cache.DefaultExpiration)
+				// track it
+				h.unHealthyList[failedHost.ID] = true
+				// Call the custom callback hook
+				go h.failureCallback(failedHost)
 			}
 			go h.pingCallback(failedHost)
 
--- a/host_checker_manager.go
+++ b/host_checker_manager.go
@@ -282,7 +282,7 @@ func (hc *HostCheckerManager) IsHostDown(urlStr string) bool {
 	_, err = hc.store.GetKey(PoolerHostSentinelKeyPrefix + u.Host)
 
 	// Found a key, the host is down
-	return err != nil
+	return err == nil
 }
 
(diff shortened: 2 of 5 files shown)


buger:
tip: better to review with white space check turned off https://github.com/TykTechnologies/tyk/pull/635/files?w=1

host_checker.go
```diff
@@ -104,23 +104,23 @@ func (h *HostUptimeChecker) HostReporter() {
...
-					log.Debug("[HOST CHECKER] [HOST WARNING]: ", failedHost.CheckURL)
-					// Reset the count
-					go h.sampleCache.Set(failedHost.ID, 1, cache.DefaultExpiration)
-					// track it
-					h.unHealthyList[failedHost.ID] = true
-					// Call the custom callback hook
-					go h.failureCallback(failedHost)
-				}
+			newVal := cachedHostCount.(int)
+			newVal++
```
mvdan: You could instead do:

```
newVal := 1
if count, found := h.sampleCache.Get(failedHost.ID); found {
    newVal = count.(int) + 1
}

File: host_checker.go [hunk 2]
```diff
@@ -104,23 +104,23 @@ func (h *HostUptimeChecker) HostReporter() {
...
+			go h.sampleCache.Set(failedHost.ID, newVal, cache.DefaultExpiration)
+
+			if newVal >= h.sampleTriggerLimit {
+				log.Debug("[HOST CHECKER] [HOST WARNING]: ", failedHost.CheckURL)
+				// Reset the count
+				go h.sampleCache.Set(failedHost.ID, 1, cache.DefaultExpiration)
```
mvdan: What if the `Set` above happens to run after the one here? Do we really need to fire the two goroutines, ever?

(I know this is not your code and it's not modified in this PR, but still)

host_checker_manager.go
```diff
@@ -282,7 +282,7 @@ func (hc *HostCheckerManager) IsHostDown(urlStr string) bool {
 	_, err = hc.store.GetKey(PoolerHostSentinelKeyPrefix + u.Host)
 
 	// Found a key, the host is down
```
mvdan: Ha! The comment was right, the code was wrong :)

host_checker_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	wg.Wait()
+
+	if GlobalHostChecker.IsHostDown("http://127.0.0.1:16500") {
+		t.Error("Should not mark as down")
+	}
+
+	if !GlobalHostChecker.IsHostDown("http://127.0.0.1:16501") {
+		t.Error("Should mark as down")
+	}
```
mvdan: Perhaps I'm blind, but what is making `:16500` up and `:16501` down?

mvdan: Oh duh, this is the port I used for the test http server that replaced httpbin.

I suggest you use these constants:

```
testHttpListen = "127.0.0.1:16500"
testHttpAny  = "http://" + testHttpListen
```

Otherwise it's not obvious from reading just this code, and it's also fragile since it will break if the port ever changes.

buger: Yeah, I made it hardcoded to avoid variable replace or templates.

mvdan: Sorry, I don't understand. In my opinion reusing the constants is better in every way.

buger:
@mvdan made the changes

host_checker_test.go
```diff
@@ -0,0 +1,138 @@
...
+        "listen_path": "/",
+        "enable_load_balancing": true,
+        "check_host_against_uptime_tests": true,
+        "target_list": [
+            "{{.ActiveHost}}",
+            "{{.InactiveHost}}"
+        ],
+        "strip_listen_path": true
+    },
+    "active": true
```
mvdan: please indent with tabs for consistency

host_checker_test.go
mvdan: Bump

buger:
Should be good now

JasonSheedy:
Awesome guys. Thanks for fixing this.

PR #646, opened 2017-03-29
Minor enhancements on 

matiasinsaurralde:
Added some minor enhancements, some of them follow the vetshadow suggestions.
The default permissions for bundle directory were "too open".


api.go
```diff
@@ -650,7 +650,7 @@ func handleAddOrUpdateApi(apiID string, r *http.Request) ([]byte, int) {
 		return createError("Marshalling failed"), 500
 	}
 
-	if err := ioutil.WriteFile(defFilePath, asByte, 0644); err != nil {
```
mvdan: `vetshadow` was suggesting these? IMO these changes should not be made. It makes these pieces of code depend on `err` being declared previously in the block.

command_mode.go
```diff
@@ -34,35 +34,12 @@ func handleCommandModeArgs(arguments map[string]interface{}) {
 }
 
 func handleBluePrintMode(arguments map[string]interface{}) {
-	doCreate := arguments["--create-api"]
+	doCreate := arguments["--create-api"].(bool)
```
mvdan: This will panic if it's not a bool type, e.g. if it's actually just nil. Are you 100% certain that it will always be of bool type? Have you tried to get it to crash by using weird parameters like `--create-api=foo`?

mvdan:
Bump @matiasinsaurralde

command_mode.go
matiasinsaurralde: No, it doesn't panic, it expects an argument:
```
% ./tyk --create-api=foo
--create-api must not have an argument
Usage:
		tyk [options]
% ./tyk --create-api foo
Usage:
		tyk [options]
```

PR #661, opened 2017-04-06
Don't flood logs on management nodes, plus regression test

mvdan:


mvdan:
Also note that I split the tiny fix into a separate commit, since we don't want to (nor could we) backport the test to stable.

PR #664, opened 2017-04-06
Allow Key API fallback to default storage if API with api_id not found

### LINKED ISSUE #663: /keys/* does not work if `api_id` param is provided but API not loaded on this node
bug
`api_id` param is not a filter in our case, more like a hint to our API so it will know in which storage engine it should search for the key data.

In some cases users may want to create multiple versions of the same API but with different tags, for example `stage`, `prod`. When we `tag` gateway it loads only API with given tag. It can be that our Dashboard will talk to this node, which does not load all the API's.

For querying key dashboard use `/:api_id/keys/:key` (which is translated to `/tyk/keys/:key?api_id=:api_id` on gateway side), and what is more important, UI always query data with the first API in a row.

If dashboard is configured to talk with tagged instance and this instance does not loaded the API, which is send via "api_id" request, key search will fail. On the other hand if `api_id` is not set at all, search will work as expected, considering the fact that key is stored in default (redis) storage (which is true for 99% users).

Key related API's should work for all gateway nodes, tagged or not. 

1. Use Gateway with `prod` tag to talk with Dashboard instance
1. Create API1 without any tags.
1. Create API2 with tag `prod`
1. Create Key for API2
1. Try to search in dashboard for just created key

It will fail, because search will be made in context of first API1, and it is not loaded to Gateway Node.

All

buger:
Should fix #663

### DIFF
--- a/api.go
+++ b/api.go
@@ -199,11 +199,9 @@ func GetKeyDetail(key, apiID string) (SessionState, bool) {
 	sessionManager := FallbackKeySesionManager
 	if apiID != "" {
 		spec := GetSpecForApi(apiID)
-		if spec == nil {
-			log.Error("No API Spec found for this keyspace")
-			return SessionState{}, false
+		if spec != nil {
+			sessionManager = spec.SessionManager
 		}
-		sessionManager = spec.SessionManager
 	}
 
 	return sessionManager.GetSessionDetail(key)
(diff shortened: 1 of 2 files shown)


buger:
@lonelycode i'm not sure that fallback here is the "proper" solution, but it can definitely work.

The proper solution will be probably re-thinking how search is made in Dashboard, and it should search keys in multiple Tyk nodes if needed. Or look at the API tags, compare them with Gateway Node tags (not sure if this information available), and then query Node which 100% should have access to the key. 

Current way of always querying using "first" API id is indeed a bit confusing, and works only in simple cases (for most of users it is acceptable).

lonelycode:
Agreed - this is a stopgap solution. The dashboard shouod direct the request to the correct gateway - the issue is it has no knowledge of which back-end to use. So we may need to cache metadata in mongo to make it smarter.

This is actually on the roadmap.

PR #665, opened 2017-04-07
Fix spelling in log. "rate imiter" -> "rate limiter"

waterimp:
I keep noticing in the logs. Might as well just fix it. :)


buger:
Thank you!

PR #673, opened 2017-04-13
Update default uptime test configuration

### LINKED ISSUE #668: Change default values for uptime tets in default tyk.config
Feature
uptime_tests”: {
       “disable”: false,
       “config”: {
           “failure_trigger_sample_size”: 3,
           “time_wait”: 300,
           “checker_pool_size”: 50,
           “enable_uptime_analytics”: true
       }
   },

time_wait value is 300 seconds (5 minutes) 
In order to improve host checking we should set lower value in default tyk configuration

All

We need to update documentation and default tyk configuration files on all distributions (packages, docker images..)

buger:
Currently, default gateway config assume that it should wait 20 minutes to mark host as down.
Reduced to 20 seconds.

Fix #668

### DIFF
--- a/install/data/tyk.self_contained.conf
+++ b/install/data/tyk.self_contained.conf
@@ -55,8 +55,8 @@
     "disable": false,
     "config": {
       "enable_uptime_analytics": false,
-      "failure_trigger_sample_size": 3,
-      "time_wait": 300,
+      "failure_trigger_sample_size": 2,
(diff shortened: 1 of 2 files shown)


buger:
@lonelycode i set 2 retries with 10 seconds between them, so 20 seconds till host marked is down. Seems sane.

PR #674, opened 2017-04-19
Distributed Quota branch, cherry-picked against master

lonelycode:
Ok this is a complicated one.

Distributed Quota feature (lets call it a beta): instead of having the quota enforced by a shared counter in redis, the counter is in-memory and counters are shared across a message bus to other cluster gateways, then merged, to create a distributed total. This means no centralised state and no one-to-one linkage with Redis.

**What we are trying to achieve:**

1. Stop having a one-to-one relationship with redis, when a counter-based quota is enforced, every inbound request means a roundtrip to redis, this is expensive and adds latency, even if done async it floods the redis connection counter pool and means performance degrades as traffic increases
2. Make counter operations faster: Since these are in-memory and are merged in the background, the counter is entirely in-memory and no round-trip or lookup is required. This should improve throughput for metered connections
3. The new DQ Quota object means that actually we can introduce new metering metrics because it is an abstraction
4. Managed redis writes (or any persistence), since the distributed cluster, with a single leader, will only write to the store every `x` interval (default to 3 seconds),  so no matter how high the load, the writes speed isn't a linear relation to throughput.

To enable set the following in `tyk.conf`:

- `use_distributed_counter` to `true`: This enables the feature 
- `distributed_quota_set_master` to `true`: Since we only have redis as a pub/sub mechanism at the moment, we need all nodes to be masters, this causes all nodes to persist the CRDT list and is therefore fault tolerant. This is optional, obviously, if you can be completely sure that your master will not fail - so long as one node is set as master we are clear.

It will work in single-host mode just like a regular quota.

There is a unit test, but multi-host testing was done manually, as well a failure conditions:

*Manual testing procedure:*

This is best done with a docker env because you need to generate a network of gateways.

1. Start 3 gateways against a dashboard and configure an API that uses token auth
2. Set up a gateway in CE mode and create an open API that round robins to the 3 other gateways, we are basically forcing distributed access via the CE gateway
3. Create a token with a quota inthe distributed nodes (dashboard)
4. Send traffic to the CE gateway APi that is round-robinning to the 3 nodes
5. Watch the returned headers of the response, there will be an X-Ratelimit-Remaining header that should count down
6. Ensure that an error is returned once you reach the quota

Note: The countdown is only partially accurate due to the distributed persistence, *however* the quota *will* be enforced accurately.

Further testing should then be the above procedure but with the following changes:

1. Start the cluster, then stop all instances mid-test, then start again
2. Start the cluster, kill one of the 3 nodes and then start it again while sending traffic (the requests will fail when the request hits a dead server, but that doesn;t matter, the quota should count down and continue to be enforced when it remains down AND when it is brought back online).


dq.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	DQFlusherPool.Open()
+
+	if err := QuotaHandler.Start(); err != nil {
+		log.Fatal(err)
+	}
+
+	// Give us time to catch up
+	time.Sleep(time.Millisecond * 100)
```
mvdan: Why is this necessary? If we're waiting for something to be ready, I'd rather use a proper synchronization method.

dq.go
```diff
@@ -0,0 +1,212 @@
...
+	QuotaHandler.InitQuota(int(currentSession.QuotaMax),
+		used,
+		key,
+		md)
+
+	if QuotaHandler.IncrBy(key, 1) == dq.Quota_violated {
+		return true
+	}
+
+	return false
```
mvdan: `if foo { return true }; return false` -> `return foo`

session_state.go
```diff
@@ -71,6 +72,20 @@ type SessionState struct {
 
 var murmurHasher = murmur3.New32()
 
+func (s *SessionState) IsExpired() bool {
+	if time.Now().After(time.Unix(s.Expires, 0)) {
+		return true
+	}
+	return false
```
mvdan: `if foo { return true }; return false` -> `return foo`

session_state.go
```diff
@@ -71,6 +72,20 @@ type SessionState struct {
...
+}
+
+func (s *SessionState) IsQuotaExpired() bool {
+	if time.Now().After(time.Unix(s.QuotaRenews, 0)) {
+		return true
+	}
+	return false
```
mvdan: `if foo { return true }; return false` -> `return foo`

middleware_rate_limiting.go
```diff
@@ -76,32 +77,54 @@ func (k *RateLimitAndQuotaCheck) handleQuotaFailure(r *http.Request, authHeaderV
...
 		!k.Spec.DisableQuota)
 
-	// If either are disabled, save the write roundtrip
-	if !k.Spec.DisableRateLimit || !k.Spec.DisableQuota {
-		// Ensure quota and rate data for this session are recorded
-		if !config.UseAsyncSessionWrite {
-			k.Spec.SessionManager.UpdateSession(authHeaderValue, sessionState, GetLifetime(k.Spec, &sessionState))
-			context.Set(r, SessionData, sessionState)
+	// If either are disabled, or the session has been modified, save the write roundtrip
+	if k.Spec.DisableRateLimit == false || k.Spec.DisableQuota == false {
```
mvdan: Why did we go from `!foo` to `foo == false` here?

mvdan: And in a few more places, actually.

middleware_rate_limiting.go
```diff
@@ -76,32 +77,54 @@ func (k *RateLimitAndQuotaCheck) handleQuotaFailure(r *http.Request, authHeaderV
...
+		// If we are using the distributed counter, we don;t need to write data here,
+		// the flusher will take care of it
+		if config.UseDistributedQuotaCounter == false {
+			// Ensure quota and rate data for this session are recorded
+			k.doSessionWrite(r, authHeaderValue, &sessionState)
 		} else {
-			go k.Spec.SessionManager.UpdateSession(authHeaderValue, sessionState, GetLifetime(k.Spec, &sessionState))
-			go context.Set(r, SessionData, sessionState)
+			// For DRL Users, only write if there's been a change in the quota renewal
+			if sessionState.QuotaRenews != oldQuotaRenewal {
```
mvdan: Can be merged into the line above, i.e. `else if ... {`

main.go
```diff
@@ -1246,8 +1246,28 @@ func startDRL() {
...
 	startRateLimitNotifications()
+
+	if config.UseDistributedQuotaCounter {
+		// Start the distributed quota system
+		StartDQ(DecideLeaderMechanism())
+	}
+}
+
+// In case we want to use a channel or some other leadership checker
+func DecideLeaderMechanism() GetLeaderStatusFunc {
```
mvdan: unexport?

dq.go
```diff
@@ -0,0 +1,212 @@
...
+				}
+			}
+		})
+	}
+
+	return nil
+
+}
+
+func StartDQ(statusFunc GetLeaderStatusFunc) {
```
mvdan: unexport?

dq.go
```diff
@@ -0,0 +1,212 @@
...
+		if len(config.DBAppConfOptions.Tags) > 0 {
+			tags := strings.Join(config.DBAppConfOptions.Tags, ".")
+			topic = topic + "." + tags
+		}
+	}
+
+	return topic
+}
+
+func DQErrorHandler(e error) {
```
mvdan: unexport?

dq.go
```diff
@@ -0,0 +1,212 @@
...
+	log.WithFields(logrus.Fields{
+		"prefix": "main.DQ",
+	}).Error(e)
+}
+
+var dummyAPISpec APISpec = APISpec{APIDefinition: &apidef.APIDefinition{SessionLifetime: 0}}
+
+func DQFlusher(d map[string]*dq.Quota) error {
```
mvdan: unexport?

.travis.yml
```diff
@@ -25,6 +25,8 @@ services:
   - redis-server
 
 install:
+  - go get github.com/TykTechnologies/dq
+  - go get github.com/TykTechnologies/tyk-cluster-framework/...
```
lonelycode: This will be vendored when that library is stable, they are currently codependent, one informs the utility of the other somewhat, for now I didn't want to introduce any vendoring until I'm happy with the lib across Gateway, Dashboard and Pump

dq.go
```diff
@@ -0,0 +1,212 @@
...
+
+			// We will track all the session handlers for this key
+			processedSpecs := map[SessionHandler]struct{}{}
+
+			// Let's go through all the API IDs in the metadata so we capture all the handlers
+			apis := v.Meta.(map[string]interface{})["Apis"]
+
+			// I hate this
+			var exp int64
+			switch v.Meta.(map[string]interface{})["QuotaRenewal"].(type) {
```
lonelycode: Never seen this syntax, so not sure what you mean.

dq.go
lonelycode: Not sure what you mean - feel free to change.

middleware_rate_limiting.go
lonelycode: I prefer to be explicit, makes it easier to read.

lonelycode:
This feature encompasses multiple projects and a multi-phase swap out of the pub/sub and storage mechanism... there's no way these PRs will be small...

dq.go
mvdan: I mean this sleep is either useless or hiding a concurrency bug, like in our old hot reload code. What does it do?

dq.go
mvdan: Instead of

```
switch v.(type) {
case string:
    _ = v.(string)
```

do

```
switch x := v.(type) {
case string:
    _ = x
```

.travis.yml
mvdan: SGTM, then if this gets merged before 2.4 we need to remember to pin/vendor before the release

mvdan:
Did you try rebasing? Somehow the initial commits have been duplicated, and this PR also includes commits from master (no idea how that could have happened).

Otherwise, perhaps squash into a single commit?

lonelycode:
> Did you try rebasing? 

Endless conflicts

> Otherwise, perhaps squash into a single commit?

Seems to also cause endless conflicts

dq.go
lonelycode: IIRC it makes sure that the flusher and the CDRT notifier has time to initialise and do it's thing. There are two systems working in tandem here, one is a batch writer and the other is a distributed in-memory merge, the two are eventually consistent.

This function is only called at startup, so this pause is ok.

lonelycode:
Wil squash and merge when the time comes. Spending more time pruning GH than actually getting meaningful work done.

lonelycode:
@ilijabojanovic Can you run a manual test of this PR please - need a seocnd set of eyes on it to make sure it functions robustly.

lonelycode:
Closing this in favor of #696

PR #675, opened 2017-04-19
config: fatal on config parse errors, try multiple files, add tests

### LINKED ISSUE #501: Should recognize ./tyk.conf by default if configuration option not specified
bug
By default, it searches configuration at `/etc/tyk/tyk.conf` location, and if it is not found, it sets default one, and writes it to `./tyk.conf`. If there is already `tyk.conf` in current directory, it will overwrite it :(

If there is local tyk.conf, it should not override it, and should actually use it.

Create local `tyk.conf` with some custom configuration. Run `tyk` without `--conf` option. `tyk.conf` will be overridden with default config.

All

The proper fix will be searching tyk.conf in multiple locations.

mvdan:
Split in a few commits to help code review.

For now, we only read `tyk.conf` before `/etc/tyk/tyk.conf`. More paths can be easily added later, and the precedence can also be easily modified.

Fixes #501.

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -130,7 +130,7 @@ func TestMain(m *testing.M) {
 	go func() {
 		panic(testServer.ListenAndServe())
 	}()
-	writeDefaultConf(&config)
+	writeDefaultConf("", &config)
 	config.Storage.Database = 1
 	if err := emptyRedis(); err != nil {
 		panic(err)
@@ -146,6 +146,7 @@ func TestMain(m *testing.M) {
 	config.EnableJSVM = true
 	config.Monitor.EnableTriggerMonitors = true
 	config.AnalyticsConfig.NormaliseUrls.Enabled = true
+	afterConfSetup(&config)
 	initialiseSystem(nil)
 	if analytics.GeoIPDB == nil {
 		panic("GeoIPDB was not initialized")
--- a/config.go
+++ b/config.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"strings"
@@ -277,7 +278,7 @@ func writeDefaultConf(conf *Config) {
 // LoadConfig will load the configuration file from filePath, if it can't open
 // the file for reading, it assumes there is no configuration file and will try to create
 // one on the default path (tyk.conf in the local directory)
-func loadConfig(filePath string, conf *Config) {
+func loadConfig(filePath string, conf *Config) error {
 	configuration, err := ioutil.ReadFile(filePath)
 	if err != nil {
 		if !runningTests {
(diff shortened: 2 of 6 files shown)


gateway_test.go
```diff
@@ -129,6 +129,11 @@ func TestMain(m *testing.M) {
 	config.EnableJSVM = true
 	config.Monitor.EnableTriggerMonitors = true
 	config.AnalyticsConfig.NormaliseUrls.Enabled = true
+	config.SlaveOptions.CallTimeout = 30
+	config.SlaveOptions.PingTimeout = 60
+	GlobalRPCPingTimeout = time.Second * time.Duration(config.SlaveOptions.PingTimeout)
+	GlobalRPCCallTimeout = time.Second * time.Duration(config.SlaveOptions.CallTimeout)
```
buger: Why you added lines above?

buger:
@ilijabojanovic ready for testing. 

The Proper test will be comparing default configs, generated by tyk, before and after this PR, in order to ensure that they are identical. To make this work, delete local tyk.conf file, and /etc/tyk/tyk.conf files, and run tyk without --conf option. 

And in second, create local `tyk.conf` and ensure that tyk will use it and not override if you run it without `--conf` option. Next remove local `tyk.conf` and create `/etc/tyk/tyk.conf`, and ensure that tyk will use this path, if local one not found. 

And in third, ensure that `--conf` option works with custom paths. 

👌

gateway_test.go
mvdan: Because the tests would fail without them - it's explained in its commit message. To simplify the logic of what runs and what doesn't regarding the config code in `go test`, now `loadConfig` isn't ran at all. And the last thing it does is set these up.

mvdan:
Most of the scenarios that @buger explained (except `--conf`, that's more like an integration test) are already covered by the unit tests, so hopefully it should all be OK.

gateway_test.go
buger: I'll be frank, this change does not look that good, especially we if we change config options, we will have to modify them there too. 

What about moving default config into a variable outside `writeDefaultConf` function: 
```
var defaultConf = Config{
 		ListenPort:     8080,
                ....
}
```

And then use this variable in `writeDefaultConf` and test initialization: `config = defaultConf`.

mvdan: I agree that it's not ideal, but changing how the default config won't work. Note that these settings are applied *after* the config file is read, not before. Perhaps a method like `ensureConfigSanity` to do these last-step tweaks in both prod and test.

mvdan: ^ to clarify, what I mean is that I can do these changes on this PR if you want.

buger: Yeah, lets try to do it.

mvdan: done, PTAL

mvdan:
Rebased after a few dozen commits and some minor conflicts.

mvdan:
Okay, I give up on coveralls actually doing its job.

ilijabojanovic:
Verified and approved for merge @buger

PR #679, opened 2017-04-19
Reduce number of Go files, cleanup some code

mvdan:
Updates #334.


middleware.go
```diff
@@ -147,9 +147,7 @@ func GetResponseProcessorByName(name string) (TykResponseHandler, error) {
 
 }
 
-type ResponseChain struct{}
-
-func (r ResponseChain) Go(chain []TykResponseHandler, rw http.ResponseWriter, res *http.Response, req *http.Request, ses *SessionState) error {
+func handleChain(chain []TykResponseHandler, rw http.ResponseWriter, res *http.Response, req *http.Request, ses *SessionState) error {
```
buger: Can you call it `handleResponseChain` to keep the same meaning as original one?

mvdan: done and rebased

PR #682, opened 2017-04-20
main: remove zero conf wait tail call

mvdan:
This function could recurse into itself forever, leading to increasing
memory usage or a panic. Use a simple for instead.

Also get rid of the helper func, since it's just three lines and
self-explanatory anyway.

### DIFF
--- a/main.go
+++ b/main.go
@@ -193,13 +193,6 @@ func setupGlobals() {
 
 }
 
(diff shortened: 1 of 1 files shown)


mvdan:
Also, this code was very probably harmless but could still be a worthy candidate for 2.3.x.

PR #684, opened 2017-04-21
Replace lonelycode/go-uuid with satori/go.uuid

### LINKED ISSUE #524: Swap uuid library
We use https://github.com/nu7hatch/gouuid. Last commit more than 3 years ago, lacking many useful features like `UUID.Bytes()`.

Let's use https://github.com/satori/go.uuid instead. Actually maintained, much more popular and with many more features, including the method above.

We can still use UUID v4 as specified by RFC 4122, so there should be no changes in behaviour. Some code would get simpler too.

Can do this myself if there are no objections. /cc @buger @lonelycode

mvdan:
Another outdated uuid library. Unify with the main one we use.

go-uuid's NewUUID actually returns v1 instead of v4, but that was simply
a misunderstanding - when the library was initially added, the intention
was to use v4 like in the rest of the codebase.

The old String method behaves just like the new one. The format and
length of the strings is the same.

Fixes #524.

### DIFF
--- a/command_mode.go
+++ b/command_mode.go
@@ -6,7 +6,7 @@ import (
 	"io/ioutil"
 	"strings"
 
-	"github.com/lonelycode/go-uuid/uuid"
+	"github.com/satori/go.uuid"
 
 	"github.com/TykTechnologies/tyk/apidef"
 )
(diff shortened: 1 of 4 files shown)


mvdan:
Re v1 vs v4, I asked Martin about this a couple of weeks ago and that's what he said. So he approves of the switch.

I can't remove the vendored lib yet because another outdated, vendored lib depends on it. Will do it later.

buger:
@mvdan what about tyk-analytics, does it use uuid1 or uuid4? Is there any possible conflicnts you can imagine?

mvdan:
@buger analytics uses v4 only directly, so if anything this should help with consistency.

The differences with each UUID version is not in the format, length or anything else. It's just in the sources of randomness. So there should be no conflicts. Almost everyone uses v4, as that's the only one that's really just a random number.

See https://en.wikipedia.org/wiki/Universally_unique_identifier#Versions

PR #687, opened 2017-04-21
Make C files in the root package symlinks again

mvdan:
These used to be links but were unintentionally replaced by full copies
in 15888757.

There seems to be no way around this symlink, as CGO will only compile C
files that are in the same directory as the package it's compiling.


mvdan:
Ah, tested coprocess-lua but forgot about coprocess-python :(

mvdan:
Also, the fuckup probably happened on a platform where symlinks are not a thing, like Windows. We should make sure it doesn't happen again.

buger:
@matiasinsaurralde up to you

mvdan:
bump @matiasinsaurralde

matiasinsaurralde:
I will be checking this today!

mvdan:
Thanks!

PR #695, opened 2017-04-26
jsvm: kill copies after a timeout (goroutine leak) plus test

mvdan:
To show progress until it's ready, and in case anyone wants to start reviewing any of it.

The actual fix isn't part of this yet, so CI should be red.


mvdan:
Also, for this first fix I'll hard-code the timeout to something like `5s`. If we want to make it configurable, we should do that later to not drag the extra code and design into the bugfix backport.

mvdan:
Also, the coverage bump is because all this js stuff had near zero coverage. To have another nice bump, we should add a test that modifies the response somehow (a header, etc) and checks it did work. But that's for another PR.

PR #696, opened 2017-04-26
Squashed distributed quota branch

lonelycode:
(This is a squashed version of #674, which is closed)

Ok this is a complicated one.

Distributed Quota feature (lets call it a beta): instead of having the quota enforced by a shared counter in redis, the counter is in-memory and counters are shared across a message bus to other cluster gateways, then merged, to create a distributed total. This means no centralised state and no one-to-one linkage with Redis.

**What we are trying to achieve:**

1. Stop having a one-to-one relationship with redis, when a counter-based quota is enforced, every inbound request means a roundtrip to redis, this is expensive and adds latency, even if done async it floods the redis connection counter pool and means performance degrades as traffic increases
2. Make counter operations faster: Since these are in-memory and are merged in the background, the counter is entirely in-memory and no round-trip or lookup is required. This should improve throughput for metered connections
3. The new DQ Quota object means that actually we can introduce new metering metrics because it is an abstraction
4. Managed redis writes (or any persistence), since the distributed cluster, with a single leader, will only write to the store every `x` interval (default to 3 seconds),  so no matter how high the load, the writes speed isn't a linear relation to throughput.

To enable set the following in `tyk.conf`:

- `use_distributed_counter` to `true`: This enables the feature 
- `distributed_quota_set_master` to `true`: Since we only have redis as a pub/sub mechanism at the moment, we need all nodes to be masters, this causes all nodes to persist the CRDT list and is therefore fault tolerant. This is optional, obviously, if you can be completely sure that your master will not fail - so long as one node is set as master we are clear.

It will work in single-host mode just like a regular quota.

There is a unit test, but multi-host testing was done manually, as well a failure conditions:

*Manual testing procedure:*

This is best done with a docker env because you need to generate a network of gateways.

1. Start 3 gateways against a dashboard and configure an API that uses token auth
2. Set up a gateway in CE mode and create an open API that round robins to the 3 other gateways, we are basically forcing distributed access via the CE gateway
3. Create a token with a quota inthe distributed nodes (dashboard)
4. Send traffic to the CE gateway APi that is round-robinning to the 3 nodes
5. Watch the returned headers of the response, there will be an X-Ratelimit-Remaining header that should count down
6. Ensure that an error is returned once you reach the quota

Note: The countdown is only partially accurate due to the distributed persistence, *however* the quota *will* be enforced accurately.

Further testing should then be the above procedure but with the following changes:

1. Start the cluster, then stop all instances mid-test, then start again
2. Start the cluster, kill one of the 3 nodes and then start it again while sending traffic (the requests will fail when the request hits a dead server, but that doesn;t matter, the quota should count down and continue to be enforced when it remains down AND when it is brought back online).

### DIFF
--- a/dq.go
+++ b/dq.go
@@ -1,0 +1,210 @@
+package main
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/TykTechnologies/tyk-cluster-framework/client"
+	"github.com/TykTechnologies/tyk-cluster-framework/encoding"
+
+	"github.com/jeffail/tunny"
+
+	"github.com/Sirupsen/logrus"
+
+	"github.com/TykTechnologies/dq"
+	"github.com/TykTechnologies/tyk/apidef"
+)
+
+var DQFlusherPool *tunny.WorkPool = tunny.CreatePoolGeneric(10)
+var QuotaHandler *dq.DistributedQuota
+
+type GetLeaderStatusFunc func() bool
+
+func getDQTopic() string {
+	topic := "tyk.dq"
+	if config.DBAppConfOptions.NodeIsSegmented {
+		if len(config.DBAppConfOptions.Tags) > 0 {
+			tags := strings.Join(config.DBAppConfOptions.Tags, ".")
+			topic += "." + tags
+
+		}
+	}
+
+	return topic
+}
+
+func dqErrorHandler(e error) {
+	log.WithFields(logrus.Fields{
+		"prefix": "main.DQ",
+	}).Error(e)
+}
+
+var dummyAPISpec = APISpec{APIDefinition: &apidef.APIDefinition{SessionLifetime: 0}}
+
+func dqFlusher(d map[string]*dq.Quota) error {
+	for k, v := range d {
+		DQFlusherPool.SendWork(func() {
+			// Ignore deleted flags
+			if v.Delete {
+				log.Warning("Key has been tagged for deletion, not writing counter")
+				return
+			}
+
+			// We will track all the session handlers for this key
+			processedSpecs := map[SessionHandler]struct{}{}
+
+			// Let's go through all the API IDs in the metadata so we capture all the handlers
+			apis := v.Meta.(map[string]interface{})["Apis"]
+
+			// I hate this
+			var exp int64
+			switch x := v.Meta.(map[string]interface{})["QuotaRenewal"].(type) {
+			case int64:
+				exp = x
+			case float64:
+				exp = int64(x)
+			}
+
+			expT := time.Unix(exp, 0)
+			for _, aid := range apis.([]interface{}) {
+				apiID := aid.(string)
+
+				// This will grab the session handler
+				spec := GetSpecForApi(apiID)
+				if spec == nil {
+					log.Warning("Can't find back-end for this API, skipping")
+					break
+				}
+
+				// Have we processed on this handler before (many APIs may use the same handler)?
+				_, processedOnSH := processedSpecs[spec.SessionManager]
+				if processedOnSH {
+					continue
+				}
+
+				// This handler hasn't been used yet for the API
+				// Get the session data
+				s, f := spec.SessionManager.GetSessionDetail(k)
+
+				// If it was found, lets process it for this handler
+				if !f {
+					// No longer in session store, delete
+					QuotaHandler.TagDelete(k)
+					continue
+				}
+
+				if !f || time.Now().After(expT) || s.IsExpired() {
+					QuotaHandler.TagDelete(k)
+					continue
+				}
+
+				qr := int64(v.Max - v.Counter.Count())
+				if qr < 0 {
+					qr = 0
+				}
+
+				// Only write on count difference
+				if qr != s.QuotaRemaining {
+					s.QuotaRemaining = qr
+					spec.SessionManager.UpdateSession(k, &s, getLifetime(&dummyAPISpec, &s))
+					// Since we've written the token, we don't need to re-do it at the end of the middleware chain
+					s.SetFirstSeenHash()
+					// We've performed a write on this SH now, lets tag that so we don't do it again
+					processedSpecs[spec.SessionManager] = struct{}{}
+				}
+
+			}
+		})
+	}
+
+	return nil
+
+}
+
+func startDQ(statusFunc GetLeaderStatusFunc) {
+	log.WithFields(logrus.Fields{
+		"prefix": "DQuota",
+	}).Info("Using Distributed Quota")
+	p := strconv.Itoa(config.Storage.Port)
+	cs := fmt.Sprintf("redis://%v:%v", config.Storage.Host, p)
+	c1, err := client.NewClient(cs, encoding.JSON)
+	if err != nil {
+		log.WithFields(logrus.Fields{
+			"prefix": "DQuota",
+		}).Error("Failed to create distributed quota client: ", err)
+	}
+
+	QuotaHandler = dq.NewDQ(dqFlusher, dqErrorHandler, NodeID)
+	broadcastTimer := time.Millisecond * 100
+	QuotaHandler.BroadcastWith(c1, broadcastTimer, getDQTopic())
+
+	// We always need a leader because otherwise we can;t persist data
+	QuotaHandler.SetLeader(statusFunc())
+
+	QuotaHandler.FlushInterval = time.Second * 3
+	if config.DistributedQuotaFlushIntervalInMS != 0 {
+		QuotaHandler.FlushInterval = time.Millisecond * time.Duration(config.DistributedQuotaFlushIntervalInMS)
+	}
+
+	DQFlusherPool.Open()
+
+	if err := QuotaHandler.Start(); err != nil {
+		log.Fatal(err)
+	}
+
+	// Give us time to catch with the cluster
+	time.Sleep(broadcastTimer)
+}
+
+func (l SessionLimiter) IsDistributedQuotaExceeded(currentSession *SessionState, key string) bool {
+
+	// Are they unlimited?
+	if currentSession.QuotaMax == -1 {
+		// No quota set
+		return false
+	}
+
+	// store the old expiry so we propagate the right value
+	md := map[string]interface{}{
+		"QuotaRenewal": currentSession.QuotaRenews,
+	}
+
+	// Handle renewal
+	RenewalDate := time.Unix(currentSession.QuotaRenews, 0)
+	if time.Now().After(RenewalDate) {
+		// The renewal date is in the past, we should update the quota!
+		// Also, this fixes legacy issues where there is no TTL on quota buckets
+		log.Warning("Incorrect key expiry setting detected, correcting / or renewing quota")
+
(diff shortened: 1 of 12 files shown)


mvdan:
You probably want to swap the author of the commit to be you instead of me :) I forgot to do that.

lonelycode:
> You probably want to swap the author of the commit to be you instead of me :) I forgot to do that.

Does it really matter?

mvdan:
Not really, but don't want to be confused later on. Just fixed it, will do another code review pass.

session_state.go
```diff
@@ -71,6 +72,17 @@ type SessionState struct {
 
 var murmurHasher = murmur3.New32()
 
+func (s *SessionState) IsExpired() bool {
+	if time.Now().After(time.Unix(s.Expires, 0)) {
+		return true
+	}
+	return false
```
mvdan: `return time.Now...`

middleware_rate_limiting.go
```diff
@@ -76,32 +76,52 @@ func (k *RateLimitAndQuotaCheck) handleQuotaFailure(r *http.Request, authHeaderV
...
-			go context.Set(r, SessionData, sessionState)
+	// If either are disabled, or the session has been modified, save the write roundtrip
+	if k.Spec.DisableRateLimit == false || k.Spec.DisableQuota == false {
+		// If we are using the distributed counter, we don;t need to write data here,
+		// the flusher will take care of it
+		if config.UseDistributedQuotaCounter == false {
+			// Ensure quota and rate data for this session are recorded
+			k.doSessionWrite(r, authHeaderValue, &sessionState)
+		} else if sessionState.QuotaRenews != oldQuotaRenewal {
+			k.doSessionWrite(r, authHeaderValue, &sessionState)
```
mvdan: both branches do the same? Perhaps collapse them

dq.go
```diff
@@ -0,0 +1,212 @@
...
+
+}
+
+func startDQ(statusFunc GetLeaderStatusFunc) {
+	log.WithFields(logrus.Fields{
+		"prefix": "DQuota",
+	}).Info("Using Distributed Quota")
+	p := strconv.Itoa(config.Storage.Port)
+	cs := fmt.Sprintf("redis://%v:%v", config.Storage.Host, p)
+	c1, _ := client.NewClient(cs, encoding.JSON)
```
mvdan: we ignore the error here?

dq.go
```diff
@@ -0,0 +1,212 @@
...
+					skip = true
+				}
+
+				if s.IsExpired() {
+					// Remove expired data too
+					QuotaHandler.TagDelete(k)
+					skip = true
+				}
+
+				if !skip {
```
mvdan: You can get rid of this `if` and just use `continue` in all the `skip = true` lines.

dq.go
```diff
@@ -0,0 +1,212 @@
...
+			}
+
+			expT := time.Unix(exp, 0)
+			for _, aid := range apis.([]interface{}) {
+				apiID := aid.(string)
+
+				// This will grab the session handler
+				spec := GetSpecForApi(apiID)
+				if spec == nil {
+					log.Warning("Can't find back-end for this API, skippingc")
```
mvdan: typo in `skipping`

dq.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	DQFlusherPool.Open()
+
+	if err := QuotaHandler.Start(); err != nil {
+		log.Fatal(err)
+	}
+
+	// Give us time to catch up
+	time.Sleep(time.Millisecond * 100)
```
mvdan: Continuing discussion from the original PR:

> IIRC it makes sure that the flusher and the CDRT notifier has time to initialise and do it's thing. There are two systems working in tandem here, one is a batch writer and the other is a distributed in-memory merge, the two are eventually consistent.
> This function is only called at startup, so this pause is ok.

I realise that it's not hugely important as far as start-up time goes, but this can be racy and bite us in the future really easily. Is there no way to wait for exactly what we're after, not just an arbitrary amount of time?

session_state.go
lonelycode: Done

middleware_rate_limiting.go
lonelycode: Going to leave this, it's tempting to merge them, but they deal with different configuration cases.

dq.go
lonelycode: Fixed

dq.go
lonelycode: Done

dq.go
lonelycode: Added a log

dq.go
lonelycode: > I realise that it's not hugely important as far as start-up time goes, but this can be racy and bite us in the future really easily. Is there no way to wait for exactly what we're after, not just an arbitrary amount of time?

Have made this directly linked to the broadcast interval, so the node waits for at least 1 CRDT to arrive to fill it's cache.

middleware_rate_limiting.go
mvdan: ack

dq.go
```diff
@@ -0,0 +1,215 @@
...
+var QuotaHandler *dq.DistributedQuota
+
+type GetLeaderStatusFunc func() bool
+
+func getDQTopic() string {
+	topic := "tyk.dq"
+	if config.DBAppConfOptions.NodeIsSegmented {
+		if len(config.DBAppConfOptions.Tags) > 0 {
+			tags := strings.Join(config.DBAppConfOptions.Tags, ".")
+			topic = topic + "." + tags
```
mvdan: nit: `topic += ...`

dq.go
```diff
@@ -0,0 +1,215 @@
...
+				return
+			}
+
+			// We will track all the session handlers for this key
+			processedSpecs := map[SessionHandler]struct{}{}
+
+			// Let's go through all the API IDs in the metadata so we capture all the handlers
+			apis := v.Meta.(map[string]interface{})["Apis"]
+
+			// I hate this
```
mvdan: I take it we should enforce QuotaRenewal to always be int64, and somewhere in the code we do float64 instead by mistake?

dq.go
```diff
@@ -0,0 +1,215 @@
...
+				if time.Now().After(expT) {
+					QuotaHandler.TagDelete(k)
+					continue
+				}
+
+				if s.IsExpired() {
+					// Remove expired data too
+					QuotaHandler.TagDelete(k)
+					continue
+				}
```
mvdan: You could just do:

```
if !f || time.Now().After(expT) || s.IsExpired() {
    QuotaHandler.TagDelete(k)
    continue
}
```

gateway_test.go
```diff
@@ -750,6 +750,50 @@ func TestQuota(t *testing.T) {
...
+	param := make(url.Values)
+	req, err := http.NewRequest("GET", param.Encode(), nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	req.Header.Add("authorization", keyId)
+
+	config.DQSetMaster = true
+	config.UseDistributedQuotaCounter = true
+	config.DistributedQuotaFlushIntervalInMS = 100
```
mvdan: these are not reset, so they modify the global config for the rest of the tests. You should either set them globally, or set back to the original values in this test via a `defer`.

gateway_test.go
```diff
@@ -750,6 +750,50 @@ func TestQuota(t *testing.T) {
...
+func TestDistributedQuotaSingleNode(t *testing.T) {
+	spec := createSpecTest(t, nonExpiringDefNoWhiteList)
+	session := createQuotaSession()
+	keyId := testKey(t, "key")
+	spec.SessionManager.UpdateSession(keyId, session, 60)
+	defer spec.SessionManager.ResetQuota(keyId, session)
+
+	recorder := httptest.NewRecorder()
+	param := make(url.Values)
+	req, err := http.NewRequest("GET", param.Encode(), nil)
```
mvdan: What does `param` do here, exactly?

dq.go
lonelycode: It's actually a deserialisation issue - this loop handles both objects from the local node and those that come in from the broker. For some reason it deserialises to a float when it should be an int.

gateway_test.go
lonelycode: Nothing probably - copypasta, will fix.

dq.go
mvdan: Oh, this is likely JSON then. It should consistently be `float64` here, so you should be able to use a one-liner.

Also see https://golang.org/pkg/encoding/json/#Decoder.UseNumber

mvdan: (to clarify - this is because in the JSON spec all numbers are float64, and if Go is just given a map instead of a struct, it defaults to float64 for compatibility)

lonelycode:
> So I understand you enable UseDistributedQuotaCounter on all gateways and DQSetMaster in exactly one, and it just works? What if you set two to be master? What if you set none?

If you set none to be master quota doesn't work. 

If you set more than one it doesn't matter, the stored value isn't used except initialising a fresh node that has never seen the token and reporting. More than one will also cause more writes to your data store which we are trying to avoid, but if using redis, more than one should be master for HA.

However, with a raft based k/v there is automatic leader election and detection, and in the Tcf framework it's plug and play as part of the store package, so master can be dynamically adjusted on node failure once integrated.

> I'm still concerned about the initial sleep. What would happen if we removed it? What would happen if the network was slow and nothing happened in those 100ms?

For a single node it is irrelevant, it matters for new nodes joining a cluster as part of a scaling event, they need enough time to gather CRDT data from other nodes to initialise their own counters so they can accurately enforce a quota without resorting to the DB (because the stored value can be out of date by whatever the flush interval is).

I tried removing it, if the node gets traffic immediately the you run the risk of accidentally resetting the quota for the user. We could double it, or add a channel to wait for data, but if this is the first node of a cluster, it will block. So we'd end up with a  wait anyway. 

However, if the network is slow, that's ok, the worst case scenario is that a quota is not enforced correctly, which isn't terrible, just not ideal.

dq.go
lonelycode: It gets trickier, this deserialisation is handled at the framework level in `tyk-cluster-framework`, since the deserialisation is handled there, I'm not sure I want to force using numbers when the same framework sends floats for things like DRL notifications. Would need to expose the `UseNumber()` somehow on the `DecodeMessage`, but that's pretty specific.

Could eventually refactor to offer a raw message so that decode can be exported, but defeats purpose of a managed payload interface and structure.

gateway_test.go
lonelycode: Have added a `defer()`

lonelycode:
Aaaaand for some reason the ***** commits are duplicated... gah! I give up.

mvdan:
Back to one commit, squashed and rebased :)

lonelycode:
@ilijabojanovic over to you...

mvdan:
I broke mergeability of this with a merged PR of mine, so I've rebased and fixed conflicts.

mvdan:
More conflicts, this time in config.go - fixed.

mvdan:
Rebased once more.

buger:
Will close it but keep branch for reference

Issue #683, opened 2017-04-21
Replace gorilla/context with std's context


mvdan:
`gorilla/context` appeared before Go's `context` appeared in 1.7. Since we only support 1.7 and 1.8, we can now do the jump.

That should remove an external dependency (which is already semi-abandoned because of this) and should simplify our code, as `net/http` already has methods like `WithContext`.

Not pressing, but would be nice to do.

buger:
If we cover it with tests I'm ok with it :)

mvdan:
We depend quite heavily on either context package for stuff like session info, so quite a lot of things would break if I screw it up. Should be easy to test.

mvdan:
Unfortunately, this is impossible right now because of how we're doing HTTP middlewares right now.

`gorilla/context` and `context` work a bit differently in how the former uses global storage, and the latter uses local storage. In Gorilla, you do:

```
context.Set(r, Key, val)
[...]
val := context.Get(r, Key)
[...]
context.Clear(r)
```

Note that the `Clear` line must exist, otherwise the globally stored context will be a memory leak.

On the other hand, `std`'s context is like:

```
r = r.WithContext(context.WithValue(r.Context(), Key, val)))
[...]
val := r.Context().Value(Key)
```

Note that we no longer have to clean up globals. However, `WithContext` returns a modified `*http.Request`, one with a new layer of `context.Context` with what we need.

Unfortunately, the way we do middlewares right now, `*http.Request` isn't passed from one middleware to the next. So it's impossible to have this locally-stored context instead of a global one, unless we first refactor middlewares.

To summarise, our middlewares right now are in the form of:

```
func (m *SomeMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Request) (error, int) {
    // do work
    return nil, 200
}
```

When they should be:

```
func (m *SomeMiddleware) ServeHTTP(w http.ResponseWriter, r *http.Request) {
    // do work
    m.Next.ServeHTTP(w, r)
}
```

Note that the second version *can* pass a modified `*http.Request` along very easily. This is the very structure of middlewares that `pressly/chi` uses. A very simple example:

https://github.com/pressly/chi/blob/35ec20b432d9c9d57c5da99623ebffd172433da5/middleware/request_id.go#L63

mvdan:
Pushing back a milestone since this is blocked by other refactors. @buger this is a big point in favor of moving to something like `chi`, in my opinion. Not pressing, but should happen at some point.

[...]

PR #698, opened 2017-04-27
all: get rid of context.GetOk usage

mvdan:
std's context has no "value, ok", version of a getter, so to transition
to it we first need to rewrite these into the simpler "value" return.

This should be a safe change since we only ever set non-nil values on
any context.

Updates #683.


mvdan:
Also, I didn't add tests here on purpose - I can't test all of this because none of these funcs are self-contained or small enough to just test the context values easily.

I'll add proper tests once I factor out the session stuff into separate funcs (e.g. to deal with SessionData uniformly) but I can't do that as long as we mix `Get` and `GetOk`.

PR #702, opened 2017-04-28
Enable the gateway to use a dashboard (or cluster) as a pub/sub server instead of redis

lonelycode:
To enable the pub/sub client, set `"pubsub_master_connectionstring": "mangos://tyk-dashboard:9001"` in the `tyk.conf file, make sure the dashboard or dashboards are running. The hostnmae can be a single host or a load balancer.

This sits on top of the dq branch.


lonelycode:
@mvdan might need your help to squash this thing down.

mvdan:
We should do that once the dq branch is merged, as this one seems to depend on that.

lonelycode:
Yep it's based off the original dq cherry pick which is why it's funky. I just wanted a PR in place so that I can start on the k/v piece.

mvdan:
Sure - let's get the dq one merged first. I'll hold off reviewing this one for now.

mvdan:
This one is getting a bit out of control with the conflicts, so I'll do a squash+rebase on monday. @lonelycode do you want to keep any history? Can I do a single commit like the dq branch?

lonelycode:
No go for it - squash away

buger:
Will close it but keep branch for reference

PR #706, opened 2017-04-28
api: list all addresses in user_ip log again

mvdan:
I unintentionally broke this when cleaning up and fixing our GetIPFromRequest code. Only in master though, so we're good.

Commits contain more info.


mvdan:
Note: Part of my screw-up is because of the confusing name of `user_ip`. I assume we're OK with having a list in there, and addresses (e.g. ports, domains) instead of just IPs.

PR #708, opened 2017-05-02
api: rework code that filters by hostname

### LINKED ISSUE #670: Hostname overrides control_api_hostname
BUG

“hostname”: “api.tyk-test.com”,
   “enable_api_segregation”: true,
   “control_api_hostname”: “www.tyk-test.com“,

Tyk admin endpoint does not work on www.tyk-test.com/tyk/

hostname value in tyk.conf should not override control api (/tyk/)

1. Add hostname in gateway configuration file
2. Add control_api_hostname in in gateway configuration file
3. Start gateway 
4. curl http://{control_api_hostname}/tyk/apis

2.3.4 and master 2.4

mvdan:
Make the routers filter by hostname in loadApps and loadAPIEndpoints.
The former has to filter by HostName if non-empty, and the latter has to
filter by either HostName or ControlAPIHostname, the latter taking
priority if both are set.

Do it in the funcs that actually add handlers to the routers as they are
called from many places, and this way we can be sure that the hostnames
are always set up correctly and they only affect the sub-router they
should be affecting.

In the case of the Control API, we were using getHostName() which just
used HostName, ignoring ControlAPIHostname. This resulted in
ControlAPIHostname not being used properly.

Fixes #670.

### DIFF
--- a/api_loader.go
+++ b/api_loader.go
@@ -538,6 +538,13 @@ func (d *DummyProxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 // Create the individual API (app) specs based on live configurations and assign middleware
 func loadApps(apiSpecs []*APISpec, muxer *mux.Router) {
+	hostname := config.HostName
+	if hostname != "" {
+		muxer = muxer.Host(hostname).Subrouter()
+		log.WithFields(logrus.Fields{
+			"prefix": "main",
+		}).Info("API hostname set: ", hostname)
+	}
 	ListenPathMap = cmap.New()
 	// load the APi defs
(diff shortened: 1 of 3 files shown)


mvdan:
Also note that this means that all handlers should be set up in either `loadApps` or `loadAPIEndpoints`. I looked at the code and that seems to be true. The only exception is the `/debug/pprof/...` stuff, but that wasn't being filtered by hostname before either.

mvdan:
Also, @ilijabojanovic when testing this, be sure to test both before and after a reload. The code that sets up the routers is different, but hostnames should work as expected in both cases.

ilijabojanovic:
@buger approved for merge

PR #714, opened 2017-05-02
only strip first instance of listen path

drpebcak:
Changes the call to strings.Replace to only replace 1 (the first) instance of matching listen path.

This is to solve an issue where your listen path might be something like `/api/`, and your application has a route that also matches `/api/`.

For example: your application has a route like `example.com/api/endpoint`).
In tyk, you configure the listen path to be `/api/` and the target url is `http://example.com/`. The way this currently functions, if you make a request to `your-tyk.com/api/api/endpoint`, both of the instances of `api` get trimmed off, leaving you with a malformed request.

I think most people would agree, when setting the listen path, the expected behavior is for it to only match against the first instance of that phrase.

### DIFF
--- a/res_handler_header_transform.go
+++ b/res_handler_header_transform.go
@@ -62,7 +62,7 @@ func (h HeaderTransform) HandleResponse(rw http.ResponseWriter,
 			if len(h.Spec.target.Path) != 0 {
 				newHeaderValue = strings.Replace(
 					newHeaderValue, h.Spec.target.Path,
-					h.Spec.Proxy.ListenPath, -1)
+					h.Spec.Proxy.ListenPath, 1)
 			} else {
 				newHeaderValue = strings.Replace(
 					newHeaderValue, req.URL.Path,


mvdan:
Actually, while you're at it, perhaps fix the rest of the `strings.Replace(x, y, -1)` calls nearby.

matiasinsaurralde:
What about `strings.TrimPrefix(path, listenPath)`?

mvdan:
Maybe - I don't know the nature of what this code is supposed to do very well, so I kept the suggestion a bit generic :)

lonelycode:
@mvdan this code normalises a URL so that when the proxy sends the request upstream it doesn't contain the slug used to hit the gateway.

E.g. If the gateway listens on /super-api/

The target might actually be serving on root (/), so if we proxies /super-api/{id} it would 404, so we strip the section out of the URL before proxying it so that you can essentially generate a facade.

I'd say replacing the first occurrence is better than prefix, because we can't guarantee that there isn't another proxy downstream doing some of its own shenanigans, and so can't guarantee that the listen path will start at index 0.

drpebcak:
@mvdan I _could_ change the other instances of `strings.Replace()` , but I'm not sure if any of those are in scope for the issue I am trying to fix here.

I think `strings.Replace()` is still the right answer, as long as we are trimming just that first instance. To @lonelycode's point, we can't say for sure that there isn't another proxy downstream, but we can say that tyk is only going to _match_ against the first instance of listen path when it decides how to route - it makes sense then that only the first instance would be eligible for removal.

mvdan:
First match sounds good then. I still think we should be consistent in the other `Replace` calls if possible, but it's up to someone who knows the code well like @lonelycode.

buger:
It looks good, but can you confirm that this fix works for you?

I'm asking because code that actually stips the path is here: https://github.com/TykTechnologies/tyk/blob/master/handler_success.go#L327 and here https://github.com/TykTechnologies/tyk/blob/master/handler_success.go#L362. And seems like it already strips only first instance. 

Code you modified executed only if API spec have special "response_processors" field with list of available processors and their options: "header_injector", "response_body_transform", "header_transform". 

Can you confirm if you are using "response_body_transform" processor?

Thank you!

drpebcak:
Interestingly enough, I'm not using any of those things.

My api definition is super basic - https://gist.github.com/drpebcak/50b3d1e1fcbcfd88e31605528b5d9b68

With this configuration, I'm seeing logged entries from tyk that show requests being munged as they pass through tyk...
`http://my-tyk-gateway.com/sites/sites/1234` is being sent through to the backend as `http://example.com:8080/api1234`

drpebcak:
And for completeness sake, I am using version 2.3.4 of the gateway, and (if it matters) version 1.3.4 of the dashboard.

buger:
Was fixed in 2.3.7, appologies for delay!

PR #717, opened 2017-05-03
api: add "block" url parameter to /tyk/reload

### LINKED ISSUE #703: Proposal: a version of the /tyk/reload endpoint that blocks
`/tyk/reload` currently just queues a reload (if none is already queued) and always returns 200 immediately.

I suggest adding another endpoint, like `/tyk/reload-block` or whatever name we fancy, that doesn't return 200 until a reload has been finished.

If a reload was already queued, it waits for that one to finish. If none was queued, it waits for the one it queued to finish.

It would still always return 200, because reloads don't currently catch any errors. But if in the future they do, this blocking version may return other status codes.

This is especially useful for scripting and our tests. Right now, if you want to fire a reload and wait for it, you have no option but to sleep for an arbitrary amount of time. Currently, our tests do this in a very hacky way, by first doing the `GET` and then just doing a reload manually:

```
 {method: "GET", path: "/tyk/reload/", adminAuth: true, code: 200, afterFn: func() { doReload() }},
```

This can lead to two reloads that are racy too.

CC @buger @lonelycode

mvdan:
The endpoint blocks even if a reload is already queued, in which case it waits for the queued reload to finish.

See the commits for details.

Fixes #703.

### DIFF
--- a/api.go
+++ b/api.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Sirupsen/logrus"
@@ -865,9 +866,19 @@ func groupResetHandler(w http.ResponseWriter, r *http.Request) {
 	doJSONWrite(w, code, obj)
 }
 
+// resetHandler will try to queue a reload. If fn is nil and block=true
+// was in the URL parameters, it will block until the reload is done.
+// Otherwise, it won't block and fn will be called once the reload is
+// finished.
 func resetHandler(fn func()) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
+		var wg sync.WaitGroup
+		if fn == nil && r.URL.Query().Get("block") == "true" {
+			wg.Add(1)
+			fn = wg.Done
+		}
 		obj, code := handleURLReload(fn)
+		wg.Wait()
 		doJSONWrite(w, code, obj)
 	}
 }
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -413,9 +413,6 @@ func TestGetAPISpecsDashboardSuccess(t *testing.T) {
 		}
 	}
 	handleRedisEvent(msg, handled, wg.Done)
-	if len(reloadChan) != 1 {
-		t.Fatal("Should trigger reload")
-	}
 
 	// Since we already know that reload is queued
 	reloadTick <- time.Time{}
--- a/api_test.go
+++ b/api_test.go
@@ -599,10 +599,8 @@ func TestGroupResetHandler(t *testing.T) {
 func TestHotReloadSingle(t *testing.T) {
 	oldRouter := mainRouter
 	var wg sync.WaitGroup
-	if !reloadURLStructure(wg.Done) {
-		t.Fatal("reload wasn't queued")
-	}
 	wg.Add(1)
+	reloadURLStructure(wg.Done)
 	reloadTick <- time.Time{}
 	wg.Wait()
 	if mainRouter == oldRouter {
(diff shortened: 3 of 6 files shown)


mvdan:
I completely forgot about reloads that aren't queued... need to fix that and test it.

mvdan:
I've tried two approaches, but so far not happy with any. Not as easy as it might sound.

* If we use a global lock, the reload worker will need to do regular polling, wasting time and CPU
* If we use a single channel like now, there is no way (that I can think of) to "attach" to the queued channel without a race

Need to think more about the design.

lonelycode:
Could `handleURLReload` take an additional argument `onComplete chan struct{}`, then when the reload handler completes it asynchronously notifies this channel (so it can't block) and closes it. 

The caller, if interested, can then either block on the chan to confirm completion (with timeout) or just ignore it altogether / read it non-blocking / or call `handleURLReload` with `nil` instead of a callback chan?

mvdan:
@lonelycode that's the same that we have right now with `func handleURLReload(fn func())`. A callback func is basically the same, async-wise, than a channel send.

That would still leave us with the problem that this would notify goroutines that queued a job, but not the ones that didn't queue one.

lonelycode:
Gotcha, so basically we need a publisher instead of just a channel.

So maybe don't send a channel, instead register a channel with a notification queue / list / map, and have a dedicated reload notifier channel that sends data to a listening long running goroutine, this routine just copies the message to each registered channel and then closes and removes it.

That means any method that needs to know if a reload has happened just registers with the reload listener?

mvdan:
I spent most of the day yesterday playing with non-polling solutions. I did come to one that worked with just one channel (still have it on disk), but it had major disadvantages over this one:

* The endpoint was not testable, as the test could not "peek" into the queue like it does now by just grabbing the lock
* Twice the amount of code, and fairly complex for something that should be simple

The solution I've just uploaded is just a global struct with a `sync.Mutex`. It doesn't have those two issues, but it does function differently:

* The reload goroutine will poll for work every second, wasting a tiny bit of CPU
* Since we use a polling mechanism, doing a single hot reload might now take 0-1s to start as opposed to ~0s

I assume neither of those are a problem. After all, hot reloads slept for a number of seconds in 2.3.x :)

One alternative could be to have the endpoint block only when a hot reload starts because of it, returning 200. If the endpoint was used again and a reload was already queued, it would return a non-200 status code immediately, without blocking. But I feel like this approach leaks internal details and could be confusing in gateways with multiple users firing reloads separately.

mvdan:
Also, if anyone has an idea on how to do it, please speak up. Martin proposed one above, but I honestly didn't understand it.

The idea is to make all this work without polling. To make things clearer, in your proposed solutions please clarify how many worker goroutines you would use and what they would do (mine uses 1, the reload goroutine that polls for a queued reload).

We can always swap the implementation later on, since the HTTP API would be the same.

lonelycode:
Mine requires two goroutine a:

1. The reloader that waits for a reload signal and actually does he work
2. The publisher that Listens for signals from the reloader when it's done

Each reload caller (sender to the reloader) (api, redis signal, other) can register a one off listen channel with a global map (essentially a subscription to a topic)

When the reload completes, it sends a signal on the reload channel

The reload channel notifies the publisher

The publisher loops through each registered subscription channel to tell those callers that the reload is complete

No polling is required, it's entirely push-based

The caller can either chose to block on the subscription channel or skip it.

It's basically a pub/sub structure with channels

mvdan:
I'll give two goroutines a shot, thanks! This isn't pressing and I'm a bit bored of this so I'll get back to it after the weekend.

mvdan:
@lonelycode you were absolutely correct, two goroutines is the way to go. I've gone ahead and done that in a second commit, to better separate what each of the two commits does. It seems to work well, and it's actulaly not complicated.

No extra tests are needed to check that duplicate reload requests still block, as `TestHotReloadMany` already checks that - see the second commit message.

As for QA, it's not particularly easy to test this right now because reloads are executed at most every second. If it were changed to, say, one minute, then it would be far more obvious when reloads are ignored because one is already queued. I believe the unit tests cover these scenarios well enough.

mvdan:
Rebased due to conflicts. Any takers for a code review?

mvdan:
Rebased once more.

ilijabojanovic:
@buger @mvdan 
Tested on on-prem, hybrid and mdcb. Also i tested on prem with group of gateways. Looking good

mvdan:
(Martin approved in a code review in France next to me)

PR #721, opened 2017-05-08
plugin: marshal body into json as []byte

### LINKED ISSUE #351: Non-UTF8 in request bodies is broken when using a JS plugin
Uploading multi-part/form binary data via the gateway causes file corruption.

Regular text files work correctly, but binary ones (like images) get corrupted (size increment).

As a sample this is before Tyk

```
    -----------------------------1664174991698023324981845245
    Content-Disposition: form-data; name="file"; filename="mz1N2xGt_400x400.jpeg"
    Content-Type: application/octet-stream

    <FF><D8><FF><E0>^@^PJFIF^@^A^A^@^@^A^@^A^@^@<FF><E2>^B<A0>ICC_PROFILE^@^A^A^@^@^B<90>lcms^D0^@^@mntrRGB XYZ ^G<DF>^@^B^@@
    ^@'^@9acspAPPL^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@<F6><D6>^@^A^@^@^@^@<D3>-lcms^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
    [para^@^@^@^@^@^C^@^@^@^Bff^@^@<F2><A7>^@^@^MY^@^@^S<D0>^@^@
```

and this after Tyk

```
    -----------------------------16059868141680811620898218070
    Content-Disposition: form-data; name="file"; filename="mz1N2xGt_400x400.jpeg"
    Content-Type: application/octet-stream

    <EF><BF><BD><EF><BF><BD><EF><BF><BD><EF><BF><BD>^@^PJFIF^@^A^A^@^@^A^@^A^@^@<EF><BF><BD><EF><BF><BD>^B<EF><BF><BD>ICC_PROFILE^@^A^A^@^@^B<EF><BF><BD>lcms^D0^@
    ^@'^@9acspAPPL^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@<EF><BF><BD><EF><BF><BD>^@^A^@^@^@^@<EF><BF><BD>-lcms^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@
    [para^@^@^@^@^@^C^@^@^@^Bff^@^@<EF><BF><BD><EF><BF><BD>^@^@^MY^@^@^S<EF><BF><BD>^@^@
    [para^@^@^@^@^@^C^@^@^@^Bff^@^@<EF><BF><BD><EF><BF><BD>^@^@^MY^@^@^S<EF><BF><BD>^@^@
```

Relevant Community Forum Discussion:  https://community.tyk.io/t/uploaded-binary-file-thouth-tyk-corrupted-mangled/989

``

mvdan:
The encoding/json godoc reads:

	String values encode as JSON strings coerced to valid UTF-8,
	replacing invalid bytes with the Unicode replacement rune.

	... []byte encodes as a base64-encoded string, ...

We need to marshal the body as a []byte, not a string. Since JSON needs
to be valid UTF-8, we can't just keep the bytes as-is.

We can't simply quote the string, as JSON does not have a way to escape
bytes like \xff. It does have \u, but that is only for valid unicode
characters, which binary bytes may not be.

Using []byte in JSON means encoding to base64 in Go, so we must
introduce functions in the JSVM to decode and encode base64. This is so
that JS plugins can replace or modify the body in plaintext, not base64.
This is required for backwards compatibility.

This may or may not work in some JS implementations depending on whether
or not their strings can contain non-UTF8. Luckily, Otto supports that
just fine. We check this via the added test, which has ASCII, non-ASCII
unicode and non-unicode characters in one string that is modified.

Fixes #351.

### DIFF
--- a/js/tyk.js
+++ b/js/tyk.js
@@ -17,6 +17,7 @@ TykJS.TykMiddleware.MiddlewareComponentMeta.prototype.ProcessRequest = function(
 };
 
 TykJS.TykMiddleware.MiddlewareComponentMeta.prototype.DoProcessRequest = function(request, session, config) {
+    request.Body = b64dec(request.Body)
     var processed_request = this.ProcessRequest(request, session, config);
 
     if (!processed_request) {
@@ -26,6 +27,7 @@ TykJS.TykMiddleware.MiddlewareComponentMeta.prototype.DoProcessRequest = functio
     
     // Reset the headers object
     processed_request.Request.Headers = {}
+    processed_request.Request.Body = b64enc(processed_request.Request.Body)
 
     return JSON.stringify(processed_request)
 };
--- a/plugins.go
+++ b/plugins.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"io/ioutil"
@@ -27,7 +28,7 @@ type MiniRequestObject struct {
 	Headers         map[string][]string
 	SetHeaders      map[string]string
 	DeleteHeaders   []string
-	Body            string
+	Body            []byte
 	URL             string
 	Params          map[string][]string
 	AddParams       map[string]string
@@ -90,7 +91,7 @@ func (d *DynamicMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reques
 		Headers:        r.Header,
 		SetHeaders:     make(map[string]string),
 		DeleteHeaders:  make([]string, 0),
-		Body:           string(originalBody),
+		Body:           originalBody,
 		URL:            r.URL.Path,
 		Params:         r.URL.Query(),
 		AddParams:      make(map[string]string),
@@ -182,7 +183,7 @@ func (d *DynamicMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reques
 		r.Body = ioutil.NopCloser(bytes.NewBuffer(originalBody))
 	} else {
 		r.ContentLength = int64(len(newRequestData.Request.Body))
-		r.Body = ioutil.NopCloser(bytes.NewBufferString(newRequestData.Request.Body))
+		r.Body = ioutil.NopCloser(bytes.NewBuffer(newRequestData.Request.Body))
 	}
 
 	r.URL.Path = newRequestData.Request.URL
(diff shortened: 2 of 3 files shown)


mvdan:
I know doing work in JS is ugly, as is encoding and decoding to base64 twice for each body. But it's the only way I could get it to work reliably without breaking backwards compatibility.

Some alternative solutions:

* Only do this round-tripping to base64 if the body is not UTF8 to save some work (at the cost of some more code complexity)
* If the body is not UTF8, make `IgnoreBody` default to `true` (body won't be overwritten by js by default)
* Tell users to set `IgnoreBody` to `true` if an endpoint may take binary bodies

I didn't do the first because I believe JS plugins are already quite costly. Adding this work-around for non-UTF in JSON is probably a drop in the bucket. If we're concerned about performance of these (I think not because then users should use coprocess or Go plugins), we could benchmark and see if skipping base64 would be worth the effort.

lonelycode:
This is good - it should be fully backwards compatible with JS that users already have. Not so worried about the additional encoding.

mvdan:
Awesome. The added test has ascii, unicode and binary, so we should be fully backwards compatible.

@ilijabojanovic over to you, no hurry :)

To reproduce the original issue, simply have an API set up with a js plugin and send binary bodies (with non-utf8 bytes) through it. Before this, non-utf8 chars should be replaced. After, the body should be left intact. The added test does basically this, but in a more unit-test way.

buger:
To be consistent rename this functions to `btoa` and `atob` - standard browser methods to work with base64

mvdan:
I thought about it, but I preferred not to in case they function slightly differently to my implementations. Naming them `btoa` and `atob` could also mean that users might think using those funcs is officially supported and backwards compatible in 2.4 and beyond.

Are you positive on those two points?

mvdan:
And of course, there's the fact that `b64enc` is a much better name than `atob` :)

buger:
We do officially support them, after this PR.  I'm generally ok with this names, just do not forget to update documentation.

mvdan:
Do we? As long as we don't document them, I'd assume we don't officially support them. I would hide the funcs from user code if I knew how, but our current JS setup doesn't allow for that (I think).

buger:
I provided example in another issue why base64 may be useful, and one of the customers wanted them. So I would make them public.

mvdan:
Ah, sure.

mvdan:
To clarify - if anyone wants to look into `atob` and `btoa`, we can reuse those names if we're 100% sure it behaves the same in all situations (e.g. decoding invalid base64).

We could always leave it for later, aliasing `atob` and `btoa` to `b64enc` and `b64dec`.

buger:
Well, if to turn on :nerd_face: mode, `atob` and `btoa` functions is not a package of ECMA standard, and even `v8` does not ship them, not talking about `node` https://github.com/nodejs/node/issues/3462. These functions are browser extensions.

So we are free to name them as we want, and do not break any rules :)

mvdan:
Ah, good point - I vote for keeping my names then.

mvdan:
Rebased once more.

buger:
@ilijabojanovic still ready to be tested :)

mvdan:
This branch was ages behind master again, so I rebased and fixed the tests due to conflicts.

ilijabojanovic:
@buger approved for merge

mvdan:
:champagne: :tada:

PR #733, opened 2017-05-11
all: rework some return logic

mvdan:
* Return nil more often when there is an error
* Use less naked returns
* Use more error early returns

### DIFF
--- a/tyk_reverse_proxy_clone.go
+++ b/tyk_reverse_proxy_clone.go
@@ -37,30 +37,30 @@ func GetURLFromService(spec *APISpec) (*apidef.HostList, error) {
 		sd := ServiceDiscovery{}
 		sd.New(&spec.Proxy.ServiceDiscovery)
 		data, err := sd.GetTarget(spec.Proxy.ServiceDiscovery.QueryEndpoint)
-		if err == nil {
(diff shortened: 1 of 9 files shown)


tyk_reverse_proxy_clone.go
```diff
@@ -37,30 +37,30 @@ func GetURLFromService(spec *APISpec) (*apidef.HostList, error) {
...
+			return spec.LastGoodHostList, nil
 		}
+
+		ServiceCache.Set(spec.APIID, data, cache.DefaultExpiration)
+		// Stash it too
+		spec.LastGoodHostList = data
+		spec.HasRun = true
 		spec.ServiceRefreshInProgress = false
-		return nil, err
+		return data, nil
```
buger: I'm concerned about this line, you completely switched logic here, nope?

mvdan: Note the old `if err == nil` at the top, that is now `if err != nil`. We were using a non-error early return, which was weird. We now use a regular error early return. So this is no longer the "there was an error" branch, this is the "all went ok" branch.

This is perhaps confusing because both have a `spec.ServiceRefreshInProgress = false` before the return.

PR #735, opened 2017-05-12
Move more context logic to helper funcs

mvdan:
Only `OrgSessionContext` and `RetainHost` remain, but I'm bored for today.

### DIFF
--- a/api.go
+++ b/api.go
@@ -1849,3 +1849,17 @@ func ctxSetSession(r *http.Request, s *SessionState) {
 	}
 	context.Set(r, SessionData, s)
 }
+
+func ctxGetAuthToken(r *http.Request) string {
+	if v := context.Get(r, AuthHeaderValue); v != nil {
+		return v.(string)
+	}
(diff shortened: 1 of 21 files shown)


api.go
```diff
@@ -1849,3 +1849,67 @@ func ctxSetSession(r *http.Request, s *SessionState) {
...
+}
+
+func ctxSetTrackedPath(r *http.Request, p string) {
+	if p == "" {
+		panic("setting a nil context TrackThisEndpoint")
+	}
+	context.Set(r, TrackThisEndpoint, p)
+}
+
+func ctxGetDoNotTrack(r *http.Request) bool {
```
buger: Probably should be named `ctxIsDoNotTrack`

mvdan: I prefer `Get` for consistency. Also "is do not track" seems like the wrong sentence, since the argument is a request. Perhaps "has do not track". But I still prefer get for consistency with the rest of the funcs like it.

PR #737, opened 2017-05-13
api: rework to vastly simplify json logic

mvdan:
Each "handler" func would return raw JSON instead of an object to be
marshaled. This means that the logic to write a JSON response was
copy-pasted in dozens of places, including what to do when marshaling
fails.

Instead, have these return an interface{} and have the JSON be handled
in fewer places, and with a proper helper func.

This also lets us get rid of a lot of "success" logic clutter.

### DIFF
--- a/api.go
+++ b/api.go
@@ -33,21 +33,20 @@ type APIErrorMessage struct {
 	Error  string `json:"error"`
 }
 
-func createError(errorMsg string) []byte {
-	errorObj := APIErrorMessage{"error", errorMsg}
-	responseMsg, err := json.Marshal(&errorObj)
-
-	if err != nil {
-		log.Error("Couldn't marshal error stats: ", err)
-	}
-
-	return responseMsg
+func apiOk(msg string) APIErrorMessage {
+	return APIErrorMessage{"ok", msg}
 }
 
-func doJSONWrite(w http.ResponseWriter, code int, responseMessage []byte) {
+func apiError(msg string) APIErrorMessage {
+	return APIErrorMessage{"error", msg}
+}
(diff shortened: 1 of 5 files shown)


mvdan:
This had been bugging me for months, and it's something that will lead to endless backporting pain, so I think it's important to have in 2.4.

There is more work to do, so I think I'll still be able to remove another 200 lines from `api.go`. But this patch is big enough as it is.

api.go
```diff
@@ -33,21 +33,26 @@ type APIErrorMessage struct {
...
+}
 
-	return responseMsg
+func apiError(msg string) APIErrorMessage {
+	return APIErrorMessage{"error", msg}
 }
 
-func doJSONWrite(w http.ResponseWriter, code int, responseMessage []byte) {
+func doJSONWrite(w http.ResponseWriter, code int, obj interface{}) {
+	// TODO(mvdan): remove a couple of weeks after the api refactor,
```
buger: Can we merge without this comment?:)

I mean can you ensure that all dojwonwrite calls will not cause panic?

mvdan: Without the comment, or without the comment and panic?

I can't ensure it easily right now. I guess I could statically, but I would need to write up a tool to do static analysis and check all the possible types for the interface parameter. I've been playing with `guru` for a while now, but I can't get it to work properly - it seems to think the possible types are three, so it's clearly wrong.

A far easier option is to have this panic for a little bit and see if any of us run into a panic, which I would quickly fix. Is that a problem?

mvdan: Proven statically and removed, PTAL

PR #743, opened 2017-05-15
mw: don't use a goroutine for ctxSetSession

mvdan:
This func is very cheap and non-blocking. There is no reason to run it
in the background. This is unlike the SessionManager call, which might
do a call on our key-value store and do network work.

In fact, if further down the line the session is fetched from the
request context, it might not yet have been saved if this goroutine
hasn't been ran by any chance.

Also flip the if/else to remove the '!'.


mvdan:
Damn you, travis.

PR #746, opened 2017-05-16
api: deduplicate API listen paths if necessary

### LINKED ISSUE #658: When the gateway detects a listen path collision, it should modify the listen paths
bug/feature

When two APIs are created that share a listen path, the colision causes them both to not be loaded by the gateway, which can have side-effects such as the gateway not being able to generate tokens for that ID.

The gateway should still load the API, but should modify the listen paths so that other operations can continue - it should throw an error to inform the user what has happened.

1. Create an API with the listen path `/test`
2. Create another with the same listenm path
3. Both APIs will not load in the gateway

All

mvdan:
First add "-$appid", and keep trying if necessary by adding '_'
characters. Thus, if we have two API definitions with the path "/foo"
and IDs 1 and 2, their paths will end up being "/foo-1" and "/foo-2".

Fixes #658.

Logs look like:

```
[May 16 17:32:05] ERROR main: Listen path collision, changed to /v1-3 api_id=3 org_id=default
[May 16 17:32:05] ERROR main: Listen path collision, changed to /v1-2 api_id=2 org_id=default
```

### DIFF
--- a/api_loader.go
+++ b/api_loader.go
@@ -75,17 +75,6 @@ func skipSpecBecauseInvalid(referenceSpec *APISpec) bool {
 		return true
 	}
 
-	domainHash := generateDomainPath(referenceSpec.Domain, referenceSpec.Proxy.ListenPath)
-	val, listenPathExists := ListenPathMap.Get(domainHash)
-	if listenPathExists && val.(int) > 1 {
-		log.WithFields(logrus.Fields{
-			"prefix": "main",
-			"org_id": referenceSpec.APIDefinition.OrgID,
-			"api_id": referenceSpec.APIDefinition.APIID,
-		}).Error("Listen path is a duplicate: ", domainHash)
-		return true
-	}
-
 	return false
 }
 
@@ -137,6 +126,35 @@ func processSpec(referenceSpec *APISpec,
 		}).Warning("Skipped!")
 		chainDef.Skip = true
 		return &chainDef
+	}
+
+	pathModified := false
+	for {
+		hash := generateDomainPath(referenceSpec.Domain, referenceSpec.Proxy.ListenPath)
+		if v, e := ListenPathMap.Get(hash); !e || v.(int) < 2 {
+			// not a duplicate
+			break
+		}
+		if !pathModified {
+			prev := ApiSpecRegister[referenceSpec.APIID]
+			if prev != nil && prev.Proxy.ListenPath == referenceSpec.Proxy.ListenPath {
+				// if this APIID was already loaded and
+				// had this listen path, let it keep it.
+				break
+			}
+			referenceSpec.Proxy.ListenPath += "-" + referenceSpec.APIID
+			pathModified = true
+		} else {
+			// keep adding '_' chars
+			referenceSpec.Proxy.ListenPath += "_"
+		}
+	}
+	if pathModified {
+		log.WithFields(logrus.Fields{
+			"prefix": "main",
+			"org_id": referenceSpec.APIDefinition.OrgID,
+			"api_id": referenceSpec.APIDefinition.APIID,
+		}).Error("Listen path collision, changed to ", referenceSpec.Proxy.ListenPath)
 	}
 
 	// Set up LB targets:
--- a/api_test.go
+++ b/api_test.go
@@ -203,6 +203,61 @@ func TestApiHandlerPost(t *testing.T) {
 	}
 }
 
+func TestApiHandlerPostDupPath(t *testing.T) {
+	specs := func() (res []*APISpec) {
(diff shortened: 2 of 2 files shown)


mvdan:
@buger note that I ended up changing all paths, not just all but the first. This is because the current code fills a map with hashes first, and only checks for it later and concurrently. With this mechanism it would be tricky to leave the first unmodified - it would require extra code.

I don't think we care much either way, especially because this is something the user should fix as soon as possible.

mvdan:
Also, it might not be obvious but the panic found by @lghiur should also be fixed by this. This is because we used to skip API specs that had conflicting paths, and now we don't. It's the API specs that we skip that may be left uninitialized and have weird behaviour.

api_loader.go
```diff
@@ -139,6 +128,29 @@ func processSpec(referenceSpec *APISpec,
 		return &chainDef
 	}
 
+	pathModified := false
+	for {
+		hash := generateDomainPath(referenceSpec.Domain, referenceSpec.Proxy.ListenPath)
+		if v, e := ListenPathMap.Get(hash); !e || v.(int) < 2 {
```
buger: What "v" variable means in this context?

mvdan: Same as the original `val`, just an interface value. It's quite common as a single-char name, like `i` and `s`.

mvdan: The `< 2` is also ported from the original code. It's the count of API specs with said hash. So 1 means no dups.

buger:
I think it still make sense load first API without a prefix. Because otherwise it is really easy to fuck up. Imagine that you have working production API, and someone creates a duplicate, and production API stops working because path changed. 

You, know I realized that there is more consistent way to sort this specs. First of all it should check if spec are currently already loaded by Tyk, eg. working API, prioritize it, and count rest as duplicates. After it should sort by mongo ID: its first 4 bytes is a unix timestamp, so it acts as "created_at". If mongo id is not available (file based mode), sort by whatever we have, like ID. 

Main idea that adding a duplicate by accident should not affect existing working API's.

mvdan:
Fair enough, I'll update the patch. I was trying to go for the smallest diff to help with backporting - not sure how big it will get with these changes.

I assume that what you mean by Mongo ID is just the same APIID. I don't see any other ID field or anything related to Mongo.

Note that `loadApps` already sorts the APIs. It's a bit funny because it sorts them first by host and then by listen path, so I'm fairly sure the former is useless.

Why do we sort by listen path? There are no comments and this is an old commit by @lonelycode with a vague commit message.

Do we want to entirely swap the sorting to APIID? Or do we want to first sort by APIID, look for duplicates, then sort by listen path for the rest of `loadApps`?

lonelycode:
We sort by host and then listen path because we need to make sure that the routes are longest-first, and listen paths *can* be duplicated between domains, but not within them.

If we don't have longest-first, we end up with a shorter route capturing the traffic for another route, so:

/api/
/api/v2

V2 will never be hit because the first route will always be matched.

@buger Sorting by mongo ID is not a good idea, since with CE there is no Mongo In the stack and IDs are arbitrary. We should use api_id (not id), but since this is an arbitrary string in CE file based mode we can't guarantee that it will be of any specific length. It might actually be better to just use a counter.

mvdan:
> If we don't have longest-first, we end up with a shorter route capturing the traffic for another route

Good point, I'll keep that there then. But sorting by host does not work like you expect, as the sorting by path is not stable. I'm not sure why we need to sort by host first - we don't seem to need it, otherwise stuff would be broken already.

> We should use api_id (not id), but since this is an arbitrary string in CE file based mode we can't guarantee that it will be of any specific length.

What @buger was suggesting is a way to be consistent about which API will get to keep the duplicate path. I agree that we might chose the wrong one if we sort by APIID - for example if the files use numbers and `123` comes before `2`. But I don't know what else we could use that isn't over-complicating things.

There's always the current state of this PR, which doesn't keep any of the original paths. I still think it's an OK solution given how we want the users to fix the conflict as soon as possible.

> It might actually be better to just use a counter.

All of the solutions need counters to remember which duplicate is the first. What do you mean here?

mvdan:
@buger now if an API was already loaded and its listen path wasn't modified in the last reload, it won't get its path modified. PTAL

I tried to cover all possible scenarios in the test. The change in `loadApps` is rather small.

buger:
Looks good, @ilijabojanovic goes up to you

mvdan:
Just to clarify, the postfix is the APIID, not a number starting from 1. This is to make the postfixes stable across different scenarios.

@buger could you approve the changes on here, to clarify that this PR is just waiting for QA?

ilijabojanovic:
Verified and approved for merge @buger

PR #750, opened 2017-05-19
coprocess: unindent some C python code

mvdan:
No need for an else after a return.


buger:
@matiasinsaurralde up to you

PR #752, opened 2017-05-19
api: replace cmap with a regular old map

mvdan:
The way we use this map is that we populate it once synchronously, and
then we query it from multiple goroutines concurrently. Fortunately for
us, a regular Go map already supports this. It's only map writes that
have to be synchronous.

Recent Go releases already panic without the race detector if a map has
concurrent writes or a write concurrent to a read, so if the assumption
was wrong (or the code was buggy) we would notice.

As a bonus point, remove the cmap library. All it is is a
map[string]interface{} with a lock, so it's a fairly useless library. We
can roll that sort of stuff ourselves if we need to, or just use the new
sync.Map.


buger:
Looking at the code I see that `loadApps` can't be run concurrently because it gets called only inside `doReload` function, which we control, and `listen` which called only once.

If you confirm this as well, should be safe to merge.

mvdan:
The whole point of `loadApps` is that it's called sequentially, yes. If it was ran concurrently, plenty of other things would break first. For example, the way we use and then overwrite `ApiSpecRegister`. That would be a race panic too.

PR #754, opened 2017-05-19
websocket: handle request transmission errors

### LINKED ISSUE #471: Sloppy error handling in handler_websocket.go
@lonelycode could you please clarify this? (in a func that returns an error)

```
errc := make(chan error, 2)
cp := func(dst io.Writer, src io.Reader) {
        _, err := io.Copy(dst, src)
        errc <- err
}
go cp(d, nc)
go cp(nc, d)

<-errc

return nil, nil
```

Should we ignore errors, or should we treat them? Right now they're ignored, but the code seems like a broken attempt at handling them.

mvdan:
These could be read/write errors, for example. If anything happened,
let's log and record it.

Fixes #471.

### DIFF
--- a/handler_websocket.go
+++ b/handler_websocket.go
@@ -98,9 +98,19 @@ func (ws *WSDialer) RoundTrip(req *http.Request) (*http.Response, error) {
 	go cp(d, nc)
 	go cp(nc, d)
 
-	<-errc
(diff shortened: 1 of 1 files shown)


mvdan:
Also note that such a change was already approved by Martin in the issue.

PR #756, opened 2017-05-22
api: deduplicate API listen paths if necessary (backport)

### LINKED ISSUE #658: When the gateway detects a listen path collision, it should modify the listen paths

mvdan:
First add "-$appid", and keep trying if necessary by adding '_'
characters. Thus, if we have two API definitions with the path "/foo"
and IDs 1 and 2, their paths will end up being "/foo-1" and "/foo-2".

As suggested by Leon, let APIs that were already loaded with unmodified
listen paths keep them without changes. This way, if an API definition
is added with a conflicting listen path by mistake, the existing API it
conflicts with won't suddenly disappear.

Fixes #658.

### DIFF
--- a/api_loader.go
+++ b/api_loader.go
@@ -77,19 +77,6 @@ func skipSpecBecauseInvalid(referenceSpec *APISpec) bool {
 		return true
 	}
 
-	domainHash := generateDomainPath(referenceSpec.Domain, referenceSpec.Proxy.ListenPath)
-	val, listenPathExists := ListenPathMap.Get(domainHash)
-	if listenPathExists {
-		if val.(int) > 1 {
-			log.WithFields(logrus.Fields{
-				"prefix": "main",
-				"org_id": referenceSpec.APIDefinition.OrgID,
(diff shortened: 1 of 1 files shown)


mvdan:
Note that I left out the tests because they heavily depended on stuff that's in master but not in 2.3. So this will need confirming by QA that the backport didn't break the fix.

There were just minor conflicts, so I hope that won't be the case.

mvdan:
Err, fuck, why did I close this?

mvdan:
Reping @buger, looks like I closed it by mistake (perhaps I deleted the branch forgetting what it was)

Issue #749, opened 2017-05-19
Python error when using a plugin


ConsM:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
User reported getting the following error when using a plugin
```
Fatal Python error: drop_gil: GIL is not locked
SIGABRT: abort
PC=0x7f7a35f88c37 m=0
signal arrived during cgo execution

goroutine 417 [syscall, locked to thread]:
runtime.cgocall(0xbe74e0, 0xc42051a068, 0x0)
	/usr/local/go/src/runtime/cgocall.go:131 +0x110 fp=0xc42051a038 sp=0xc420519ff8
main._Cfunc_Python_HandleMiddlewareCache(0x33034a0)
	??:0 +0x41 fp=0xc42051a068 sp=0xc42051a038
main.(*PythonDispatcher).HandleMiddlewareCache(0xc42037c500, 0xc4205f2f60, 0xc4203ca910, 0x4b)
	/src/github.com/TykTechnologies/tyk/coprocess_python.go:223 +0x43 fp=0xc42051a090 sp=0xc42051a068
main.(*Bundle).AddToSpec(0xc4205f2f20)
	/src/github.com/TykTechnologies/tyk/coprocess_bundle.go:109 +0x99 fp=0xc42051a0c0 sp=0xc42051a090
main.loadBundle(0xc420559d60)
	/src/github.com/TykTechnologies/tyk/coprocess_bundle.go:321 +0x112e fp=0xc42051a7c0 sp=0xc42051a0c0
main.processSpec(0xc420559d60, 0xc4203a6500, 0x20, 0xc42039f720, 0xc42039f740, 0xc42039f760, 0xc4201bdd40, 0xc4201bdda0, 0xc4203a6500, 0xc4202416e8)
	/src/github.com/TykTechnologies/tyk/api_loader.go:216 +0x8ee0 fp=0xc42051be50 sp=0xc42051a7c0
main.loadApps.func1(0xc4202ffad0, 0xc42002c510, 0xc42039f720, 0xc42039f740, 0xc42039f760, 0xc4201bdd40, 0xc4201bdda0, 0xc4201bde00, 0xc420559d60, 0x20, ...)
	/src/github.com/TykTechnologies/tyk/api_loader.go:627 +0x10a fp=0xc42051bf38 sp=0xc42051be50
runtime.goexit()
[...]

PR #758, opened 2017-05-22
coprocess_python: enforce main OS thread lock when reloading bundles

matiasinsaurralde:
This fixes the issue described [here](https://community.tyk.io/t/python-gil-error/1624/4), I thought about making this more general and following this approach on the functions that call `HandleMiddlewareCache` (this is part of the dispatcher interface), however this is specific to the Python runtime.

### DIFF
--- a/coprocess_python.go
+++ b/coprocess_python.go
@@ -171,6 +171,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"unsafe"
 
@@ -212,8 +213,11 @@ func (d *PythonDispatcher) Reload() {
 
 // HandleMiddlewareCache isn't used by Python.
 func (d *PythonDispatcher) HandleMiddlewareCache(b *apidef.BundleManifest, basePath string) {
-	CBundlePath := C.CString(basePath)
-	C.Python_HandleMiddlewareCache(CBundlePath)
+	go func() {
+		runtime.LockOSThread()
+		CBundlePath := C.CString(basePath)
+		C.Python_HandleMiddlewareCache(CBundlePath)
+	}()
 }
 
 // PythonInit initializes the Python interpreter.


mvdan:
Is this the right fix? Looks like more of a workaround to me.

LGTM if there is no better way to fix it, though.

matiasinsaurralde:
It's the simplest solution, we need to ensure that this gets called from
the main thread, so the GIL locks correctly. The GIL itself is a big
limitation for the Python support, but it's something we have to deal with.

mvdan:
Please add "Fixes #749." to the end of the commit message and I'll merge.

matiasinsaurralde:
@mvdan Updated!

buger:
@matiasinsaurralde can you pls create an issue in Tyk repo, and describe how/when it happens. And if possible, add here some details on how to test it.

matiasinsaurralde:
See #761, could be related to recent changes in the API loader. I've found that forcing it to run in the main thread fixes the issue. Also if multiple bundles are loaded concurrently, we can't really call `HandleMiddlewareCache` in the same way, so we **need to lock the GIL**.

matiasinsaurralde:
@mvdan Modified

buger:
Yes, @ilijabojanovic pls check

ilijabojanovic:
@buger i can confirm that system working fine with this fix, but i am not capable for in-deep testing

mvdan:
@matiasinsaurralde by the way, I just realised that you patched a method that has a godoc saying `HandleMiddlewareCache isn't used by Python`. Is the godoc wrong, or did you patch some code that should never run?

mvdan:
bump @matiasinsaurralde ^

matiasinsaurralde:
@mvdan That was an old comment!

mvdan:
nice - could you send a PR updating it? :)

Issue #762, opened 2017-05-24
As a user I should be able to override non-error CP responses


matiasinsaurralde:
**Do you want to request a *feature* or report a *bug*?**
Feature/enhancement

**What is the current behavior?**
Currently we have a `ReturnOverrides` structure that allows you to set a custom HTTP code and error message for CP responses. In the middleware logic we check if this HTTP code is larger than 400, if this condition is true, we use this value but the body is ignored.

**What is the expected behavior?**
It would be useful to override non-error CP responses, scenario:
- A user writes a Python / gRPC middleware that performs a request in a special/custom way.
- The middleware passes this response (e.g. HTTP 200, and some JSON body) to Tyk and the request flow ends there.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
All versions.

buger:
@matiasinsaurralde do not forget to update "Project" of the issue, and move it though pipeline.

ilijabojanovic:
Looking good

PR #763, opened 2017-05-24
Extend ReturnOverrides

matiasinsaurralde:
See #762.

As a sample plugin (test_app.py)

```python
from tyk.decorators import *

@Hook
def MyCustomMiddleware(request, session, spec):
    print("my_middleware: MyCustomMiddleware")
    request.object.return_overrides.headers['content-type'] = 'application/json'
    request.object.return_overrides.response_code = 200
    request.object.return_overrides.response_error = "{\"key\": \"value\"}\n"
    return request, session
```

With manifest:
```json
{
  "file_list": [
    "test_app.py"
  ],
  "custom_middleware": {
    "post": [
      {"name": "MyCustomMiddleware"}
    ],
    "driver": "python"
  }
}
```

Curl output:
```
% curl localhost:8080/test-api/headers -v
*   Trying ::1...
* Connected to localhost (::1) port 8080 (#0)
> GET /test-api/headers HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 200
< Content-Type: application/json
< Date: Wed, 24 May 2017 08:32:50 GMT
< Content-Length: 17
< 
{"key": "value"}
* Connection #0 to host localhost left intact
```


coprocess.go
```diff
@@ -285,6 +285,15 @@ func (m *CoProcessMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Requ
 		return errors.New("Key not authorised"), int(returnObject.Request.ReturnOverrides.ResponseCode)
 	}
 
+	if returnObject.Request.ReturnOverrides.ResponseCode > 0 {
+		for h, v := range returnObject.Request.ReturnOverrides.Headers {
+			w.Header().Set(h, v)
+		}
+		w.WriteHeader(int(returnObject.Request.ReturnOverrides.ResponseCode))
+		w.Write([]byte(returnObject.Request.ReturnOverrides.ResponseError))
+		return nil, 666
```
mvdan: What does `666` mean here?

lonelycode: 666 is a magic return code that makes the gateway halt MW processing and respond

mvdan: Ah, I see we have `1666` also - we should use constants for these to make them more obvious, like `return nil, mwStatusRespond`. I'll ticket that up to do once this is merged.

lonelycode: That's a fantastic idea, let's do that! So much clearer

matiasinsaurralde:
@mvdan Just updated the tests, let's wait for #764 to replace the integers with constants.
I'm also extending the rest of the CP tests, trying to enhance them, will be a separate PR.

coprocess_test.go
```diff
@@ -247,6 +247,33 @@ func TestCoProcessAuth(t *testing.T) {
...
+	spec := createSpecTest(t, basicCoProcessDef)
+
+	chain := buildCoProcessChain(spec, "hook_test_return_overrides", coprocess.HookType_Pre, apidef.MiddlewareDriver("python"))
+
+	session := createNonThrottledSession()
+	spec.SessionManager.UpdateSession("abc", session, 60)
+
+	recorder := httptest.NewRecorder()
+
+	req, err := http.NewRequest("GET", "/headers", strings.NewReader(""))
```
mvdan: You can use a `nil` body

coprocess_test_helpers.go
```diff
@@ -125,6 +125,7 @@ func TestTykTriggerEvent(eventName, eventPayload string) {
 
 //export applyTestHooks
 func applyTestHooks(objectPtr unsafe.Pointer) {
+
```
mvdan: rogue empty line?

coprocess_test.go
matiasinsaurralde: A `nil` body panics 😮

mvdan: Something is buggy somewhere... forget I said anything for now :)

buger:
@matiasinsaurralde can we fix tests?

mvdan:
@buger tests are fixed, FYI

buger:
I'm right that this is similar to what we have now with Virtual Endpoints in our JS middleware?
Previously we already had option to add new headers, but with this functionality it is now possible to write any logic. Right?

Also, does it affect only Coprocess (Lua, Python) plugins, or gRPC ones too (because I see that you changed protobuffers files as well)?

matiasinsaurralde:
I'm not very familiar with the virtual endpoint feature. Yes, basically it's possible to take control of the request flow and return a custom body.
There is an existing feature to add headers from the CP middleware, but it's for request headers (before they hit the upstream API). The funcionality covered by this PR is specific to the response object, and the headers that are specified here are received by the client.
It affects all the plugins and will work as long as you're using the latest definitions and/or bindings provided by us.

PR #776, opened 2017-05-30
api: use x-tyk-api-expires to inform on API expiry

### LINKED ISSUE #437: Feature Request: include expiry date meta-data in header
As an API consumer, if I make a request for an API which has an expiry date, the response will include the expiry date as meta data in the header.

mvdan:
If a request matched a version within an API definition, and it wasn't
expired, this adds a header informing the user of what the expiration
date will be.

Use the RFC1123 format, as the HTTP/1.1 spec says that headers should
be written with dates in that format only.

Fixes #437.

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -981,27 +981,30 @@ func (a *APISpec) getVersionFromRequest(r *http.Request) string {
 	return ""
 }
 
-// IsThisAPIVersionExpired checks if an API version (during a proxied request) is expired
-func (a *APISpec) IsThisAPIVersionExpired(versionDef *apidef.VersionInfo) bool {
+// IsThisAPIVersionExpired checks if an API version (during a proxied
+// request) is expired. If it isn't and the configured time was valid,
+// it also returns the expiration time.
+func (a *APISpec) IsThisAPIVersionExpired(versionDef *apidef.VersionInfo) (bool, *time.Time) {
 	// Never expires
 	if versionDef.Expires == "" || versionDef.Expires == "-1" {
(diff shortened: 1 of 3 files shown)


mvdan:
Note that I'm no fan of adding an extra return value to `IsRequestValid`. But I saw no other way to do this without re-doing the `GetVersionData` and `IsThisAPIVersionExpired` calls.

buger:
Yeah, this looks ugly :(

What do you think about making `IsThisAPIVersionExpired` receive `request` object and call `GetVersionData` by itself: eg. making it independent from `isRequestValid`. And more over move it out of `isRequestValid` scrope, and call this check separately in `ProcessRequest` function. 

It should be ok, since `IsRequestValid` called only in one place.

mvdan:
Note that `IsRequestValid` uses results from `GetVersionData` elsewhere, such as `versionPaths`, so that wouldn't work as you described.

A simpler option that just occured to me is abusing the `meta interface{}` return value. It's already used only once for a `*apidef.EndpointMethodMeta`. But only if the status was `StatusRedirectFlowByReply`. So in the `StatusOk*` ones, we can just have it be `*time.Time`.

I realise this is also not particularly pretty either, but there isn't much we can do now without refactoring `IsRequestValid`. Right now it acts as a filter for all this data, so we must get it through somehow. I can leave in a TODO and we can tackle this separately at a later time.

buger:
Ok, lets continue with TODO.

mvdan:
@buger done and TODO added

PR #787, opened 2017-06-06
mw: remove copy-pasted boilerplate

mvdan:
Take advantage of how embedding structs also inherits methods and how they can be overriden.

@lonelycode this solves one of my main concerns regarding our middleware structure, since we no longer have so much unnecessary boilerplate code - but with absolutely no API changes :)

### DIFF
--- a/coprocess.go
+++ b/coprocess.go
@@ -3,8 +3,6 @@
 package main
 
 import (
-	"github.com/mitchellh/mapstructure"
-
 	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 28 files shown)


mvdan:
Also note that I removed some config methods and types as they're not actually used at all in their `ProcessRequest` counterparts.

mvdan:
The decrease in code coverage is sort of expected, as there are dozens of covered lines of code that have been deduplicated.

PR #788, opened 2017-06-06
analytics: remove config enabled check

mvdan:
All NormalisePath call sites (handler_success and handler_error) wrap
the call in the very same if, so this is redundant.


buger:
Adding milestone and label, because it may require changes in docs.

mvdan:
Err, why? This doesn't change how the software works at all.

PR #795, opened 2017-06-09
test: use ServeHTTP for all api endpoint tests

mvdan:
This makes sure that we test the router and the routes set up through
it, instead of going straight to the handlers. In other words, this
makes the tests go through gorilla/mux too, which will be necessary to
take advantage of gorilla's support of vars and other features.

### DIFF
--- a/api.go
+++ b/api.go
@@ -1016,7 +1016,7 @@ func createKeyHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	obj := APIModifyKeySuccess{
-		Action: "create",
+		Action: "added",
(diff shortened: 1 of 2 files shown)


mvdan:
I'm slightly uneasy about using `mainRouter` in the tests, but it works and I see no other best way to test the overall, main router including all the routes and sub-routers.

mvdan:
Also note that I did realise there is a lot of copy-pasting with the auth header stuff, but I want to take care of that in another commit and also deduplicate stuff along with it. Wanted to keep this PR small.

buger:
👍

PR #796, opened 2017-06-09
Use gorilla/mux for filtering of http methods

mvdan:
See the commit messages.

### DIFF
--- a/api.go
+++ b/api.go
@@ -561,22 +561,11 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
 		log.Debug("Creating new definition file")
 		obj, code = handleAddOrUpdateApi(apiID, r)
 	case "PUT":
-		if apiID != "" {
-			log.Debug("Updating existing API: ", apiID)
-			obj, code = handleAddOrUpdateApi(apiID, r)
-		} else {
-			obj, code = apiError("Must specify an apiID to update"), 400
-		}
+		log.Debug("Updating existing API: ", apiID)
+		obj, code = handleAddOrUpdateApi(apiID, r)
 	case "DELETE":
-		if apiID != "" {
-			log.Debug("Deleting API definition for: ", apiID)
-			obj, code = handleDeleteAPI(apiID)
-		} else {
-			obj, code = apiError("Must specify an apiID to delete"), 400
-		}
-	default:
-		// Return Not supported message (and code)
-		obj, code = apiError("Method not supported"), 405
+		log.Debug("Deleting API definition for: ", apiID)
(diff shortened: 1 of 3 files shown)


mvdan:
Note that misusing an endpoint will now be a `404` instead of a `400`. This is by design, as the route will no longer match unless both the path and the method match the filters.

I think this is an okay tradeoff to make. After all, this is common practice and these errors aren't common.

I found no way to keep the old behaviour while still making proper use of gorilla. Perhaps we could have sub-routers for each endpoint, defaulting to a "method unsupported" if the method filter didn't match, but we'd be back to extra code for each endpoint again.

buger:
@mvdan regarding your TODO. Unfortunately when we set route to "/apis/keys", gorilla/mux do not understand "/apis/keys/" path. So we have to do both.

buger:
I'm talking about your TODO comment added in this PR https://github.com/TykTechnologies/tyk/pull/796/files#diff-7ddfb3e035b42cd70649cc33393fe32cR307

Previously we had only "/api/keys/" route, and a lot of our users was confused because querying "/api/keys" (without trailing slash) did not work, so we added the separate route.

mvdan:
Yes - but the TODO is not about why we have to get both working. The TODO is about me not getting gorilla to do this with a single line. I have to add both routes manually and duplicate the rest of the line. Otherwise, only one of the two versions works.

mvdan:
To clarify, this TODO is in case we figure out how to do it with gorilla in the future, or to look at it again if we ever swap the HTTP router library.

buger:
Right 👍

mvdan:
I got confused with your first comment because I hadn't realised the duplication was already in place, and I thought I was introducing it here. Even then, I think the TODO should be there since we want to clean this up eventually.

buger:

buger:
@mvdan for documentation and testing reasons can you list here all the changes made in this PR. I mostly mean list of paths and HTTP methods, and how http codes changed.

mvdan:
The only change is that using the wrong methods will now be a 404 instead of a 405. Some examples:

* `POST /tyk/health/`
* `GET /tyk/keys/create`
* `DELETE /tyk/reload`

buger:
Well, you change lot of internal code, handling of HTTP codes, and move some of this logic to the router. And potentially /keys/* endpoint can have regression.

mvdan:
Of course, existing endpoints should keep on working. I assumed that was always done as part of regression testing.

mvdan:
I'm going to re-do this with our own method filtering middleware, keeping our current 405 status codes.

PR #798, opened 2017-06-12
test: deduplicate http.NewRequest logic

mvdan:
Handle both errors and []byte/string readers in the helper. This greatly
reduces copy-pasting of code.


mvdan:
More deduplication to follow, including authorization headers and JSON marshaling.

PR #805, opened 2017-06-15
jsvm: fix goroutine leak in the timeout fix

### LINKED ISSUE #804: JSVM leak introduced in 2.3.5
If you have an API that uses a JS middleware and hammer it with requests, Tyk's memory usage will increase over time. This is due to a goroutine leak.

Reproduced on both master and 2.3.5/2.3.6. Not present on 2.3.4.

This was very likely introduced by myself when fixing #496.

mvdan:
The code was buggy, in how we would end up sending twice to the channel
most of the time. The defer executes if we don't reach the timeout, so
the goroutine would get blocked at the second send.

Instead of trying to fix it by removing the second send, use a buffered
channel and send zero or one times. This way, it's easy to prove that
the goroutine will exit - it can't possibly block when sending zero or
one values to a buffered channel of size one.

Added a test that makes use of leaktest to check if the leak is still
there. Before the patch, it correctly finds the leaked goroutine:

```
--- FAIL: TestJSVMProcessLeak (1.04s)
        leaktest.go:90: leaktest: timed out checking goroutines
        leaktest.go:108: leaktest: leaked goroutine: goroutine 58 [chan send]:
                github.com/TykTechnologies/tyk.(*DynamicMiddleware).ProcessRequest.func1.1(0xc4201cfc20)
                        /home/mvdan/go/src/github.com/TykTechnologies/tyk/plugins.go:149 +0x5e
                github.com/TykTechnologies/tyk.(*DynamicMiddleware).ProcessRequest.func1(0xc4201cfc20, 0xc420a4a950, 0xd734d4, 0x7, 0xc42022a120, 0xd8, 0x112, 0xc4203b2500, 0x222, 0x24a)
                        /home/mvdan/go/src/github.com/TykTechnologies/tyk/plugins.go:153 +0x1a3
                created by github.com/TykTechnologies/tyk.(*DynamicMiddleware).ProcessRequest
                        /home/mvdan/go/src/github.com/TykTechnologies/tyk/plugins.go:140 +0xd69
```

Vendor leaktest too.

Fixes #804.

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -136,17 +136,15 @@ func (d *DynamicMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reques
 	log.WithFields(logrus.Fields{
 		"prefix": "jsvm",
 	}).Debug("Running: ", middlewareClassname)
-	ret := make(chan otto.Value)
+	// buffered, leaving no chance of a goroutine leak since the
+	// spawned goroutine will send 0 or 1 values.
+	ret := make(chan otto.Value, 1)
 	go func() {
 		defer func() {
 			// the VM executes the panic func that gets it
 			// to stop, so we must recover here to not crash
 			// the whole Go program.
 			recover()
-			// send a dummy value to the ret channel to
-			// signal that we died, since a panic will mean
-			// the regular send won't happen.
-			ret <- otto.Value{}
 		}()
 		returnRaw, _ := vm.Run(middlewareClassname + `.DoProcessRequest(` + string(asJsonRequestObj) + `, ` + string(sessionAsJsonObj) + `);`)
 		ret <- returnRaw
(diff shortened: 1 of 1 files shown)


mvdan:
I also tested this manually end-to-end with an API definition that had a silly JS middleware that did nothing, and hammering the endpoint with `curl`. Having `main.go` do something like:

```
go func() {
    for {
        println(runtime.NumGoroutine())
        time.Sleep(time.Second)
    }
}()
```

Before the fix, this number would steadily increase past the initial 40-80 number. With the fix, it stays around those values.

mvdan:
Hmm, it seems like leaktest is flakey on Travis due to the tests starting goroutines elsewhere.

mvdan:
I've removed the test for now.

buger:
@ilijabojanovic pls test that JS plugins still work

ilijabojanovic:
@buger approved for merge

PR #806, opened 2017-06-15
mw: rename all files to shorten names

### LINKED ISSUE #724: Proposal: rename all middlware_*.go files to mw_*.go
This is mainly because it's tedious to have to write `middleware_foo` instead of just `mw_foo` before being able to hit "tab" and get the completion for the proper file. Doesn't help that there is a `middleware` directory.

Some of these filenames are also fairly long, like `middleware_organisation_activity.go` or `middleware_key_expired_check.go`. Neither have a test file, so it can get even worse like `middleware_key_expired_check_test.go`.

CC @matiasinsaurralde @buger

mvdan:
Fixes #724.


buger:
For some reason lot of conflicts here.

mvdan:
Oh, this is because of the headers PR. Will fix.

mvdan:
@buger PTAL

PR #809, opened 2017-06-16
all: replace gorilla/context with std's context

### LINKED ISSUE #683: Replace gorilla/context with std's context

mvdan:
Instead of using a global map, store the context data inside each
http.Request.

This removes an external dependency and removes the necessity to use
gorilla/context.Clear to avoid memory leaks.

Most importantly, not using a hidden global map means that Go plugins as
middlewares will be able to modify request contexts easily.

See the TODO as to why we modify a request's context in-place.

We can't remove gorilla/context from vendor/ as gorilla/mux still
imports it (even though it's not used in Go 1.7+).

Fixes #683.

### DIFF
--- a/api.go
+++ b/api.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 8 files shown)


mvdan:
Rebased to fix conflicts.

mvdan:
@matiasinsaurralde merged, so let me know how it works with the Go plugins.

PR #810, opened 2017-06-16
all: make more use of the StorageHandler interface

mvdan:
This will be useful to swap out the driver in the future, and to keep
the code easier to modify and maintain.

Added a couple of storage methods that weren't in the interface yet but
were needed.


mvdan:
Also, I was thinking that we should do something similar with the pub/sub handler - we should have an interface for it.

PR #817, opened 2017-06-19
jsvm: remove config from middleware

mvdan:
Like in the coprocess one, this is never actually used.

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -10,7 +10,6 @@ import (
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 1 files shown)


buger:
Are you sure that it is not used? 

See https://community.tyk.io/t/virtual-endpoint-config/1291 and http://support.tyk.io/helpdesk/tickets/1120

mvdan:
@buger now that we have `config_data` tests, I've rebased on master to be extra sure that this change is safe.

PR #818, opened 2017-06-19
api: filter allowed methods in an HTTP middleware

mvdan:
Removes quite a lot of boilerplate.

Unfortunately, we can't use gorilla/mux's Methods() as that one results
in 404s instead of 405s. This is because it keeps the routes from
matching, resulting in no matching route being a 404.


mvdan:
@buger as promised earlier this morning, a better version of #796.

buger:
👍  merging since behavior does not changed

PR #819, opened 2017-06-20
api: use the router for URL arguments

mvdan:
We can do that now that the tests go through gorilla/mux, instead of
going to the handlers directly.

Cleans up the code a bit and removes the confusing use of {rest:.*}.


mvdan:
I had this saved on disk to send after the HTTP methods pull request, but couldn't send until now due to conflicts.

As with the other one, no behaviour changes.

PR #820, opened 2017-06-20
virtendp: re-add config_data handling

mvdan:
I removed the old GetConfig method here since ProcessRequest made no use
of the config parameter.

What I missed is that GetConfig was actually used in ServeHTTPForCache.

Re-add the code. And use a separate method name to clarify that it's not
the same GetConfig method seen in other middlewares. This way we also
make sure it won't be removed by mistake in the future, as the new
configData method is not part of the base middleware type.

### DIFF
--- a/mw_virtual_endpoint.go
+++ b/mw_virtual_endpoint.go
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/mitchellh/mapstructure"
+
 	"github.com/TykTechnologies/tyk/apidef"
 )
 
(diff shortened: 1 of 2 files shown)


buger:
Looks good to me. Can you create an issue related to this PR and attach it to 2.3.7 milestone with needs documenting label, because we have no right docs for this feature.

mvdan:
I assume you mean 2.4? This is a regression in master, not stable (thankfully).

buger:
Ah, right

mvdan:
done ^

PR #827, opened 2017-06-21
mw: don't copy entire request to get a cookie

mvdan:
Cookies always sit in headers, so it's safe to reuse the same request as
there is no chance that we will read the body or otherwise modify the
request like when accessing form values.


mvdan:
I found this while researching how we read and copy entire request and response bodies.

This doesn't fix any issue in particular, but it's low hanging fruit and an easy win.

PR #828, opened 2017-06-21
Mutual TLS

### LINKED ISSUE #357: Protect APIs with client certificates
As a user I should be able to protect my application with an approved set of client certificates:

For Auth:

- The client cert should be aligned with a token or a policy so that we can set rate limits - this may require a generating an internal representation of the token using: OrgID + hash(certificate) to store the metadata
- The certificates should be encrypted or protected in some way while at rest in redis
- When a request comes in, presented client cert should be used to identify the underlying token / identity to apply rate limits
- Should follow the same rules as a standard auth middleware (i.e. support chaining)

For management:

- I should be able to add a CA Cert to the gateway using the REST API, it should reply with a key ID I can use to manage this certificate
- I should be able to look up a certificate's session data by just querying using the certificate itself

buger:
Fixes #357

### Overview

In most of the cases when you try to access secured HTTPS/TLS endpoint, you experience only client-side check of the server certificate. Purpose of this check is to ensure that no fraud involved and data transfer between client and the server is encrypted.

In fact, TLS standard allows specifying client TLS certificate, so the server can accept connections only from clients which certificates registered at server certificate authority. In other words, clients are required to provide certificate and this certificate should be whitelisted on the server. This is what we call “Mutual TLS”, e.g. both sides require and verify certificates.

This change introduce Mutual TLS on the following layers:
* Authorization (whitelisting certificates on API level)
* Authentification (creating keys based on certificates)
* Upstream access (including JSVM)
* Control API
* Dashboard and MDCB API

### Basic architecture

There is 2 types of certificates: with and without private keys. Certificates without public keys used for authorization and authentification. Certificates with private keys used for upstream access, and server certificates: in other words when we need sign and encrypt request or response.

We support only certificates in PEM format. Nice bonus of PEM that it allows having multiple entries inside same file. It helps simplify the logic:  certificate with private keys stored in the same file.

This PR adds new `certs` package, introducing `CertificateManager` API, which handle all the certificate parsing, storage and retrieval logic. Certificates can be stored inside Redis or plain files. Certificates stored in Redis identified by their SHA256 fingerprint. Worth noticing that x509 certificate format implies that fingerprint is already embed into certificate, same as information about algorithm which was used to generate fingerprint. Tyk certificate storage do not use embed certificate fingeprints, and insted always use SHA256 algorithm all the time.

### Certificate management API
Using new Gateway API you can create, remove, list, and see info about certificates.

* Create: `POST /tyk/certs` with PEM body. Returns `{"id": "<cert-id>", ...}`
* Delete: `DELETE /tyk/certs/<cert-id>`
* Get info: `GET /tyk/certs/<cert-id>`. Return meta info about certificate, smth like this: `{"id": "<cert-id>", "has_private_key": false, "subject": {"common_name": "<cn>"...}...}`.
* Get info about multiple certificate: `GET /tyk/certs/<cert-id1>,<cert-id2>,<cert-id3>`. Returns array of meta info objects, similar to above.
* List all certificate IDs: `GET /tyk/certs`. Returns: `{"certs": ["<cert-id1>", "<cert-id2>", ...]}`

Note that this API works only with Redis certificates, and will not list of get info for file-based certs.

In addition, you may notice that you can't get raw certificate back, only its meta information. It is expected to improve security.

Certificates with private keys have special treatment and encoded before storing in redis. On certificate creation process, it first analyzes all PEM records, and if private certificate found, it decompose PEM, encode Private key with AES256, and encoded to PEM back. For encoding secret it use `security.private_certificate_encoding_secret` from tyk.conf, and if it is empty, fallbacks to `secret` value.

### Authorization

Added 2 new fields to API definition. First is `use_mutual_tls_auth`, boolean field, which enable client certificate check for given API. Second is `client_certificates`, which is a string array, with list of certificate ids. As mentioned previously, ids can be either SHA256 fingerprints, if certificate stored in Redis, or file paths.

When `use_mutual_tls_auth` set to `true` Gateway starts requiring and verifying client certificates, against list defined in `client_certificates` field. Client without certificate or with wrong certificate will be rejected.

It is imporant to mention that rejecting client during TLS handshake require that each API have a separate domain. However Tyk provides fallback when API have no separate domain, and additionaly verify client certificate on HTTP level. Using such fallback enables Mutual TLS for multiple APIs, with different listen paths, on the same domain.

In addition you can set global list of client certificates, available to each API, using `security.certificates.apis` string array option, in tyk configuration file.

### Authentification

You can create a key based on a provided certificate, and user will be authentificated based on client certificate. From technical point of view this is extension of "Auth token" authentication mode. You need to enable certificate check, by setting API definition `auth.use_certificate` boolean variable to `true`. After, when creating a key, you need to set `certificate` field to the existing certificate ID or path. Such keys has special treatment, and key ID will be generated as `OrgID + SHA256(certificate)`. "Auth token" middleware, if `auth.use_certificate` is enabled, and key is not provided via headers or url, will try to find key based on client certificate.

### Upstream access

You can specify certificates IDs per hostname, on API or global level.
In both cases, it is a hash when a key is a hostname, and value is certificate ID.
For API you should set `upsteam_certificates` to following format: `{"example.com": "<cert-id>"}`. For the global level it is same, but should be specified via `security.certificates.upstream` field in tyk.conf, same format.

For each HTTP call including reverse proxy, or JSVM http call, or Dashboard API call, Tyk checks agains maps mentioned above, and apply client certificate.

It is possible to specify `*` instead of hostname name, to match all domains.

Worth noticing that hostname name can include port if request made via non-standard HTTP port.

### Control API

You can protect Control API with mutual TLS, separately from APIs. Similar to authentification it will work in both cases when control API bind on separate domain or on the same as rest of APIs. In the last case it will fall back to validation on HTTP level.

Allowed certificate IDs should be specified via `security.certificates.control_api` configuration option in tyk.conf.

### Dashboard and MDCB API

Dashboard and MDCB APIs can be protected with mutual TLS, and you can specify client certificates used for talking with this APIs. Here tyk utilize upstream functionality mentioned above, so you just need to specify this certificates via `security.certificates.upstream` configuration option.

### Changes in server certificates

Right now to specify certificates for tyk server itself, you need to use `http_server_options.certificates` config option in following format: `[{"domain_name": "<domain>", "cert_file": "<path>", "key_file": "<path>"}]`.

In order to follow new architecture, and support certificates via PEM format, and certificates stored in Redis, you now can use new `http_server_options.ssl_certificates` option, which is just list of certificate IDs, similar to functionality described in previous paragraphs. Note that "domain_name" field is gone, and now it follows the guidelines, and just reads "common_name" or "dns_names" fields from certificate itself.

The old configuration option is still supported and will be marked as legacy.

### Technical implementation
Go 1.8 added a handy helper for `tls.Config` called `GetConfigForClient`, which allow loading different TLS configurations for each client based on TLS handshake info. The only attribute of handshake payload can be used in our case is `ServerName`.
Note: TLS is separate protocol from HTTP, so it does not include the notion of hostname (only IP), and `ServerName`  is actually SNI extension added to TLS quite long ago and supported by most of the clients.

By having access to ServerName we can enable Mutual TLS per API, based on APISpec.Domain. The only requirement is that each API which uses Mutual TLS should have own domain.

In cases when API domain is not specified or SNI extension is not provided by the client (so we do not know hostname), Tyk fallback to HTTP validation middleware, which checks client certificate via `request.TLS.PeerCertificates` variable.

### Testing

You can create self-signed client and server certificates with this command: `openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365`

For the server in common name specify domain, or just pass `-subj "/CN=localhost"` to OpenSSL command. After, follow https://www.tyk.io/docs/security/tls-and-ssl/ guide.

To get certificate SHA256 fingerprint use following command:
`openssl x509 -noout -fingerprint -sha256 -inform pem -in <cert>`

If you testing using `curl` command may look like: `curl  —cert client_cert.pem --key client_key.pem https://localhost:8181`

Be aware that Chrome 58 and after, may require a bit more complex commands, since it requires server certificate to have SNI DNS extension, and the command will look like:
```
openssl req \
    -newkey rsa:2048 \
    -x509 \
    -nodes \
    -keyout server.pem \
    -new \
    -out server.pem \
    -subj /CN=localhost \
    -reqexts SAN \
    -extensions SAN \
    -config <(cat /System/Library/OpenSSL/openssl.cnf \
        <(printf '[SAN]\nsubjectAltName=DNS:localhost')) \
    -sha256 \
    -days 3650
```
Where you should replayce /System/Library/OpenSSL/openssl.cnf to your path, and `localhost` to your domain.

Since all certificate ID configuration options is just arrays, you can set them via environment variables, like this: `TYK_GW_SECUIRITY_CERTIFICATES_CONTROLAPI="<cert1>,<cert2>"`, separated by commas. Docker users should be really happy with such approach. In this case updating Tyk SSL certificate is just matter of doing API call to the Tyk for adding certificate, and restarting container with environment variable with new Cert ID.

### Things left to do
- [x] JSVM http calls should use upstream certificates
- [ ] MDCB sync layer
- [ ] Dashboard and MDCB API


main.go
```diff
@@ -193,13 +194,18 @@ func buildConnStr(resource string) string {
 // Pull API Specs from configuration
 var APILoader = APIDefinitionLoader{}
 
-func getAPISpecs() []*APISpec {
-	var apiSpecs []*APISpec
+// Currently loaded specs
+var APISpecs []*APISpec
+var apiSpecSyncMutex sync.Mutex
```
mvdan: Are you certain a lock for writing only is enough? What if there's a read while you're writing with the lock grabbed?

main.go
```diff
@@ -193,13 +194,18 @@ func buildConnStr(resource string) string {
 // Pull API Specs from configuration
 var APILoader = APIDefinitionLoader{}
 
-func getAPISpecs() []*APISpec {
-	var apiSpecs []*APISpec
+// Currently loaded specs
+var APISpecs []*APISpec
```
mvdan: Any reason why this is exported?

main.go
```diff
@@ -209,31 +215,35 @@ func getAPISpecs() []*APISpec {
...
+			APISpecs[i].SessionProvider = config.AuthOverride.SessionProvider
 		}
 	}
-
-	return apiSpecs
 }
 
-func getPolicies() {
+var Policies = map[string]Policy{}
+var policySyncMux sync.Mutex
```
mvdan: Same as my other `Mutex` question - certain that locking only on writes is enough?

main.go
mvdan: Also, if you undo the unexport the diff will get smaller as you don't have to modify all the names below

buger:
@ilijabojanovic change is big, and you can start testing it. The dashboard comes soon.

buger:
@mvdan ready for you :)

api_loader.go
```diff
@@ -297,6 +297,11 @@ func processSpec(spec *APISpec, apisByListen map[string]int,
 		mwAppendEnabled(&chainArray, &VersionCheck{BaseMiddleware: baseMid})
 		mwAppendEnabled(&chainArray, &RequestSizeLimitMiddleware{baseMid})
 		mwAppendEnabled(&chainArray, &TrackEndpointMiddleware{baseMid})
+
+		if spec.UseMutualTLSAuth {
```
mvdan: define `EnabledForSpec` instead

api_loader.go
```diff
@@ -360,6 +364,14 @@ func processSpec(spec *APISpec, apisByListen map[string]int,
 		useCoProcessAuth := EnableCoProcess && mwDriver != apidef.OttoDriver && spec.EnableCoProcessAuth
 		useOttoAuth := !useCoProcessAuth && mwDriver == apidef.OttoDriver && spec.EnableCoProcessAuth
 
+		if spec.UseMutualTLSAuth {
```
mvdan: same as above, and you can either remove the log or move it into the mw's `EnabledForSpec`

apidef/api_definitions.go
```diff
@@ -279,8 +279,12 @@ type APIDefinition struct {
 		UseCookie      bool   `mapstructure:"use_cookie" bson:"use_cookie" json:"use_cookie"`
 		CookieName     string `mapstructure:"cookie_name" bson:"cookie_name" json:"cookie_name"`
 		AuthHeaderName string `mapstructure:"auth_header_name" bson:"auth_header_name" json:"auth_header_name"`
+		UseCertificate bool   `mapstructure:"use_certificate" bson:"use_certificate" json:"use_certificate"`
 	} `bson:"auth" json:"auth"`
 	UseBasicAuth            bool                 `bson:"use_basic_auth" json:"use_basic_auth"`
+	UseMutualTLSAuth        bool                 `bson:"use_mutual_tls_auth" json:"use_mutual_tls_auth"`
```
mvdan: just a thought - perhaps we should simply use `len(spec.ClientCertificates) > 0`

cert.go
```diff
@@ -0,0 +1,116 @@
...
+// dummyGetCertificate needed because TLSConfig require setting Certificates array or GetCertificate function from start, even if it get overriden by `getTLSConfigForClient`
+func dummyGetCertificate(*tls.ClientHelloInfo) (*tls.Certificate, error) {
+	return nil, nil
+}
+
+func getTLSConfigForClient(baseConfig *tls.Config, listenPort int) func(hello *tls.ClientHelloInfo) (*tls.Config, error) {
+
+	// Supporting legacy certificate configuration
+	serverCerts := make([]tls.Certificate, 0)
+	certNameMap := make(map[string]*tls.Certificate)
```
mvdan: Since you're using `make`, you might as well init the `cap`s to `len(config.Global.HttpServerOptions.Certificates)`

cert.go
```diff
@@ -0,0 +1,116 @@
...
+	for _, certData := range config.Global.HttpServerOptions.Certificates {
+		cert, err := tls.LoadX509KeyPair(certData.CertFile, certData.KeyFile)
+		if err != nil {
+			log.Fatalf("Server error: loadkeys: %s", err)
```
mvdan: I believe `log.Fatal` will exit the program. You either want `Error` and continue, or just `Fatal`.

cert.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	for _, cert := range CertificateManager.List(config.Global.HttpServerOptions.SSLCertificates, certs.CertificatePrivate) {
+		serverCerts = append(serverCerts, *cert)
+	}
+
+	baseConfig.Certificates = serverCerts
+
+	baseConfig.BuildNameToCertificate()
+	for k, v := range certNameMap {
```
mvdan: `for name, cert :=`

cert.go
```diff
@@ -0,0 +1,116 @@
...
+		if certID, err = CertificateManager.Add(content); err != nil {
+			doJSONWrite(w, 403, apiError(err.Error()))
+			return
+		}
+
+		doJSONWrite(w, 200, &APICertificateStatusMessage{certID, "ok", "Certificate added"})
+	case "GET":
+		if certID == "" {
+			certIds := CertificateManager.ListAllIds()
+			doJSONWrite(w, 200, &APIAllCertificates{certIds})
```
mvdan: break and unindent the rest

cert_test.go
```diff
@@ -0,0 +1,770 @@
...
+
+func genCertificate(template *x509.Certificate) ([]byte, []byte, []byte, tls.Certificate) {
+	priv, _ := rsa.GenerateKey(rand.Reader, 2048)
+
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, _ := rand.Int(rand.Reader, serialNumberLimit)
+	template.SerialNumber = serialNumber
+	template.BasicConstraintsValid = true
+	template.NotBefore = time.Now()
+	template.NotAfter = time.Now().Add(time.Hour)
```
mvdan: nit: don't call `time.Now()` multiple times (what if the program executes reeeeeally slowly?)

cert_test.go
```diff
@@ -0,0 +1,770 @@
...
+
+	config.Global.HttpServerOptions.UseSSL = true
+	config.Global.ListenPort = 0
+
+	dir, _ := ioutil.TempDir("", "certs")
+
+	defer func() {
+		os.RemoveAll(dir)
+		config.Global.HttpServerOptions.UseSSL = false
+		config.Global.ListenPort = 8080
```
mvdan: Don't hard-code the defaults please, save and restore them instead

mvdan: (this already bit us once with the `apps/` one)

cert_test.go
```diff
@@ -0,0 +1,770 @@
...
+	}()
+
+	t.Run("Without certificates", func(t *testing.T) {
+		ln, _ := generateListener(0)
+		baseURL := "https://" + strings.Replace(ln.Addr().String(), "[::]", "localhost", -1)
```
mvdan: Why is the replace necessary?

cert_test.go
```diff
@@ -0,0 +1,770 @@
...
+
+	dir, _ := ioutil.TempDir("", "certs")
+
+	defer func() {
+		os.RemoveAll(dir)
+		config.Global.HttpServerOptions.UseSSL = false
+		config.Global.ListenPort = 8080
+	}()
+
+	t.Run("Without certificates", func(t *testing.T) {
```
mvdan: I would avoid spaces in the names, and use camelcase instead like `WithoutCerts` to be consistent with the parent test names

certs/manager.go
```diff
@@ -0,0 +1,384 @@
...
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha256"
+	"crypto/tls"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/hex"
+	_ "encoding/json"
```
mvdan: ?

certs/manager.go
```diff
@@ -0,0 +1,384 @@
...
+
+// StorageHandler is a standard interface to a storage backend,
+// used by AuthorisationManager to read and write key values to the backend
+type StorageHandler interface {
+	GetKey(string) (string, error)
+	SetKey(string, string, int64) error
+	GetKeys(string) []string
+	DeleteKey(string) bool
+	DeleteScanMatch(string) bool
+}
```
mvdan: <3 the small interface

certs/manager.go
```diff
@@ -0,0 +1,384 @@
...
+	if key, err := x509.ParseECPrivateKey(der); err == nil {
+		return key, nil
+	}
+
+	return nil, errors.New("tls: failed to parse private key")
+}
+
+func isSHA256(value string) bool {
+	// 32 SHA256 size * 2 Hex encoding
+	if len(value) != 64 {
```
mvdan: please use `sha256.Size * 2` for clarity

mvdan: Actually, use `if hex.DecodedLen(value) != sha256.Size` and you can get rid of the comment, as it's obvious

certs/manager.go
```diff
@@ -0,0 +1,384 @@
...
+		return false
+	}
+
+	for _, c := range value {
+		if (c < 'a' || c > 'z') && (c < '0' || c > '9') {
+			return false
+		}
+	}
```
mvdan: Replace with `_, err := hex.DecodeString(value); return err == nil`

(perhaps a bit slower, but much simpler and self-explanatory)

certs/manager.go
```diff
@@ -0,0 +1,384 @@
...
+
+	return true
+}
+
+func HexSHA256(cert []byte) string {
```
mvdan: please unexport

certs/manager.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+func (c *CertificateManager) GetRaw(certID string) (string, error) {
+	return c.storage.GetKey("raw-" + certID)
+}
+
+func (c *CertificateManager) Add(certData []byte) (string, error) {
+	var certBlocks [][]byte
+	var keyPEM, keyRaw []byte
+	var block *pem.Block
```
mvdan: nit, but `block` can be defined inside the loop

certs/manager.go
```diff
@@ -0,0 +1,384 @@
...
+	c.storage.DeleteKey("raw-" + certID)
+	c.cache.Delete(certID)
+}
+
+func (c *CertificateManager) CertPool(certIDs []string) *x509.CertPool {
+	pool := x509.NewCertPool()
+
+	c.logger.Warn("CERT POOL REQUEST:", certIDs)
+	for _, cert := range c.List(certIDs, CertificatePublic) {
+		c.logger.Warn("Adding cert:", string(cert.Leaf.Extensions[0].Value))
```
mvdan: Is a warning good for when nothing went wrong? Maybe Info?

certs/manager.go
```diff
@@ -0,0 +1,384 @@
...
+			return nil
+		}
+	}
+
+	return errors.New("Certificate with SHA256 " + certID + " not allowed")
+}
+
+func (c *CertificateManager) FlushCache() {
+	c.cache.Flush()
+}
```
mvdan: unexport?

certs/manager_test.go
```diff
@@ -0,0 +1,198 @@
+package certs
+
+import (
+	"bytes"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	_ "crypto/tls"
```
mvdan: ?

apidef/api_definitions.go
buger: I wanted to make it consistent with rest of authentification methods. While from technical point of view we can replace it like this, from UX perspective, keeping separate attribute makes more sense, since it follows existing conventions on how we configure authentification.

cert.go
buger: It is possible but note that this is a static code which runs only on gateway start.

api_loader.go
buger: I'm not sure what you mean here if you look at the whole function this is a standard way how we conditionally include middleware: `if spec.UseBasicAuth`, `if spec.EnableSignatureChecking`, `if spec.EnableJWT` and etc.

certs/manager.go
buger: Actually, it used a lot both in tests and code, outside of package. For example `mw_auth_key.go`

certs/manager.go
buger: Used outside package, needed for tests.

apidef/api_definitions.go
mvdan: Sounds good - was just a thought.

cert.go
mvdan: Sure, but I was just thrown off by your explicit setting of `0` as the capacity. If you don't care about performance, you can omit it or use `[]tls.Certificate{}` and `map[string]*tls.Certificate{}`.

api_loader.go
mvdan: Then all of them are wrong :) all these should use `EnabledForSpec`, as per the interface contract.

This was probably done this way because the contract wasn't used properly before, which I fixed recently. Feel free to leave it like this though, as it won't be the only one to fix.

buger: Yeah, will leave it for you to refactor one day :)

mvdan: Added to my ever growing list of stuff to clean up ;)

cert_test.go
buger: TLS domain check works only if you pass it via URL. Inside this test actually both situations. First is when domain is provided in url, SNI turned on, and verification happens on TLS level, using domain. And second when domain is unknown.

cert_test.go
buger: I would prefer staying like this because this is one of the features I like with using `t.Run`. For me, it is the replacement of comments in the code, so I can in human language describe the high-level goal of the test. And in the end, it works fine when you run the single nested test from the command line too, it just replaces spaces with underlines.

certs/manager.go
buger: Nice!

cert_test.go
mvdan: We agree to disagree :) I simply don't like the name `CamelCase/With_many_underscores`, it seems unnecessary to mix the two styles.

PR #832, opened 2017-06-22
sessmgr: use an enum for session failure reasons

mvdan:
Fixes the TODO and stops using confusing integer values all over the
place.

Remove the boolean return value too, as it was true iff the int was 0.
Use sessionFailNone instead.

Also add a TODO in mw_organisation_activity.go since it seems to handle
only one of the failure reasons.


mvdan:
(assigning to Martin since it was his TODO)

buger:
Regarding TODO, organization limits apply only to quotas, it does not handle rate limiting. So should be good as it is.

mvdan:
Ah, I'll add a comment instead then.

mvdan:
@buger done, PTAL

PR #840, opened 2017-06-23
api: force gorilla/mux to ignore trailing slashes

mvdan:
So that we can easily support /foo/bar and /foo/bar/ for all static
endpoints without having to define them twice nor redirecting one to the
other (e.g. via mux's StrictSlash, which does redirects).

Do this by using {_:/?}, which matches the regex /? and discards it as
a path argument.

### DIFF
--- a/main.go
+++ b/main.go
@@ -299,17 +299,17 @@ func loadAPIEndpoints(muxer *mux.Router) {
 	}).Info("Initialising Tyk REST API Endpoints")
 
 	// set up main API handlers
-	r.HandleFunc("/reload/group", allowMethods(groupResetHandler, "GET"))
-	r.HandleFunc("/reload/", allowMethods(resetHandler(nil), "GET"))
+	r.HandleFunc("/reload/group{_:/?}", allowMethods(groupResetHandler, "GET"))
+	r.HandleFunc("/reload{_:/?}", allowMethods(resetHandler(nil), "GET"))
 
 	if !isRPCMode() {
 		r.HandleFunc("/org/keys/{keyName:.*}", allowMethods(orgHandler, "POST", "PUT", "GET", "DELETE"))
 		r.HandleFunc("/keys/policy/{keyName:.*}", allowMethods(policyUpdateHandler, "POST"))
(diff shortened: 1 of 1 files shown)


buger:
I'm not sure about part which removes PUT and DELETE, because it breaks compatibility in 2 ways: different status codes and, which is more important, we not show a good error message.

mvdan:
I'll revert that part and keep the current behavior then.

mvdan:
@buger PTAL - I've joined the two routes instead.

main.go
```diff
@@ -299,17 +299,16 @@ func loadAPIEndpoints(muxer *mux.Router) {
...
 	if !isRPCMode() {
 		r.HandleFunc("/org/keys/{keyName:.*}", allowMethods(orgHandler, "POST", "PUT", "GET", "DELETE"))
 		r.HandleFunc("/keys/policy/{keyName:.*}", allowMethods(policyUpdateHandler, "POST"))
-		r.HandleFunc("/keys/create", allowMethods(createKeyHandler, "POST"))
-		r.HandleFunc("/apis", allowMethods(apiHandler, "GET", "POST", "PUT", "DELETE"))
-		r.HandleFunc("/apis/{apiID:.*}", allowMethods(apiHandler, "GET", "POST", "PUT", "DELETE"))
-		r.HandleFunc("/health/", allowMethods(healthCheckhandler, "GET"))
-		r.HandleFunc("/oauth/clients/create", allowMethods(createOauthClient, "POST"))
+		r.HandleFunc("/keys/create{_:/?}", allowMethods(createKeyHandler, "POST"))
+		r.HandleFunc("/apis{_:/?}{apiID:.*}", allowMethods(apiHandler, "GET", "POST", "PUT", "DELETE"))
```
buger: Just a thought, since "/" is optional here, does it means that "/api{apiID}" (without slash between path and id) will be valid? It is not the critical, just want to confirm.

mvdan: You're absolutely right, I'm a dumbass.

mvdan: Fixed.

PR #846, opened 2017-06-30
RPC client should send pool ID on connection

buger:
Now we generate pool ID to unique identify Gateway instance, so MDCB
can use it instead of API address.

It is implemented on TCP level, by overriding Dial function. Once we
have connection object, we write handshake “register” message, so Sink
can distinguish legacy clients from new ones. Next, we send the length of ID
(max 256, or single byte), and the third message is ID itself.

Note that tests include simple MDCB mock, which shows how we read the ID and apply it.

Part of https://github.com/TykTechnologies/tyk-sink/pull/11

### DIFF
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -479,7 +479,7 @@ func (ln *customListener) Accept() (conn io.ReadWriteCloser, clientAddr string, 
 		return
 	}
 
-	handshake := make([]byte, 8)
+	handshake := make([]byte, 6)
 	if _, err = io.ReadFull(c, handshake); err != nil {
 		return
 	}
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -476,24 +476,18 @@ func (ln *customListener) Accept() (conn io.ReadWriteCloser, clientAddr string, 
 	}
 
 	handshake := make([]byte, 8)
-	_, err = c.Read(handshake)
-
-	// Let gorpc handle it
-	if err != nil {
+	if _, err = io.ReadFull(c, handshake); err != nil {
 		return
 	}
 
 	idLenBuf := make([]byte, 1)
-	_, err = c.Read(idLenBuf)
-	if err != nil {
+	if _, err = io.ReadFull(c, idLenBuf); err != nil {
 		return
 	}
 
 	idLen := uint8(idLenBuf[0])
 	id := make([]byte, idLen)
-	_, err = c.Read(id)
-
-	if err != nil {
+	if _, err = io.ReadFull(c, id); err != nil {
 		return
 	}
 
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -317,6 +317,8 @@ func startRPCMock(dispatcher *gorpc.Dispatcher) *gorpc.Server {
 
 	server := gorpc.NewTCPServer(":9090", dispatcher.NewHandlerFunc())
 	server.Listener = &customListener{}
+	server.LogError = gorpc.NilErrorLogger
+
 	globalConf.SlaveOptions.ConnectionString = server.Addr
 
 	go server.Serve()
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"io"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"sync"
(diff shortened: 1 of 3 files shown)


api_definition_test.go
```diff
@@ -437,3 +435,71 @@ func TestRoundRobin(t *testing.T) {
...
+		return
+	}
+
+	handshake := make([]byte, 8)
+	_, err = c.Read(handshake)
+
+	// Let gorpc handle it
+	if err != nil {
+		return
+	}
```
mvdan: From the doc:

> Read reads up to len(p) bytes into p. It returns the number of bytes read (0 <= n <= len(p)) and any error encountered.

So you could get `n, err == 4, nil` for example.

I believe you want `io.ReadFull`, which will make sure `n == 8`, even if multiple `Read` calls are necessary:

> ReadFull reads exactly len(buf) bytes from r into buf.

api_definition_test.go
```diff
@@ -437,3 +435,71 @@ func TestRoundRobin(t *testing.T) {
...
+		return err
+	}
+	if err := tcpConn.SetKeepAlivePeriod(30 * time.Second); err != nil {
+		return err
+	}
+	return nil
+}
+
+type customListener struct {
+	L net.Listener
```
mvdan: If you make this an embedded field, you could avoid defining `ListenAddr` and `Close` and just use `Addr` and `Close` from `net.Listener`.

rpc_storage_handler.go
```diff
@@ -145,22 +147,41 @@ func (r *RPCStorageHandler) Connect() bool {
...
+			conn, err = dialer.Dial("tcp", addr)
+		}
+
+		if err != nil {
+			return
+		}
+
+		conn.Write([]byte("register"))
+		conn.Write([]byte{byte(len(RPCClientSingletonConnectionID))})
+		conn.Write([]byte(RPCClientSingletonConnectionID))
```
mvdan: So this sends `register`, the length as a byte, and the ID itself? I assume the first two are so that we can change this "protocol" in the future more easily.

api_definition_test.go
```diff
@@ -437,3 +435,71 @@ func TestRoundRobin(t *testing.T) {
...
+		return
+	}
+
+	idLen := uint8(idLenBuf[0])
+	id := make([]byte, idLen)
+	_, err = c.Read(id)
+
+	if err != nil {
+		return
+	}
```
mvdan: same `ReadFull` comment as above

api_definition_test.go
```diff
@@ -437,3 +435,71 @@ func TestRoundRobin(t *testing.T) {
...
+	// Let gorpc handle it
+	if err != nil {
+		return
+	}
+
+	idLenBuf := make([]byte, 1)
+	_, err = c.Read(idLenBuf)
+	if err != nil {
+		return
+	}
```
mvdan: I'm not sure if the `ReadFull` thing applies here because you're only reading one byte, but you might as well do it out of consistency.

api_definition_test.go
mvdan: Also, you don't actually check that the string here is? It would also be more readable if you didn't hard-code the number `8` here, like:

```
const want = "register"
handshake := make([]byte, len(want))
[...]
```

rpc_storage_handler.go
mvdan: Also, this is perhaps a bit silly, but I would add something like:

```
if len(RPCClientSingletonConnectionID) > 255 {
    panic("RPCClientSingletonConnectionID is too long")
}
```

The silent overflow is a bit scary, even if we're sure that we won't touch this very often.

api_definition_test.go
buger: If client wrote 8 bytes on the wire, it will be always possible to read them in 1 single read. If it is not possible, it means that client wrote less than 8 bytes.

buger: Yeah we should introduce const for this.

buger: Right, I misunderstood what ReadFull does, you are right we can re-use it.

mvdan: Exactly - if the client wrote less, you want to error, not accept it :)

rpc_storage_handler.go
buger: > the first two are so that we can change this "protocol" in the future more easily.
Exactly

> The silent overflow is a bit scary, even if we're sure that we won't touch this very often.
Good point, because if we would like to change protocol, we for sure forgot about this limitation.

api_definition_test.go
buger: I just realized that I can't just embedd the field, because I need to override Accept function. Also `ListenAddr` is a go-rpc function, and this is actually is not net.Listener but gorpc.Listener interface, which differs.

mvdan: You can override methods from an embedded field - that's exactly what embedding is for. But you're right, I assumed the two interfaces were the same.

ilijabojanovic:
@buger approved for merge

Issue #829, opened 2017-06-22
JSVM: fix config_data, add a test and document it with examples (JS plugins and virtual paths)


mvdan:
This was broken unintentionally on master, because there were no tests and not even any docs for this feature.

In https://github.com/TykTechnologies/tyk/pull/820 I'm already fixing it and adding the test. This issue is to track adding documentation for it, including an example.

My test there should be a good start for an example.

mvdan:
Fix done, only the documentation remains.

buger:
Needs to be fixed for JSVM as well.

mvdan:
(it never existed nor worked for the JSVM, so this is a new feature for that middleware)

PR #847, opened 2017-06-30
jsvm: send config_data to JS plugins too

mvdan:
jsvm: send config_data to JS plugins too

Like in the JS code for virtual paths. JS plugins can now take 2 or 3
parameters, depending on whether or not they need config_data. Since JS
functions don't check the number of parameters received, the third one
for config_data is "optional".

Updates #829.

### DIFF
--- a/js/tyk.js
+++ b/js/tyk.js
@@ -11,13 +11,13 @@ var TykJS = {
         }
 };
 
-TykJS.TykMiddleware.MiddlewareComponentMeta.prototype.ProcessRequest = function(request, session) {
+TykJS.TykMiddleware.MiddlewareComponentMeta.prototype.ProcessRequest = function(request, session, config) {
     log("Process Request Not Implemented");
     return request;
 };
 
-TykJS.TykMiddleware.MiddlewareComponentMeta.prototype.DoProcessRequest = function(request, session) {
-    var processed_request = this.ProcessRequest(request, session);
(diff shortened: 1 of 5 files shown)


mvdan:
Examples of JS plugins taking 2 and 3 parameters are in `plugins_test.go`, see the two `const js = ...` lines.

js/tyk.js
```diff
@@ -40,7 +40,13 @@ TykJS.TykMiddleware.NewMiddleware.prototype = Object.create(TykJS.TykMiddleware.
 TykJS.TykMiddleware.NewMiddleware.prototype.constructor = TykJS.TykMiddleware.NewMiddleware;
 
 TykJS.TykMiddleware.NewMiddleware.prototype.NewProcessRequest = function(callback) {
-    this.ProcessRequest = callback;
+    if (callback.length == 2) { // backwards compatibility with two-arg type
```
buger: Actually, I think it is totally safe to remove this compatibility check. If you pass 3 arguments to the function that have only 2, everything should be fine.

mvdan: You're right, removed this bit and simplified the commit message.

mvdan:
@buger flaky RPC again, now green

buger:
Pls, fix the test. Looks good so far. Feel free to open backport PR as well.

PR #848, opened 2017-06-30
jsvm: add rawlog func to print directly to log.Out

### LINKED ISSUE #844: As a developer I should be able print directly to the console without logger
feature
Currently, users can use `log` function, and it formats the content, adds date, escapes special characters and etc.
Goal of the new `print` function is to avoid any formatting and print directly to gateway log

mvdan:
Without going through logrus and the pretty formatter. This means that
there will be no timestamp and no labels, and that your string won't be
escaped at all.

Call it raw to describe those, and also to have a scary name so that
people wishing to use it are warned that they should know what they're
doing.

The test is a bit convoluted because of the timestamps and because
background goroutines might use the logger while the test is running.

Fixes #844.

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


mvdan:
Also, I went for `rawlog` instead of `rawprint` because the output still goes to the log output, which is not always stdout/stderr.

buger:
Can you backport it as well?

Issue #753, opened 2017-05-19
Update our fork of httputil's reverse proxy


mvdan:
We forked it in 2014. There have been numerous changes to it since, including fixes. We should update it.

If possible, we should try to keep our modifications separate from the forked code so that updating the fork in the future would be easier.

Ideally, we wouldn't need to fork. We should look into whether we can use httputil directly now.

Step 1 is comparing our copy with that of 2014 to see what changes were made, since it was all added in a single commit.

PR #867, opened 2017-07-05
Backport httputil.ReverseProxy changes from early 2015 to mid 2016

mvdan:
We forked in late 2014. This backports the changes from all of 2015 and half of 2016. Only the rest of 2016 and 2017 remain - this takes care of about half of the changes.

I've backported nearly every change, excluding some minor doc changes and a buffer pool thing that is a feature rather than a fix - and which we could roll out on our own if we needed to.

I've also excluded Trailers support (https://go-review.googlesource.com/c/11940/) as we've touched the code that it changes and it was a bit risky. And it's also more of a feature than a fix, which we can come back to if we ever need.

Updates #753.

### DIFF
--- a/tyk_reverse_proxy_clone.go
+++ b/tyk_reverse_proxy_clone.go
@@ -326,6 +326,7 @@ func copyHeader(dst, src http.Header) {
 // http://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html
 var hopHeaders = []string{
 	"Connection",
+	"Proxy-Connection", // non-standard but still sent by libcurl and rejected by e.g. google
 	"Keep-Alive",
 	"Proxy-Authenticate",
 	"Proxy-Authorization",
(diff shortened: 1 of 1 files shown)


buger:
Are there any changes we can use somehow, even potentially?

Also `User-Agent` line looks supricious, will it still be proxied?

mvdan:
> Are there any changes we can use somehow, even potentially?

Well, some of the backported commits here are fixes, so they'll probably be useful. I'm not sure what you mean by that.

> Also User-Agent line looks supricious, will it still be proxied?

It will be proxied only if the original request had one. This just prevents the reverse proxy from adding Go's user-agent if none was set, as `net/http` does by default. That's why I said it updates the user-agent bug (but doesn't fix it - these lines will need changing to fix it).

buger:
Fair enough 👌

PR #869, opened 2017-07-05
Fix all remaining "go test -race" races

mvdan:
I ran it a few times and nothing remains.

There are probably still a couple hidden - I need to abuse `-race -count 100` before enabling this on CI.


mvdan:
The drop in coverage is probably because `testHttp` used to do more in its calls of `initialiseSystem`. But if we want to test `initialiseSystem` properly, we should do it separately in its own test.

PR #873, opened 2017-07-06
CI: also run go test -race

### LINKED ISSUE #363: CI: also run tests with -race

mvdan:
Separately from the other "go test" run, for various reasons:

* It can be slow, so no need to run it in all Go versions
* If a test fails only when in race mode, that's useful info
* No need to run the race detector if the tests failed without it

Fixes #363.

### DIFF
--- a/utils/ci-test.sh
+++ b/utils/ci-test.sh
@@ -32,9 +32,11 @@ for pkg in $PKGS; do
 done
 
 if [[ ! $LATEST_GO ]]; then
-	echo "Skipping linting and coverage report"
+	echo "Skipping race, checks, and coverage report"
 	exit 0
 fi
+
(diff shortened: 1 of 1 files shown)


mvdan:
@buger PTAL

I ran both `go test -race` and `go test -tags coprocess -race` with large counts locally and hit no races. I'm fairly confident we don't have any in master with the current test coverage.

Of course there can still be some flakiness - if that happens, we can take care of it as it would likely be an actual bug.

PR #876, opened 2017-07-06
Do not run letsencrypt watcher unless explicitly enabled

buger:
LetsEncrypt is a third party service, and it is a bad idea to depend on it unless the user really wants.

While it is run in separate goroutine, it still can cause memory leaks, panics, and etc.


mvdan:
Typo? :)

buger:
Made change via Github UI 🙈, will squash.

mvdan:
I was ready to point that out ;)

PR #877, opened 2017-07-06
revproxy: simplify switchTargets logic

mvdan:
Instead of an extra variable and a bool, just swap the target value
directly.


mvdan:
Also, the simpler logic should help with figuring out the right fix for the `RetainHost` stuff.

PR #878, opened 2017-07-06
all: protect apisByID with a rwmutex

mvdan:
Like in the case of the global policies map, reloads may swap this map
while other goroutines are using it. Add a test too.

Fixes the following race report from the test without the fix:

	WARNING: DATA RACE
	Write at 0x0000018a9fd8 by goroutine 494:
	  github.com/TykTechnologies/tyk.loadApps()
	      /home/mvdan/go/src/github.com/TykTechnologies/tyk/api_loader.go:622 +0xa0d
	  github.com/TykTechnologies/tyk.loadSampleAPI()
	      /home/mvdan/go/src/github.com/TykTechnologies/tyk/api_test.go:45 +0xd5
	  github.com/TykTechnologies/tyk.TestReloadApisRace()
	      /home/mvdan/go/src/github.com/TykTechnologies/tyk/gateway_test.go:973 +0x7c
	  testing.tRunner()
	      /home/mvdan/tip/src/testing/testing.go:754 +0x16c

	Previous read at 0x0000018a9fd8 by goroutine 394:
	  github.com/TykTechnologies/tyk.TestReloadApisRace.func1()
	      /home/mvdan/go/src/github.com/TykTechnologies/tyk/gateway_test.go:969 +0x4e

### DIFF
--- a/api.go
+++ b/api.go
@@ -67,6 +67,8 @@ func allowMethods(next http.HandlerFunc, methods ...string) http.HandlerFunc {
 }
 
 func GetSpecForOrg(apiID string) *APISpec {
+	apisMu.RLock()
+	defer apisMu.RUnlock()
 	var aKey string
 	for k, v := range apisByID {
 		if v.OrgID == apiID {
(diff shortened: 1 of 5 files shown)


buger:
Let's try to backport it as well. Maybe instead of automatic merge better to do it manually.

PR #889, opened 2017-07-07
mw: rename New to Init

mvdan:
Like in the event handler interface, this is a method that initialises -
it doesn't create and return a new object.

While at it, remove some statements that just set struct fields to their
zero values in init funcs - that's unnecessary.


mvdan:
`-race` flake, reported above and restarted the build.

PR #892, opened 2017-07-07
all: simplify funcs that take method and url

mvdan:
Have them take an http.Request instead. Since all methods and urls came
from r.Method and r.URL.Path anyway, this simplifies the call sites too.


buger:
^ conflict

mvdan:
Fixed.

PR #896, opened 2017-07-07
Finish backporting reverse proxy fixes from httputil

### LINKED ISSUE #753: Update our fork of httputil's reverse proxy

mvdan:
Fixes #753.

Again, skipped a few - small doc updates, a feature to change the logger, a feature to add a response modifier function and a minor request copying optimization.

### DIFF
--- a/tyk_reverse_proxy_clone.go
+++ b/tyk_reverse_proxy_clone.go
@@ -338,27 +338,6 @@ var hopHeaders = []string{
 	"Upgrade",
 }
 
-type requestCanceler interface {
-	CancelRequest(*http.Request)
-}
-
-type runOnFirstRead struct {
-	io.Reader // optional; nil means empty body
-
-	fn func() // Run before first Read, then set to nil
-}
-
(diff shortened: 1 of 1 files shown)


buger:
So you backported commit by commit, or by overall diff? 

Asking because in first PR you backported `CloseNotifier` support, and in this PR it was kinda removed and replaced by new version.

mvdan:
Commit by commit. The original commits also reverted each other like this.

I preferred to not squash to make it easier for us to know what was backported in the future.

Will fix conflicts when I'm back on my laptop.

mvdan:
Conflicts resolved.

PR #902, opened 2017-07-10
api: fix /tyk/cache/{apiID}/ (with trailing slash)

### LINKED ISSUE #900: Delete cache endpoint returns 404 error
Branch: master
ENV: All

Reproduction steps:

1. Create api definiton 
2. Trigger cache invalidation 
```
Method: DELETE 
Endpoint: /tyk/apis/{api_id}
```
Actual:
In response we have this error:
404 page not found
Response code: 404

Expected:
We should got: Message: OK

mvdan:
We supported this since we were using ".*" as a regex, and removing all
slashes from the apiID later. This is sloppy and confusing, hence why I
broke it.

Do it the proper way with a route matching regex, and add a test.

Fixes #900.

### DIFF
--- a/api.go
+++ b/api.go
@@ -1377,7 +1377,7 @@ func invalidateCacheHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func handleInvalidateAPICache(apiID string) error {
-	keyPrefix := "cache-" + strings.Replace(apiID, "/", "", -1)
+	keyPrefix := "cache-" + apiID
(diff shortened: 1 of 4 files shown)


mvdan:
I know the test is dumb (see the TODO) but it's better than nothing.

mvdan:
@ilijabojanovic please re-run the cache test with this fix

ilijabojanovic:
@mvdan fixed
@buger approved for merge

PR #903, opened 2017-07-10
api: re-allow empty keyName path argument in GETs

### LINKED ISSUE #901: Get API Tokens return 404 error
Branch: Master
ENV: All

Reproduction steps:

1. Create api definition 
2. Create tokens for that api definition 
3. Call:
```
Method: GET
/tyk/apis/{api_id}/keys
```
Actual:
{'Status': 'Error', 'Message': 'Could not retrieve keys', 'Meta': None}
Response code: 404

Expected:
We should got list

mvdan:
When I removed all ".*" regexes from gorilla/mux, I thought the default
regex was "[^/]*" - i.e. that an empty value was accepted. But the
default value is actually "[^/]+", where an empty value doesn't match
the route.

This broke a few endpoints that used an empty keyName in GET requests to
list all the keys.

Modify one of the tests to check this.

Fixes #901.

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -314,24 +314,26 @@ func TestKeyHandlerUpdateKey(t *testing.T) {
 }
 
 func TestKeyHandlerGetKey(t *testing.T) {
-	for _, api_id := range []string{"1", "none", ""} {
-		loadSampleAPI(t, apiTestDef)
-		createKey(t)
+	for _, pathSuffix := range []string{"/", "/1234"} {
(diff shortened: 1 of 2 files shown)


mvdan:
@ilijabojanovic please run the tests again with this fix

mvdan:
Also, the regexes fucked with the markdown formatting here - check the plaintext commit message instead.

ilijabojanovic:
@mvdan fixed
@buger approved for merge

PR #905, opened 2017-07-11
all: replace def RawData with simpler ConfigData

mvdan:
Before this commit, RawData was an entire copy of the APIDefinition as a
map[string]interface{}, all the way down. This was only used in a few
middlewares to obtain certain structured data via mapstructure.

In practice, this was only used to obtain the config_data field of type
map[string]interface{}.

Remove the entire copy of APIDefinition in memory and just add a
ConfigData field of type map[string]interface{}. This way, we can also
get rid of all the code that was gluing RawData to its final purpose.

This is in line with the changes being made in other repos to add
config_data to the API definition schema. We don't want to allow
arbitrary root fields - we only want to allow config_data, at least for
now. This makes sure that the gateway doesn't support anything else
unless we meant to.

### DIFF
--- a/api.go
+++ b/api.go
@@ -452,11 +452,9 @@ func handleGetAPIList() (interface{}, int) {
 	apisMu.RLock()
 	defer apisMu.RUnlock()
 	apiIDList := make([]*apidef.APIDefinition, len(apisByID))
-
 	c := 0
 	for _, apiSpec := range apisByID {
 		apiIDList[c] = apiSpec.APIDefinition
-		apiIDList[c].RawData = nil
 		c++
 	}
 	return apiIDList, 200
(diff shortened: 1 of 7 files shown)


mvdan:
@buger this is what I was talking to you about yesterday, when I said I could simplify `config_data` handling quite a lot.

Note the object format change in the JS code. Other options include:

* Manually re-do the boxing of the object as `{config_data: {foo: bar}}`
* Adding it at the end to support both old and new ways, like `{{foo: bar}, config_data: {foo: bar}}`

I'd rather keep it the way it is now, though. Not sure if it's best to change this (while documenting the change) in 2.3.7 or 2.4. As of 2.3.6 neither is documented, so we could do it in 2.3.7 as the "documented way", or change the official behavior in 2.4.

mvdan:
Went with the first option of re-boxing the JSON object for backwards compatibility. Left a TODO for 3.0 or later.

PR #910, opened 2017-07-11
CI: also run go 'test -race -tags coprocess'

mvdan:


matiasinsaurralde:
We'll probably tackle this in a different way after #911 and some test modifications

mvdan:
@matiasinsaurralde I didn't know you were going to do that. Should I close this?

matiasinsaurralde:
@mvdan let's keep it open for now, just wanted to reference it

mvdan:
@matiasinsaurralde said that the `coprocess` tag will disappear, so this will be done without the extra machinery.

matiasinsaurralde:
That's right!

PR #928, opened 2017-07-14
Allow urls to be rewritten in a coprocess middleware

coatezy:
This PR allows urls to be rewritten within a coprocess middleware.

For example:
```
from tyk.decorators import Hook

@Hook
def PostMiddleware(request, session, metadata, spec):
  request.object.url = '/hello/world'

```

The existing pattern in the mini request object is to define a variable prefixed with the verb describing the action to be performed such as `SetHeaders` and `DeleteParams` but all existing examples are updating dictionaries.

As this is mutating a string I'm not sure if the existing pattern makes sense?

Happy to discuss.

This PR relates to this thread that I posted on the forum. https://community.tyk.io/t/rewrite-upstream-url-in-plugin-python/1746

### DIFF
--- a/coprocess.go
+++ b/coprocess.go
@@ -145,6 +145,7 @@ func (c *CoProcessor) ObjectPostProcess(object *coprocess.Object, r *http.Reques
 		values.Set(p, v)
 	}
 
+	r.URL.Path = object.Request.Url
 	r.URL.RawQuery = values.Encode()
 }
 


matiasinsaurralde:
Hi @coatezy, I've been testing this and looks good, the approach is similar to the JSVM implementation. Mutating a string makes sense in this case, no dicts needed!

coatezy:
Hi @matiasinsaurralde,

Thanks for taking the time to test and review the PR. 😄 

Should I also submit an accompanying PR to the docs repo documenting the rewrite functionality?

buger:
@matiasinsaurralde if you want to backport it to upcoming 2.3.8, pls handle backporting, and documentation part.

matiasinsaurralde:
@buger Sure!

coatezy:
@buger @matiasinsaurralde Thanks guys! 😄

paulmangam:
Is it also possible to override upstream host thru coprocess middleware? Similar to how request.object.url overrides the path.

PR #937, opened 2017-07-24
api: redesign round-robin with sync/atomic

mvdan:
This means we no longer use a mutex around a number. The code should be
a bit faster, but most importantly it's much shorter and easier to
understand.

The only change in behavior is that when the max gets shortened, instead
of resetting back to 0, we wrap around with the remainder operation.
This is less predictable - the new position will be random - but it's
actually better. If we always reset to 0, the first elements of the list
would tend to get selected more often than the last elements.

Also test that we don't barf with a divide by zero if the length is zero
and there are no positions to select from.

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -10,7 +10,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
-	"sync"
+	"sync/atomic"
 	textTemplate "text/template"
 	"time"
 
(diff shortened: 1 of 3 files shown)


api_definition.go
```diff
@@ -1087,27 +1087,14 @@ func (a *APISpec) Version(r *http.Request) (*apidef.VersionInfo, []URLSpec, bool
...
-func (r *RoundRobin) SetLen(len int) { r.SetMax(len - 1) }
-
-func (r *RoundRobin) Pos() int {
-	cur := r.pos
-	if r.pos++; r.pos > r.max {
-		r.pos = 0
-	}
-	return cur
+	// -1 to start at 0, not 1
+	cur := atomic.AddUint32(&r.pos, 1) - 1
```
buger: What about potential integer overflow?

mvdan: It's unsigned, so it will just go back to 0.

I the number of bits was small and the list was very long, this could mean that we wouldn't get to all the elements. But given that it's 32 bits and lists will rarely ever go beyond the dozens, it's fine.

For what it's worth, the existing implementation also has overflows. But with this one it's more obvious and more predictable, as the number is always unsigned and 32 bits.

mvdan:
friendly ping @buger :)

ilijabojanovic:
@buger approved for merge

buger:
👌

PR #941, opened 2017-07-25
all: always make mainRouter == defaultRouter

### LINKED ISSUE #940: Reloading may not work if config.HttpServerOptions.OverrideDefaults is true
Related to https://github.com/TykTechnologies/tyk/issues/915, as the OverrideDefaults block uses `defaultRouter` while the other uses `mainRouter`.

mvdan:
setupGlobals sets them up to be equal, but they might diverge under some
circumstances like a reload.

In some cases we register things on defaultRouter instead of mainRouter.
If they're not the same pointers, things start breaking.

This was the case with OverrideDefaults.

Don't merge the two global vars in stable, as that change is a bit too
invasive. Simply fix the two occurences where they can diverge.

Fixes #940.

### DIFF
--- a/main.go
+++ b/main.go
@@ -727,6 +727,7 @@ func doReload() {
 	}).Info("Preparing new router")
 	newRouter := mux.NewRouter()
 	mainRouter = newRouter
+	defaultRouter = mainRouter
 
 	loadAPIEndpoints(newRouter)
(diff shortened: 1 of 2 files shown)


mvdan:
@ilijabojanovic found that this PR doesn't actually fix the issue - closing.

PR #949, opened 2017-07-26
mw: avoid double goroutine in rate limiting

mvdan:
If UseAsyncSessionWrite is true, DefaultSessionManager.UpdateSession
does the actual store write in a goroutine. Other than that, the
function has very little logic and a json marshal.

As such, it's redundant for this middleware to also use a goroutine. It
would only lead to extra CPU use and delay in writing the session to the
store, given how two goroutines need to be scheduled instead of just
one.


buger:
Why not remove it from auth_manager.go as well?

mvdan:
You mean as well, or instead? If we remove both, then that option would do nothing at all. These are its only two uses.

buger:
Is there any usage difference between rate limit and auth middlewares? My main point that, in my view, they look almost identical. And if it is redundant here, it probably can be removed from auth manager as well.

mvdan:
(we figured it out privately in chat)

PR #952, opened 2017-07-26
jsvm: allow any objects inside config_data

### LINKED ISSUE #951: config_data is restricted to map[string]string on stable
See https://community.tyk.io/t/config-data-more-complex-values/1788.

This is fixed on master, as the type there is `map[string]interface{}`.

mvdan:
Don't limit it to map[string]string. Add a test too, with a nested
object and an integer.

The only type that matters here is configDataDef, but change the others
for consistency.

Fixes #951.


mvdan:
@buger once this is merged, I'll do another PR adding a regression test on master too.

PR #958, opened 2017-07-27
CI: bump 1.9 to RC1, use it for the tools

mvdan:


mvdan:
Wow, `go vet -tags foo` is broken if you add `-v`. RC bug uncovered? :)

mvdan:
Yup, release blocker bug in 1.9 found: https://github.com/golang/go/issues/21188

PR #659, opened 2017-04-05
Add --config-files flags to fpm commands

kohrVid:
Adding flags in order to resolve issues with the upgrade script.

PR #965, opened 2017-07-28
Package config files and generalise build scripts

excieve:
This fixes and replaces PR #659. It also generalises build scripts so that they can be easily executed from any environment while still keeping the same defaults as before.

The script `utils/release_tgz.sh` is removed as it appears to be broken and not used any more. In any case `utils/dist_build.sh` and `utils/dist_push.sh` completely replace it.


utils/dist_build.sh
```diff
@@ -1,25 +1,30 @@
...
+: ${BUILDPKGS:="1"}
+BUILDTOOLSDIR=$SOURCEBINPATH/build_tools
+
 echo Set version number
 export VERSION=$(perl -n -e'/v(\d+).(\d+).(\d+)/'' && print "$1\.$2\.$3"' version.go)
 
 echo Generating key
-[[ $(gpg --list-keys | grep -w 729EA673) ]] && echo "Key exists" || gpg --import build_key.key
+
+[[ $(gpg --list-keys | grep -w $SIGNKEY) ]] && echo "Key exists" || gpg --import $BUILDTOOLSDIR/build_key.key
```
mvdan: I know this was like this before, but could this not be `foo && bar || etc` instead of `[[ $(foo) ]] && bar || etc`?

utils/dist_build.sh
```diff
@@ -117,39 +122,44 @@ tar -pczf $amd64TGZDIR/../tyk-linux-amd64-$VERSION.tar.gz tyk.linux.amd64-$VERSI
...
+
+    echo "Removing old builds"
+    rm -f *.deb
+    rm -f *.rpm
 
-echo Creating Deb Package for AMD64
-cd $amd64TGZDIR/
-fpm -n tyk-gateway --description "$DESCRIPTION" -v $VERSION  --after-install $amd64TGZDIR/install/post_install.sh --after-remove $amd64TGZDIR/install/post_remove.sh -a amd64 -s dir -t deb ./=/opt/tyk-gateway
-fpm -n tyk-gateway --description "$DESCRIPTION" -v $VERSION  --after-install $amd64TGZDIR/install/post_install.sh --after-remove $amd64TGZDIR/install/post_remove.sh -a amd64 -s dir -t rpm ./=/opt/tyk-gateway
+    echo Creating Deb Package for AMD64
```
mvdan: This code seems repetitive - perhaps we should use a loop instead. Again, I realise this was like that before, so just a suggestion :)

utils/dist_build.sh
```diff
@@ -117,39 +122,44 @@ tar -pczf $amd64TGZDIR/../tyk-linux-amd64-$VERSION.tar.gz tyk.linux.amd64-$VERSI
...
+    cd $amd64TGZDIR/
+    fpm "${FPMCOMMON[@]}" --after-install $amd64TGZDIR/install/post_install.sh --after-remove $amd64TGZDIR/install/post_remove.sh -a amd64 -t deb "${CONFIGFILES[@]}" ./=/opt/tyk-gateway
+    fpm "${FPMCOMMON[@]}" -x **/*.deb --after-install $amd64TGZDIR/install/post_install.sh --after-remove $amd64TGZDIR/install/post_remove.sh -a amd64 -t rpm "${CONFIGFILES[@]}" ./=/opt/tyk-gateway
```
excieve: FYI: Previously RPM packages also included the DEB package inside it so now it's about half in size :)

mvdan: I love you.

utils/dist_build.sh
excieve: Makes sense! Will look into it.

buger:
Note that it need be ported to `release-2.3` branch. We do not change build scripts too often, but keep in mind.

excieve:
@mvdan Re commits: it's pretty much rewritten now so should be fine to just squash I think.
@buger There's nothing really different in the release branch so this shouldn't be a problem. When done — I'll just amend the pipelines accordingly.

I'll also check out other parts' build scripts a bit later.

excieve:
@mvdan Is there anything left to do before merging this?

utils/dist_build.sh
```diff
@@ -0,0 +1,130 @@
...
+
+echo "Prepare the release directories"
+
+export SOURCEBIN=tyk
+export CLIBIN=tyk-cli
+
+declare -A ARCHTGZDIRS
+ARCHTGZDIRS[i386]=$SOURCEBINPATH/build/i386/tgz/tyk.linux.i386-$VERSION
+ARCHTGZDIRS[amd64]=$SOURCEBINPATH/build/amd64/tgz/tyk.linux.amd64-$VERSION
+ARCHTGZDIRS[arm64]=$SOURCEBINPATH/build/arm/tgz/tyk.linux.arm64-$VERSION
```
mvdan: nit, but you could do:

```
ARCHTGZDIRS=(
    [i386]=$SOURCEBINPATH/build/i386/tgz/tyk.linux.i386-$VERSION
    [amd64]=$SOURCEBINPATH/build/amd64/tgz/tyk.linux.amd64-$VERSION
    [arm64]=$SOURCEBINPATH/build/arm/tgz/tyk.linux.arm64-$VERSION
)
```

utils/dist_build.sh
```diff
@@ -0,0 +1,130 @@
...
+
+echo "Compressing"
+for arch in ${!ARCHTGZDIRS[@]}
+do
+    cd ${ARCHTGZDIRS[$arch]}/../
+    tar -pczf ${ARCHTGZDIRS[$arch]}/../tyk-linux-$arch-$VERSION.tar.gz tyk.linux.$arch-$VERSION/
+done
+
+if [ $BUILDPKGS == "1" ]; then
+    CONFIGFILES=(--config-files /opt/tyk-gateway/apps --config-files /opt/tyk-gateway/templates --config-files /opt/tyk-gateway/middleware --config-files /opt/tyk-gateway/event_handlers --config-files /opt/tyk-gateway/js --config-files /opt/tyk-gateway/policies --config-files /opt/tyk-gateway/tyk.conf)
```
mvdan: This will make for some really complex diffs - perhaps do one `--config-files foo` per line?

Also, given that the flag name is in plural, could you not specify the flag only once at the start, or use something like a comma-separated list? (just wondering, hoping that you don't have to repeat it over and over)

utils/dist_build.sh
```diff
@@ -0,0 +1,130 @@
...
+mv $CPBINNAME_PYTHON ${ARCHTGZDIRS[amd64]}/$SOURCEBIN-python
+
+echo "Compressing"
+for arch in ${!ARCHTGZDIRS[@]}
+do
+    cd ${ARCHTGZDIRS[$arch]}/../
+    tar -pczf ${ARCHTGZDIRS[$arch]}/../tyk-linux-$arch-$VERSION.tar.gz tyk.linux.$arch-$VERSION/
+done
+
+if [ $BUILDPKGS == "1" ]; then
```
mvdan: Perhaps invert the test here and exit early, to unindent the rest of the code.

utils/dist_build.sh
excieve: From the `fpm` man page:
```
--config-files CONFIG_FILES   Mark a file in the package as being a config file. This uses 'conffiles' in debs and %config in rpm. If you have multiple files to mark as configuration files, specify this flag multiple times.  If argument is directory all files inside it will be recursively marked as config files.
```

One definition per line is fine by me — will do.

utils/dist_build.sh
excieve: 👍

buger:
@excieve looks good! Pls, prepare similar change for the dashboard. 

Thanks!

utils/dist_build.sh
excieve: 👍

buger:
@ilijabojanovic list of files/folders which should not be updated on package upgrade:

```
/opt/tyk-gateway/apps
/opt/tyk-gateway/templates
/opt/tyk-gateway/middleware
/opt/tyk-gateway/event_handlers
/opt/tyk-gateway/js
/opt/tyk-gateway/policies
/opt/tyk-gateway/tyk.conf
```

PR #970, opened 2017-07-28
coprocess_python: lock to OS thread on initialization

### LINKED ISSUE #761: GIL error when using Python bundles
Bug

When the bundles are loaded or reloaded, a GIL error shows up. This happens intermittently.
This might be related to recent changes in the API loader.

The bundles should be loaded correctly.

Load a Python plugin, start/stop Tyk a few times.

All versions.

### LINKED ISSUE #969: GIL error when starting Tyk
Bug

When Tyk starts it throws a Python GIL error, when loading plugins. This happens intermittently.

This is related to #761, which was for reload errors.

Add two or three API definitions that use Python plugin bundles, run Tyk.

All

matiasinsaurralde:
Fixes #969, complementary to #758.

### DIFF
--- a/coprocess_python.go
+++ b/coprocess_python.go
@@ -39,8 +39,6 @@ static int Python_LoadDispatcher() {
 	PyObject *module_name = PyUnicode_FromString( dispatcher_module_name );
 	dispatcher_module = PyImport_Import( module_name );
 
-	Py_DECREF(module_name);
-
 	if( dispatcher_module == NULL ) {
 		PyErr_Print();
 		return -1;
@@ -86,18 +84,20 @@ static void Python_HandleMiddlewareCache(char* bundle_path) {
 }
 
 static int Python_NewDispatcher(char* middleware_path, char* event_handler_path, char* bundle_paths) {
+	PyThreadState*  mainThreadState = PyEval_SaveThread();
+	gilState = PyGILState_Ensure();
 	if( PyCallable_Check(dispatcher_class) ) {
 		dispatcher_args = PyTuple_Pack( 3, PyUnicode_FromString(middleware_path), PyUnicode_FromString(event_handler_path), PyUnicode_FromString(bundle_paths) );
 		dispatcher = PyObject_CallObject( dispatcher_class, dispatcher_args );
 
-		Py_DECREF(dispatcher_args);
-
 		if( dispatcher == NULL) {
 			PyErr_Print();
+			PyGILState_Release(gilState);
 			return -1;
 		}
 	} else {
 		PyErr_Print();
+		PyGILState_Release(gilState);
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


mvdan:
The other PR, #758, called `runtime.LockOSThread()` in a very clear new goroutine. This one, however, is not a clear new goroutine, nor does it call `UnlockOSThread`.

How do you know that you're not locking the main goroutine - or another long-lived goroutine - to the OS thread? Have you looked into this?

mvdan:
Also, doesn't this actually fix https://github.com/TykTechnologies/tyk/issues/969, the new issue?

matiasinsaurralde:
@mvdan yes, I wrote the wrong ID, it's actually 969.

mvdan:
We spoke privately, and we agreed that it would be best to separate the python GIL initialization work into its own, short-lived goroutine - which can then use `LockOSThread`. That way, it's easy to prove that we're not locking anything else to the OS thread.

matiasinsaurralde:
I've been heavily testing this and came up with some enhancements like using the thread state API. This approach gives me zero crashes on Tyk initializations and reloads, finally!
To avoid refactoring all the CP logic I added some signaling inside the initialization of the Python driver (as suggested by @mvdan), so that we can return `CoProcessInit` after the cgo calls are done.
Would be great if somebody could test this up, the original issue occurs intermittently when you load more than one Python plugin bundle.

buger:
@ilijabojanovic ^

coprocess_python.go
```diff
@@ -295,20 +296,20 @@ func NewCoProcessDispatcher() (coprocess.Dispatcher, error) {
...
+		PythonLoadDispatcher()
+		dispatcher, err = PythonNewDispatcher(middlewarePath, eventHandlerPath, bundlePaths)
+		if err != nil {
+			log.WithFields(logrus.Fields{
+				"prefix": "coprocess",
+			}).Error(err)
+		}
+		pythonInitDone <- true
+	}()
+	<-pythonInitDone
```
mvdan: You probably want to make `pythonInitDone` an error channel, so that the error can be returned or otherwise handled.

coprocess_python.go
```diff
@@ -187,6 +185,9 @@ const CoProcessName = apidef.PythonDriver
 // MessageType sets the default message type.
 var MessageType = coprocess.ProtobufMessage
 
+// pythonInitDone is used to signal the end of Python initialization step
+var pythonInitDone = make(chan bool)
```
mvdan: Is there a reason to make this a global?

If this is to reuse the channel allocation, are we sure it won't be used more than once concurrently?

coprocess_python.go
mvdan: Err, I see that you are actually returning it, but by writing to the parent's error variable instead. This is a bit unintuitive, and will break easily with data races if we ever touch this code. Better to explicitly pass the error via the channel, returning nil if there was no error.

coprocess_python.go
```diff
@@ -86,18 +84,20 @@ static void Python_HandleMiddlewareCache(char* bundle_path) {
 }
 
 static int Python_NewDispatcher(char* middleware_path, char* event_handler_path, char* bundle_paths) {
+	PyThreadState*  mainThreadState = PyEval_SaveThread();
+	gilState = PyGILState_Ensure();
```
mvdan: Why do we need both `runtime.LockOSThread` and this? Is this an improvement that's unrelated to the fix?

matiasinsaurralde: We were already using `PyEval_ReleaseLock` but this is the proper/better way of using it, and this replaces the lock operations.

matiasinsaurralde:
Updated!

matiasinsaurralde:
@mvdan I think that to test this you need the full scenario, multiple bundles, etc. So I suggest using this in its current status and add tests on my refactoring branch (which I would update with this patch as well).

mvdan:
Sounds good - I'm mostly concerned about regressions, given how this code is subtle and it even took us two tries to fix it fully. As long as the test makes it into 2.4, we're good :)

ilijabojanovic:
@buger approved for merge

PR #972, opened 2017-08-01
revproxy: set Tyk/VERSION user agent if none set

### LINKED ISSUE #422: Add default user agent config option
Enhancement
If the User Agent header is deleted in an API endpoint, the default User Agent set by the "net/http" package in Go ("Go-http-client/1.1") is added instead which can cause issues for users with strict security settings, &c. The existing work-around for users is to simultaneously remove and add the header in the endpoint.
**What is the expected behaviour?**
Some users expect to the deleted header to stay deleted. The Gateway should at least override the default the string with an empty string if one isn't set by the user. 
**If the current behaviour is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
The current behaviour is the default behaviour.
All versions are affected as this behaviour is caused by the "net/http" package

mvdan:
Note how set but empty user agents are not set to the default.

Fixes #422.

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -12,6 +12,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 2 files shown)


gateway_test.go
```diff
@@ -991,3 +992,46 @@ func TestListenPathTykPrefix(t *testing.T) {
...
+		{"", `^$`},
+		// set and not empty
+		{"SomeAgent", `^SomeAgent$`},
+	}
+	for _, tc := range tests {
+		rec := httptest.NewRecorder()
+		req := testReq(t, "GET", "/sample", nil)
+		if s, ok := tc.sent.(string); ok {
+			req.Header.Set("User-Agent", s)
+		}
```
mvdan: If anyone wonders, I tried using `*string` but it made the table-based tests impossible - you can't do `&"foo"`.

mvdan:
@buger PTAL, the CI failure was a flake (https://github.com/TykTechnologies/tyk/issues/987).

Issue #915, opened 2017-07-11
Unify mainRouter and defaultRouter


mvdan:
There are subtle differences between the two at the moment. There should be no reason to separate them, like there is with `controlRouter`.

/cc @buger

PR #982, opened 2017-08-04
Fix reloads when using config.HttpServerOptions.OverrideDefaults

### LINKED ISSUE #940: Reloading may not work if config.HttpServerOptions.OverrideDefaults is true

matiasinsaurralde:
Fixes #940 and follows the suggestion described in #915. So this contains two changes.
After a few tests I've discovered that we need to update the handler of the custom HTTP server for the new router to take effect.

### DIFF
--- a/main.go
+++ b/main.go
@@ -61,7 +61,8 @@ var ApiSpecRegister map[string]*APISpec //make(map[string]*APISpec)
 var keyGen = DefaultKeyGenerator{}
 
 var mainRouter *mux.Router
-var defaultRouter *mux.Router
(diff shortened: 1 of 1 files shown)


buger:
Closing in favor of https://github.com/TykTechnologies/tyk/pull/984. Note that I just backported `master` changes. Basic idea is that now instead of setting "handler" to the server, it sets `DefaultServeMux`

PR #983, opened 2017-08-04
Add option to configure default proxy timeout

buger:
Should replace https://github.com/TykTechnologies/tyk/pull/976


ilijabojanovic:
@buger approved for merge

PR #984, opened 2017-08-04
Fix reloads when http.override_defaults turned on

buger:
Ported similar changes from master


ilijabojanovic:
@buger approved for merge. I will update auto tests

PR #999, opened 2017-08-08
Lua fix for Go 1.8 builds

matiasinsaurralde:
This was reported by @excieve.
It's a small fix because of cgo changes.

### DIFF
--- a/coprocess_lua.go
+++ b/coprocess_lua.go
@@ -146,7 +146,7 @@ func (d *LuaDispatcher) Reload() {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


mvdan:
Yep, this is the same fix that we did back in January in the master branch.

mvdan:
Though, I wonder, does this matter? I thoght we were building 2.3.X with Go 1.7.X for the official binaries. Bumping to 1.8 for a bugfix release seems unnecessary.

excieve:
@mvdan This was a suggestion by @buger to use the latest 1.8.x on staging pipelines.

PR #1004, opened 2017-08-09
RawLogs should not print to stdout in tests

### LINKED ISSUE #1002: JSVM log tests now spit output to terminal
```
$ go test
foo
{"x": "y"}
PASS
ok      github.com/TykTechnologies/tyk  2.136s
```

@buger this is likely your recent `rawlog` fix. I might have time for this today.

buger:
Fix #1002

@mvdan Since this affects only tests, I will keep it in master without backporting.

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -242,6 +242,7 @@ type JSVM struct {
 	VM      *otto.Otto
 	Timeout time.Duration
 	Log     *logrus.Logger // logger used by the JS code
+	RawLog  *logrus.Logger // logger used by `rawlog` func to avoid formatting
 }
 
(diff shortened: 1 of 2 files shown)


plugins.go
```diff
@@ -242,6 +242,7 @@ type JSVM struct {
 	VM      *otto.Otto
 	Timeout time.Duration
 	Log     *logrus.Logger // logger used by the JS code
+	RawLog  *logrus.Logger // logger used by the JS code
```
mvdan: these two comments probably need clarification now, i.e. that one is used in `log` and the other in `rawlog`

PR #1007, opened 2017-08-10
servdisc: unindent and simplify code

mvdan:
We can simplify:

	if x {
	} else {
		if y {
		} else {
		}
	}

Into:

	if x {
	} else if y {
	} else {
	}

To unindent the second and third blocks of code.

Also remove an unnecessary nil check around a slice range.


mvdan:
ping @buger

PR #1008, opened 2017-08-10
fire AuthFailed if HMAC middleware fails auth

### LINKED ISSUE #963: mw: HMAC failure does not trigger an EventAuthFailure
bug

If I call an endpoint with an invalid `Authorization` token it does not trigger a EventAuthFailure event.

I would expect an EventAuthFailure event to fire as is the case for the `JWT` or the `Auth Token` (although I might be wrong about this so please correct me before I make a PR for this).

If I configure a WebHook for the AuthFailure event it triggers for `JWT` and `Auth Token` but not for `HMAC` if I send requests with invalid Authorization headers 

I am using the latest Tyk gateway from docker hub (https://hub.docker.com/r/tykio/tyk-gateway/tags/)

othane:
this fixes #963 so that any failure in the HMAC authorization will
fire a AuthFailed event so users can write webhooks to log possible
attacks

### DIFF
--- a/mw_hmac_test.go
+++ b/mw_hmac_test.go
@@ -10,8 +10,12 @@ import (
 	"net/url"
 	"regexp"
 	"strings"
+	"sync"
 	"testing"
 	"time"
+
+	"github.com/TykTechnologies/tyk/apidef"
+	"github.com/TykTechnologies/tyk/config"
 
 	"github.com/justinas/alice"
 )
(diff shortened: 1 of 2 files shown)


mvdan:
You were able to write the tests - awesome :) will be reviewing this now.

mw_hmac_test.go
```diff
@@ -69,10 +73,48 @@ func getHMACAuthChain(spec *APISpec) http.Handler {
...
 	session := createHMACAuthSession()
 
+	// Should not receive an AuthFailure event
+	var eventWG sync.WaitGroup
+	var eventAuthFailState bool = false
+	eventWG.Add(1)
+	cb := func(em config.EventMessage) {
+		eventAuthFailState = true
+		eventWG.Done()
+	}
```
mvdan: You could simplify the logic here by having `waitTimeout` return whether the waitgroup finished or timed out. In the former case, it's the same as if `eventAuthFailState` was set to true.

So you can get rid of the bool, and use `if !waitTimedOut(&eventWG, 20*time.Millisecond) { ...` below.

othane: good point :) ... I have updated the PR ... thanks :+1:

mvdan:
Thanks again!

othane:
Cool, thanks for the help, it was nice to play with Go :) 

Any idea which release this is likely be in ?

mvdan:
2.3.8 is going out any moment now, and it includes many important fixes. We've been testing it for a bit already though, so this change isn't critical enough to warrant delaying the release.

It will for sure be included in 2.4, which should come out in the next month or two.

PR #1018, opened 2017-08-21
all: remove the last few direct r.RemoteAddr uses

### LINKED ISSUE #1017: Remote address context variable should be aware of proxies
Bug

Currently `remote_addr` context variable for header/body transformations only accounts for the last direct hop (or whatever golang http request's `RemoteAddr` provides). This doesn't serve much purpose when the gateway is accessed through a reverse proxy or a load balancer (such as on Tyk Cloud).

When `X-Real-Ip` or `X-Forwarded-For` headers are present the `remote_addr` context variable should contain the first hop that is not on a private network if it's the latter and just the whole value if it's the former. It may be beneficial to make this configurable since those headers are easy to spoof if the source is not trusted.

For the Tyk Cloud, infrastructure should also make sure the real IP is forwarded from the load balancers.

1. On Tyk Cloud or any other environment behind a load balancer create an API with a request header transformation, which adds a header with `$tyk_context.remote_addr` value. The API must display request headers (e.g. httpbin's `headers` resource).
2. Make sure context variables are enabled for this API.
3. Make a request to it and observe custom headers. The added header will show the IP of last hop before the gateway instead of the client's IP.

2.3.8 and most likely all the previous ones.

mvdan:
They should all be using either requestIP or requestAddrs, depending on
whether they want the real IP or all the addresses of all the hops.

This fixes an issue found in the context vars code. While at it, fix a
couple of other occurences.

Fixes #1017.

### DIFF
--- a/instrumentation_handlers.go
+++ b/instrumentation_handlers.go
@@ -50,7 +50,7 @@ func InstrumentationMW(next http.Handler) http.Handler {
 
 		next.ServeHTTP(w, r)
 		job.EventKv("called", health.Kvs{
-			"from_ip":  r.RemoteAddr,
+			"from_ip":  requestIP(r),
 			"method":   r.Method,
 			"endpoint": r.URL.Path,
 			"raw_url":  r.URL.String(),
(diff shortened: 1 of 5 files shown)


buger:
@excieve feel free deploy to stage and confirm that it fixes your issue.

mvdan:
(note that `X-Real-IP` won't work just yet, that'd be #1019)

buger:
@excieve i see issue with deploy, any details on this PR?

mvdan:
@buger we decided to try again today, under the assumption that it was buddy being flaky yesterday. so still ongoing.

excieve:
@mvdan @buger Deployed it on staging and it appears to work fine now.

mvdan:
Thanks for testing!

PR #1020, opened 2017-08-21
revproxy: don't include port in X-Forwarded-For

mvdan:
In f5e90899 we reintroduced the function that joined all the hops in a
request. However, a mistake was made - the hops are just IPs, not
addresses with a port.

Get the code back to how it was in 2.3.x, which removed the port from
r.RemoteAddr before appending it to the list. And rename the helper
function to clarify what it does.


mvdan:
(found this while working on the other IP stuff PR, hence the reviewers)

PR #1022, opened 2017-08-22
all: unindent some more bits of code

mvdan:
Found with mvdan.cc/unindent. Mostly merging chained ifs.


buger:
Nice, happy to see your library in the wild :)

PR #1027, opened 2017-08-24
session: undo MetaData map[string]string switch

mvdan:
This was seen as a safe change because the gateway never uses values
other than strings, nor does it have tests that use it any other way.

However, the dashboard does use nested objects with strings in them.
Thus this change broke the dashboard.

Revert the change, updating the old map conversion TODO and adding
another one in the fmt.Sprintf call, since we probably want something
stricter and more predictable.

There is still work to do here, as this still allows for too many types.
However, reverting is the right decision now to un-break the dashboard
for 2.4.

### DIFF
--- a/mw_jwt.go
+++ b/mw_jwt.go
@@ -235,7 +235,7 @@ func (k *JWTMiddleware) processCentralisedJWT(r *http.Request, token *jwt.Token)
 
 		if err == nil {
 			session = newSession
-			session.MetaData = map[string]string{"TykJWTSessionID": sessionID}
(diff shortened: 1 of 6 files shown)


mvdan:
@buger this is very similar to your fix for stable, but smaller and having in mind other cleanups that happened in master since.

ilijabojanovic:
@buger approved for merge

ilijabojanovic:
Original TASK:
https://github.com/TykTechnologies/tyk-analytics/issues/309

PR #1028, opened 2017-08-24
all: disable and deprecate health check

### LINKED ISSUE #784: Deprecate HealthCheck API
HealthCheck API is unreliable and should be deprecated.

It should be turned off by default, official config should not mention this options, and docs should explicitly say that it was deprecated.

mvdan:
Disable it by default and remove from the sample config. Also show a
warning if used.

Fixes #784.


buger:
👍 ready for testing

ilijabojanovic:
[36mtyk-gateway_1  |[0m time="Sep  4 13:45:42" level=debug msg="Health Checker initialised." 
[36mtyk-gateway_1  |[0m time="Sep  4 13:45:42" level=warning msg="The Health Checker is deprecated and we do no longer recommend its use." 

@buger approved for merge

PR #1029, opened 2017-08-24
revproxy: return a 503 if all hosts are down

### LINKED ISSUE #836: Modify the load balancer to return a proper error if all hosts are down

If all hosts are down, the load balancer will route to any of them.

Instead of that, return a meaningful error response like "all hosts are down, please try again later".

Affects 2.3.x and master.

Note that I haven't encountered this nor reproduced this, but it's fairly obvious from the code that this is what we're doing - and we shouldn't do it.

/cc @buger

mvdan:
Instead of directing to any one of the hosts.

Due to the nature of how the proxy is implemented, the load balancing
happens in the director. At that stage, we can only change the target
URL, not directly write to the response.

Hence, we must start up a local http server that will simply write the
503 and an error message. Use sync.Once to make sure it's not started
multiple times, and reuse it.

Fixes #836.

### DIFF
--- a/host_checker_test.go
+++ b/host_checker_test.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"bytes"
+	"net/http/httptest"
+	"net/url"
 	"sync"
 	"testing"
 	"text/template"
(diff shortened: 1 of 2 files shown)


mvdan:
Note that we probably could refactor this to avoid the silly http server for internal use. But for now, this isn't too much added complexity and does the job. Once we have the test, if we have time, we can rework it.

buger:
👍 

ilijabojanovic:
@buger approved for merge

PR #1042, opened 2017-08-29
Response overrides

lonelycode:
This just adds support to override the response from the JSVM in terms of the Body, Headers and Status code, it adds more support than the current version which just allows overriding an error.

@matiasinsaurralde I'll need your help with the CP integration, I think the ReturnOverrides are only used in the AuthChecker at the moment, and not in Otto, so they will start diverging in behaviour?

### DIFF
--- a/coprocess_id_extractor.go
+++ b/coprocess_id_extractor.go
@@ -39,7 +39,7 @@ func (e *BaseExtractor) ExtractAndCheck(r *http.Request) (sessionID string, retu
 	log.WithFields(logrus.Fields{
 		"prefix": "idextractor",
 	}).Error("This extractor doesn't implement an extraction method, rejecting.")
-	return "", ReturnOverrides{403, "Key not authorised"}
+	return "", ReturnOverrides{ResponseCode:403, ResponseError:"Key not authorised"}
 }
 
 // PostProcess sets context variables and updates the storage.
--- a/coprocess_id_extractor.go
+++ b/coprocess_id_extractor.go
@@ -39,7 +39,7 @@ func (e *BaseExtractor) ExtractAndCheck(r *http.Request) (sessionID string, retu
 	log.WithFields(logrus.Fields{
 		"prefix": "idextractor",
 	}).Error("This extractor doesn't implement an extraction method, rejecting.")
-	return "", ReturnOverrides{ResponseCode:403, ResponseError:"Key not authorised"}
+	return "", ReturnOverrides{ResponseCode: 403, ResponseError: "Key not authorised"}
 }
 
 // PostProcess sets context variables and updates the storage.
--- a/mw_virtual_endpoint.go
+++ b/mw_virtual_endpoint.go
@@ -174,19 +174,19 @@ func (d *VirtualEndpoint) ServeHTTPForCache(w http.ResponseWriter, r *http.Reque
 
 	log.Debug("JSVM Virtual Endpoint execution took: (ns) ", time.Now().UnixNano()-t1)
 
-	copiedResponse, code := ForceResponse(w, r, &newResponseData, d.Spec, session, false)
+	copiedResponse := forceResponse(w, r, &newResponseData, d.Spec, session, false)
 
-	go d.sh.RecordHit(r, 0, code, copiedRequest, copiedResponse)
+	go d.sh.RecordHit(r, 0, copiedResponse.StatusCode, copiedRequest, copiedResponse)
 
 	return copiedResponse
 
 }
 
-func ForceResponse(w http.ResponseWriter,
+func forceResponse(w http.ResponseWriter,
 	r *http.Request,
 	newResponseData *VMResponseObject,
 	spec *APISpec,
-	session *SessionState, isPre bool) (*http.Response, int) {
+	session *SessionState, isPre bool) *http.Response {
 	responseMessage := []byte(newResponseData.Response.Body)
 
 	// Create an http.Response object so we can send it tot he cache middleware
@@ -232,10 +232,10 @@ func ForceResponse(w http.ResponseWriter,
 	newResponse.Body = ioutil.NopCloser(&bodyBuffer)
 	copiedRes.Body = ioutil.NopCloser(bodyBuffer2)
 
-	HandleForcedResponse(w, newResponse, session)
+	handleForcedResponse(w, newResponse, session)
 
 	// Record analytics
-	return copiedRes, newResponse.StatusCode
+	return copiedRes
 }
 
 // ProcessRequest will run any checks on the request on the way through the system, return an error to have the chain fail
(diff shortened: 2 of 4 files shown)


buger:
Can you pls link original issue or forum post?

lonelycode:
There isn't one, see slack.

lonelycode:
I've created one here: #1043

coprocess_id_extractor.go
```diff
@@ -39,7 +39,7 @@ func (e *BaseExtractor) ExtractAndCheck(r *http.Request) (sessionID string, retu
 	log.WithFields(logrus.Fields{
 		"prefix": "idextractor",
 	}).Error("This extractor doesn't implement an extraction method, rejecting.")
-	return "", ReturnOverrides{403, "Key not authorised"}
+	return "", ReturnOverrides{ResponseCode:403, ResponseError:"Key not authorised"}
```
mvdan: this looks like it's missing `gofmt`

mw_virtual_endpoint.go
```diff
@@ -174,6 +174,19 @@ func (d *VirtualEndpoint) ServeHTTPForCache(w http.ResponseWriter, r *http.Reque
...
 
+	copiedResponse, code := ForceResponse(w, r, &newResponseData, d.Spec, session, false)
+
+	go d.sh.RecordHit(r, 0, code, copiedRequest, copiedResponse)
+
+	return copiedResponse
+
+}
+
+func ForceResponse(w http.ResponseWriter,
```
mvdan: this doesn't need exporting

mw_virtual_endpoint.go
```diff
@@ -223,13 +232,10 @@ func (d *VirtualEndpoint) ServeHTTPForCache(w http.ResponseWriter, r *http.Reque
...
 
-	d.HandleResponse(w, newResponse, session)
+	HandleForcedResponse(w, newResponse, session)
 
 	// Record analytics
-	go d.sh.RecordHit(r, 0, newResponse.StatusCode, copiedRequest, copiedResponse)
-
-	return copiedRes
-
+	return copiedRes, newResponse.StatusCode
```
mvdan: Do you need the two return values? `StatusCode` is also part of the response copy.

mw_virtual_endpoint.go
```diff
@@ -245,7 +251,12 @@ func (d *VirtualEndpoint) ProcessRequest(w http.ResponseWriter, r *http.Request,
 }
 
 func (d *VirtualEndpoint) HandleResponse(rw http.ResponseWriter, res *http.Response, ses *SessionState) error {
+	// Externalising this from the MW so we can re-use it elsewhere
+	return HandleForcedResponse(rw, res, ses)
 
+}
+
+func HandleForcedResponse(rw http.ResponseWriter, res *http.Response, ses *SessionState) error {
```
mvdan: This can also be unexported

plugins.go
```diff
@@ -224,10 +225,24 @@ func (d *DynamicMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reques
 		"prefix": "jsvm",
 	}).Debug("JSVM middleware execution took: (ns) ", time.Now().UnixNano()-t1)
 
-	if newRequestData.Request.ReturnOverrides.ResponseCode != 0 {
+	if newRequestData.Request.ReturnOverrides.ResponseCode != 0 && newRequestData.Request.ReturnOverrides.ResponseCode > 300 {
```
mvdan: all values `> 300` are `!= 0`, so the first half can go.

Also, if you meant to do this for all error status codes, you probably want `>= 400`. Otherwise, I'm not sure why the `> 300` was chosen.

plugins.go
```diff
@@ -224,10 +225,24 @@ func (d *DynamicMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reques
...
+		responseObject := VMResponseObject{
+			Response: ResponseObject{
+				Body: newRequestData.Request.ReturnOverrides.ResponseError,
+				Code: newRequestData.Request.ReturnOverrides.ResponseCode,
+				Headers: newRequestData.Request.ReturnOverrides.ResponseHeaders,
+			},
+		}
+
+		ForceResponse(w, r, &responseObject, d.Spec, session, d.Pre)
+		return nil, 666
```
mvdan: Use the new named constant instead of `666`, to make this clearer

plugins.go
```diff
@@ -224,10 +225,24 @@ func (d *DynamicMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reques
 		"prefix": "jsvm",
 	}).Debug("JSVM middleware execution took: (ns) ", time.Now().UnixNano()-t1)
 
-	if newRequestData.Request.ReturnOverrides.ResponseCode != 0 {
+	if newRequestData.Request.ReturnOverrides.ResponseCode != 0 && newRequestData.Request.ReturnOverrides.ResponseCode > 300 {
 		return errors.New(newRequestData.Request.ReturnOverrides.ResponseError), newRequestData.Request.ReturnOverrides.ResponseCode
 	}
 
+	if newRequestData.Request.ReturnOverrides.ResponseCode != 0 && newRequestData.Request.ReturnOverrides.ResponseCode < 300 {
```
mvdan: Also, this confuses me. All values `> 300` are discarded above, and here only `< 300` are handled. Why do we treat 300, `StatusMultipleChoices`, differently?

lonelycode:
Have updated with suggested changes.

bmudda:
+1

lonelycode:
@mvdan bump

mw_virtual_endpoint.go
```diff
@@ -245,7 +251,12 @@ func (d *VirtualEndpoint) ProcessRequest(w http.ResponseWriter, r *http.Request,
 }
 
 func (d *VirtualEndpoint) HandleResponse(rw http.ResponseWriter, res *http.Response, ses *SessionState) error {
+	// Externalising this from the MW so we can re-use it elsewhere
+	return handleForcedResponse(rw, res, ses)
+
+}
 
+func handleForcedResponse(rw http.ResponseWriter, res *http.Response, ses *SessionState) error {
```
mvdan: This func never errors. Either remove the return value, or actually pass on errors like the one from `io.Copy`.

plugins.go
```diff
@@ -224,10 +225,24 @@ func (d *DynamicMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reques
 		"prefix": "jsvm",
 	}).Debug("JSVM middleware execution took: (ns) ", time.Now().UnixNano()-t1)
 
-	if newRequestData.Request.ReturnOverrides.ResponseCode != 0 {
+	if newRequestData.Request.ReturnOverrides.ResponseCode >= 400 {
 		return errors.New(newRequestData.Request.ReturnOverrides.ResponseError), newRequestData.Request.ReturnOverrides.ResponseCode
 	}
 
+	if newRequestData.Request.ReturnOverrides.ResponseCode != 0 && newRequestData.Request.ReturnOverrides.ResponseCode <= 300 {
```
mvdan: I still don't understand why we treat `300` differently from other 3xx codes here. Don't you mean `< 300`, to capture all 2xx and 1xx codes?

lonelycode:
Have addressed the two review issues... will merge when ready

mvdan:
Please squash-merge, there are quite a few fixup commits

plugins.go
buger: Actually a good point, because our co-process functionality (which should be synced with this one, just use `> 0` https://github.com/TykTechnologies/tyk/blob/master/coprocess.go#L274

mvdan: My point was simply about the `3xx` inconsistency, but your point seems correct too. @lonelycode?

bitsofinfo:
Please merge! :)

buger:
@bitsofinfo this was merged long ago

bitsofinfo:
duh, late night :)

PR #1044, opened 2017-08-29
Jq transform

klyr:
Hi,

As discussed by email, here is the pull request in order to facilitate the discussion.
This is not rebased on master, and this is not production read at all. It's just meant as a starting point as discussion.

This adds a new transformation middleware, with used the well known [jq](https://stedolan.github.io/jq/) JSON processor library.

The current implementation takes a jq filter and does the transformation, it can put variables from the body to the contextVars in order to be use in other middlewares.

In the response transformation middleware, one can also use the headers from the original request and from the response.

Please feel free to comment.

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -20,6 +20,7 @@ import (
 	"github.com/TykTechnologies/tyk/config"
 
 	"github.com/klyr/go-jq"
+	"sync"
 )
 
 const (
@@ -114,6 +115,7 @@ type TransformSpec struct {
 
 type TransformJQSpec struct {
 	apidef.TransformJQMeta
+	sync.Mutex
 	JQFilter *jq.JQ
 }
 
--- a/api_definition.go
+++ b/api_definition.go
@@ -19,7 +19,7 @@ import (
 	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/config"
 
-	"github.com/mgood/go-jq"
+	"github.com/klyr/go-jq"
 )
 
 const (
@@ -516,7 +516,8 @@ func (a APIDefinitionLoader) compileTransformPathSpec(paths []apidef.TemplateMet
 }
 
 func (a *APIDefinitionLoader) compileJQFilter(filter string) (*jq.JQ, error) {
-	return jq.NewJQ(filter)
+	v, err := jq.NewJQ(filter)
+	return v, err
 }
 
 func (a *APIDefinitionLoader) compileTransformJQPathSpec(paths []apidef.TransformJQMeta, stat URLStatus) []URLSpec {
--- a/api_definition.go
+++ b/api_definition.go
@@ -18,6 +18,8 @@ import (
 
 	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/config"
+
+	"github.com/mgood/go-jq"
 )
 
 const (
@@ -37,9 +39,11 @@ const (
 	BlackList
 	Cached
 	Transformed
+	TransformedJQ
 	HeaderInjected
 	HeaderInjectedResponse
 	TransformedResponse
+	TransformedJQResponse
 	HardTimeout
 	CircuitBreaker
 	URLRewrite
@@ -65,6 +69,8 @@ const (
 	StatusCached                   RequestStatus = "Cached path"
 	StatusTransform                RequestStatus = "Transformed path"
 	StatusTransformResponse        RequestStatus = "Transformed response"
+	StatusTransformJQ              RequestStatus = "Transformed path with JQ"
+	StatusTransformJQResponse      RequestStatus = "Transformed response with JQ"
 	StatusHeaderInjected           RequestStatus = "Header injected"
 	StatusMethodTransformed        RequestStatus = "Method Transformed"
 	StatusHeaderInjectedResponse   RequestStatus = "Header injected on response"
@@ -82,26 +88,33 @@ const (
 // path is on any of the white, plack or ignored lists. This is generated as part of the
 // configuration init
 type URLSpec struct {
-	Spec                    *regexp.Regexp
-	Status                  URLStatus
-	MethodActions           map[string]apidef.EndpointMethodMeta
-	TransformAction         TransformSpec
-	TransformResponseAction TransformSpec
-	InjectHeaders           apidef.HeaderInjectionMeta
-	InjectHeadersResponse   apidef.HeaderInjectionMeta
-	HardTimeout             apidef.HardTimeoutMeta
-	CircuitBreaker          ExtendedCircuitBreakerMeta
-	URLRewrite              apidef.URLRewriteMeta
-	VirtualPathSpec         apidef.VirtualMeta
-	RequestSize             apidef.RequestSizeMeta
-	MethodTransform         apidef.MethodTransformMeta
-	TrackEndpoint           apidef.TrackEndpointMeta
-	DoNotTrackEndpoint      apidef.TrackEndpointMeta
+	Spec                      *regexp.Regexp
+	Status                    URLStatus
+	MethodActions             map[string]apidef.EndpointMethodMeta
+	TransformAction           TransformSpec
+	TransformResponseAction   TransformSpec
+	TransformJQAction         TransformJQSpec
+	TransformJQResponseAction TransformJQSpec
+	InjectHeaders             apidef.HeaderInjectionMeta
+	InjectHeadersResponse     apidef.HeaderInjectionMeta
+	HardTimeout               apidef.HardTimeoutMeta
+	CircuitBreaker            ExtendedCircuitBreakerMeta
+	URLRewrite                apidef.URLRewriteMeta
+	VirtualPathSpec           apidef.VirtualMeta
+	RequestSize               apidef.RequestSizeMeta
+	MethodTransform           apidef.MethodTransformMeta
+	TrackEndpoint             apidef.TrackEndpointMeta
+	DoNotTrackEndpoint        apidef.TrackEndpointMeta
 }
 
 type TransformSpec struct {
 	apidef.TemplateMeta
 	Template *textTemplate.Template
+}
+
+type TransformJQSpec struct {
+	apidef.TransformJQMeta
+	JQFilter *jq.JQ
 }
 
 type ExtendedCircuitBreakerMeta struct {
@@ -502,6 +515,38 @@ func (a APIDefinitionLoader) compileTransformPathSpec(paths []apidef.TemplateMet
 	return urlSpec
 }
 
+func (a *APIDefinitionLoader) compileJQFilter(filter string) (*jq.JQ, error) {
+	return jq.NewJQ(filter)
+}
+
+func (a *APIDefinitionLoader) compileTransformJQPathSpec(paths []apidef.TransformJQMeta, stat URLStatus) []URLSpec {
+	urlSpec := []URLSpec{}
+
+	log.Debug("Checking for JQ tranform paths ...")
+	for _, stringSpec := range paths {
+		newSpec := URLSpec{}
+		a.generateRegex(stringSpec.Path, &newSpec, stat)
+		newTransformSpec := TransformJQSpec{TransformJQMeta: stringSpec}
+
+		var err error
+		newTransformSpec.JQFilter, err = a.compileJQFilter(stringSpec.Filter)
+
+		if stat == TransformedJQ {
+			newSpec.TransformJQAction = newTransformSpec
+		} else {
+			newSpec.TransformJQResponseAction = newTransformSpec
+		}
+
+		if err == nil {
+			urlSpec = append(urlSpec, newSpec)
+		} else {
+			log.Error("JQ Filter load failure! Skipping transformation: ", err)
+		}
+	}
+
+	return urlSpec
+}
+
 func (a APIDefinitionLoader) compileInjectedHeaderSpec(paths []apidef.HeaderInjectionMeta, stat URLStatus) []URLSpec {
 	// transform an extended configuration URL into an array of URLSpecs
 	// This way we can iterate the whole array once, on match we break with status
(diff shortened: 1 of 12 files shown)


mw_transform_jq.go
```diff
@@ -0,0 +1,120 @@
...
+	bodyBuffer.Write(body)
+	bodyBuffer.WriteString("]")
+
+	err = t.JQFilter.HandleJson(bodyBuffer.String())
+	if err != nil {
+		return errors.New("Input is not a valid JSON")
+	}
+
+	// First element is the transformed body
+	if t.JQFilter.Next() {
```
buger: I'm not sure why we have 2 steps here (t.JQFilter.Next()). Can you clarify it?

buger:
Sounds exciting! Can you add some of the documentation here, as a comment.

mw_transform_jq.go
klyr: The first element is the actual transformed body transformed by JQ.

If the JQ filter provides a second element (another JSON object) with specific "output_headers" and/or "output_vars" object, it will be given back to tyk in order to update he headers of the tyk query/reponse, or feed the context vars.

For example, let's take the following jq filter:

    {mynewfield: .data}

it will return a JSON object with only the data field.

Now, if the filter is:

    {mynewfield: .data}, {output_headers: {x_name: .name}, output_vars: {age: .userAge}}

it will add the `x_name` header to the http query or response, and add the `age` variable to the context vars, which will be available with the `$tyk_context.jq_output_var_age` in url rewrites for example.

Thinking of it, I now think that the output_headers is useless, and that the output_vars mechanism can be used to add specific HTTP headers (if the modify_headers middleware is context vars aware).

klyr:
Hi @buger,

Are you missing some more documentation ? Do you have any comments or idea, in order to improve it, or make it useful to integrate into Tyk ?

Just to ping you :)

mw_transform_jq.go
buger: Missed your message! 

Thank you for clarification, I was confused because our transformation middleware supports only response or request body modification. And your one also supports headers and context variables.

I'm not sure if fully understand use case of setting context variables or headers based on the body. Is it something you actually use by your own?

klyr: Yes, it's a use case we can have. I understand it's very specific to our usecase.
For example, we can imagine having a "deviceID" field inside the JSON body, that we want to put as a X-DEVICE-ID http header for the upstream server.

buger: Additional queston, in your examples you do: `. as [$reqContext, $resHeaders, $body] `, which is a bit cryptic. Can we use proxy object instead, smth like .tyk.body, .tyk.context and etc?

klyr: This is a way to give metadata (like the headers or the request context) other than the body itself, to the jq filter.

So instead of just giving the actual body to the jq filter, Tyk gives the body, inside a 3 elements array, along with the requestContext and the response headers.

That is why, the jq filter have to begin with `. as [$reqContext, $resHeaders, $body]` in order to assign each element of the array to a variable that we can further use in the jq filter.

I agree it's a trick, and I found no other way to do it at the moment.

buger: Not sure if you got my idea. `.` in jq treated as a `this` or input data. What if instead of passing array to it, we will pass `{ "body": .., "context": ... , "headers": ... }` object, so our jq filter will look like:
```
.body + {\"TRANSFORMED-REQUEST-BY-JQ\": true, path: .context.path, user_agent: .context.headers_User_Agent}```

Comment (buger) [reply mw_transform_jq.go]: Also, just to clarify, in jq syntax "+" means merge objects, right?

Comment (klyr) [reply mw_transform_jq.go]: Yes, "+" means merge.

Regarding your idea, yes, we can do that as well. But at the end, it's the same idea, except that instead of an array (an maybe a cryptic syntax to do variable assignment), it's a object.

I'll update my code, to use an object instead of an array, it will make things easier to understand.

Good idea !

**Comment by buger** [comment]
@mvdan do you want to add smth here? 

For sure there some rough edges, like code duplication, but overall looks nice. And functionality introduced here can be very powerful. Would like to merge it soon and include into 2.4

**Comment by buger** [comment]
Probably the most dangerous part here is jq bindings, they are in the early state. It uses CGO, and it needs to be verified for thread safety. The good thing here, that this binding should be fast because operate directly with C structures. Bindings code is very tiny, so it should not be a problem to solve bugs. Anyway, I would treat this feature as experimental, for 2.4, similar as gRPC bindings. 

@klyr any reason why you are using your own fork?

File: mw_transform_jq.go [hunk 2]
```diff
@@ -0,0 +1,115 @@
...
+func transformJQBody(r *http.Request, t *TransformJQSpec, contextVars bool) error {
+	// Read the body:
+	defer r.Body.Close()
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return err
+	}
+
+	var bodyObj map[string]interface{}
+	if err := json.Unmarshal(body, &bodyObj); err != nil {
```
mvdan: Use `json.NewDecoder(r.Body).Decode` and drop the `ReadAll`

mw_transform_jq.go
```diff
@@ -0,0 +1,115 @@
...
+		return err
+	}
+
+	jqObj := map[string]interface{}{
+		"body":       bodyObj,
+		"reqContext": ctxGetData(r),
+	}
+
+	if err = t.JQFilter.Handle(jqObj); err != nil {
```
mvdan: nit: use `:=` here

mw_transform_jq.go
```diff
@@ -0,0 +1,115 @@
...
+		return err
+	}
+
+	// First element is the transformed body
+	if t.JQFilter.Next() {
```
mvdan: do `if !x { return errors.New(...) }` to simplify the flow and unindent the other code.

mw_transform_jq.go
```diff
@@ -0,0 +1,115 @@
...
+		r.ContentLength = int64(bodyBuffer.Len())
+	} else {
+		return errors.New("Errors while applying JQ filter to input")
+	}
+
+	// Second optional element is an object like:
+	// { "output_headers": {"header_name": "header_value", ...},
+	//   "outputs_vars":   {"var_name_1": "var_value_1", ...}
+	// }
+	if t.JQFilter.Next() {
```
mvdan: You can also do the same here, to unindent all the rest.

res_handler_jq_transform.go
```diff
@@ -0,0 +1,90 @@
...
+	"strconv"
+)
+
+type ResponseTransformJQOptions struct {
+	//FlushInterval time.Duration
+}
+
+type ResponseTransformJQMiddleware struct {
+	Spec   *APISpec
+	config ResponseTransformJQOptions
```
mvdan: If the options are empty, remove it, the type and the code to decode it.

res_handler_jq_transform.go
```diff
@@ -0,0 +1,90 @@
...
+	defer res.Body.Close()
+	body, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return err
+	}
+
+	t := meta.(*TransformJQSpec)
+
+	var bodyObj map[string]interface{}
+	if err := json.Unmarshal(body, &bodyObj); err != nil {
```
mvdan: As before, use `r.Body` directly.

mvdan: Actually, the other two comments apply here as well, as there is some degree of copy-pasting.

mw_transform_jq.go
```diff
@@ -0,0 +1,115 @@
...
+	"encoding/json"
+	"errors"
+	"github.com/Sirupsen/logrus"
+	"github.com/mitchellh/mapstructure"
+	"io/ioutil"
+	"net/http"
+)
+
+type TransformJQMiddleware struct {
+	*BaseMiddleware
```
mvdan: We changed things a bit in master, so this should be without the pointer now.

api_definition.go
```diff
@@ -517,6 +530,39 @@ func (a APIDefinitionLoader) compileTransformPathSpec(paths []apidef.TemplateMet
 	return urlSpec
 }
 
+func (a *APIDefinitionLoader) compileJQFilter(filter string) (*jq.JQ, error) {
+	v, err := jq.NewJQ(filter)
+	return v, err
+}
```
mvdan: This func is purely `jq.NewJQ` - just remove the func and use that instead.

api_definition.go
```diff
@@ -517,6 +530,39 @@ func (a APIDefinitionLoader) compileTransformPathSpec(paths []apidef.TemplateMet
...
+
+func (a *APIDefinitionLoader) compileTransformJQPathSpec(paths []apidef.TransformJQMeta, stat URLStatus) []URLSpec {
+	urlSpec := []URLSpec{}
```
mvdan: Call this something in plural like `urlSpecs`

mvdan:
Oh, and of course, please fix the conflicts :)

klyr:
@mvdan: Thank you very much for the code review, I'm new to Go, so, still learning. I just rebased on top of master.

@buger: Regarding go-jq, I use my own fork because the library was missing a few error checks, so I added them. The pull request is pending... This library was one of the most simple, that is the reason of this choice, but it can of course be changed by something else.

Regarding the fact that it uses CGO, I totally agree, it's a weak point, I won't argue against this :) I can even say that it will make your build/check/CI workflow a little bit more complex, as it needs to have libjq development files installed, and libjq becomes a requirement at runtime. Maybe it's possible to make it a conditional option at build time ?

Anyway, I will refine it.

Thanks.

mvdan:
The current default build of Go has no special build nor runtime dependencies besides just Go and the Go vendored libraries. So yes, we definitely don't want to depend on libjq by default.

We have a similar situation with our use of Python and Lua via CGO, which depend on the python and lua libraries to be installed at both build and run time too. For those, we use a `coprocess` build tag. I would prefer to not add more build tags. Perhaps we could merge all of these CGO and dynamic linking dependencies into a single build tag, and call it something like `extradeps` or `extended`.

I'll leave that call for @buger and @matiasinsaurralde.

Regarding the library fork, could you point us at the pull request that has yet to be merged with your fixes?

mw_transform_jq.go
```diff
@@ -53,15 +53,9 @@ func (t *TransformJQMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Re
 }
 
 func transformJQBody(r *http.Request, t *TransformJQSpec, contextVars bool) error {
-	// Read the body:
-	defer r.Body.Close()
```
mvdan: did you delete this on purpose?

res_handler_jq_transform.go
```diff
@@ -35,17 +35,11 @@ func (h *ResponseTransformJQMiddleware) HandleResponse(rw http.ResponseWriter, r
 		return nil
 	}
 
-	// Read the body:
-	defer res.Body.Close()
```
mvdan: same as before

res_handler_jq_transform.go
klyr: I don't understand what I need to do here

mw_transform_jq.go
klyr: If you are asking, I guess it's because I shouldn't have. I thought the json decoder was doing the close()

res_handler_jq_transform.go
mvdan: This field is never used. It can be removed, as well as the mapstructure stuff that decodes into it, and the type that defines `ResponseTransformJQOptions`.

mw_transform_jq.go
mvdan: The json decoder can't (or rather shouldn't) do the closing as it receives an `io.Reader`, not an `io.ReadCloser`.

mvdan:
Also see the CI failure, which is because of the missing library. We install other stuff on CI already, for example for Python.

As said before, we need to make a decision on the build tags before merging this. We can't have the default build mode depend on an exotic library to be installed in the system.

klyr:
@mvdan: Here is the, old, pull request: https://github.com/mgood/go-jq/pull/3
I can switch to another library, but at the end, it will still be a wrapper to the C library.

Another solution can be to use a native Go implementation of JQ like this one https://github.com/savaki/jq, but the problem is that this implementation has only a subset of features of JQ.

res_handler_jq_transform.go
```diff
@@ -0,0 +1,82 @@
...
+	t := meta.(*TransformJQSpec)
+
+	var bodyObj interface{}
+	dec := json.NewDecoder(res.Body)
+	if err := dec.Decode(&bodyObj); err != nil {
+		return err
+	}
+	jqObj := map[string]interface{}{
+		"body":       bodyObj,
+		"reqContext": ctxGetData(req),
```
buger: should be called `_tyk_context` to be similar to follow same guidelines https://tyk.io/docs/concepts/context-variables/

res_handler_jq_transform.go
```diff
@@ -0,0 +1,82 @@
...
+		"resHeaders": res.Header,
```
buger: Should be called `response_headers`

mw_transform_jq.go
```diff
@@ -0,0 +1,120 @@
...
+	"net/http"
+)
+
+type TransformJQMiddleware struct {
+	BaseMiddleware
+}
+
+type JQTransformOptions struct {
+	OutputHeaders map[string]string `mapstructure:"output_headers"`
+	OutputVars    map[string]string `mapstructure:"output_vars"`
```
buger: Should be called `tyk_context`

mw_transform_jq.go
```diff
@@ -0,0 +1,120 @@
...
+	"io/ioutil"
+	"net/http"
+)
+
+type TransformJQMiddleware struct {
+	BaseMiddleware
+}
+
+type JQTransformOptions struct {
+	OutputHeaders map[string]string `mapstructure:"output_headers"`
```
buger: Should be called `rewrite_headers`

mw_transform_jq.go
```diff
@@ -0,0 +1,120 @@
...
+
+	var bodyObj interface{}
+	dec := json.NewDecoder(r.Body)
+	if err := dec.Decode(&bodyObj); err != nil {
+		return err
+	}
+
+	jqObj := map[string]interface{}{
+		"body":       bodyObj,
+		"reqContext": ctxGetData(r),
```
buger: Should be called `_tyk_context`

mw_transform_jq.go
```diff
@@ -0,0 +1,120 @@
...
+	// Replace header in the request
+	for hName, hValue := range opts.OutputHeaders {
+		r.Header.Set(hName, hValue)
+	}
+
+	if contextVars {
+		// Set variables in context vars
+		contextDataObject := ctxGetData(r)
+		for k, v := range opts.OutputVars {
+			contextDataObject["jq_output_var_"+k] = v
```
buger: Set them without prefix. Just  `contextDataObject[k]`

res_handler_jq_transform.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	t.Lock()
+	defer t.Unlock()
+	if err := t.JQFilter.Handle(jqObj); err != nil {
+		return errors.New("Response returned by upstream server is not a valid JSON")
```
buger: Should extend error message with an original error `err`.

buger:
Apologies for such delay, this feature is finally ready to be included in 2.5. I requested some final changes, but we can do it by ourselves if needed.

klyr:
@buger:  I'm working on new modifications (WIP) on an alternate branch (https://github.com/klyr/tyk/tree/jq-transform-no-dep). Removing the dependency of go-jq, and putting what is relevant into tyk directly.
Moreover, now, instead of returning a coma-separated json value (the body and the output headers) only one json object is required to be returned, of the following form:
```json
{
  "body": THE_TRANSFORMED_BODY,
  "output_headers": {"header1_name": "header1_value", ...},
  "output_vars": {"var1_name": "var1_value", ...}
}
```

Also fixed a race condition by using a mutex to allow only one goroutine at a time to use a JQ context.

I'll update the code in regards to your comments.

buger:
@klyr I would love to see your changes! Any updated on this?

klyr:
@buger : The new changes are on a new branch, you can see the diff on https://github.com/TykTechnologies/tyk/compare/TykTechnologies:master...klyr:jq-transform-no-dep

Do you prefer that I do a pull request ?

buger:
Sure, let's close this one in favor of new one. Feel free to open a new PR. Thanks!

Would love to get it merged in 2.5.

Issue #1050, opened 2017-08-30
Updating key, errors on master branch


lghiur:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When trying to update a key from Dashboard, I get the following error:

Gateway: 
```
ERROR api: Could not add key for this API ID, API doesn't exist. api_id=e2b7e190866145fd42fcd053fd0d3f16 key=5989653286341f6a160639cfb8960814d0f448789fc05bebafaad593 org_id=5989653286341f6a160639cf path=-- server_name=system user_id=system user_ip=--
```

Analytics:
```
ERROR API Request failed: {"status":"error","message":"Failed to create key, ensure security settings are correct."}

[Aug 30 10:44:20] ERROR Failed to update session object to Tyk
```

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

1. Go to the Dashboard and create a new key;
2. Try to update the key you just created with new data

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**

mvdan:
(this works fine on stable)

mvdan:
This is indeed a gateway regression of some sort. The underlying issue was that on stable, if we failed to update a key, we silently returned 200, while on master, we return a 500.

stable:

```
addUpdateErr := doAddOrUpdate(keyName, newSession, suppress_reset)
if addUpdateErr != nil {
        success = false
        responseMessage = createError("Failed to create key, ensure security settings are correct.")
}
```

(note how `code` isn't set, so it remains at the default `200`; `success = false` doesn't set the code either)

master:

```
if err := doAddOrUpdate(keyName, &newSession, suppressReset); err != nil {
        return apiError("Failed to create key, ensure security settings are correct."), 500
}
```

To make this more obvious, the actual cause of the error is seen on both stable and master:

```
ERROR api: Could not add key for this API ID, API doesn't exist. api_id=22ad7a2b8b194ba374a389e1e3682c95 key=638fdf169f7342f140612b7b6b01b99a org_id= path=-- server_name=system user_id=system user_ip=--
```

What causes the error is that the key is attached to an ApiID that isn't known by the gateway, i.e. that isn't in the global `apisByID` map. I'm not exactly sure why this would happen. Perhaps because the Dashboard stores the api definitions on its own and the gateway doesn't have them?

[...]

PR #1056, opened 2017-09-01
api: revert key update error to a 200

mvdan:
To unblock the policies and keys work and their testing, for now. We can
figure out the proper fix later, before 2.4 is out.

Updates #1050.

### DIFF
--- a/api.go
+++ b/api.go
@@ -255,7 +255,10 @@ func handleAddOrUpdate(keyName string, r *http.Request) (interface{}, int) {
 	}
 	suppressReset := r.FormValue("suppress_reset") == "1"
 	if err := doAddOrUpdate(keyName, &newSession, suppressReset); err != nil {
-		return apiError("Failed to create key, ensure security settings are correct."), 500
+		// TODO: revisit this 200 before 2.4, since we probably
+		// want it to be an error code like 500. The code is 200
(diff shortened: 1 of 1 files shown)


lghiur:
@mvdan @buger  with this change, updating keys data works again

buger:
I do not like merging it as a PR, maybe @lghiur can make this change locally in this own branch?

lghiur:
@buger I personally don't need this change, but if this doesn't work then we can't release multiple policies issue as @ilijabojanovic won't be able to test it. This is the main concern. 

Otherwise @mvdan s changes on multiple policies are working from my perspective.

mvdan:
I don't like it either, but I don't see the problem with reverting back to stable behavior on master as a temporary measure.

buger:
It will not help @ilijabojanovic as well because this change does not fix the issue, just ignore it on UI.

If goal is to release this feature, the only way it to fix origin issue.

mvdan:
You're technically right, which is why I opened the issue above. But in practice, stable works and this is a separate issue, so I was simply trying to unblock the other, larger issue.

I'm going to close this for now. I'll look at the regression again, but I'm not sure when we're going to be able to unblock multiple policies.

PR #1070, opened 2017-09-04
main: honor --log-instrumentation flag again

### LINKED ISSUE #1066: Tyk always logs error on startup: "Instrumentation is enabled, but no connectionstring set for statsd"
The error mentioned inside the title is always logged and cannot be disabled. "main.go" always calls `SetupInstrumentation(true)` (with the hardcoded `true`), regardless of any settings.

Seen with tyk-gateway 2.3.8.

mvdan:
The flag started being ignored last december for unknown reasons. Use
it again.

Fixes #1066.

### DIFF
--- a/instrumentation_handlers.go
+++ b/instrumentation_handlers.go
@@ -13,13 +13,12 @@ import (
 var applicationGCStats = debug.GCStats{}
 var instrument = health.NewStream()
 
-// SetupInstrumentation handles all the intialisation of the instrumentation handler
-func SetupInstrumentation(enabled bool) {
-	if os.Getenv("TYK_INSTRUMENTATION") == "1" {
-		enabled = true
-	}
-
-	if !enabled {
+// setupInstrumentation handles all the intialisation of the instrumentation handler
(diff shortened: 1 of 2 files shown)


instrumentation_handlers.go
```diff
@@ -13,13 +13,12 @@ import (
...
-	if os.Getenv("TYK_INSTRUMENTATION") == "1" {
-		enabled = true
-	}
-
-	if !enabled {
+// setupInstrumentation handles all the intialisation of the instrumentation handler
+func setupInstrumentation(arguments map[string]interface{}) {
+	switch {
+	case arguments["--log-instrumentation"] == true:
+	case os.Getenv("TYK_INSTRUMENTATION") == "1":
```
buger: If we going to make it right, let's also introduce configuration variable.

mvdan: How so? Not all configs have flags, and not all flags have configs. For example, the log level doesn't have a config option - and they're fairly similar flags.

mvdan: bump

PR #1071, opened 2017-09-04
health: remove extra goroutine

mvdan:
StoreCounterVal does almost no synchronous work; the actual store write
is done in a goroutine. So calling it in a goroutine is completely
unnecessary.

While at it, simplify the value report calls.


mvdan:
This could help with a panic test flake that @excieve found on Buddy.

buger:
conflict ^

PR #1074, opened 2017-09-05
jsvm: don't ignore otto's Run errors

### LINKED ISSUE #959: Investigate whether the JSVM can give proper errors
Right now, if there is an error in someone's JS code, we log this:

```
ERROR jsvm: Failed to decode middleware request data on return from VM: invalid character 'u' looking for beginning of value
```

This is extremely confusing. This is because we're trying to unmarshal JSON from an `undefined` string.

Surely, Otto has a better error for us in most or all of the cases. We should do a better job at logging the correct error - ideally, with a source position.

mvdan:
They often have helpful information about what went wrong, which greatly
helps with debugging. For example, given the following JS middlware:

	var errMid = new TykJS.TykMiddleware.NewMiddleware({});

	errMid.NewProcessRequest(function(request, session) {
	       log(forgot_to_define);
	       return errMid.ReturnData(request, session.meta_data);
	});

We would previously log:

	ERROR jsvm: Failed to decode middleware request data on return from VM: invalid character 'u' looking for beginning of value

Now we log:

	ERROR jsvm: Failed to run JS middleware: ReferenceError: 'forgot_to_define' is not defined

Don't add a test because all of this code ignores errors, and making the
tests aware of the errors would be a major refactor. The patch is
trivial and ignoring errors is usually a red flag, so it's very unlikely
that a regression test will be particularly useful.

Fixes #959.

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -138,6 +138,7 @@ func (d *DynamicMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reques
 	// buffered, leaving no chance of a goroutine leak since the
 	// spawned goroutine will send 0 or 1 values.
 	ret := make(chan otto.Value, 1)
+	errRet := make(chan error, 1)
 	go func() {
 		defer func() {
 			// the VM executes the panic func that gets it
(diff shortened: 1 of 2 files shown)


mvdan:
(steps to reproduce/test are right there in the commit message)

buger:
👍

PR #1078, opened 2017-09-06
test: force ipv4 when listening on :0

mvdan:
To work around a docker issue that affects Buddy. This is fixed in Go
1.9, since it treats systems with half-broken ipv6 better and can handle
this without it breaking.

For now, since we need 1.8 to do releases, work around it here.


buger:
nice find 👌

PR #1085, opened 2017-09-07
all: collect more Otto.Run errors

mvdan:
In particular, when loading the JS core, when loading people's
middlewares and when loading a virtual endpoint js program. There are
plenty of ways for users to get errors, such as missing files or bad
syntax, so they're very useful.

While at it, avoid the unnecessary use of ioutil.ReadFile. Otto.Run
supports an io.Reader too, so there's no need to read the entire file
into memory.

No tests because errors in these scenarios aren't fatal at the moment.
Adding regression tests would require a major refactor for little gain.


mvdan:
@buger this will greatly help with support queries too, as these errors are completely ignored at the moment, leading to weird crashes/errors when a request comes in.

PR #1087, opened 2017-09-07
all: remove a bunch of ReadFile/ReadAll calls

mvdan:
The standard library and plenty of libraries tend to work with
io.Readers as well as string/[]byte, so most of the time we don't need
to read entire bodies and files into memory.

This also simplifies the code a bit, since we remove an extra step in
some cases.

While at it, stop ignoring some unmarshalling and parsing errors that
were spotted while refactoring.

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -204,11 +204,6 @@ func (a APIDefinitionLoader) MakeSpec(def *apidef.APIDefinition) *APISpec {
 	return spec
 }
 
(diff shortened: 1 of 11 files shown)


mvdan:
Would be good if we could get @ilijabojanovic to run his end-to-end test suite on this PR, just in case.

buger:
Yes, looks good, @ilijabojanovic pls test

ilijabojanovic:
@buger approved tests are green, and quick checking looking good

PR #1094, opened 2017-09-08
main: get rid of ServeMux and use mux.Router as the default handler.

### LINKED ISSUE #1093: Skip URL cleaning isn't working
Bug

Enable "skip_url_cleaning".
Create an API

Latest version.

matiasinsaurralde:
This fixes #1093.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1386,11 +1386,8 @@ func listen(l net.Listener, err error) {
 				Addr:         ":" + targetPort,
 				ReadTimeout:  time.Duration(ReadTimeout) * time.Second,
 				WriteTimeout: time.Duration(WriteTimeout) * time.Second,
+				Handler:      defaultRouter,
 			}
-
-			newServeMux := http.NewServeMux()
-			newServeMux.Handle("/", defaultRouter)
-			http.DefaultServeMux = newServeMux
 
 			// Accept connections in a new goroutine.
 			go s.Serve(l)
@@ -1450,11 +1447,8 @@ func listen(l net.Listener, err error) {
 				Addr:         ":" + targetPort,
 				ReadTimeout:  time.Duration(ReadTimeout) * time.Second,
 				WriteTimeout: time.Duration(WriteTimeout) * time.Second,
+				Handler:      defaultRouter,
 			}
-
-			newServeMux := http.NewServeMux()
-			newServeMux.Handle("/", defaultRouter)
-			http.DefaultServeMux = newServeMux
 
 			log.WithFields(logrus.Fields{
 				"prefix": "main",


buger:
How it fixes the issue? Before and after it use router `newServeMux.Handle("/", defaultRouter)`, and router code handle SkipClean option.

buger:
I think the actual issue happen on this line https://github.com/TykTechnologies/tyk/blob/release-2.3/main.go#L738

When we assign new router but do not set SkipClean option, so after first reload, this functionality gets disabled. We should add smth like this to `doReload`:
```
if globalConf.HttpServerOptions.OverrideDefaults {
   mainRouter.SkipClean(globalConf.HttpServerOptions.SkipURLCleaning)
}
```

mvdan:
I agree with @buger - not sure this is the right fix (without having reproduced the issue and tested this fix, though).

The cleanup is nice, however. This should be merged even if it's not a bugfix, as long as it doesn't break `override_defaults`.

mvdan:
Also, answering @lonelycode's question on whether this could break the hot reloads code - I doubt it would. That code sits on top of this, so I would be surprised. And it has decent tests too.

mvdan:
I just realised that this is patch is against stable - so my comment about the cleanup still being desirable would apply if this was done in `master`, obviously.

matiasinsaurralde:
Hi, I will add tests for this. `SkipClean` is ignored even before a hot reload, so I'm not sure if the code suggested by @leon is enough.
To debug this I modified `mux` to print every time its `ServeHTTP` call was called, when sending a URL with double-slashes, `mux` wasn't being called and the redirect was automatically triggered (ignoring the configuration flag), this redirect was triggered by the `ServeMux`.

matiasinsaurralde:
I've been testing this again and I can confirm that it doesn't break on reloads, also:
- This patch only changes the behavior when using `OverrideDefaults`.

The scenario is:
- User enables `OverrideDefaults`
- User sets `skip_url_cleaning` to `true`.

The URL cleanup is still done all the time, after the gateway starts, the original issue *isn't related to reloads at all*.

Removing `DefaultServeMux` doesn't fix the issue, but removing the `ServeMux` code completely works fine, even after reloads.
I suggest trying this patch with the standard scenario I've mentioned here and in the issue, then running something like:

```
$ curl http://localhost:8080/test-api/http://abc.com
```

It's hard to add tests because we only call `initialiseSystem` for tests, couldn't come up with a good solution for abstracting this part (we don't have any tests for the `OverrideDefaults` flags in general).

matiasinsaurralde:
As I mentioned before, to debug this I've added a call to the `mux` `ServeHTTP` method, we expect this to be called in all the scenarios because it would react to the `SkipClean` flag, however its `ServeHTTP` method isn't called when sending a "dirty" URL, the `ServeMux` triggers the redirect before it.

buger:
Closing in favor of https://github.com/TykTechnologies/tyk/pull/1105

Issue #739, opened 2017-05-14
jsvm: bundle js/tyk.js as a string literal in the source code


mvdan:
`js/tyk.js` has glue/helper code that is used by JS plugins. The JSVM runs it as a string, so it doesn't need it to be in a file on disk. The code is also internal, so there is no reason why it should be easily accessible and modifyiable by a gateway user.

I propose that we move it to a string literal and simplify things. This would also mean we could get rid of the `js` directory in the root of the repo, since there doesn't seem to be any other use for it.

@matiasinsaurralde mentioned that the file could be useful for users to run some global JS code prior to initializing their JS plugins. Do we support that officially? I could not find any docs around it.

If the "global" code only has to run before a plugin is set up, it could also just be added a the beginning of a plugin JS file.

/cc @lonelycode @buger

buger:
Recent example: one of the users wanted to use base64 encoding inside middleware to make a HTTP request to URL protected by basic auth. Otto itself does not have such built-in function, so he had to embed it into middleware itself. 

So, there are multiple solutions to this issue.
1) Leave it as it is
2) Hide Tyk specific stuff into the string, and bundle to binary, and add an alternative solution for loading external code: for example load *all* files in js directory. 
3) Hide this stuff as in step 2, and tell user embed all needed functions right into the middleware. 

In my view, 3) is easiest to support and maintain.

mvdan:
Agreed on 3. That's what I meant by "added at the beginning of a plugin JS file".

Also, coincidentally in https://github.com/TykTechnologies/tyk/pull/721 I'm adding newly exposed base64 funcs. Maybe we should make then "official". Or we can just leave them as is, undocumented, and we don't have to worry about supporting them and backwards compat.

buger:
Note that this task will probably require tuning our release script, so it will stop including this file into distribution.

nickReyn:
@ilijabojanovic any reason you moved this back to in progress? Trying to figure out what is needed for this issue

mvdan:
@nickReyn because he believes he found an issue - see https://github.com/TykTechnologies/tyk/pull/1097. I replied there.

ilijabojanovic:
@nickReyn @mvdan responded, on all tasks where i found some issues am moving them back to in progress since they are not finished

nickReyn:
Cool, thanks 👍  sorry was reviewing quickly while on planning call and didn't see the PR

mvdan:
[...]

PR #1097, opened 2017-09-11
js: bundle js/tyk.js, TykJSPath now user's code

mvdan:
As discussed in the issue below, moved the file into a constant embedded
into the binary. While at it, indent with tabs and remove the
unnecessary semicolons.

A few other bits will need updating, such as our release repositories
and scripts as well as our documentation.

For the use case where people were using TykJSPath to define code for
all JS plugins, that path will still be loaded after our TykJS core. Add
a test for that scenario.

Updates #739.


mvdan:
(made the PR an "updates" instead of "fixes" since other things need doing)

mvdan:
Also note that removing the field from the config is fine; the JSON parser doesn't error on unknown fields.

buger:
I think you need to add compatibility layer. A lot of our users already have code embed into tyk.js, and after upgrading the app will break. I propose to leave config option, and if the file in such path is found, load it.

Actually maybe make sense transform this task, by saying that all the built-in tyk functions is embedded to our JS code, and all the global JS functions, can be specified in tyk.js file. So this is similar to what I proposed in #739 

>Hide Tyk specific stuff into the string, and bundle to binary, and add an alternative solution for loading external code: for example load all files in js directory.

This makes code both compatible with previous versions, we move our internal code into binary, and user has official option to set his global funcs.

mvdan:
I'm a bit surprised by this, since we seemed to agree on option 3 - `Hide this stuff as in step 2, and tell user embed all needed functions right into the middleware.`

Trying to keep backwards compatibility is a good idea, but I'm not sure that this would be enough. If people just upgrade, `js/tyk.js` would also include our core JS stuff - potentially with old code, given that they'd keep the same file if they modified it.

So the old code could be a problem. But even if it's the same core Tyk code, we would be loading it twice. We would have to ensure that it won't break anything.

That's why I was going with the route of telling people to just load their global code from each of the JS files. After all, there is no global state available, so having a global JS file is even unintuitive. Whatever we do, there's no way to keep full backwards compatibility, as it will require some degree of action from the users.

buger:
Well, we should do it backward compatible, there is no another option here. The only place where we can drop tyk.js support is version 3. So my proposal was on how to get benefits from this situation.

Having an old code in Tyk.js is not a problem if we first load tyk.js and after our internal code: it still will use newest code.

mvdan:
> Having an old code in Tyk.js is not a problem if we first load tyk.js and after our internal code: it still will use newest code.

That is still assuming many things - that the overwrite will be clean and that, for example, the user's code didn't depend on the core code at all.

I propose then that we keep the option, and tell people to only keep their custom code in `js/tyk.js` when upgrading to 2.4. I don't think that's backwards incompatible - we have required config/setup changes in major releases before, and I don't think it's a lot to ask.

I'd still load our core first, because otherwise it would be impossible to use it in `js/tyk.js`, potentially breaking people's code.

buger:
> I'd still load our core first, because otherwise, it would be impossible to use it in js/tyk.js, potentially breaking people's code.

Yes, you are right. But we still need to load it.

buger:
bump

mvdan:
@buger done, added test. once this is merged, the docs will need updating and the release notes will have to tell people to either remove `tyk_js_path` from the config if they didn't add any code, or to just keep their own code in the file if they do (removing the old `TykJS` stuff).

config/config.go
```diff
@@ -272,7 +272,6 @@ func WriteDefault(path string, conf *Config) {
 		ListenPort:     8080,
 		Secret:         "352d20ee67be67f6340b4c0605b044b7",
 		TemplatePath:   "templates",
-		TykJSPath:      "js/tyk.js",
```
buger: I assume you should rollback this config option?

mvdan: No - this sets the default. We no longer ship or have `js/tyk.js`, so this needs to go. If a user wants to use global JS code, they need to set the config option to whatever path they like.

mvdan:
I just realised that the release scripts mention the file - will update them too.

config/config.go
buger: Right

buger:
@ilijabojanovic ready for testing. 

The basic idea that we are removing  "js/tyk.js" file from our distribution, and do not set `tyk_js_path` default config value. We still support loading Js files through `tyk_js_path` config variable, and our previous users should not have compatibility issues since they have old configs. 

All the code loaded from file specified in `tyk_js_path` should be accessible to all JS middleware, and the easiest way to test it is to create function with like this: `function foo(){ return "bar" }` and use it in VirtualPath

mvdan:
Rebased.

buger:
@ilijabojanovic this pr is getting old :)

ilijabojanovic:
@buger @mvdan 
I found one issue during testing this PR:

Env: On-prem

Reproduction steps:
1. Leave "tyk_js_path": "./js/tyk.js" in gateway configuration
2. Over dashboard import api definition which have virtual path

**Actual:
Gateway panics and shutdowns**

```
tyk-gateway_1    | time="Oct  3 10:39:02" level=error msg="Could not load user's TykJS: open ./js/tyk.js: no such file or directory"
tyk-gateway_1    | time="Oct  3 10:39:02" level=debug msg="INITIALISING EVENT HANDLERS"
tyk-gateway_1    | time="Oct  3 10:39:02" level=debug msg="Checking for transform paths..."
tyk-gateway_1    | time="Oct  3 10:39:02" level=debug msg="Checking for transform paths..."
tyk-gateway_1    | time="Oct  3 10:39:02" level=debug msg="Loading JS blob"
tyk-gateway_1    | panic: runtime error: invalid memory address or nil pointer dereference
tyk-gateway_1    | [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xb67fc4]
tyk-gateway_1    |
tyk-gateway_1    | goroutine 223 [running]:
tyk-gateway_1    | main.preLoadVirtualMetaCode(0xc420748b98, 0xc4207a6370)
tyk-gateway_1    | 	/develop/go/src/github.com/TykTechnologies/tyk/mw_virtual_endpoint.go:76 +0x2b4
tyk-gateway_1    | main.APIDefinitionLoader.compileVirtualPathspathSpec(0xc4203171e0, 0x2, 0x4, 0xc, 0xc4207a6300, 0x0, 0x0, 0x0)
tyk-gateway_1    | 	/develop/go/src/github.com/TykTechnologies/tyk/api_definition.go:672 +0x18f
tyk-gateway_1    | main.APIDefinitionLoader.getExtendedPathSpecs(0xc420217990, 0x7, 0x0, 0x0, 0x13731b8, 0x0, 0x0, 0x13731b8, 0x0, 0x0, ...)

tyk-gateway_1    | 	/develop/go/src/github.com/TykTechnologies/tyk/api_definition.go:722 +0x454
tyk-gateway_1    | main.APIDefinitionLoader.MakeSpec(0xc420318000, 0xc42070f7e0)
tyk-gateway_1    | 	/develop/go/src/github.com/TykTechnologies/tyk/api_definition.go:193 +0xb1c
tyk-gateway_1    | main.APIDefinitionLoader.FromDashboardService(0xc420767890, 0x25, 0xc4201db1a0, 0x20, 0x0, 0x0, 0x0)
tyk-gateway_1    | 	/develop/go/src/github.com/TykTechnologies/tyk/api_definition.go:286 +0x927
tyk-gateway_1    | main.getAPISpecs(0x4c257e, 0x3b, 0x40)
tyk-gateway_1    | 	/develop/go/src/github.com/TykTechnologies/tyk/main.go:194 +0x647
tyk-gateway_1    | main.doReload()
tyk-gateway_1    | 	/develop/go/src/github.com/TykTechnologies/tyk/main.go:554 +0x3c
tyk-gateway_1    | main.reloadLoop(0xc420588c60)
tyk-gateway_1    | 	/develop/go/src/github.com/TykTechnologies/tyk/main.go:604 +0x102
tyk-gateway_1    | created by main.start
tyk-gateway_1    | 	/develop/go/src/github.com/TykTechnologies/tyk/main.go:1081 +0x180
tyk-dashboard_1  | time="Oct  3 10:39:02" level=debug msg="Sending notification{NoticeDashboardZeroConf {\"DashboardConfig\":{\"Hostname\":\"www.tyk-test.com:8181\",\"Port\":3000,\"UseTLS\":false},\"TimeStamp\":1507027142} }"
compose_tyk-gateway_1 exited with code 2
```
**NOTE:
This issue is not visible when we remove "tyk_js_path": "./js/tyk.js" from configuration** 

Here is testing gateway configuration
```
{
    "listen_address": "",
    "listen_port": 8181,
    "secret": "352d20ee67be67f6340b4c0605b044b7",
    "node_secret": "352d20ee67be67f6340b4c0605b044b7",
    "template_path": "./templates",
    "tyk_js_path": "./js/tyk.js",
    "middleware_path": "./middleware",
    "policies": {
        "policy_source": "service",
        "policy_connection_string": "http://tyk-dashboard:3000",
        "policy_record_name": "tyk_policies",
        "allow_explicit_policy_id": true
    },
    "use_db_app_configs": true,
    "db_app_conf_options": {
        "connection_string": "http://tyk-dashboard:3000",
        "node_is_segmented": false,
        "tags": ["test"]
    },
    "disable_dashboard_zeroconf": false,
    "app_path": "./test_apps/",
    "storage": {
        "type": "redis",
        "host": "",
        "port": 0,
        "hosts": {
            "tyk-redis": "6379"
        },
        "username": "",
        "password": "",
        "database": 0,
        "optimisation_max_idle": 3000,
        "optimisation_max_active": 5000,
        "enable_cluster": false
    },
    "enable_separate_cache_store": false,
    "cache_storage": {
        "type": "redis",
        "host": "",
        "port": 0,
        "hosts": {
            "localhost": "6379"
        },
        "username": "",
        "password": "",
        "database": 0,
        "optimisation_max_idle": 3000,
        "optimisation_max_active": 5000,
        "enable_cluster": false
    },
       "enable_analytics": true,
    "analytics_config": {
        "type": "mongo",
        "ignored_ips": [],
        "enable_detailed_recording": true,
        "enable_geo_ip": true,
        "geo_ip_db_path": "./GeoLite2-City.mmdb",
        "normalise_urls": {
            "enabled": true,
            "normalise_uuids": true,
            "normalise_numbers": true,
            "custom_patterns": []
        }
    },
    "health_check": {
        "enable_health_checks": false,
        "health_check_value_timeouts": 0
    },
    "enable_detailed_recording": false,
    "optimisations_use_async_session_write": true,
    "allow_master_keys": false,
    "hash_keys": true,
    "suppress_redis_signal_reload": false,
    "suppress_default_org_store": false,
    "use_redis_log": true,
    "sentry_code": "",
    "use_sentry": false,
    "use_syslog": true,
    "use_graylog": false,
    "use_logstash": false,
    "graylog_network_addr": "",
    "logstash_network_addr": "",
    "syslog_transport": "UDP",
    "logstash_transport": "",
    "syslog_network_addr": "192.168.1.50",
    "enforce_org_data_age": true,
    "enforce_org_data_detail_logging": false,
    "enforce_org_quotas": true,
    "experimental_process_org_off_thread": true,
    "enable_non_transactional_rate_limiter": true,
    "enable_sentinel_rate_limiter": false,
    "management_node": false,
    "Monitor": {
        "enable_trigger_monitors": false,
        "configuration": {
            "method": "",
            "target_path": "",
            "template_path": "",
            "header_map": null,
            "event_timeout": 0
        },
        "global_trigger_limit": 0,
        "monitor_user_keys": false,
        "monitor_org_keys": false
    },
    "oauth_refresh_token_expire": 0,
    "oauth_token_expire": 0,
    "oauth_redirect_uri_separator": ";",
     "slave_options": {
        "use_rpc": false,
        "connection_string": "",
        "rpc_key": "",
        "api_key": "",
        "enable_rpc_cache": false,
        "bind_to_slugs": false,
        "disable_keyspace_sync": false,
        "group_id": ""
    },
    "disable_virtual_path_blobs": false,
    "local_session_cache": {
        "disable_cached_session_state": true,
        "cached_session_timeout": 0,
        "cached_session_eviction": 0
    },
    "http_server_options": {
    	"skip_url_cleaning": true,
        "override_defaults": true,
        "read_timeout": 0,
        "write_timeout": 0,
        "use_ssl": false,
        "use_ssl_le": false,
        "enable_websockets": true,
        "certificates": [],
        "server_name": "",
        "min_version": 0,
        "flush_interval": 0
    },
    "service_discovery": {
        "default_cache_timeout": 0
    },
    "close_connections": true,
    "auth_override": {
        "force_auth_provider": false,
        "auth_provider": {
            "name": "",
            "storage_engine": "",
            "meta": null
        },
        "force_session_provider": false,
        "session_provider": {
            "name": "",
            "storage_engine": "",
            "meta": null
        }
    },
    "uptime_tests": {
        "disable": false,
        "config": {
            "failure_trigger_sample_size": 1,
            "time_wait": 2,
            "checker_pool_size": 50,
            "enable_uptime_analytics": true
        }
    },
    "hostname": "",
    "enable_api_segregation": false,
    "control_api_hostname": "",
    "enable_custom_domains": true,
    "enable_jsvm": true,
    "enable_coprocess": false,
    "hide_generator_header": false,
    "event_handlers": {
        "events": {}
    },
    "event_trigers_defunct": {},
    "pid_file_location": "./tyk-gateway.pid",
    "allow_insecure_configs": true,
    "public_key_path": "/home/tyk/go/src/github.com/lonelycode/tyk-cloud-config-v2/keys/public.pem",
    "close_idle_connections": false,
    "allow_remote_config": true,
    "enable_bundle_downloader": false
}
```

Here is testing api definition:
```
{
    "id": "59b84a07840e2a01cc414f27",
    "name": "test",
    "slug": "test",
    "api_id": "6a7a99485dc845dc52ba58e43275b2ac",
    "org_id": "59b849ee840e2a01cc414f23",
    "use_keyless": true,
    "use_oauth2": false,
    "use_openid": false,
    "openid_options": {
        "providers": [],
        "segregate_by_client": false
    },
    "oauth_meta": {
        "allowed_access_types": [],
        "allowed_authorize_types": [],
        "auth_login_redirect": ""
    },
    "auth": {
        "use_param": false,
        "param_name": "",
        "use_cookie": false,
        "cookie_name": "",
        "auth_header_name": ""
    },
    "use_basic_auth": false,
    "enable_jwt": false,
    "use_standard_auth": false,
    "enable_coprocess_auth": false,
    "jwt_signing_method": "",
    "jwt_source": "",
    "jwt_identity_base_field": "",
    "jwt_client_base_field": "",
    "jwt_policy_field_name": "",
    "notifications": {
        "shared_secret": "",
        "oauth_on_keychange_url": ""
    },
    "enable_signature_checking": false,
    "hmac_allowed_clock_skew": -1,
    "base_identity_provided_by": "",
    "definition": {
        "location": "header",
        "key": "x-api-version"
    },
    "version_data": {
        "not_versioned": true,
        "versions": {
            "Default": {
                "name": "Default",
                "expires": "",
                "paths": {
                    "ignored": [],
                    "white_list": [],
                    "black_list": []
                },
                "use_extended_paths": true,
                "extended_paths": {
                    "virtual": [
                        {
                            "response_function_name": "testVirtData",
                            "function_source_type": "blob",
                            "function_source_uri": "ZnVuY3Rpb24gdGVzdFZpcnREYXRhKHJlcXVlc3QsIHNlc3Npb24sIGNvbmZpZykgewogICAgdmFyIHJlc3AgPSB7CiAgICAgICAgQm9keTogcmVxdWVzdC5Cb2R5ICsgIiBhZGRlZCBib2R5IiwgCiAgICAgICAgSGVhZGVyczogewogICAgICAgICAgICAiZGF0YS1mb28iOiBjb25maWcuY29uZmlnX2RhdGEuZm9vCiAgICAgICAgfSwKICAgICAgICBDb2RlOiAyMDIKICAgIH0KICAgIHJldHVybiBUeWtKc1Jlc3BvbnNlKHJlc3AsIHNlc3Npb24ubWV0YV9kYXRhKSAgIAp9",
                            "path": "/ilija2",
                            "method": "POST",
                            "use_session": false
                        },
                        {
                            "response_function_name": "batchTest",
                            "function_source_type": "blob",
                            "function_source_uri": "ZnVuY3Rpb24gYmF0Y2hUZXN0IChyZXF1ZXN0LCBzZXNzaW9uLCBjb25maWcpIHsKICAgICAgICAvLyBTZXQgdXAgYSByZXNwb25zZSBvYmplY3QKICAgICAgICB2YXIgcmVzcG9uc2UgPSB7CiAgICAgICAgICAgIEJvZHk6ICIiCiAgICAgICAgICAgIEhlYWRlcnM6IHsKICAgICAgICAgICAgICAgICJ0ZXN0IjogInZpcnR1YWwtaGVhZGVyLTEiLAogICAgICAgICAgICAgICAgInRlc3QtMiI6ICJ2aXJ0dWFsLWhlYWRlci0yIiwKICAgICAgICAgICAgICAgICJjb250ZW50LXR5cGUiOiAiYXBwbGljYXRpb24vanNvbiIKICAgICAgICAgICAgfSwKICAgICAgICAgICAgQ29kZTogMjAwCiAgICAgICAgfQogICAgCiAgICAgICAgLy8gQmF0Y2ggcmVxdWVzdAogICAgICAgIHZhciBiYXRjaCA9IHsKICAgICAgICAgICAgInJlcXVlc3RzIjogWwogICAgICAgICAgICAgICAgewogICAgICAgICAgICAgICAgICAgICJtZXRob2QiOiAiR0VUIiwKICAgICAgICAgICAgICAgICAgICAiaGVhZGVycyI6IHsKICAgICAgICAgICAgICAgICAgICAgICAgIngtdHlrLXRlc3QiOiAiMSIsCiAgICAgICAgICAgICAgICAgICAgICAgICJ4LXR5ay12ZXJzaW9uIjogIjEuMiIsCiAgICAgICAgICAgICAgICAgICAgICAgICJhdXRob3JpemF0aW9uIjogIjFkYmM4M2I5YzQzMTY0OWQ3Njk4ZmFhOTc5N2UyOTAwZiIKICAgICAgICAgICAgICAgICAgICB9LAogICAgICAgICAgICAgICAgICAgICJib2R5IjogIiIsCiAgICAgICAgICAgICAgICAgICAgInJlbGF0aXZlX3VybCI6ICJodHRwOi8vaHR0cGJpbi5vcmcvaGVhZGVycyIKICAgICAgICAgICAgICAgIH0sCiAgICAgICAgICAgICAgICB7CiAgICAgICAgICAgICAgICAgICAgIm1ldGhvZCI6ICJHRVQiLAogICAgICAgICAgICAgICAgICAgICJoZWFkZXJzIjoge30sCiAgICAgICAgICAgICAgICAgICAgImJvZHkiOiAiIiwKICAgICAgICAgICAgICAgICAgICAicmVsYXRpdmVfdXJsIjogImh0dHA6Ly9odHRwYmluLm9yZy91c2VyLWFnZW50IgogICAgICAgICAgICAgICAgfQogICAgICAgICAgICBdLAogICAgICAgICAgICAic3VwcHJlc3NfcGFyYWxsZWxfZXhlY3V0aW9uIjogZmFsc2UKICAgICAgICB9CiAgICAKICAgICAgICBsb2coIltWaXJ0dWFsIFRlc3RdIE1ha2luZyBVcHN0cmVhbSBCYXRjaCBSZXF1ZXN0IikKICAgICAgICB2YXIgbmV3Qm9keSA9IFR5a0JhdGNoUmVxdWVzdChKU09OLnN0cmluZ2lmeShiYXRjaCkpCiAgICAKICAgICAgICAvLyBXZSBrbm93IHRoYXQgdGhlIHJlcXVlc3RzIHJldHVybiBKU09OIGluIHRoZWlyIGJvZHksIGxldHMgZmxhdHRlbiBpdAogICAgICAgIHZhciBhc0pTID0gSlNPTi5wYXJzZShuZXdCb2R5KQogICAgICAgIGZvciAodmFyIGkgaW4gYXNKUykgewogICAgICAgICAgICBhc0pTW2ldLmJvZHkgPSBKU09OLnBhcnNlKGFzSlNbaV0uYm9keSkKICAgICAgICB9CiAgICAKICAgICAgICAvLyBXZSBuZWVkIHRvIHNlbmQgYSBzdHJpbmcgb2JqZWN0IGJhY2sgdG8gVHlrIHRvIGVtYmVkIGluIHRoZSByZXNwb25zZQogICAgICAgIHJlc3BvbnNlLkJvZHkgPSBKU09OLnN0cmluZ2lmeShhc0pTKQogICAgCiAgICAgICAgcmV0dXJuIFR5a0pzUmVzcG9uc2UocmVzcG9uc2UsIHNlc3Npb24ubWV0YV9kYXRhKQogICAgCiAgICB9CiAgICBsb2coIkJhdGNoIFRlc3QgaW5pdGlhbGlzZWQiKQ==",
                            "path": "/ilija",
                            "method": "GET",
                            "use_session": false
                        }
                    ]
                },
                "global_headers": {},
                "global_headers_remove": [],
                "global_size_limit": 0,
                "override_target": ""
            }
        }
    },
    "uptime_tests": {
        "check_list": [],
        "config": {
            "expire_utime_after": 0,
            "service_discovery": {
                "use_discovery_service": false,
                "query_endpoint": "",
                "use_nested_query": false,
                "parent_data_path": "",
                "data_path": "",
                "port_data_path": "",
                "target_path": "",
                "use_target_list": false,
                "cache_timeout": 60,
                "endpoint_returns_list": false
            },
            "recheck_wait": 0
        }
    },
    "proxy": {
        "preserve_host_header": false,
        "listen_path": "/test/",
        "target_url": "http://httpbin.org/",
        "strip_listen_path": true,
        "enable_load_balancing": false,
        "target_list": [],
        "check_host_against_uptime_tests": false,
        "service_discovery": {
            "use_discovery_service": false,
            "query_endpoint": "",
            "use_nested_query": false,
            "parent_data_path": "",
            "data_path": "hostname",
            "port_data_path": "port",
            "target_path": "/api-slug",
            "use_target_list": false,
            "cache_timeout": 60,
            "endpoint_returns_list": false
        }
    },
    "disable_rate_limit": false,
    "disable_quota": false,
    "custom_middleware": {
        "pre": [],
        "post": [],
        "post_key_auth": [],
        "auth_check": {
            "name": "",
            "path": "",
            "require_session": false
        },
        "response": [],
        "driver": "",
        "id_extractor": {
            "extract_from": "",
            "extract_with": "",
            "extractor_config": {}
        }
    },
    "custom_middleware_bundle": "",
    "cache_options": {
        "cache_timeout": 60,
        "enable_cache": true,
        "cache_all_safe_requests": false,
        "cache_response_codes": [],
        "enable_upstream_cache_control": false
    },
    "session_lifetime": 0,
    "active": true,
    "auth_provider": {
        "name": "",
        "storage_engine": "",
        "meta": {}
    },
    "session_provider": {
        "name": "",
        "storage_engine": "",
        "meta": null
    },
    "event_handlers": {
        "events": {}
    },
    "enable_batch_request_support": true,
    "enable_ip_whitelisting": false,
    "allowed_ips": [],
    "dont_set_quota_on_create": false,
    "expire_analytics_after": 0,
    "response_processors": [],
    "CORS": {
        "enable": false,
        "allowed_origins": [],
        "allowed_methods": [],
        "allowed_headers": [],
        "exposed_headers": [],
        "allow_credentials": false,
        "max_age": 24,
        "options_passthrough": false,
        "debug": false
    },
    "domain": "",
    "do_not_track": false,
    "tags": [],
    "enable_context_vars": false,
    "config_data": {
        "foo": "bar"
    }
}
```

mvdan:
@ilijabojanovic this went under my radar somehow, sorry.

In any case, this is sort of expected, given the error:

```
tyk-gateway_1    | time="Oct  3 10:39:02" level=error msg="Could not load user's TykJS: open ./js/tyk.js: no such file or directory"
```

It's true that we could somehow avoid the crash, but it would be a nasty error when loading the API definition one way or another. The root of the problem is that you're pointing the path to a file that does not exist.

You could argue that I should make the error bubble up better, but that would be a major refactor of how APIs are loaded. At the moment, errors are not handled at all and most of these issues are let through.

mvdan:
ping @ilijabojanovic - have you seen my comment above?

ilijabojanovic:
@mvdan yes, @buger approved for merge

Issue #495, opened 2017-02-09
Cache DNS to prevent DNS Spamming


lonelycode:
**UPDATED by @buger**

Go by default do not provide any DNS caching mechanism and on each connection does a DNS query. In the case when upstream does not support keep-alive connections, DNS query will be made on each HTTP request, which makes the huge load on DNS server. 

One of the cases to fix it on OS level by installing caching DSN proxy, but it is not possible in all the cases. So we need to introduce application level logic to cache DNS queries.

On first DNS query, for a given domain, it adds an entry to the cache. After we have a background job, which periodically goes through all cache entries and updates records in the background. Period of this job should be a configurable option, calling it TTL probably makes sense, however its job a bit different, so maybe we can find another name. 

Additionally, for removing stale records, we may also record last “query” time for a given domain and remove cache entries which were not used for last N seconds. 

## Dealing with multiple DSN records

DNS itself can return multiple records, and Go function which does DNS lookup returns array of IPs as well, see `net.DefaultResolver.LookupIPAddr`. So our cache should cache multiple entries as well. 

If we look at how Go internally works
https://golang.org/src/net/dial.go?s=10090:10150#L318, by default, it goes through list one by one, until a successful connection established. 

Another approach is so-called Round Robin DNS, where you obviously round robin through multiple records. 

I think we should actually support both, and add configurable option to specify DNS strategy: sequential or round-robin.

Interesting discussion on multiple DNS records https://webmasters.stackexchange.com/questions/10927/using-multiple-a-records-for-my-domain-do-web-browsers-ever-try-more-than-one

## Enable on API level
Not all APIs require such logic, only which has issues with DNS, so I think it makes sense enable this feature on API level while keeping cache TTL configured on gateway level. Should be added new API options called “enable_dns_caching”, and “dns_caching_strategy”, 

## Testing 

We include special DNS mock for writing unit tests which require multiple domains. In case of this task, this mock should be extended, to actually pass callback which customizes DNS server response, to return multiple records or similar, depending on test logic. 

See `initDNSMock` function in helpers_test.go

mvdan:
Perhaps something like https://github.com/viki-org/dnscache, but that package seems abandoned.

https://golang.org/src/net/dnsclient_unix.go also has a TODO about a small cache.

buger:
From what I see it should not be a problem to implement. Resolving DNS record can be done using `net/#LookupAddr` and caching itself can me smth like:

```
[...]

PR #1101, opened 2017-09-11
revproxy: refactor default transport

mvdan:
revproxy: refactor default transport

Remove the TykTransporter wrapper around the http.Transport. Its only
purpose was SetTimeout, which was only used once and was very limited -
it did not override the Dial family of methods.

Second, be consistent about overriding and using DialContext instead of
Dial. Dial is defined as deprecated, and many libraries including std
will ignore it if DialContext is set. For now, use context.TODO() when
calling, which will equal to the previous behavior.

Third, sync the default transport with http.DefaultTransport - namely,
adding fields like MaxIdleConns and DualStack.

Finally, use a separate http.Transport object per proxied request.
Before we were modifying and using a global in-place, which meant that
concurrent requests could run into race conditions and misbehave if
requiring different transport tweaks.

This also fixes one of the last two staticcheck warnings, stemming from
handler_websocket's use of net.Dial.

Updates #495 - we will need a consistent default transport to cache the
DNS queries, and this one seems logical.

### DIFF
--- a/handler_websocket.go
+++ b/handler_websocket.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"crypto/tls"
 	"errors"
 	"io"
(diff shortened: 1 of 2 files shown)


buger:
^ conflict

mvdan:
fixed, thanks

PR #1102, opened 2017-09-11
revproxy: greatly simplify singleJoiningSlash

mvdan:
Add tests to make sure we didn't break it. They pass both before and
after the reimplementation.

### DIFF
--- a/reverse_proxy.go
+++ b/reverse_proxy.go
@@ -317,42 +317,13 @@ var TykDefaultTransport = &TykTransporter{http.Transport{
 	TLSClientConfig:     &tls.Config{},
 }}
 
-func cleanSlashes(a string) string {
-	endSlash := strings.HasSuffix(a, "//")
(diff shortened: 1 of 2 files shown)


mvdan:
I've never liked this func, and finally got to redoing it.

Note that this brings it back to almost the same as `net/http/httputil`'s, from which this one forked. But it's slightly different:

```
func singleJoiningSlash(a, b string) string {
        aslash := strings.HasSuffix(a, "/")
        bslash := strings.HasPrefix(b, "/")
        switch {
        case aslash && bslash:
                return a + b[1:]
        case !aslash && !bslash:
                return a + "/" + b
        }
        return a + b
}
```

In other words, in `httputil`'s, `foo// + //bar` results in `foo///bar` while in ours it results in `foo/bar`. We deduplicate all slashes, they just deduplicate one.

PR #1105, opened 2017-09-12
Fix "skip_url_cleaning" by not using DefaultServeMux

### LINKED ISSUE #1093: Skip URL cleaning isn't working

buger:
The issue itself happen because we wrap `gorilla.Mux` router through `http.ServeMux`, which is kinda wrong, because `Mux` is already a `ServeMux` implementation. So we processing routes twice.

Default `ServeMux` implementation have own route sanitization logic, and can’t be configured with `SkipClean` option, like gorilla implementation.

The solution here is to stop using ServeMux and use Gorilla directly.

Problem with this approach is that we need to reload, eg. replace, the router inside `doReload` function. Neither gorilla or ServeMux support replacing the routes, so replacing the whole router object is the only solution. When we pass raw Gorilla Router object to the `Server.Handler`, and replace `mainRouter` in `doReload`, server do not see the changes: thats why we had issues when API reloads was not working.

There are 2 ways to solve this issue: expose server object as a global variable, and dynamically replace `Handler` attribute, or write simple `http.Handler` wrapper around `mainRouter`, which always calls the current router.

I’m not sure about safety of this approach but since previously we replaced `DefaultServeMux` directly, and had no any issues, I may assume that it should be thread safe.

Should fix #1093

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/justinas/alice"
 
+	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/config"
 )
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -563,18 +563,15 @@ func doReload() {
 	log.WithFields(logrus.Fields{
 		"prefix": "main",
 	}).Info("Preparing new router")
-	newRouter := mux.NewRouter()
-	mainRouter = newRouter
+	mainRouter = mux.NewRouter()
+	if globalConf.HttpServerOptions.OverrideDefaults {
```
buger: This is also a new addition, which I proposed in https://github.com/TykTechnologies/tyk/pull/1094.
Since we creating a new router, we should re-apply same router settings.

main.go
```diff
@@ -1190,6 +1187,13 @@ func startDRL() {
 	startRateLimitNotifications()
 }
 
+// The only purpose is to allow mainRouter to be dynamically replaced
+type appHandler struct{}
```
mvdan: I'd call this `mainHandler` to clarify that it's always linked to `mainRouter`.

Also, please use the godoc format as in `mainHandler's only purpose is...`

gateway_test.go
```diff
@@ -1029,3 +1030,48 @@ func TestProxyUserAgent(t *testing.T) {
...
+		specBytes, _ := json.Marshal(spec)
+		specFilePath := filepath.Join(globalConf.AppPath, spec.APIID+strconv.Itoa(i)+".json")
+		ioutil.WriteFile(specFilePath, specBytes, 0644)
+		defer os.Remove(specFilePath)
+	}
+
+	doReload()
+}
+
+func TestSkipUrlCleanning(t *testing.T) {
```
mvdan: `Cleaning`

gateway_test.go
```diff
@@ -1029,3 +1030,48 @@ func TestProxyUserAgent(t *testing.T) {
 		}
 	}
 }
+
+func buildAndLoadAPI(apiGens ...func(spec *APISpec)) {
```
mvdan: I'd make this use a temporary AppPath directory instead - we already do this in other tests. It's the best way to ensure that nothing else is in the dir, and that even if we panic, we won't leave anything behind. And you can delete with a single `os.RemoveAll` call.

gateway_test.go
```diff
@@ -1029,3 +1030,48 @@ func TestProxyUserAgent(t *testing.T) {
...
+	for i, gen := range apiGens {
+		spec := &APISpec{APIDefinition: &apidef.APIDefinition{}}
+		json.Unmarshal([]byte(sampleAPI), spec)
+		gen(spec)
+		specBytes, _ := json.Marshal(spec)
+		specFilePath := filepath.Join(globalConf.AppPath, spec.APIID+strconv.Itoa(i)+".json")
+		ioutil.WriteFile(specFilePath, specBytes, 0644)
```
mvdan: Check for an error here, just in case.

gateway_test.go
```diff
@@ -1029,3 +1030,48 @@ func TestProxyUserAgent(t *testing.T) {
 		}
 	}
 }
+
+func buildAndLoadAPI(apiGens ...func(spec *APISpec)) {
+	for i, gen := range apiGens {
+		spec := &APISpec{APIDefinition: &apidef.APIDefinition{}}
+		json.Unmarshal([]byte(sampleAPI), spec)
```
mvdan: Unmarshal into `spec.APIDefinition`, just in case

main.go
mvdan: I wonder how this ever worked properly before this change.

gateway_test.go
buger: Not sure what you mean here

mvdan: `sampleAPI` is an `APIDefinition`, not an `APISpec`. One is embedded into the other, so in practice it's probably fine, but it could cause issues if any field names ever conflict or if the JSON string ever includes fields directly from `APISpec`.

So I'm saying that you should do `json.Unmarshal([]byte(sampleAPI), spec.APIDefinition)`, to make sure none of those are possible.

buger: 👌

main.go
buger: It did not worked :)

buger:
Fixed comments

ilijabojanovic:
@buger approved for merge

PR #1110, opened 2017-09-12
all: use http.DefaultClient more often

mvdan:
Using &http.Client{} means an extra allocation and a bit of added
complexity. Using the same client more often will also simplify changing
the global client in certain ways.

### DIFF
--- a/plugins.go
+++ b/plugins.go
@@ -425,8 +425,6 @@ func (j *JSVM) LoadTykJSApi() {
 		u.Path = hro.Resource
 		urlStr := u.String() // "https://api.com/user/"
 
-		client := &http.Client{}
-
 		var d string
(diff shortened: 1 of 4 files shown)


plugins.go
```diff
@@ -444,7 +442,7 @@ func (j *JSVM) LoadTykJSApi() {
 			r.Header.Set(k, v)
 		}
 		r.Close = true
-		resp, err := client.Do(r)
```
buger: I think in future we will use custom client, in order to override TLS settings, but for now make sense to merge for consistency.

mvdan: Yes - also because `http.DefaultClient` may be modified by something else, e.g. a badly written library.

PR #1111, opened 2017-09-12
Fixes #971 Adds initial capability to add a gRPC Proxy to Tyk using bundler or direct API Def config

### LINKED ISSUE #971: Enable initial REST to gRPC proxying
Feature 

We can't proxy to gRPC endpoints directly.

1. I should be able to upload, or use a CLI command to generate a Go-plugin from a `protoc` file.
2. Reference this plugin name from my API Definition in a `grpc_endpoints` section in the MW.
3. Add paths to the `grpc_endpoints` configuration section that map to the services that are being proxied.

We could use [this library](https://github.com/grpc-ecosystem/grpc-gateway) to provide the base functionality of plugin creation since it already generates a `gw` object that just needs binding to a muxer.

lonelycode:
Fixes #971


.travis.yml
```diff
@@ -25,6 +25,9 @@ services:
   - redis-server
 
 install:
+  # gRPC Proxy as plugins means we can't have these vendored
+  - go get "github.com/grpc-ecosystem/grpc-gateway/runtime"
+  - go get "golang.org/x/net/trace"
```
mvdan: This is now fixed and they're vendored - remove?

api_loader.go
```diff
@@ -539,6 +544,17 @@ func processSpec(spec *APISpec,
 		}).Info("Loaded: ", spec.Name)
 	}
 
+	if spec.CustomMiddleware.GRPCProxy.Path != "" {
+		pluginPath := spec.CustomMiddleware.GRPCProxy.Path
+		pathPrefix := globalConf.MiddlewarePath
+		if spec.CustomMiddlewareBundle != "" {
+			pathPrefix = path.Join(getTykBundlePath(), spec.APIID+"-"+spec.CustomMiddlewareBundle)
+		}
+		filePath := path.Join(pathPrefix, pluginPath)
```
mvdan: `filepath.Join`

api_loader.go
```diff
@@ -539,6 +544,17 @@ func processSpec(spec *APISpec,
 		}).Info("Loaded: ", spec.Name)
 	}
 
+	if spec.CustomMiddleware.GRPCProxy.Path != "" {
+		pluginPath := spec.CustomMiddleware.GRPCProxy.Path
+		pathPrefix := globalConf.MiddlewarePath
+		if spec.CustomMiddlewareBundle != "" {
+			pathPrefix = path.Join(getTykBundlePath(), spec.APIID+"-"+spec.CustomMiddlewareBundle)
```
mvdan: `filepath.Join`

grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,118 @@
...
+	"os"
+	"plugin"
+	"strings"
+
+	"github.com/akutz/gpds/lib"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+)
+
+type conf struct{}
```
mvdan: this global name throws me off a bit - reading its declaration, I have no idea what it's for.

grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,118 @@
...
+
+func LoadGRPCProxyPlugin(path string, targetURL string, gRPCProxyMux *runtime.ServeMux) error {
+	var err error
+	var p *plugin.Plugin
+
+	// Load our plugin
+	if p, err = plugin.Open(path); err != nil {
```
mvdan: Replace these three lines with

```
p, err := ...
if err != nil {
```

grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,118 @@
...
+		log.Fatal(err)
+		return err
+	}
+
+	// lookup the plug-in's Types symbol; it's the type map used to
+	// register the plug-in's modules
+	tmapObj, err := p.Lookup("Types")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: failed to lookup type map: %v\n", err)
+		os.Exit(1)
```
mvdan: A func that errors but may also exit? (here and below)

grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,118 @@
...
+	// register the plug-in's modules
+	for k, v := range tmap {
+		lib.RegisterModule(k, v)
+	}
+
+	// Instantiate a copy of the module registered by the plug-in.
+	modGo := lib.NewModule("register")
+
+	ctx := context.Background()
+	ctx, _ = context.WithCancel(ctx)
```
mvdan: What's the point of this line if you discard the `cancel` func?

grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,118 @@
...
+		return err
+	}
+
+	return nil
+}
+
+type v2Config struct {
+	ctx  context.Context
+	mux  *runtime.ServeMux
+	e    string
```
mvdan: `e` is a bit too short to know what it is

plugin_build/README.md
```diff
@@ -0,0 +1,160 @@
...
+make check
+sudo make install
+```
+
+Then, `go get -u` as usual the following packages:
+
+```sh
+go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
+go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger
+go get -u github.com/golang/protobuf/protoc-gen-go
```
mvdan: I thought we vendored them?

lonelycode: Not sure my last comment saved - if building the pb.go files manually you need to do this otherwise you can't actually generate the templates

grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,106 @@
...
+	p, err := plugin.Open(path)
+	if err != nil {
+		return err
+	}
+
+	// lookup the plug-in's Types symbol; it's the type map used to
+	// register the plug-in's modules
+	tmapObj, err := p.Lookup("Types")
+	if err != nil {
+		return fmt.Errorf("error: failed to lookup type map: %v", err)
```
mvdan: All the errors have an `error: ` prefix, that can be removed

grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// assert that the Types symbol is a *map[string]func() interface{}
+	tmapPtr, tmapOk := tmapObj.(*map[string]func() interface{})
+	if !tmapOk {
+		return fmt.Errorf("error: invalid type map: %T", tmapObj)
+	}
+
+	// assert that the type map pointer is not nil
+	if tmapPtr == nil {
+		return fmt.Errorf("error: nil type map: type=%[1]T val=%[1]v", tmapPtr)
```
mvdan: This looks like bad formatting, even github highlights the `%` chars in red

grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,106 @@
...
+
+type v2Config struct {
+	ctx        context.Context
+	mux        *runtime.ServeMux
+	entrypoint string
+	opts       []grpc.DialOption
+}
+
+// Get returns the value for the specified key
+func (c *v2Config) Get(ctx context.Context, key string) interface{} {
```
mvdan: Have you come up with the `Get/Set` interface? If so, why not make it type-safe, like:

```
func (c *v2Config) Ctx() *context.Context { return c.ctx }
```

(or just exporting the fields, which is even easier)

plugin_build/README.md
mvdan: I thought we would rely on our vendored copies - otherwise plenty of things can go wrong, e.g. if those libs ever have any breaking changes

For example, we could tell people to use our `vendor/` as their `GOPATH`

plugin_build/wrap.go.tmpl
```diff
@@ -0,0 +1,50 @@
...
+		config.Get(ctx,"ctx").(context.Context),
+		config.Get(ctx,"mux").(*runtime.ServeMux),
+		config.Get(ctx,"e").(string),
+		getOpts())
+
+	return nil
+}
+
+// Do not edit this function
+func changeMe(ctx context.Context, mux *runtime.ServeMux, endpoint string, opts []grpc.DialOption) (err error) {
```
mvdan: `changeMe` that says `do not edit`? :)

reverse_proxy.go
```diff
@@ -534,23 +546,43 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
...
-			} else if res.StatusCode == 500 {
-				breakerConf.CB.Fail()
+
+	// Handle gRPC Proxy
+	log.Debug("===> GRPC Server service name: ", p.TykAPISpec.CustomMiddleware.GRPCProxy.Name)
+	if p.TykAPISpec.CustomMiddleware.GRPCProxy.Name != "" {
+		log.Debug("====> Calling: ", outreq.URL.String())
+		if breakerEnforced {
+			if breakerConf.CB.Ready() {
+				rec := httptest.NewRecorder()
```
mvdan: Are we sure that using `httptest` outside of a test is safe?

plugin_build/wrap.go.tmpl
```diff
@@ -0,0 +1,50 @@
+package main
```
mvdan: I suggest making these files `.go` files again, but with a `+build ignore` so that:

* they don't break the regular build
* tools like `gofmt` and `goimports` still fix them

mvdan: There's an example on how to do it here: https://golang.org/pkg/go/build/#hdr-Build_Constraints

plugin_build/README.md
```diff
@@ -0,0 +1,162 @@
...
+## Step 6: Create the gRPC Proxy plugin for Tyk:
+
+Copy the wrapper files from the [Tyk Github Repo](https://github.com/Tyktechnologies/tyk), they are:
+
+- `plugin_build/opts.go`
+- `plugin_build/wrap.go`
+
+Copy these to the same directory as your stubs.
+
+Open the *.pb.gw.go and *.pb.go files, you must change the package name from whatever was generated, to `main`, so the
```
mvdan: whenever you use `*` you should quote it, otherwise the format gets fucked up

plugin_build/README.md
```diff
@@ -0,0 +1,162 @@
...
+   protoc -I/usr/local/include -I. \
+     -I$GOPATH/src \
+     -I$GOPATH/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
+     --grpc-gateway_out=logtostderr=true:. \
+     path/to/your_service.proto
+   ```
+
+It will generate a reverse proxy `path/to/your_service.pb.gw.go`.
+
+Note: After generating the code for each of the stubs, in order to build the code, you will want to run ```go get .``` from the directory containing the stubs.
```
mvdan: I think for a single line, a single `` ` `` is enough?

grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// register the plug-in's modules
+	for k, v := range tmap {
+		lib.RegisterModule(k, v)
+	}
+
+	// Instantiate a copy of the module registered by the plug-in.
+	modGo := lib.NewModule("register")
+
+	ctx := context.Background()
```
mvdan: Related to the question below on the `Get/Set` interface - why is a context in the mix? We don't seem to make any use of it. And it seems unnecessary for getter/setters.

plugin_build/README.md
```diff
@@ -0,0 +1,162 @@
+# GRPC Proxy How To
```
mvdan: I understand `plugin_build` is just a guide. Is this going to move to the docs?

plugin_build/README.md
```diff
@@ -0,0 +1,162 @@
...
+
+**This is currently a very manual process due to how Golang's plugin mechanism works. We are working on a bakery-style
+docker image that will make this process more automated**
+
+*Edit* To make this easier, use the `tykio/bakery` container, it will generate a plugin file from your proto
+files without any additional steps.
```
mvdan: These two notes seem to contradict each other a bit.

plugin_build/README.md
lonelycode: It doesn't affect these instructions, this command generates the plugins needed by `protoc` in the previous step to do the generating, so the stuff in our vendor dir won;t be part of the go bin path. 

They would need to change their GOPATH *after* when they actually build the plugin.

grpcproxy/grpcproxy.go
lonelycode: I really don't want to mess with these because it's part of the plugin fix for type passing - it's literally a lift and shift from the `github.com/akutz/gpds/lib` code.

grpcproxy/grpcproxy.go
lonelycode: See the last comment, this mechanism solves a whole bunch of type conversion headaches. I'm not confident enough to redesign it, so if you want to take a shot, be my guest.

plugin_build/README.md
lonelycode: Eventually, this is alpha, so keeping a readme here is a good idea until we formalize

reverse_proxy.go
lonelycode: I do not know... I haven't found a better way to dump the request so we can use it after.

lonelycode:
@mvdan If you can suggest a good alternative to `httptest.NewRecorder()` would be happy to change it.

grpcproxy/grpcproxy.go
mvdan: I've had a look and it seems like context isn't used there at all. Are we going to make use of it? If not, I'd leave it out of the API.

reverse_proxy.go
mvdan: So if we're using the grpc proxy, we're buffering the entire response? That seems pretty bad to me. If someone wishes to use a grpc API that writes tons of data or writes data incrementally or over time, this won't work at all.

I assume you did this because you only have `ServeHTTP` to use from grpc. We should be calling that directly with the user's `ResponseWriter`, instead of reading the whole response into memory first like this.

Note that this is different from what `RoundTrip` does - that one does the request, but it doesn't buffer the entire response body into memory.

plugin_build/README.md
```diff
@@ -0,0 +1,161 @@
...
+Open the ``*.pb.gw.go` and `*.pb.go` files, you must change the package name from whatever was generated, to `main`, so the
+header for all files should read:
+
+    ```go
+    package main
+    ```
+
+Now you must edit the `opts.go` file:
+
+1. For specific gRPC options, edit the `getOpts()`` return value with the gRPC Options you need for your gateway
```
mvdan: one too many `` ` `` this time I think

plugin_build/README.md
```diff
@@ -0,0 +1,161 @@
...
+2. Change the name of `var doRegister = changeMe` to whatever is in the `*.pb.gw.go` file, it will be something like:
+
+    ```RegisterYourServiceHandlerFromEndpoint```
```
mvdan: If you indent something with four spaces or more, it's already a blockquote - no need for the quotes

mvdan:
The `httptest` thing seems like a release blocker to me, but otherwise this looks okay to me - you probably want another set of eyes as well, like Leon's.

reverse_proxy.go
lonelycode: Tyk has response processors - that process the response on the way back. We could completely bypass those but then the cache will also be unavailable.

What we basically need is a copy of the response object to make that work.

I don't know how else to proceed, I don't know enough about RoundTrip.

mvdan: We can leave things as is, but it's going to be a bit limited with the buffering. I guess we can address that if it ever becomes a concern.

mvdan:
Also, I just noticed no tests? I realise that a plugin test would be a nightmare, but you could decouple the plugin loading stuff from the actual important code and test that part.

lonelycode:
I have changed the context part, a context is needed by the function that gets called to register the endpoint in the gRPC Proxy (added a comment to that effect). 

`ctx` is also part of the module library that solves the other type sharing headaches. Unless we want to fork and modify that,m I'm leaving it as is as it doesn't hurt. Instead I am passing the Init() func a nil value which seems to be ok since it is not used).

Have added tests.

reverse_proxy.go
lonelycode: I can't see another way of grabbing the response, unless we can write a matching interface that will partially buffer like RoundTrip does... but that's beyond my ken.

mvdan:
I don't think the latest comments from yesterday were addressed - and remember to remove the `tunny` vendor copy

grpcproxy/grpcproxy.go
```diff
@@ -0,0 +1,115 @@
+package grpcproxy
+
+import (
+	"fmt"
+	"plugin"
+	gr "runtime"
```
mvdan: I'd do the opposite, as `runtime` is way more commonly Go's rather than GRPC's

grpcproxy/grpcproxy_test.go
```diff
@@ -0,0 +1,62 @@
...
+	f, err := generatePlugin()
+	pluginPath := fmt.Sprintf("../plugin_build/%v.so", f)
+	defer os.Remove(pluginPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	testMux := runtime.NewServeMux()
+	err = LoadGRPCProxyPlugin(pluginPath, "http://localhost", testMux)
+	if strings.Contains(gr.Version(), "go1.8") {
```
mvdan: Just skip the test entirely at start, it will be faster and we won't require it to error for no reason.

grpcproxy/grpcproxy_test.go
```diff
@@ -0,0 +1,62 @@
...
+	scriptName := "build_plugin.sh"
+	defer os.Remove(scriptName)
+	err := ioutil.WriteFile(scriptName, []byte(buildScript), 0755)
+	if err != nil {
+		return "", err
+	}
+
+	uid := uuid.NewV4().String()
+	cmd := exec.Command(fmt.Sprintf("./%v", scriptName), uid)
+	_, err = cmd.CombinedOutput()
```
mvdan: You don't need the output, use `Run`

grpcproxy/grpcproxy_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"strings"
+
+	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"github.com/satori/go.uuid"
+)
+
+var buildScript string = `#!/bin/bash
+cd ../plugin_build
+go build --tags="dummy" --buildmode=plugin -o $1.so
+`
```
mvdan: Wait, a two-line script? This would be trivial to do in a `exec.Cmd`. You can specify a different dir with `cmd.Dir` and the args can be passed as you'd expect.

lonelycode:
Which ones in particular? Some I won't fix (like the context).

mvdan:
> Which ones in particular? Some I won't fix (like the context).

The markdown formatting ones :)

grpcproxy/grpcproxy_test.go
lonelycode: Well, you donwant to ensure that a user compiling the source with 1.8 gets a meaningful error instead of a panic, so we do actually want to test that the error is indeed there ;-)

mvdan: Ah, good point.

lonelycode:
Bump

vendor/vendor.json
```diff
@@ -6,9 +6,7 @@
 			"checksumSHA1": "9BfDgevpBgdiP4B6DT9SvcuVyLs=",
 			"path": "github.com/Jeffail/gabs",
```
mvdan: you moved the vendor copy from `Jeffail` to `jeffail`, so update this too

grpcproxy/grpcproxy_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"github.com/satori/go.uuid"
+)
+
+func generatePlugin() (string, error) {
+	pName := fmt.Sprintf("%v.so", uuid.NewV4().String())
+	fileArg := fmt.Sprintf("-o=%v", pName)
+
+	cmd := exec.Command("go", "build", "--tags='dummy'", "--buildmode=plugin", fileArg)
```
mvdan: This isn't run in a shell, so remove the quotes around `dummy` (otherwise you really are using `'dummy'` with quotes as the build tag)

buger:
Looks good! 

However, I'm not sure what "Step 3" with adding custom option does, and why it is needed https://github.com/TykTechnologies/tyk/blob/f07bd7ab0c7312b52258a3755875ef5712d1698e/plugin_build/README.md#step-3-add-a-custom-option-to-the-proto-file
It defines mapping between gRPC method and proxied URL, right?

If so, seems like bakery don't do it, and readme should mention that this step needed even with bakery.

lonelycode:
That's the gRPC proxy specific code annotations to have the proto file generate a proxy in the first place.

You're right though the  readme should be clear that yu need to do this with the bakery.

lonelycode:
Bump

buger:
@matiasinsaurralde pls do the QA of this task. Once you confirm that everything works, can be merged.

mvdan:
bump @matiasinsaurralde

lonelycode:
We may need to drop this for v2.4, happy to do so, it's hyper experimental.

PR #1112, opened 2017-09-13
Fix SkipUrlClean by not using DefaultServeMux

### LINKED ISSUE #1093: Skip URL cleaning isn't working

mvdan:
Backported from master.

The issue itself happen because we wrap `gorilla.Mux` router through
`http.ServeMux`, which is kinda wrong, because `Mux` is already a
`ServeMux` implementation. So we processing routes twice.

Default `ServeMux` implementation have own route sanitization logic, and
can’t be configured with `SkipClean` option, like gorilla
implementation.

The solution here is to stop using ServeMux and use Gorilla directly.

Problem with this approach is that we need to reload, eg. replace, the
router inside `doReload` function. Neither gorilla or ServeMux support
replacing the routes, so replacing the whole router object is the only
solution. When we pass raw Gorilla Router object to the
`Server.Handler`, and replace `mainRouter` in `doReload`, server do not
see the changes: thats why we had issues when API reloads was not
working.

There are 2 ways to solve this issue: expose server object as a global
variable, and dynamically replace `Handler` attribute, or write simple
`http.Handler` wrapper around `mainRouter`, which always calls the
current router.

I’m not sure about safety of this approach but since previously we
replaced `DefaultServeMux` directly, and had no any issues, I may assume
that it should be thread safe.

Should fix #1093

### DIFF
--- a/main.go
+++ b/main.go
@@ -726,16 +726,13 @@ func doReload() {
 	log.WithFields(logrus.Fields{
 		"prefix": "main",
 	}).Info("Preparing new router")
-	newRouter := mux.NewRouter()
-	mainRouter = newRouter
+	mainRouter = mux.NewRouter()
+	if config.HttpServerOptions.OverrideDefaults {
+		mainRouter.SkipClean(config.HttpServerOptions.SkipURLCleaning)
+	}
 
-	loadAPIEndpoints(newRouter)
-	loadApps(specs, newRouter)
-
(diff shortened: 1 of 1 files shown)


mvdan:
@ilijabojanovic please re-test the skipclean fix on this backport, just in case I messed it up with the conflicts

PR #1113, opened 2017-09-13
all: unify deafultRouter with mainRouter

### LINKED ISSUE #915: Unify mainRouter and defaultRouter

mvdan:
They are the same, especially now that we got rid of http.ServeMux.

Fixes #915.


buger:
Long awaited change! 👏

PR #1119, opened 2017-09-13
all: get rid of some unnecessary globals

mvdan:
Skipped the ones in middlewares as they actually change behavior. Since
middlewares are set up per-API and re-created at each reload, a field is
not the same as a global.


buger:
^ conflict

mvdan:
Fixed

PR #1128, opened 2017-09-17
Refactor everything into packages...

lonelycode:
This isn't for merging, it's only a PoC that it's possible to move everything to packages using Aliases in a pretty short period of time.

Some packages here are too small, others mix functionality that shouldn't be mixed, but it avoids circular imports.

But... it runs :-)


buger:
@mvdan what is the best way to merge it (note that it is for 2.5 not 2.4)

mvdan:
Well, I understood this isn't to be merged. I seem to remember that @lonelycode did this as an exercise to prove that it can be split, and to get ideas of what chunks would be easiest to do in master first.

@lonelycode is that correct? what chunks of code did you see that would be easy to do for 2.5, or even 2.4?

lonelycode:
@mvdan @buger correct - this was done in a morning to see if it's possible, some of the packages will need to better extraction.

Given that by 2.5 master will have dramatically changed, it might be worth me doing it again closer to the time.

This PR is huge because the goal was to pull out apispec, it might actually be easier to inverse what this PR does and just try to do one or two of the smaller packages over time (outside in) instead of focussing on apispec.

mvdan:
Storage is somewhat doable at the moment, with the exception of the RPC one which pulls in some unrelated stuff. Pulling out the interface alone and always using the interface would already be a big load off apispec.

PR #1129, opened 2017-09-18
Applying `global_headers_remove` to X-Forwarded-For

### LINKED ISSUE #1149: `global_headers_remove` do not remove `X-Forwarded-For`
bug
Tyk automatically adds `X-Forwarded-For` to all requests, even if `global_headers_remove` include it. The issue is that this header is forbidden for some services (for example AWS API Gateway), and there is no way to remove it.
`X-Forwarded-For` should respect `global_headers_remove` and `versions.extended_paths.transform_headers` 
Remove `X-Forwarded-For` header on global or path level in API designer. Make a request, it will still contain it.

all

danp3d:
Fixed #1149

My first PR here (in fact, first time looking at the code). Tried to follow the same style I saw around, but please let me know if anything too strange :)
# Overview
## What is it doing:
Applying the `global_headers_remove` rule to `X-Forwarded-For`, too.
### TL;DR version
`Tyk` -> `AWS API Gateway w/ custom URL` doesn't work because of `X-Forwarded-For` header; trying to get around it. 
### Long explanation
I'm having to add a Serverless API (AWS API Gateway + Lambda) to Tyk. This API gateway has a custom URL (ie not the standard `execute-api` URL from AWS).
The issue is that it generates a CloudFront distribution which can't be configured, and by default it doesn't allow the `X-Forwarded-For` header, replying with a `307 - Temporary Redirection` and a `Location` header pointing to the custom URL (i.e. trying to bypass Tyk).
Since it's not configurable on the AWS side, I'm creating this PR to prevent X-Forwarded-For to be set.
## Why `global_headers_remove`
Because it makes sense. When I looked at the code, I wondered if I should add a new config flag (it makes sense considering how the code implemented it), but as an user I would be very confused.
I actually tried adding `X-Forwarded-For` to `global_headers_remove` as my first attempt, which made a lot of sense to me.

### DIFF
--- a/reverse_proxy.go
+++ b/reverse_proxy.go
@@ -371,6 +371,17 @@ func (p *ReverseProxy) CheckHardTimeoutEnforced(spec *APISpec, req *http.Request
 	return false, globalConf.ProxyDefaultTimeout
 }
 
+func (p *ReverseProxy) CheckHeaderAllowed(hdr string, spec *APISpec, req *http.Request) bool {
+	vInfo, _, _, _ := spec.Version(req)
+	for _, gdKey := range vInfo.GlobalHeadersRemove {
+		if strings.ToLower(gdKey) == strings.ToLower(hdr) {
+			return false
+		}
+	}
+
+	return true
+}
+
 func (p *ReverseProxy) CheckCircuitBreakerEnforced(spec *APISpec, req *http.Request) (bool, *ExtendedCircuitBreakerMeta) {
 	if !spec.CircuitBreakerEnabled {
 		return false, nil
@@ -487,7 +498,10 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 	}
 
 	addrs := requestIPHops(req)
-	outreq.Header.Set("X-Forwarded-For", addrs)
+	forwadedForHdr := "X-Forwarded-For"
+	if p.CheckHeaderAllowed(forwadedForHdr, p.TykAPISpec, req) {
+		outreq.Header.Set(forwadedForHdr, addrs)
+	}
 
 	// Circuit breaker
 	breakerEnforced, breakerConf := p.CheckCircuitBreakerEnforced(p.TykAPISpec, req)
--- a/reverse_proxy.go
+++ b/reverse_proxy.go
@@ -371,23 +371,25 @@ func (p *ReverseProxy) CheckHardTimeoutEnforced(spec *APISpec, req *http.Request
 	return false, globalConf.ProxyDefaultTimeout
 }
 
-func (p *ReverseProxy) CheckHeaderAllowed(hdr string, spec *APISpec, req *http.Request) bool {
-	log.Debug("Checking if header allowed: ", hdr)
+func (p *ReverseProxy) CheckHeaderInRemoveList(hdr string, spec *APISpec, req *http.Request) bool {
 	vInfo, versionPaths, _, _ := spec.Version(req)
-	deleteHeaders := vInfo.GlobalHeadersRemove
-
-	if found, meta := spec.CheckSpecMatchesStatus(req, versionPaths, HeaderInjected); found {
-		hmeta := meta.(*apidef.HeaderInjectionMeta)
-		deleteHeaders = append(deleteHeaders, hmeta.DeleteHeaders...)
-	}
-
-	for _, gdKey := range deleteHeaders {
+	for _, gdKey := range vInfo.GlobalHeadersRemove {
 		if strings.ToLower(gdKey) == strings.ToLower(hdr) {
-			return false
+			return true
 		}
 	}
 
-	return true
+	// Check path config
+	if found, meta := spec.CheckSpecMatchesStatus(req, versionPaths, HeaderInjected); found {
+		hmeta := meta.(*apidef.HeaderInjectionMeta)
+		for _, gdKey := range hmeta.DeleteHeaders {
+			if strings.ToLower(gdKey) == strings.ToLower(hdr) {
+				return true
+			}
+		}
+	}
+
+	return false
 }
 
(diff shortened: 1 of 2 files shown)


buger:
Hello! 

Thank you for a great work! 

Looks good so far, and make sense. I think that for consistency reasons, it should check not only "global_remove_headers", but also per "path" configuration: "versions.extended_paths.transform_headers". Imagine if you are refactoring your code base, and splitting monolith to microservices, and replace urls one by one: in this case you will want remove this header only from few urls (which, for example, point to custom API Gateway upstream urls).

danp3d:
@buger - Thanks mate.
Commited a few changes to ensure the rule is also applied to path configurations.
I'm not overly familiar with path configs (only ever used it in very basic scenarios) but I believe I'm covering what's needed in there.
Cheers!

reverse_proxy.go
```diff
@@ -371,6 +371,25 @@ func (p *ReverseProxy) CheckHardTimeoutEnforced(spec *APISpec, req *http.Request
...
 }
 
+func (p *ReverseProxy) CheckHeaderAllowed(hdr string, spec *APISpec, req *http.Request) bool {
+	log.Debug("Checking if header allowed: ", hdr)
+	vInfo, versionPaths, _, _ := spec.Version(req)
+	deleteHeaders := vInfo.GlobalHeadersRemove
+
+	if found, meta := spec.CheckSpecMatchesStatus(req, versionPaths, HeaderInjected); found {
+		hmeta := meta.(*apidef.HeaderInjectionMeta)
+		deleteHeaders = append(deleteHeaders, hmeta.DeleteHeaders...)
```
buger: I would prefer having 2 `for` cycles for both global and request specific headers, to avoid additional memory allocations.

reverse_proxy.go
```diff
@@ -371,6 +371,25 @@ func (p *ReverseProxy) CheckHardTimeoutEnforced(spec *APISpec, req *http.Request
 	return false, globalConf.ProxyDefaultTimeout
 }
 
+func (p *ReverseProxy) CheckHeaderAllowed(hdr string, spec *APISpec, req *http.Request) bool {
```
buger: If feel like "allowed" is not the right word here, because we talking about headers which should be removed.

reverse_proxy.go
```diff
@@ -487,7 +506,14 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 	}
 
 	addrs := requestIPHops(req)
-	outreq.Header.Set("X-Forwarded-For", addrs)
+	forwadedForHdr := "X-Forwarded-For"
+	if p.CheckHeaderAllowed(forwadedForHdr, p.TykAPISpec, req) {
+		log.Debug("Setting HDR: ", forwadedForHdr)
```
buger: Do not forget to remove log debug calls, and I think you can pass header name directly instead of using `forwadedForHdr` variable. Also, I think you can remove `outreq.Header.Del` at all, because our header transformation middleware will delete it, so you do the double work: goal of this functionality is not to override/add new heaver value.

reverse_proxy.go
danp3d: Yeah, I'm struggling a little bit with this one.
I wanted to keep consistent with other similar methods which are all called `Check<...>`.
But something like `CheckHeaderInRemoveList` means it will have to be negated in most cases (ie `if !CheckHeaderInRemoveList...`) which some people are a bit uncomfortable with (I personally have no problems with that).

I'm committing it as `CheckHeaderInRemoveList`, but name suggestions here are very welcome :smile:

buger:
@mvdan what do you think?

reverse_proxy_test.go
```diff
@@ -132,3 +132,137 @@ func TestRequestIP(t *testing.T) {
...
+					"not_versioned": true,
+					"versions": {
+						"Default": {
+							"name": "Default",
+							"use_extended_paths": true,
+							"global_headers_remove": [ "X-Random-Header" ]
+						}
+					}
+				}
+			}`,
```
mvdan: You could simplify this test by just having the version definition in each of these strings, like:

```
version: `{
"name": "Default",
"use_extended_paths": true,
"global_headers_remove": [ "X-Random-Header" ]
 }
```

The rest is always the same, so you can just use a `text/template.Template` to put it together.

reverse_proxy_test.go
```diff
@@ -132,3 +132,137 @@ func TestRequestIP(t *testing.T) {
...
+func TestCheckHeaderInRemoveList(t *testing.T) {
+	tests := []struct {
+		header   string
+		spec     string
+		expected bool
+	}{
+		{
+			header:   "X-Forwarded-For",
+			spec:     "",
+			expected: false,
```
mvdan: zero values (`""`, `false`) don't need to be put in each case

reverse_proxy_test.go
```diff
@@ -132,3 +132,137 @@ func TestRequestIP(t *testing.T) {
...
+			expected: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(fmt.Sprintf("%s:%t", tc.header, tc.expected), func(t *testing.T) {
+			rp := &ReverseProxy{}
+			r := new(http.Request)
+			r.Method = "GET"
+			r.URL = &url.URL{Path: "test"}
```
mvdan: Use `http.NewRequest` instead (if it needs a full URL, just use `http://test`)

buger:
@danp3d are you still here, and willing to do the last changes?

if you do not have time, just tell, and we fix it by our own.

Thank you!

danp3d:
@buger Sorry man, past few days have been seriously crazy.
I'm starting on the changes now - I'll let you guys know if I end up running out of time here.

Thanks!

danp3d:
Right-o, if you guys wanna take over from here that would be awesome.
I'm already using this in production BTW (been running for about a week now), and no signs of issues so far :+1:.

buger:
Perfect! Will be included into 2.4

PR #1132, opened 2017-09-18
config: move global conf here

mvdan:
This will enable sub-packages to access the global config without
needing circular dependencies into main.

Initially this wasn't done since the proper way to organize the project
would be to not have such a global state. However, there's no need for
this to block splitting main into multiple packages - this change has no
disadvantages.


mvdan:
I had a look, and this is the only thing blocking splitting the storage stuff into a sub-package.

lonelycode:
LGTM.

PR #1137, opened 2017-09-20
Swagger imports use tracked endpoints instead of white list, fixes #643

### LINKED ISSUE #643: Make swagger imports generate tracked URLs instead of white listed URLs
Feature

When I import a swagger definition, it generates a base API with all the paths listed as white lists, thbis is not ideal as it locks down the API unnecessarily.

The paths that are generated should be "tracked paths" instead as this is more useful and serves as a better start

This is only really possible since v2.2 or so because that's when tracked paths were introduced.

**NOTE**: This must be ported to the dashboard too.

lonelycode:


lonelycode:
There is a dashboard dependency on this, this only affects the gateway CLI command mode.

PR #1138, opened 2017-09-20
Add API-level rate limiter, fixes #643

### LINKED ISSUE #643: Make swagger imports generate tracked URLs instead of white listed URLs

lonelycode:
Uses the same rate limiting mechanism as per-token limits but at the api level. Adds RL section to the apidef:

```
"global_rate_limit": {
		"rate": 3,
		"per": 1
	}
```

Will require FE work as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .idea/
 .vagrant/
 temp/
+/middleware/bundles
 /middleware/e1d21f942ec746ed416ab97fe1bf07e8/
 /middleware/e1d21f942ec746ed416ab97fe1bf07e8_IGNORED/
 testapps/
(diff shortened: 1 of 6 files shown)


mvdan:
Is this PR pulling in all the swagger stuff by mistake?

mvdan:
Otherwise looks good - glad to see you coding again :)

lonelycode:
Swagger stuf removed

ilijabojanovic:
During testing this PR i found issue:
ENV: On-prem
Dashboard branch: TykTechnologies/tyk-analytics#453

When we set rate limit on api level and update rate limit after some time, old limit is applied on api def. Rate limit changes are visible only when we restart gateway, on gateway reload gateway does not propagate new rate limit values.

lonelycode:
@ilijabojanovic https://github.com/TykTechnologies/tyk/pull/1208

Issue #790, opened 2017-06-06
Allow to specify custom analytics tags using headers


buger:
**Do you want to request a *feature* or report a *bug*?**
feature
**What is the current behavior?**
It will be useful to allow users to tag requests with custom values, like api version, country, or maybe Request-ID. Right now it is set of fixed values like `api-*`, `org-*` and etc. 

With this custom tags, it will be possible to use standard log viewer to view advanced request details, without enabling detailed logging.

My proposal is to introduce a new config option, where we specify list of headers which will be extracted to be tags. For example if we specify `X-Request-Id` header to act as tag, it will generate  `x-request-id-<header_value>` tag.

lghiur:
Define a list of header in Apis designer (Advanced Options), name of section TBD.

nickReyn:
![](https://github.trello.services/images/mini-trello-icon.png) [Allow to specify custom analytics tags using headers](https://trello.com/c/alO1EibF/468-allow-to-specify-custom-analytics-tags-using-headers)

lonelycode:
Is there a back-end PR for this? This has a gateway dependency.

ilijabojanovic:
Verified and approved for merge

PR #1139, opened 2017-09-20
Added value tagging for headers, addresses #790

lonelycode:
Fixes https://github.com/TykTechnologies/tyk/issues/790

### DIFF
--- a/analytics_test.go
+++ b/analytics_test.go
@@ -79,3 +79,47 @@ func TestURLReplacer(t *testing.T) {
 		t.Error(recordCust.Path)
 	}
 }
+
+func TestTagHeaders(t *testing.T) {
+	req := testReq(t, "GET", "/tagmeplease", nil)
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("X-Tag-Me", "1")
+	req.Header.Set("X-Tag-Me2", "2")
+	req.Header.Set("X-Tag-Me3", "3")
+	req.Header.Set("X-Ignore-Me", "4")
+
+	existingTags := []string{"first", "second"}
+	existingTags = tagHeaders(req, map[string]interface{}{
+		"x-tag-me":  0,
+		"x-tag-me2": 0,
+		"x-tag-me3": 0},
+		existingTags)
+
+	if len(existingTags) == 2 {
+		t.Fatal("Existing tags have not been expanded")
+	}
+
+	if len(existingTags) != 5 {
+		t.Fatalf("Wrong number of tags, got %v, wanted %v", len(existingTags), 5)
+	}
+
+	check := []string{
+		"x-tag-me-1",
+		"x-tag-me2-2",
+		"x-tag-me3-3",
+	}
+
+	found := 0
+	for _, t := range existingTags {
+		for _, c := range check {
+			if t == c {
+				found += 1
+			}
+		}
+	}
+
+	if found != 3 {
+		t.Fatalf("Header values not proerly set, got: %v", existingTags)
+	}
+
+}
--- a/analytics_test.go
+++ b/analytics_test.go
@@ -89,10 +89,10 @@ func TestTagHeaders(t *testing.T) {
 	req.Header.Set("X-Ignore-Me", "4")
 
 	existingTags := []string{"first", "second"}
-	existingTags = tagHeaders(req, map[string]interface{}{
-		"x-tag-me":  0,
-		"x-tag-me2": 0,
-		"x-tag-me3": 0},
+	existingTags = tagHeaders(req, []string{
+		"x-tag-me",
+		"x-tag-me2",
+		"x-tag-me3"},
 		existingTags)
 
 	if len(existingTags) == 2 {
--- a/analytics_test.go
+++ b/analytics_test.go
@@ -103,23 +103,21 @@ func TestTagHeaders(t *testing.T) {
 		t.Fatalf("Wrong number of tags, got %v, wanted %v", len(existingTags), 5)
 	}
 
-	check := []string{
-		"x-tag-me-1",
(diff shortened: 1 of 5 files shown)


handler_success.go
```diff
@@ -75,6 +92,10 @@ func (s *SuccessHandler) RecordHit(r *http.Request, timing int64, code int, requ
 			alias = session.Alias
 		}
 
+		if len(s.Spec.TagHeaders) > 0 {
```
buger: I assume we should add tagging to error handler too.

lonelycode: See the same code in [error_handler.go](https://github.com/TykTechnologies/tyk/pull/1139/files#diff-e0d46d3f416ee63c099afead1b191696R115)?

buger: my bad

handler_success.go
```diff
@@ -43,6 +45,21 @@ type SuccessHandler struct {
 	BaseMiddleware
 }
 
+func tagHeaders(r *http.Request, th map[string]interface{}, tags []string) []string {
+	for k, v := range r.Header {
+		cleanK := strings.ToLower(k)
```
mvdan: I would use `net/http.CanonicalHeaderKey` here. Unless we need lowercase for a reason in particular?

In fact, the map is already using `CanonicalHeaderKey` internally, so you don't need to convert it at all.

analytics_test.go
```diff
@@ -79,3 +79,47 @@ func TestURLReplacer(t *testing.T) {
...
+		for _, c := range check {
+			if t == c {
+				found += 1
+			}
+		}
+	}
+
+	if found != 3 {
+		t.Fatalf("Header values not proerly set, got: %v", existingTags)
+	}
```
mvdan: This is a bit pedantic, but this test is incorrect - you're checking for three matches, not for all three strings above. If you got `"x-tag-me-1"` three times, the test would still pass.

A better, easier way to do it would be a `remaining := map[string]bool{...}` and then deleting from the map as you iterate, erroring at the end if there map is not empty.

handler_success.go
mvdan: Whatever we go for - canonical keys or lowercase keys - we should error if the user gives us a `TagHeaders` that contains keys that don't comply to the format. Or convert them ourselves.

lonelycode: I'll force lowercase, I don't want to leave it up to the user and converting to CanonicalHeaderKey seems unnecessary.

buger:
@ConsM it is ready for building the UI on top of it

buger:
@lonelycode why are you using map[string]interface{} here? Don't you think using just array of headers will be enough? Not sure what is the purpose of map.

buger:
It's not about performance, it's more about API definition design. Map values here do not play any role, at least from what I see.

mvdan:
I know, and I agree - my point was that he might have gone for maps for performance reasons. And I'm saying that the same can be accomplished by converting the list to a map when loading the API.

lonelycode:
I chose a map so that I didn't need to iterate an array on each request, that is correct.

I also thought that a map interface would mean that in future, if we want to extend this feature to do more than just copy the value of the header into the analytics, then we can add per-header config operations to it, so for example:

```
"tag_headers": {
    "x-role": {
    	"count": true,
    	"category_set": {
    		"user": "foo",
    		"admin": "bar",
    		"badger": "baz"
    	}
    }
}
```

This could then create:

```
x-role-N #Where n is an incremented counter
x-role-category-set-foo #if the value of x-role header was "user"
```

etc.

In terms of the API design, I realise it's arbitrary at the moment, but I'd rather be able to extend this than lock it down with an array to then have to change it to a map later.

buger:
Well, in fact, slices are faster than maps when number of items < 10 :)

Right now to me, this looks kind of ugly and complex. How would we understand the user in the documentation that value argument of the map means nothing? If the feature is not obvious to use, it becomes very confusing. 

I would say that if we want such advanced logic, it would make more sense exposing new context variable/array, which represents custom tags, and can be set via middleware.

lonelycode:
Fair enough - wanted to go with simplest option first. Instead will now have another nested loop instead... which IMO is uglier

Not sure how:

    tag_headers": {
        "x-role": true
        "x-banana" true
    }

Is confusing for a user - even more so when it's done behind the scenes in a UI :-)

lonelycode:
Actually, come to think of it, having this as a map isn't a great idea - but not because of the reasons you raised. 

This will need encoding into bson, and mongo throws an error if you use a field name that has reserved characters in it like a dot. So we'd need to base64 encode it like we do with the version names and a couple of other fields.

Now dots might not be standard header chars, but plenty of users use non-standard headers, so that *will* be a problem once it comes to integrating with the dashboard. Might as well save ourselves that headache.

Might make in array of objects, then we can still extend the objects if we need to down the line, I.e.

    tag_headers: [
         {name: "foo"}
    ]

buger:
If you going with using array of objects, it is a perfectly safe to use plain strings right now and when we want to introduce objects, update `string` type to `interface{}` and do some simple type checking.

buger:
Thank you! And sorry for being too picky 🙈 

@ConsM pls continue with the UI

buger:
@ilijabojanovic ready to be tested. Feature consist of 3 PRs: this, TykTechnologies/tyk-analytics-ui#171 and TykTechnologies/tyk-analytics#438

Go the advanced tab in API designer, header tags section. Add some tags, some of them should match header names, some not. Next, if you have header tag `X-Custom-Header`, and you make a request to your API with this header, its value will be extracted and turned to tag in analytics. So when you go to log browser, you will see smth like `x-custom-header-<value>`.

ilijabojanovic:
@mvdan approved for merge

PR #1146, opened 2017-09-22
Add extended routing capability, fixes #639

### LINKED ISSUE #639: As a user I should be able to define routing rules based on a combination of header, query param, content and path components
Feature 

Currently I can only redirect requests based on their inbound URL (using the URL rewriter), however if I want to route based on other factors (e.g. in a microservices pattern where a facade may expose a wide variety of underlying containers), then I can;t do this easily without resorting to virtual endpoints.

I should be able to route based on combinations of factors. 

**There are two components:**

For complex routing scenarios, I should probably use the JSVM and JSVM middleware, since I can just program it all, but currently when I return a different path, i cannot alter the target hostname. So an extension to the JSVM API might be the way to go so that I can return `"new_target": "foo"`, or add a host to the returned URL and Tyk will interpret it as a host-based reroute.

This can actually be triggered easily, since the URL rewriter just sets a context variable that sets a new target.

For simpler routes, we could extend the URL rewriter to include options such as:

```
"headers_must_contain": {"header": "{regex}"}
"query_param_must_contain"" {"variable": "{regex}"}
```

Then I can write a capture path, a regex to interpret for rewriting, and then only execute when the headers and query params are present. 

Open to more ideas on this?

lonelycode:
Fixed #639 

This PR adds rewrite "triggers" on top of the existing rewrite functionality. These are essentially sub-matches to allow a rewrite rule to have variations.

For example, for the URL `/foo/bar/baz` I want to rewrite to `/fooble/barble/bazble` if the query string variable `culprit` is set to `kronk`, but I want to redirect to `/foozle/barzle/bazzle` if the variable `culprit` is set to `yzma`.

I can do this with the new URL Rewriter as follows:

```
{
	"url_rewrites": [
		{
			"path": "/foo/bar/baz",
			"method": "GET",
			"match_pattern": "/foo/bar/baz",
			"rewrite_to": "/foo/bar/baz",
			"triggers": [
				{
					"on": "any",
					"options": {
						"query_val_matches": {
							"culprit": {
								match_rx": "kronk"
							}
						}
					}
					"rewrite_to": "/fooble/barble/bazble"
				}
				{
					"on": "any",
					"options": {
						"query_val_matches": {
							"culprit": {
								match_rx": "yzma"
							}
						}
					}
					"rewrite_to": "/foozle/barzle/bazzle"
				}
			]
		}
	]
}
```

Here if there is no trigger match, the rwrite will fallback to the parent `rewrite_to`, but if either of the other two are triggered, the rewrite target is changed.

Each trigger also sets a context variable for each match it finds, these context vars can then be used in the rewrites, so the above example with context vars could be rewritten to:

```
{
	"url_rewrites": [
		{
			"path": "/foo/bar/baz",
			"method": "GET",
			"match_pattern": "/foo/bar/baz",
			"rewrite_to": "/foo/bar/baz",
			"triggers": [
				{
					"on": "any",
					"options": {
						"query_val_matches": {
							"culprit": {
								match_rx": "kronk"
							}
						}
					}
					"rewrite_to": "/fooble/barble/bazble?victim=$tyk_context.trigger-0-culprit-0"
				}
				{
					"on": "any",
					"options": {
						"query_val_matches": {
							"culprit": {
								match_rx": "yzma"
							}
						}
					}
					"rewrite_to": "/foozle/barzle/bazzle?victim=$tyk_context.trigger-1-culprit-0"
				}
			]
		}
	]
}
```

Trigger contexts take the format: `$tyk_context.trigger-{n}-{name}-{i}` where `n` is the trigger index in the array, `nam` is the rx matcher name and `i` is the index of that match (since querystrings and headers can be arrays of values). 

The Trigger functionality supports:

1. Header matches
2. Query string variable/value matches
3. Path part matches, i.e. components of the path itself
4. Session meta data values (edit)
5. Payload matches

For each trigger, the trigger can either use the `on: any` or `on: all` formatting, for `any`, if any one of the options in the trigger is true, the rewrite rule is fired. for `all`, all the options must be satisfied. This is limited to triggers, not groups of triggers, these will be evaluated one by one.

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -129,6 +129,7 @@ type StringRegexMap struct {
 type RoutingTriggerOptions struct {
 	HeaderMatches   map[string]StringRegexMap `bson:"header_matches" json:"header_matches"`
 	QueryValMatches map[string]StringRegexMap `bson:"query_val_matches" json:"query_val_matches"`
+	PathPartMatches map[string]StringRegexMap `bson:"path_part_matches" json:"path_part_matches"`
 	PayloadMatches  StringRegexMap            `bson:"payload_matches" json:"payload_matches"`
 }
 
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -3,6 +3,8 @@ package apidef
 import (
 	"encoding/base64"
 
+	"regexp"
+
 	"github.com/lonelycode/osin"
 	"gopkg.in/mgo.v2/bson"
 )
@@ -20,6 +22,7 @@ type MiddlewareDriver string
 type IdExtractorSource string
 type IdExtractorType string
 type AuthTypeEnum string
+type RoutingTriggerOnType string
 
 const (
 	NoAction EndpointMethodAction = "no_action"
@@ -53,6 +56,11 @@ const (
 	OIDCUser      AuthTypeEnum = "oidc_user"
 	OAuthKey      AuthTypeEnum = "oauth_key"
 	UnsetAuth     AuthTypeEnum = ""
+
+	// For routing triggers
+	All    RoutingTriggerOnType = "all"
+	Any    RoutingTriggerOnType = "any"
+	Ignore RoutingTriggerOnType = ""
 )
 
 type EndpointMethodMeta struct {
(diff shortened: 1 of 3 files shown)


buger:
Nice! 

@lghiur @nickReyn UI for this feature can be non-obvious, and require planning.

buger:
There is no dashboard API needed here. The example you see here is part of API definitions url_rewrites object, mentioned by you.

nickReyn:
@lonelycode @buger @lghiur I've sketched out the below for how I see this could work. Can you guys take a look and feedback thoughts pls.
screen shot 2017-12-08 at 15 40 43

In terms of how a user get's to this, i see a few options. Number 1 is my preferred. 

1. We add a button close to the `Match pattern` field, something like `Triggers` which opens a modal with the above in. This is my preferred option because the UI of the endpoint designer is already very complex, and adding more might just add complexity when we could split this functionality into a diff UI element. e.g.
screen shot 2017-12-08 at 15 32 48

2. We add this UI to underneath the `Match pattern` and `Rewrite to` fields. However, as i said above, this just adds more complexity to an already very complex UI piece.

I'm not sure if there are different fields that need to show up if a user selects one of the other `Match types`, can you advise please? e.g. 

> Header matches
> Query string variable/value matches
> Path part matches, i.e. components of the path itself
> Session meta data values (edit)
> Payload matches

nickReyn:
Round two of this based on feedback from @lghiur and @buger.

This will be another plugin on the Endpoint designer, because it's too complex to create it as an extension of the current URL rewrites functionality, as discussed with Leon.

Check out a simple flow here: https://www.figma.com/proto/JJXxlc0Cr7yKaHgAYWSx2kRR/Routing-rules?scaling=contain&node-id=49%3A18&redirected=1

@lonelycode

nickReyn:
@buger @lghiur  and I had some slack convos about this and decided it's not ideal to have another plugin for this. I propose we have it as a separate section below the standard url rewrites which is activated on checkbox, labelled "Enable advanced url rewrites".

The flow is here: https://www.figma.com/proto/JJXxlc0Cr7yKaHgAYWSx2kRR/Routing-rules?scaling=contain&node-id=86%3A2

lghiur:
@lonelycode @buger 
I'm getting some errors when trying to build this branch

```
 github.com/TykTechnologies/tyk
./mw_url_rewrite.go:33: meta.Triggers undefined (type *apidef.URLRewriteMeta has no field or method Triggers)
./mw_url_rewrite.go:42: meta.Triggers undefined (type *apidef.URLRewriteMeta has no field or method Triggers)
./mw_url_rewrite.go:45: undefined: apidef.Any
./mw_url_rewrite.go:229: path.Triggers undefined (type apidef.URLRewriteMeta has no field or method Triggers)
./mw_url_rewrite.go:292: undefined: apidef.StringRegexMap
./mw_url_rewrite.go:322: undefined: apidef.StringRegexMap
./mw_url_rewrite.go:352: undefined: apidef.StringRegexMap
./mw_url_rewrite.go:380: undefined: apidef.StringRegexMap
./mw_url_rewrite.go:410: undefined: apidef.StringRegexMap
```

buger:
Will check

buger:
@lghiur works fine for me. How do you run it, using dev env via docker?

For some reason, it does not see changes in "apidef". If you trying to use this new fields in dashboard, you should govendor new `apidef` package changes. 

Going to merge this server-side PR.

chasebish:
How can I add this feature to my APIs?

lonelycode:
The documentation will be released with v2.5

PR #1148, opened 2017-09-24
Fix hmac bug when timezone have no code, only numeric representation

buger:
Not all timezones have abbreviations. I found this bug while testing
in my location, and since my timezone has no abbreviation, even if
date template tells use TZ code, Go replaced it with number value (in
my case -5). Because of this, all HMAC tests was failing on my machine.

Go date parsing mechanism do not support parsing timezone both with
numeric and abbreviation timezones, so we have to manually fallback to
it.


buger:
@mvdan can you pls take a look?

PR #1153, opened 2017-09-27
Devenv fix

joshblakeley:
If no connection was made while setting up the docker containers the response object was nil so it would panic. Quick fix by separating out the logic into an else-if.


mvdan:
Nice find!

mvdan:
You've added a dump file - please remove it.

buger:
👍  Congratulations with the first contribution :)

mvdan:
Also, I'm not sure I understand how this fix works. In the old code, we would use `resp.StatusCode` if and only if `err == nil`. And I'm fairly sure that `Client.Do` will only return nil responses when there is an error.

Are you certain you were running master? Could you show the panic message?

buger:
Actually, good point, missed it. At least one argument resp or err should be non-nil..

joshblakeley:
here's what I was getting:

tyk-gateway_1    | panic: runtime error: invalid memory address or nil pointer dereference
tyk-gateway_1    | [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xb115cc]
tyk-gateway_1    |
tyk-gateway_1    | goroutine 1 [running]:
tyk-gateway_1    | main.(*HTTPDashboardHandler).Register(0xc42025a870, 0x0, 0x0)
tyk-gateway_1    |  /develop/go/src/github.com/TykTechnologies/tyk/dashboard_register.go:84 +0x10c
tyk-gateway_1    | main.handleDashboardRegistration()
tyk-gateway_1    |  /develop/go/src/github.com/TykTechnologies/tyk/main.go:1154 +0x1fa
tyk-gateway_1    | main.listen(0x127c520, 0xc420166020, 0x0, 0x0, 0x12708a0, 0xc4200160a0)
tyk-gateway_1    |  /develop/go/src/github.com/TykTechnologies/tyk/main.go:1202 +0x31b
tyk-gateway_1    | main.main()
tyk-gateway_1    |  /develop/go/src/github.com/TykTechnologies/tyk/main.go:967 +0x1fd
compose_tyk-gateway_1 exited with code 2

mvdan:
Please add debug prints like:

```
fmt.Println(resp)
fmt.Println(err)
fmt.Println(resp.StatusCode)
```

Unfortunately the compiler doesn't use column numbers yet, so it's hard to tell exactly what in that line failed.

joshblakeley:
The resp object is printing as nil so calling resp.StatusCode caused the panic.

mvdan:
This is really weird. And `err` is not nil?

joshblakeley:
Error is "Get http://tyk-dashboard:3000/register/node: dial tcp 172.19.0.6:3000: getsockopt: connection refused"

mvdan:
Hang on, I misread. The issue is that we use `resp` in the body, not in the condition. You are absolutely right.

mvdan:
Remove the dump file (an extra commit is fine) and I'll merge :)

mvdan:
You also undid your other changes :)

joshblakeley:
ah, yes. I until them while i did that debugging. (Idiot)

mvdan:
Note that I squashed all your commits into one and used a better title than "devenv fix". You should try to do those things in the future. But it can always be fixed when merging on the github UI as I did, so it's not a big problem.

For the commits, you can probably find some guide by googling around. Most of the time, you'll want `git commit --amend` to update the last commit instead of creating a new one.

joshblakeley:
Thanks Dan, appreciated!

Issue #1151, opened 2017-09-26
`active` configuration flag not honored in API definition


klyr:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

- Only using Tyk gateway (no dashboard)
- Using .json configuration files

Creating a API with the "active: false" does not disable the API. Endpoints are still server regardless of the `active` flag being set or not.

**What is the expected behavior?**

When not using the dashboard,
When using .json configuration files for API definitions,

When the `active` flag is `false`, I expect the API to be disabled, even if the API configuration file is present.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**

Not working on current `master` (6a87c9cce0f8cd044f777f34a5f8d1e5ba91f421)
Was not working on (4b1a502bcf7a8cda41f2891fc64e0979630e901f)

In Tyk Gateway code, I cannot see any occurrence of code checking the value of the `Active` field of an API definition.

I suspect it's only used by the Dashboard, in this case, https://tyk.io/docs/tyk-rest-api/api-definition-object-details/ should be updated.

Cheers

MangoMM:
I think I have managed to recreate this issue:

Inactive API
```
curl -X POST \
  http://localhost:8080/tyk/apis/ \
  -H 'cache-control: no-cache' \
  -H 'x-tyk-authorization: 352d20ee67be67f6340b4c0605b044b7' \
  -d '{
    "name": "Inactive",
    "slug": "inactive-api",
    "api_id": "inactive-api",
    "org_id": "53ac07777cbb8c2d53000002",
    "use_keyless": true,
    "definition": {
        "location": "header",
        "key": "x-api-version"
    },
    "version_data": {
        "not_versioned": true,
        "versions": {
            "Default": {
                "name": "Default",
                "use_extended_paths": true
            }
        }
    },
    "proxy": {
        "listen_path": "/inactive-api/",
        "target_url": "http://httpbin.org/",
        "strip_listen_path": true
    },
    "active": false,
    "domain": ""
 }'
```

Active API
```
curl -X POST \
  http://localhost:8080/tyk/apis/ \
  -H 'cache-control: no-cache' \
  -H 'x-tyk-authorization: 352d20ee67be67f6340b4c0605b044b7' \
  -d '{
    "name": "Active Api",
    "slug": "active-api",
    "api_id": "active-api",
[...]

PR #1162, opened 2017-09-28
Skip inactive apis

MangoMM:
Resolves https://github.com/TykTechnologies/tyk/issues/1151 

Implements a filter to ensure inactive APIs are not loaded by the gateway.

### DIFF
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -381,7 +381,7 @@ func TestGetAPISpecsDashboardSuccess(t *testing.T) {
 	// Mock Dashboard
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.URL.Path == "/system/apis" {
-			w.Write([]byte(`{"Status": "OK", "Nonce": "1", "Message": [{"api_definition": {}}]}`))
+			w.Write([]byte(`{"Status": "OK", "Nonce": "1", "Message": [{"api_definition": {"active":true}}]}`))
 		} else {
 			t.Fatal("Unknown dashboard API request", r)
 		}
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -381,7 +381,7 @@ func TestGetAPISpecsDashboardSuccess(t *testing.T) {
 	// Mock Dashboard
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.URL.Path == "/system/apis" {
-			w.Write([]byte(`{"Status": "OK", "Nonce": "1", "Message": [{"api_definition": {"active":true}}]}`))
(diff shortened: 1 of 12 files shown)


mvdan:
By the way, your git email address is linked to another github account - @AhmetS. I would suggest using one username consistenly, as otherwise it's going to make things confusing for everyone.

api_loader.go
```diff
@@ -551,10 +551,29 @@ func loadApps(specs []*APISpec, muxer *mux.Router) {
...
+			}).Info("Skipping Inactive.")
+
+			continue
+		}
+		// copy & increment index
+		specs[i] = spec
+		i++
+	}
+	specs = specs[:i]
+
```
mvdan: You can simplify this by just doing this logic in the `for i, spec := range specs` loop. All you would need is:

```
		if !spec.Active {
 			log.WithFields(logrus.Fields{
 				"prefix":   "main",
 				"api_name": spec.Name,
 				"domain":   spec.Domain,
 			}).Info("Skipping Inactive.")
                        loadList[i] = &ChainObject{Skip: true}
 			continue
 		}
```

api_definition_test.go
```diff
@@ -378,7 +381,7 @@ func TestGetAPISpecsDashboardSuccess(t *testing.T) {
 	// Mock Dashboard
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.URL.Path == "/system/apis" {
-			w.Write([]byte(`{"Status": "OK", "Nonce": "1", "Message": [{"api_definition": {}}]}`))
+			w.Write([]byte(`{"Status": "OK", "Nonce": "1", "Message": [{"api_definition": {"active":true}}]}`))
```
mvdan: nit: please space this, like `{"active": true}`

api_loader.go
```diff
@@ -7,13 +7,13 @@ import (
...
-	"github.com/gorilla/mux"
-	"github.com/justinas/alice"
-
 	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/config"
 	"github.com/TykTechnologies/tyk/coprocess"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/gorilla/mux"
+	"github.com/justinas/alice"
```
mvdan: please undo this change

api_loader.go
```diff
@@ -39,7 +39,17 @@ func prepareStorage() (*RedisClusterStorageManager, *RedisClusterStorageManager,
 	return &redisStore, &redisOrgStore, healthStore, &rpcAuthStore, &rpcOrgStore
 }
 
-func skipSpecBecauseInvalid(spec *APISpec) bool {
+func filterSpec(spec *APISpec) bool {
```
mvdan: I'd prefer `skipSpec`, as otherwise it's hard to tell what `true` is used for without looking at the code

PR #1164, opened 2017-09-29
main: register net/http/pprof on reloads too

mvdan:
We didn't need to in the past as we layered the routers, meaning that we
never fully replaced the main router. Now that we do, we must set it up.
Otherwise, its routes will be 404s after any reload.

Also a small file closing fix.

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -935,6 +935,22 @@ func TestControlListener(t *testing.T) {
 	testHttp(t, tests, true)
 }
 
+func TestHttpPprof(t *testing.T) {
+	old := httpProfile
+	defer func() { httpProfile = old }()
+
+	testHttp(t, []tykHttpTest{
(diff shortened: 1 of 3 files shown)


mvdan:
this is blocking @ilijabojanovic from sending me a profile for the gateway leak, so somewhat high priority :)

buger:
Maybe just include it into `loadAPIEndpoints` ?

mvdan:
Perhaps, but I didn't given how they're different things - Tyk Admin API and pprof endpoints. That func also sets up a hostname and sub-path, neither of which would apply. I'd rather not mix the two, personally.

buger:
My main point here that it is a internal admin api, and it should be ok expose it to control api router.

mvdan:
Ah, I see. Sure.

mvdan:
Done, and added a small test.

PR #1175, opened 2017-09-29
mw: add jsonMarshal to body transform templates

### LINKED ISSUE #1145: Response Transform middleware strips escaping characters
Bug

Response transform middleware strips backslashes from JSON values

Resulting JSON object should have escaped characters preserved.

Input:
`{ "myField": "Here is \"the\" value" }`

Transform:
`{ "newField": "{{.myField}}" }`

Result:
`{ "newField": "Here is "the" value" }`

2.3.8

mvdan:
So that one can use the body templates to change XML into JSON (or
otherwise construct JSON objects) without having to worry about properly
encoding values. The simplest case of this is strings, since using
plaintext could easily result in unescaped quote characters.

Now, instead of using the following template:

	"{{ $val }}"

One should do:

	{{ $val | jsonMarshal }}

Fixes #1145.

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -13,7 +13,7 @@ import (
 	"regexp"
 	"strings"
 	"sync/atomic"
-	textTemplate "text/template"
+	"text/template"
 	"time"
 
 	"github.com/rubyist/circuitbreaker"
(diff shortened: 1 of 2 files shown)


mvdan:
ping @buger - also note that the user agrees with my approach in the original issue

buger:
Looks good, ready for testing

ilijabojanovic:
@mvdan please add this support to dashboard BE
https://www.screencast.com/t/UHXatVKz7

ilijabojanovic:
@buger @mvdan approved for 2.4 merge

PR #1176, opened 2017-09-29
Loading API specs should skip inactive APIs.

### LINKED ISSUE #1151: `active` configuration flag not honored in API definition

MangoMM:
Resolves #1151

### DIFF
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -38,7 +38,8 @@ const sampleDefiniton = `{
 	"proxy": {
 		"listen_path": "/v1",
 		"target_url": "` + testHttpAny + `"
-	}
+	},
+	"active": true
 }`
 
 const nonExpiringDef = `{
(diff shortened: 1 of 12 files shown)


lonelycode:
@mvdan We are going to have to undo this - it will break backwards compatibility for all CE users upgrading that do not have this flag set, which is extremely likely since all our demo files do not include it for file-based.

lonelycode:
Bump to v3 or to 2.6 with *ample* warning

PR #1178, opened 2017-09-30
proxy: fix and test multi-target reverse proxy

### LINKED ISSUE #1177: NewMultiTargetProxy always returns nil on master
NewMultiTargetProxy return nil

bug bug

mvdan:
I broke it when replacing the false Init method with a regular
constructor, since I forgot to replace the nil return with the actual
object.

Add a test that covers both the default proxy and an overriding one. It
was crashing as expected. Fix the constructor, making the test pass.

While at it, clean up the code quite a bit:

* no need to parse TargetURL again
* getProxyForRequest can be much simpler
* unindent some of the code

Fixes #1177.

### DIFF
--- a/api_loader.go
+++ b/api_loader.go
@@ -243,9 +243,8 @@ func processSpec(spec *APISpec, apisByListen map[string]int,
 	}
 
 	// Already vetted
-	remote, _ := url.Parse(spec.Proxy.TargetURL)
+	spec.target, _ = url.Parse(spec.Proxy.TargetURL)
 
-	spec.target = remote
 	var proxy ReturningHttpHandler
 	if enableVersionOverrides {
 		log.WithFields(logrus.Fields{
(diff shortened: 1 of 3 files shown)


mvdan:
Should also be a nice coverage bump, since this file had zero coverage.

api_loader.go
```diff
@@ -243,9 +243,8 @@ func processSpec(spec *APISpec, apisByListen map[string]int,
 	}
 
 	// Already vetted
-	remote, _ := url.Parse(spec.Proxy.TargetURL)
+	spec.target, _ = url.Parse(spec.Proxy.TargetURL)
```
buger: I know it is not your code, but still, don't you think it can cause a data race issues because spec object is shared across requests? In most cases, it will always be single value, but if there is lot of upstream rewrites logic it potentially looks dangerous.

mvdan: I don't know what you mean. This field is only written once when loading the spec. And it is read any number of times later. There is no possibility for a data race here.

Although, looking into this made me realise that this field is useless, as we only use it like a regular variable. So I'll simply make it one.

mvdan: Err, scratch that, it is used, I just misused grep horribly.

My point above still stands. If you can reason the existance of a race, feel free to open an issue (with a race detector error, preferrably). But I don't see the point of making this PR bigger, as it is an unrelated change/fix.

buger: Yeah, I did not mean that this issue should be in the context of this PR. Just wanted your thoughts on it.

PR #1181, opened 2017-10-02
Revert "Loading API specs should skip inactive APIs."

MangoMM:
Reverts TykTechnologies/tyk#1176

Unintended consequences for existing CE users. Users who do not use the active flag will have their APIs disabled.


mvdan:
You should re-open the issue too.

Note that some of these changes can be re-used in the better PR that I hope you'll work on :) sorry that you have to revert your first contribution, but you picked a seemingly easy yet complicated issue.

PR #1184, opened 2017-10-03
loader: simplify auth chain construction

mvdan:
First, move the "enabled" checks to EnabledForSpec, which is why it's
there in the interface to begin with. Let the loader still log when a
middleware was appended by making mwAppendEnabled return a bool.

Second, remove a duplicate useCoProcessAuth if statement.

Third, greatly simplify the logic to decide whether or not to use
AuthKey. The second part of the conditional checked that no other auth
middleware was loaded. We can do this by simply checking if the auth
chain is empty, instead of using a very complex and hard to maintain
chain of bools.

### DIFF
--- a/api_loader.go
+++ b/api_loader.go
@@ -347,75 +347,53 @@ func processSpec(spec *APISpec, apisByListen map[string]int,
 
 		// Select the keying method to use for setting session states
 		var authArray []alice.Constructor
-		if spec.UseOauth2 {
-			// Oauth2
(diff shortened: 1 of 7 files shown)


mvdan:
@buger this is what I meant in the review of your auth code the other day.

buger:
Looks good. @ilijabojanovic when you perform full 2.4 integration testing, do not forget to include OpenID testing, oAuth, and simple JWT one.

PR #1188, opened 2017-10-04
storage: new package, with the interface and redis

mvdan:
Move storage_handlers.go and redis_cluster_handler.go into this package.
We will keep anything related to storage decoupled there.

We can't put rpc_storage_handler.go there yet, as it pulls in other
parts of the main package and needs some further work. For example:

	./rpc_storage_handler.go:241:14: undefined: LoadDefinitionsFromRPCBackup
	./rpc_storage_handler.go:246:4: undefined: doLoadWithBackup
	./rpc_storage_handler.go:657:4: undefined: MainNotifier
	./rpc_storage_handler.go:657:24: undefined: Notification
	./rpc_storage_handler.go:718:5: undefined: handleDeleteHashedKey
	./rpc_storage_handler.go:722:4: undefined: handleDeleteKey

For now, this is better than nothing and it will help us to further pick
the main package into pieces.

Only a couple of package members needed exporting - the former doHash
and errKeyNotFound names.


buger:
coverals ok?

Issue #336, opened 2017-01-11
Feature request: lint configuration 


lonelycode:
Tyk has many configuration options, it's really easy to break an install if they are incorrect. It would be great if Tyk had a configuration test or validation option.

mvdan:
There are various layers of checks that come to mind - how deep should we go?

* parse errors (invalid configs)
* valid configs that would make tyk just error and exit
* valid configs that definitely look wrong or would do confusing things

abeMedia:
Could be specified as jsonschema and validated with a library like https://github.com/xeipuuv/gojsonschema

lonelycode:
Schema is a good start, I think this could be done incrementally in terms of complexity. From experience we know that some users have problems installing Tyk because they missed an option or they've actually just configured something wrong (like a connection string).

So maybe we do it in bits:

- v1: Check the config against a schema to make sure no basic errors have been made
- v2: Check for common screw ups in known configuration topologies, e.g. if db based config is enabled, that the dashboard connection string is enabled in all the right places.
- v3: Sanity check some of the values, so for example: Is redis reachable and does it accept writes? Is The dashboard reachable and responding? etc.

mvdan:
@lonelycode your layered approach sounds fine. I'll develop v1 and send a PR soon.

As far as I know this would be the first tyk tool we develop. I would say a subcommand like `tyk lint` would be the best option, since we'll probably add more. We don't want to pollute the `PATH` with tons of `tyk-*` binaries, each statically linking most of Tyk.

My proposal for testing is a series of config inputs (structs) with the expected warnings, if any. This is assuming that there are already tests in place for loading a config struct from a file (plain text).

Should we get @ilijabojanovic involved in this, since it's the design stage of a new feature?

buger:
@mvdan don't you think it make sense to include lint functionality to tyk-cli project?

mvdan:
It lints the server config, not the client config, so I don't see why it should go in the client.

lonelycode:
This is definitely for Tyk gateway - it's like `nginx -t`

This will need def need @ilijabojanovic input

mvdan:
Okay, I'll wait for ilija to give input on my implementation proposal above.

@lonelycode thoughts on a subcommand versus an option? I'm inclined for the former, but perhaps you want to avoid subcommands.

lonelycode:
subcommand might be more scalable

mvdan:
[...]

PR #1190, opened 2017-10-05
lint: first working version

mvdan:
See the commits. For #336.

The tool can be used via `tyk lint`. It will use the first configuration file it finds, just like the regular Tyk would.

Its features are very limited now - all it should do is notice when the config is invalid JSON, when it has fields that we don't know about, and when values are of a type we don't expect (number instead of string and so on).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -162,51 +162,51 @@ type Config struct {
 	// was written.
 	OriginalPath string `json:"-"`
 
-	ListenAddress                     string                 `json:"listen_address"`
-	ListenPort                        int                    `json:"listen_port"`
-	Secret                            string                 `json:"secret"`
-	NodeSecret                        string                 `json:"node_secret"`
-	TemplatePath                      string                 `json:"template_path"`
-	TykJSPath                         string                 `json:"tyk_js_path"`
-	MiddlewarePath                    string                 `json:"middleware_path"`
-	Policies                          PoliciesConfig         `json:"policies"`
-	UseDBAppConfigs                   bool                   `json:"use_db_app_configs"`
-	DBAppConfOptions                  DBAppConfOptionsConfig `json:"db_app_conf_options"`
-	DisableDashboardZeroConf          bool                   `json:"disable_dashboard_zeroconf"`
-	AppPath                           string                 `json:"app_path"`
-	Storage                           StorageOptionsConf     `json:"storage"`
-	EnableSeperateCacheStore          bool                   `json:"enable_separate_cache_store"`
-	CacheStorage                      StorageOptionsConf     `json:"cache_storage"`
-	EnableAnalytics                   bool                   `json:"enable_analytics"`
-	AnalyticsConfig                   AnalyticsConfigConfig  `json:"analytics_config"`
-	HealthCheck                       HealthCheckConfig      `json:"health_check"`
-	UseAsyncSessionWrite              bool                   `json:"optimisations_use_async_session_write"`
-	AllowMasterKeys                   bool                   `json:"allow_master_keys"`
-	HashKeys                          bool                   `json:"hash_keys"`
-	SuppressRedisSignalReload         bool                   `json:"suppress_redis_signal_reload"`
-	SupressDefaultOrgStore            bool                   `json:"suppress_default_org_store"`
-	UseRedisLog                       bool                   `json:"use_redis_log"`
-	SentryCode                        string                 `json:"sentry_code"`
-	UseSentry                         bool                   `json:"use_sentry"`
-	UseSyslog                         bool                   `json:"use_syslog"`
-	UseGraylog                        bool                   `json:"use_graylog"`
-	UseLogstash                       bool                   `json:"use_logstash"`
-	GraylogNetworkAddr                string                 `json:"graylog_network_addr"`
-	LogstashNetworkAddr               string                 `json:"logstash_network_addr"`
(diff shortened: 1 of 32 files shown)


buger:
Add `additionalProperties: true` to the root and all nested object to enable typo validation.

mvdan:
> Add additionalProperties: true to the root and all nested object to enable typo validation.

See the commit message:

> The schema originates from an old version we had in anotherrepo, so it is missing some fields. That will be improved in the future, when we disallow additional (unknown) fields.

mvdan:
Also, I forgot to vendor.

buger:
Got it. As the first draft looks neat, but I would still prefer having all the possible fields we have in configuration within this PR, and typo check enabled by default. Because this is one of the most common errors. 

Also, do you think it makes sense move schema file to the file, and embed it using go-bindata? Asking because we potentially integrate this functionality to all our components, so schema should be customizable. This is not scope of this PR, but just discussion for the future.

mvdan:
> Got it. As the first draft looks neat, but I would still prefer having all the possible fields we have in configuration within this PR, and typo check enabled by default. Because this is one of the most common errors.

Well, fighting gojsonschema to get the very basics to work took all the morning already, so I would rather do this step by step. It's up to you whether to review small PRs with single commits or one big PR with many commits at once.

I thought about making the schema available, but I don't know what the best way would be. `go-bindata` is a beast, so I would rather avoid it unless we really need it. We could always expose it as a Go name, making it available to other Go code via imports.

buger:
Well, most of the stuff here is done, so the rest is checking all attributes one by one, and ensuring that they are in jsonschema. We still have time for development of this issue, so I prefer to wait.

Also, have you tries automatically generating schema from GO type https://github.com/alecthomas/jsonschema ? Looks nice and using "definitions", like it does, may simplify schema.

Regarding embedding schema to code, let's continue this discussion when we going integrate it into different components.

mvdan:
Okay, I'll just keep on adding commits here then.

> Also, have you tries automatically generating schema from GO type https://github.com/alecthomas/jsonschema ? Looks nice and using "definitions", like it does, may simplify schema.

Yes - we already talked about this library, but we discarded it because it doesn't let you specify anything more than what `encoding/json` can already do. No custom types, no dependencies, etc. So whatever we do, we have to maintain the schema separately.

buger:
I more mean use it to generate initial JSON schema, not using it in runtime.

mvdan:
@buger over to you - I'd say this PR is big enough to warrant a merge. The schema is complete now.

mvdan:
Also, the reason I did the schema mostly manually is because the tools do a rather poor job. For example, they use definitions for all the config sub-types. This makes the schema jump around definitions all the time, when in reality it only makes sense for `StorageOptionsConf`, to not duplicate it.

buger:
Nice! 

buger:
@mvdan can you pls update description with the original issue ID, and what features needs to be tested.

mvdan:
done.

ilijabojanovic:
During testing PR i found this:
When i execute ./tyk lint, we have this error in in output 

root@89fe999d29d3:/develop/go/src/github.com/TykTechnologies/tyk# ./tyk lint
issues found in tyk.conf:
Monitor: Additional property Monitor is not allowed

mvdan:
@ilijabojanovic the Monitor error should now be fixed (there is a test too). Please test again.

mvdan:
I also rebased on master and added `log_level` to the schema as per @joshblakeley's recently merged change.

mvdan:
@MangoMM @joshblakeley since @buger will be gone for a bit still, perhaps you would want to review this code? The changes that Leon didn't review are the last commit. All the other commits were modified only for the rebase and for adding `log_level` to the schema.

If you could review the rest, it would help me get this merged faster.

Of course, feel free to review everything if you wish :)

config/config.go
```diff
@@ -284,7 +284,16 @@ var Default = Config{
 	},
 }
 
-// writeDefault will create a default configuration file
+func WriteConf(path string, conf *Config) error {
+	bs, err := json.MarshalIndent(conf, "", "    ")
```
MangoMM: New standards (since nodeJS) are to indent json files with 2 spaces.

mvdan: I would honestly go for tabs, but this is a war I don't want to get into. We use four spaces for JSONs everywhere (not just the gateway), so if we wanted to change it, it would need to be a consistent change. So not the time and place, I'm afraid.

MangoMM: Yep - Was just an fyi rather than an issue.

joshblakeley:
It looks good to me from what I can see and there's nothing obviously puzzling.

main.go
```diff
@@ -961,6 +963,22 @@ func getGlobalStorageHandler(keyPrefix string, hashKeys bool) storage.Handler {
...
+		}
+		if len(lines) == 0 {
+			fmt.Printf("found no issues in %s\n", path)
+			return
+		}
+		fmt.Printf("issues found in %s:\n", path)
+		for _, line := range lines {
+			fmt.Println(line)
+		}
+		return
```
MangoMM: Shouldn't this return a non-zero exit code too?

mvdan: Ah, very well seen.

mvdan: done - it's `1` now.

buger:
@ilijabojanovic can you pls test one more time

ilijabojanovic:
@buger approved for merge

Issue #1197, opened 2017-10-05
Custom Config passed through the Spec object to CoProcess Middlewares


ArturoSerrano:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
The current behavior is that the spec object in the coprocess middleware only contains the APIID and the ORGID.

**What is the expected behavior?**
We want to be able to pass any fields through the Spec object.  We are aware of the ConfigData being available to JavaScript plugins, however, due to the map[string]interface{} type not being compatible with any protobuf3 field, that object can't be passed to CoProccess plugins.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Not a bug
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
Latest

Please see the following requests relating to this:
https://community.tyk.io/t/python-plugin-spec-fields/1925
https://community.tyk.io/t/config-data-in-a-python-middleware/1924

I am working on a PR for this.  It will create another field in the tyk.conf that will be a map[string]string which can be inserted into the protobuf3 Spec field.

PR: https://github.com/TykTechnologies/tyk/pull/1198

matiasinsaurralde:
Hi, thanks, sounds like a very useful enhancement!

PR #1198, opened 2017-10-05
Added CoProccessSpecExtras to the Tyk.Conf

ArturoSerrano:
We want to be able to pass any fields through the Spec object. We are aware of the ConfigData being available to JavaScript plugins, however, due to the map[string]interface{} type not being compatible with any protobuf3 field, that object can't be passed to CoProccess plugins.

Please see the following requests relating to this:
https://community.tyk.io/t/python-plugin-spec-fields/1925
https://community.tyk.io/t/config-data-in-a-python-middleware/1924

#1197

### DIFF
--- a/coprocess.go
+++ b/coprocess.go
@@ -111,6 +111,15 @@ func (c *CoProcessor) ObjectFromRequest(r *http.Request) *coprocess.Object {
 			"OrgID": c.Middleware.Spec.OrgID,
 			"APIID": c.Middleware.Spec.APIID,
 		}
+
+		//Append ConfigData First Layer to Spec
+		for k, v := range c.Middleware.Spec.ConfigData {
+			switch v.(type) {
+			case string:
+				object.Spec[k] = v.(string)
+			}
+		}
+
 	}
 
 	// Encode the session object (if not a pre-process & not a custom key check):
--- a/coprocess_test.go
+++ b/coprocess_test.go
@@ -288,6 +288,12 @@ const basicCoProcessDef = `{
 	"proxy": {
 		"listen_path": "/v1",
 		"target_url": "` + testHttpGet + `"
+	},
+	"config_data": {
+		"key": "value",
+		"nest_key": {
+			"k": "v"
+		}
 	}
 }`
 
@@ -318,5 +324,11 @@ const protectedCoProcessDef = `{
 	"proxy": {
 		"listen_path": "/v1",
 		"target_url": "` + testHttpGet + `"
+	},
+	"config_data": {
+		"key": "value",
+		"nest_key": {
+			"k": "v"
+		}
 	}
 }`


mvdan:
> however, due to the map[string]interface{} type not being compatible with any protobuf3 field, that object can't be passed to CoProccess plugins.

Could you elaborate why protobuf3 can't work with that type?

If what you need for protobuf3 is concrete types like `map[string]string`, what I would then do for coprocess middlewares is take `ConfigData` as `map[string]interface{}`, and filter the values which are just `string`, feeding only the resulting `map[string]string` to protobuf.

Then, we could document that `ConfigData` can have any values when using Javascript, but only string values when using coprocess, as everything else will not be passed on. This would remove the need to add an extra field, and would remove the need for users to duplicate the data if they want to use both JS and Coprocess middleware.

coprocess.go
```diff
@@ -111,6 +111,12 @@ func (c *CoProcessor) ObjectFromRequest(r *http.Request) *coprocess.Object {
 			"OrgID": c.Middleware.Spec.OrgID,
 			"APIID": c.Middleware.Spec.APIID,
 		}
+
+		//Append CoProccess extractors
+		for k, v := range c.Middleware.Spec.CoProccessSpecExtras {
+			object.Spec[k] = v
+		}
```
mvdan: This probably needs nesting in a sub-object, such as `ConfigData`. Otherwise, if you put `"APIID": "foo"` in the API definition, you'll overwrite the real APIID field.

ArturoSerrano: Hi, sure I can elaborate on the issues with protobuf3.

1) The issue is that protobuf3 doesn't have a value that directly translates to the go object of `map[string]interface{}`.  The closest that we can generate is using the `Any` value in protobuf.  Unfortunately, this doesn't create the required object in Go, but rather a `*Any` which will require using reflection and an intermediate type to assign values to it.  Additionally, even if we did an intermediate to resolve the field, we would need to implement the same for the other languages that the protobuf is responsible for.  It's a much easier solution to conform to a map[string]string since that translates fine in all the languages.

2) I like the idea of filtering out from the ConfigData object instead of adding another field.  Though I'm concerned about the performance hit requests might take to check every field in the configdata for type assertion.  I'll look into it and see if Go can do this gracefully.

3) Nesting is the original issue with the protobuf.  map[string]string only supports one layer of key/values.  I think it would be rare that someone would duplicate other config keys inside the ConfigData field.  We could add additional filtering to prevent APIID or OrgID (The 2 defaults inserted into Spec) but then that would add a precedent to filter for any other future fields.


I will now work on the suggested approach of using ConfigData and only applying first-level objects.

ArturoSerrano:
The issue is that protobuf3 doesn't have a value that directly translates to the go object of map[string]interface{}. The closest that we can generate is using the Any value in protobuf. Unfortunately, this doesn't create the required object in Go, but rather a *Any which will require using reflection and an intermediate type to assign values to it. Additionally, even if we did an intermediate to resolve the field, we would need to implement the same for the other languages that the protobuf is responsible for. It's a much easier solution to conform to a map[string]string since that translates fine in all the languages.

I like the idea of filtering out from the ConfigData object instead of adding another field. Though I'm concerned about the performance hit requests might take to check every field in the configdata for type assertion. I'll look into it and see if Go can do this gracefully.

Nesting is the original issue with the protobuf. map[string]string only supports one layer of key/values. I think it would be rare that someone would duplicate other config keys inside the ConfigData field. We could add additional filtering to prevent APIID or OrgID (The 2 defaults inserted into Spec) but then that would add a precedent to filter for any other future fields.

ArturoSerrano:
I have now changed it from using a new field to parsing out the first layer of Config Data and writing it to the Spec field.  Go had an easy way to do this, so it seems like a better solution.

@mvdan please review and let me know if I can make further improvements

javi830810:
Just wanted to throw out there that a functionality like this it's pretty much needed for Rich Plugins, currently there's no way to provide data from the API spec into the Plugins.

mvdan:
> It's a much easier solution to conform to a map[string]string since that translates fine in all the languages.

Sounds good.

> Though I'm concerned about the performance hit requests might take to check every field in the configdata for type assertion.

If this operation depends on just `APIDefinition.ConfigData`, you could work it out once when loading the API into memory, and reuse the `map[string]string` for each request. You could add the field in `APISpec`, and call it something like `coprocessConfigData`, with a comment explaining why it's necessary and why it's a different type.

I wouldn't lose too much time on this, though. Where you added the code, there already are a bunch of allocations and other non-trivial work, besides the whole protobuf shenanigan. Looping over a small map and doing reflect operations on the values is likely not going to be a significant addition. Can't harm to cache this part though, so feel free.

> Nesting is the original issue with the protobuf.

Agh, you're right. How about going one level higher? Putting it at `object.SpecConfigData`, an extra `map[string]string`. You might need to update the protobuf definitions and the bindings for this. See `coprocess/proto/`, in particular `update_bindings.sh`.

The other option is, like you say, to assume no namespace clashes. It's always an option to document this, telling the user that it's up to them to not create clashes with other keys they will need. But this complexity seems avoidable to me (if my solution above, or any other, would work).

mvdan:
@javi830810 I believe rich plugins and coprocess are the same thing. What I'm not sure of is what this change would cover - would it be just gRPC, or also Python and Lua? I'm not familiar enough with the coprocess codebase to know. @matiasinsaurralde will likely know.

ArturoSerrano:
@mvdan Rich Plugins and Coprocess are the same thing.  This change would cover all plugins under the coprocess, so gRPC, Python and Lua.  Javascript is unaffected by this.

ArturoSerrano:
@mvdan @matiasinsaurralde Hey guys, just wanted to know if any further work is required for this PR? I believe the solution we have come up with provides useful features.  thanks

buger:
@ArturoSerrano it is going to be included for 2.5, but wait for a person who responsible for the plugins: @matiasinsaurralde 

Thank you!

buger:
What if we dump config_data into json string, and will pass it like string key. 
It will make things more compatible and consistent. 

What do you think?

buger:
@ArturoSerrano @matiasinsaurralde so, what do you think about my proposal, or it is not worth doing?

ArturoSerrano:
@buger Hey,  I think that's an interesting idea.  However, it would require the middleware to continuously parse string to json for every request.  I'm open to investigating it, will hopefully get a chance to look into it over the next couple of days

buger:
@matiasinsaurralde your thoughts?

matiasinsaurralde:
The current approach looks good. gRPC, Python and Lua plugins use the handler in `coprocess.go`, except JSVM.

buger:
Closing in favor of https://github.com/TykTechnologies/tyk/pull/1347 

Will be included in upcoming 2.5

PR #1202, opened 2017-10-06
Create new user pkg, with only SessionState for now

mvdan:
See the commit messages. Once this is done, the copying of all this code in the dashboard repo will be over.


mvdan:
Also, reviewing this commit by commit will probably be the easiest.

mvdan:
Thank you both!

PR #1207, opened 2017-10-09
oauth: simplify extended interface declaration

mvdan:
We can extend an interface without listing all of its methods again, by
simply writing the interface name.


joshblakeley:
Nice cheap way to cleanup a few lines

PR #1210, opened 2017-10-10
api: use a strip slashes middleware

mvdan:
To end the gorilla/mux trailing slash madness. Do this only for the
admin API, where we were using this workaround.

While at it, remove an unnecessary subrouter, improve a test error
message, and test that nested net/http/pprof endpoints work.

Note that a new keys route was added, as the existing one requires a
slash. Before, we only supported /keys/ and /keys/foo, but not /keys.
Now all three are supported. This surfaced now given how we rewrite the
/keys/ path into /keys.

### DIFF
--- a/main.go
+++ b/main.go
@@ -279,52 +279,66 @@ func getPolicies() {
 	}
 }
 
+// stripSlashes removes any trailing slashes from the request's URL
+// path.
+func stripSlashes(next http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		path := r.URL.Path
+		if trim := strings.TrimRight(path, "/"); trim != path {
+			r2 := *r
+			r2.URL.Path = trim
+			r = &r2
+		}
+		next.ServeHTTP(w, r)
+	}
+	return http.HandlerFunc(fn)
+}
+
 // Set up default Tyk control API endpoints - these are global, so need to be added first
 func loadAPIEndpoints(muxer *mux.Router) {
 	hostname := config.Global.HostName
 	if config.Global.ControlAPIHostname != "" {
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -279,52 +279,67 @@ func getPolicies() {
...
 
+// stripSlashes removes any trailing slashes from the request's URL
+// path.
+func stripSlashes(next http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		path := r.URL.Path
+		stripped := strings.TrimRight(path, "/")
+		if stripped != path {
+			r2 := *r
+			r2.URL.Path = path[:len(path)-1]
```
mvdan: whoops.

mvdan: (fixed now)

main.go
```diff
@@ -279,52 +279,66 @@ func getPolicies() {
 	}
 }
 
+// stripSlashes removes any trailing slashes from the request's URL
+// path.
```
MangoMM: nit

mvdan: ?

MangoMM: necessary to split the line for an extra couple of characters over length to be displayed in gh window? https://github.com/golang/go/wiki/CodeReviewComments#line-length

mvdan: It's not about github - I just wrap to 72 chars as a general rule. Perhaps not necessary here, but I'm consistent and it's automated.

MangoMM: If we are doing that, then would be good to build into some kind of linter as pattern is not followed everywhere by everyone so no consistency?

mvdan: I meant consistency with myself :) Perhaps a linter would be a good idea, if it was restricted to godocs. I don't think there is a tool for that.

PR #1211, opened 2017-10-10
api: unexport all the helpers

mvdan:
While at it, simplify a few of them too.


mvdan:
Ping @MangoMM @joshblakeley?

PR #1220, opened 2017-10-12
Make logging output level a config option. Fixes #1182

### LINKED ISSUE #1182: Trigger debug mode from tyk.conf instead of the command line

feature

The gateway can be started in debug mode using --debug or TYK_LOGLEVEL=debug command. 
However a hybrid customer who can only affect their gateway instance through it's configuration file would like to be able to start debug mode with a config option.

**What is the expected behaviour?**

User set's a json field in the tyk.conf such as "debug_mode": "true", and this sets the log output to debug mode.

N/A

All versions

joshblakeley:
- log_level field added to Tyk configuration. 

- If TYK_LOGLEVEL enviroment variable is unset and no --debug flag is used when executing Tyk then the configuration file will be searched for an alternative to the default info level logging output. The log_level value can be error, warn or debug; Info is redundant as that is the default logger output set in log/log.go. Fixes #1182

### DIFF
--- a/main.go
+++ b/main.go
@@ -779,6 +779,24 @@ func initialiseSystem(arguments map[string]interface{}) error {
 		afterConfSetup(&config.Global)
 	}
 
+	if os.Getenv("TYK_LOGLEVEL") == "" && arguments["--debug"] == false {
+		level := strings.ToLower(config.Global.LogLevel)
+		switch level {
+		case "", "info":
+			// default, do nothing
+		case "error":
+			log.Level = logrus.ErrorLevel
+		case "warn":
+			log.Level = logrus.WarnLevel
+		case "debug":
+			log.Level = logrus.DebugLevel
+		default:
+			log.WithFields(logrus.Fields{
+				"prefix": "main",
+			}).Fatalf("Invalid log level %q specified in config, must be error, warn, debug or info. ", level)
+		}
+	}
+
 	if config.Global.Storage.Type != "redis" {
 		log.WithFields(logrus.Fields{
 			"prefix": "main",
--- a/config/config.go
+++ b/config/config.go
@@ -242,6 +242,7 @@ type Config struct {
 	ReloadWaitTime                    int                                   `bson:"reload_wait_time" json:"reload_wait_time"`
 	ProxySSLInsecureSkipVerify        bool                                  `json:"proxy_ssl_insecure_skip_verify"`
 	ProxyDefaultTimeout               int                                   `json:"proxy_default_timeout"`
+	LogLevel                          string                                `json:"log_level"`
 }
 
 type CertData struct {


joshblakeley:
Maybe don't need to check if log_level is not "" as that doesn't matter to the logic.

main.go
```diff
@@ -779,6 +779,20 @@ func initialiseSystem(arguments map[string]interface{}) error {
...
 
+	if os.Getenv("TYK_LOGLEVEL") == "" && arguments["--debug"] == false && config.Global.LogLevel != "" {
+		switch strings.ToLower(config.Global.LogLevel) {
+		case "error":
+			log.Level = logrus.ErrorLevel
+		case "warn":
+			log.Level = logrus.WarnLevel
+		case "debug":
+			log.Level = logrus.DebugLevel
+		}
```
MangoMM: I would consider having a default in your switch case for unsupported log levels.

joshblakeley: From what I can tell, the default case is already handled when the logger is initialised by logger.get() when creating the log var where it will always default to info level.

mvdan: Josh is right. In `--debug`, we don't set it to `InfoLevel` if the flag isn't set, so this is consistent.

mvdan:
> Maybe don't need to check if log_level is not "" as that doesn't matter to the logic.

Yes, please remove that check.

main.go
```diff
@@ -779,6 +779,20 @@ func initialiseSystem(arguments map[string]interface{}) error {
...
+		log.WithFields(logrus.Fields{
+			"prefix": "main",
+		}).Debugf("Using log level defined in config. Enabling %s-level output", strings.ToLower(config.Global.LogLevel))
```
mvdan: I'd say this is a bit silly, given how it will only do something if the debug level is debug. Remove? Or just move it to the `"debug"` case.

main.go
mvdan: However, you want to error if the config has something that is none of these, `""`, nor `"info"`. You could do:

```
switch strings.ToLower(config.Global.LogLevel) {
case "", "info":
	// default, do nothing
case "error":
	log.Level = logrus.ErrorLevel
case "warn":
	log.Level = logrus.WarnLevel
case "debug":
	log.Level = logrus.DebugLevel
default:
	log.WithFields(...).Fatal("invalid log level in the config bla bla bla")
 }
```

Otherwise, setting `"log_level": "verbose"` or `"log_level": "debgu"` and such could cause confusion.

mvdan:
GH decided to duplicate the larger comment for some god damned reason.

main.go
mvdan: On second thought, I'd say just remove. If you set the debug log level, you'll see tons of debug logs anyway.

main.go
```diff
@@ -779,6 +779,23 @@ func initialiseSystem(arguments map[string]interface{}) error {
 		afterConfSetup(&config.Global)
 	}
 
+	if os.Getenv("TYK_LOGLEVEL") == "" && arguments["--debug"] == false {
+		switch strings.ToLower(config.Global.LogLevel) {
+		case "", "info":
+			//defualt, do nothing
```
mvdan: typo, and please space like `// foo`

main.go
```diff
@@ -779,6 +779,23 @@ func initialiseSystem(arguments map[string]interface{}) error {
...
+		case "error":
+			log.Level = logrus.ErrorLevel
+		case "warn":
+			log.Level = logrus.WarnLevel
+		case "debug":
+			log.Level = logrus.DebugLevel
+		default:
+			log.WithFields(logrus.Fields{
+				"prefix": "main",
+			}).Fatal("Invalid log level in config, must be error, warn or debug.")
```
mvdan: or info

mvdan: also include the string, like `Fatalf("Invalid log level %q ...", level)`

you can reuse the var like:

```
level := strings.ToLower(config.Global.LogLevel)
switch level { ... }
```

mvdan:
ping about including the bad string? https://github.com/TykTechnologies/tyk/pull/1220#discussion_r144324562

joshblakeley:
whoops

main.go
joshblakeley: why %q and not %s in this case?

mvdan: Because `%q` will double-quote the string (see https://golang.org/pkg/fmt/)

So then, for `   confusing"` instead of `invalid log level    confusing", expected ...` you'll get `invalid log level "   confusing\"", expected...`.

mvdan: In practice both would work ok as I don't think anyone would put weird chars here. But we might as well make the fatal message more foolproof.

Issue #1221, opened 2017-10-12
PYTHONPATH isn't correctly set in some environments


matiasinsaurralde:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Tyk isn't able to initialize the Python dispatcher because the PYTHONPATH isn't correctly set (sometimes it doesn't include the Tyk root directory: "/opt/tyk-gateway").
Same occurs with the extracted bundle directories.

**What is the expected behavior?**
The dispatcher should work fine.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
It's hard to reproduce but from internal reports we know that it occurs sometimes, specially when the users create their custom initialization scripts, etc.

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
All.

ilijabojanovic:
Released

PR #1223, opened 2017-10-13
PYTHONPATH enhancement/fix

matiasinsaurralde:
This is related to #1221. I'm open to discuss about this, the main thing here is to get rid of  `os.Getwd`.
To test this:
- Create a Python plugin and serve it as specified in [this guide](https://tyk.io/docs/customise-tyk/plugins/rich-plugins/python/custom-auth-python-tutorial/).
- Use the following configuration:
```json
{
    "enable_coprocess": true,
    "python_path_prefix": "/opt/tyk-gateway"
  }
```
- Run `tyk-python` or the Tyk binary from a different working directory, like:
```
cd /tmp
/opt/tyk-gateway/tyk-python
```
- The PYTHONPATH log line should include the `python_path_prefix`.

### DIFF
--- a/config.go
+++ b/config.go
@@ -143,6 +143,7 @@ type ServiceDiscoveryConf struct {
 type CoProcessConfig struct {
 	EnableCoProcess     bool   `json:"enable_coprocess"`
 	CoProcessGRPCServer string `json:"coprocess_grpc_server"`
+	PythonPathPrefix string `json:"python_path_prefix"`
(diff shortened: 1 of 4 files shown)


mvdan:
@matiasinsaurralde you need to fix CI (and probably run gofmt too)

mvdan:
I assume you'll do the same fix in master for 2.4.

PR #1224, opened 2017-10-16
corrected "authorisation" word

jaschweder:
Simple correction in README file for "authorization" word


mvdan:
Hi @jaschweder! Thanks for contributing.

I'm afraid that the current spelling isn't incorrect however, as it's the British spelling of the word. Both are valid in broad English terms, and we tend to prefer British spellings since most of the team is European.

Thanks again and have a good day!

PR #1229, opened 2017-10-18
Python fixes

matiasinsaurralde:
I'm incorporating the changes from #1221 and #1223.


mvdan:
These were approved and tested for 2.3.10, and the changes look the same and still fine. Merging.

PR #1231, opened 2017-10-19
coprocess_python: use defer C.free for all C strings

### LINKED ISSUE #1228: Possible Leak
https://github.com/TykTechnologies/tyk/blob/c607854b9d50d40b79695e9f60a50c47a356e3f8/coprocess_python.go#L247-L254

Should the IF in line 247 evaluate true because of any reason. The release memory a few lines below wont execute, that will mean a Leak.

Also this happens many different times in the Coprocess and C integration. The correct way to use this release methods is by using a `defer C.free(....`

matiasinsaurralde:
Fixes #1228.

### DIFF
--- a/coprocess_python.go
+++ b/coprocess_python.go
@@ -200,8 +200,8 @@ func (d *PythonDispatcher) Dispatch(objectPtr unsafe.Pointer) unsafe.Pointer {
 // DispatchEvent dispatches a Tyk event.
 func (d *PythonDispatcher) DispatchEvent(eventJSON []byte) {
 	CEventJSON := C.CString(string(eventJSON))
+	defer C.free(unsafe.Pointer(CEventJSON))
 	C.Python_DispatchEvent(CEventJSON)
-	C.free(unsafe.Pointer(CEventJSON))
 }
 
 // Reload triggers a reload affecting CP middlewares and event handlers.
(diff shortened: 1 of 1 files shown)


coprocess_python.go
```diff
@@ -213,8 +213,8 @@ func (d *PythonDispatcher) Reload() {
 func (d *PythonDispatcher) HandleMiddlewareCache(b *apidef.BundleManifest, basePath string) {
 	go func() {
 		runtime.LockOSThread()
-		defer runtime.UnlockOSThread()
```
mvdan: is this on purpose? are you sure it's safe to remove?

coprocess_python.go
```diff
@@ -240,27 +240,25 @@ func PythonLoadDispatcher() error {
 // PythonNewDispatcher creates an instance of TykDispatcher.
 func PythonNewDispatcher(middlewarePath, eventHandlerPath string, bundlePaths []string) (coprocess.Dispatcher, error) {
 	CMiddlewarePath := C.CString(middlewarePath)
+	defer C.free(unsafe.Pointer(CMiddlewarePath))
 	CEventHandlerPath := C.CString(eventHandlerPath)
+	defer C.free(unsafe.Pointer(CEventHandlerPath))
 	CBundlePaths := C.CString(strings.Join(bundlePaths, ":"))
+	defer C.free(unsafe.Pointer(CBundlePaths))
```
mvdan: We had an actual leak here as there were only two frees before, correct?

mvdan:
Also, use "Fixes #n" to actually close the issue once merged.

coprocess_python.go
matiasinsaurralde: @mvdan I've updated it with a `func` that performs both operations:`C.free` and `runtime.UnlockOSThread`.

coprocess_python.go
matiasinsaurralde: That's right, this is the missing one.

matiasinsaurralde:
Just a quick note: if #1260 gets merged, this won't be necessary at all.

buger:
Closing in favor of #1260

Issue #1227, opened 2017-10-18
Ensure Python issues (#1221, #1068) are also fixed in 2.4


mvdan:
To make sure we don't forget to push the changes to master.

mvdan:
Fixed by https://github.com/TykTechnologies/tyk/pull/1229.

PR #1237, opened 2017-10-23
Fix Python bundle on hot reloads (WIP)

matiasinsaurralde:
Related to #1227.

Bug:
https://github.com/TykTechnologies/tyk/issues/1238

### DIFF
--- a/coprocess/python/dispatcher.py
+++ b/coprocess/python/dispatcher.py
@@ -9,12 +9,13 @@ from tyk.event import TykEvent, TykEventHandler
 
 from gateway import TykGateway as tyk
 
+from importlib.machinery import SourceFileLoader
+
 class TykDispatcher:
     '''A simple dispatcher'''
 
     def __init__(self, middleware_path, event_handler_path, bundle_paths):
         tyk.log( "Initializing dispatcher", "info" )
-
         self.event_handler_path = path.join(event_handler_path, '*.py')
         self.event_handlers = {}
         self.load_event_handlers()
@@ -42,18 +43,10 @@ class TykDispatcher:
         return found_middleware
 
     def load_bundle(self, base_bundle_path):
-        bundle_path = path.join(base_bundle_path, '*.py')
-        bundle_modules = self.get_modules(bundle_path)
-        sys.path.append(base_bundle_path)
-        for module_name in bundle_modules:
-            middleware = self.find_middleware(module_name)
-            if middleware:
-                middleware.reload()
-            else:
-                middleware = TykMiddleware(module_name)
-                self.middlewares.append(middleware)
+        bundle_path = path.join(base_bundle_path, 'middleware.py')
+        middleware = TykMiddleware(bundle_path)
+        self.middlewares.append(middleware)
         self.update_hook_table()
-
 
     def load_middlewares(self):
         tyk.log( "Loading middlewares.", "debug" )
@@ -142,9 +135,7 @@ class TykDispatcher:
 
     def reload(self):
         tyk.log( "Reloading event handlers and middlewares.", "info" )
-
         self.purge_event_handlers()
         self.load_event_handlers()
 
-        self.purge_middlewares()
         self.load_middlewares()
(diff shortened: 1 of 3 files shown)


coprocess/python/dispatcher.py
```diff
@@ -141,10 +134,9 @@ def dispatch_event(self, event_json):
...
 
     def reload(self):
-        tyk.log( "Reloading event handlers and middlewares.", "info" )
-
+        tyk.log( "Reloading event handlers and middlewares...", "info" )
         self.purge_event_handlers()
         self.load_event_handlers()
 
-        self.purge_middlewares()
+        # self.purge_middlewares()
```
lonelycode: Should we just get rid of this line altogether?

coprocess/python/tyk/middleware.py
```diff
@@ -16,7 +17,8 @@ def __init__(self, filepath):
         self.handlers = {}
 
         try:
-            self.module = import_module(filepath)
+            source = SourceFileLoader('middleware', self.filepath)
```
lonelycode: Is it always going to be "middleware"? (Just nitpicking)

matiasinsaurralde: This is only used internally, it could be set to "x" and would still work.

coprocess/python/dispatcher.py
```diff
@@ -42,19 +43,11 @@ def find_middleware(self, path):
...
-        bundle_modules = self.get_modules(bundle_path)
-        sys.path.append(base_bundle_path)
-        for module_name in bundle_modules:
-            middleware = self.find_middleware(module_name)
-            if middleware:
-                middleware.reload()
-            else:
-                middleware = TykMiddleware(module_name)
-                self.middlewares.append(middleware)
+        bundle_path = path.join(base_bundle_path, 'middleware.py')
```
matiasinsaurralde: @lonelycode This line will enforce `middleware.py`, it's useful to avoid collisions with other package/file names and it's also required to use the alternative `importlib` loader function that I've added in `middleware.py`, if the user loads a bundle that doesn't have `middleware.py` a Python error will show up (won't panic). All our samples use `middleware.py`.

matiasinsaurralde: In case you think this is too strict for a patch, I could try to modify the code again to "scan" all the Python files and load the appropriate one -the one containing the hooks-, just like the previous functionality.

lonelycode: Sounds ok to me

lonelycode:
Looks ok to me :-) once QA KS happy let's roll with it.

matiasinsaurralde:
To test this:
- Configure Tyk to enable Python support, add your server URL & and set the Python path:
```json
{
  "enable_bundle_downloader": true,
  "bundle_base_url": "http://localhost/",
  "coprocess_options": {
    "enable_coprocess": true,
    "python_path_prefix": "/opt/tyk-gateway"
  }
}
```
- Create a protected API with these settings:
```json
{
  "use_keyless": false,
  "enable_coprocess_auth": true,
  "custom_middleware_bundle": "bundle-v1.zip"
}
```
- Start Tyk and hit your test API:
```
$ curl localhost:8080/test-api/
```
You should get 403 and a log line from Python: "Bundle v1: auth middleware is called".

- Now update your API definition to use the "bundle-v2.zip":
```json
{
  "use_keyless": true,
  "enable_coprocess_auth": true,
  "custom_middleware_bundle": "bundle-v2.zip"
}
```
- Trigger a hot reload:
```
$ curl http://localhost:8080/tyk/reload/ -H 'x-tyk-authorization: 352d20ee67be67f6340b4c0605b044b7'
```
- You should see "Bundle v2: loaded" in the Tyk log.
- Now send a request to the API again, you should get 403 and also "Bundle v2: auth middleware is called" in the log.

I've attached two ready-to-use bundles, you can serve them using your own web server:
[bundle-v2.zip](https://github.com/TykTechnologies/tyk/files/1406170/bundle-v2.zip)
[bundle-v1.zip](https://github.com/TykTechnologies/tyk/files/1406166/bundle-v1.zip)

ilijabojanovic:
@buger approved for merge

PR #1248, opened 2017-10-28
api: ignore status when decoding lists

mvdan:
We don't ever use that field, so there is no need to decode it.


mvdan:
bump, anyone?

PR #1252, opened 2017-10-30
lint: add first custom types

mvdan:
For paths (they must exist) and addresses (they must be hosts only).
With the appropriate tests.

We must do a dance with the format errors as it is not possible for a
format check function to return a custom error or message.

### DIFF
--- a/lint/lint.go
+++ b/lint/lint.go
@@ -2,6 +2,8 @@ package lint
 
 import (
 	"encoding/json"
+	"fmt"
+	"net"
 	"os"
 
 	schema "github.com/xeipuuv/gojsonschema"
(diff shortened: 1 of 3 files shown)


buger:
Looks good. Do you think we really need `addFormats` function? `gojsonschema.FormatCheckers` is the singleton anyway, so maybe just call `gojsonschema.FormatCheckers.Add` inside `init` function, or similar?

For the reference, there is alternative way to define custom error messages by overloading `gojsonschema.Locale` object like described here https://github.com/xeipuuv/gojsonschema

Until we have lot of custom message current realization is ok.

mvdan:
> Looks good. Do you think we really need addFormats function? gojsonschema.FormatCheckers is the singleton anyway, so maybe just call gojsonschema.FormatCheckers.Add inside init function, or similar?

Sure it's a singleton, but the point is to avoid that really long expression multiple times. I could have avoided the parameter, but why not make it a bit more modular. I thought about `init`, but I would rather not touch the global `gojsonschema` unless we're actually running the linter. Using `init` could make us run into fun global conflicts with other uses of `gojsonschema` within the gateway binary.

PR #1259, opened 2017-11-01
More linter type checks

mvdan:


lint/schema.go
```diff
@@ -34,7 +35,8 @@ const confSchema = `{
 				"type": "integer"
 			},
 			"type": {
-				"type": "string"
+				"type": "string",
+				"enum": ["", "redis"]
```
buger: Does adding `enum` automatically makes field "required"? Asking because curious why you allowing empty string here.

mvdan: The default config has this as empty, and we handle empty as default (redis), hence the possible value here.

PR #1261, opened 2017-11-02
CLI package changed to Kingpin

joshblakeley:
As docopt is deprecated it is wise to use a more well supported CLI package across Tyk.


joshblakeley:
Ignore this i've got some conflicts I need to sort out first.

PR #1262, opened 2017-11-02
Change CLI package to kingpin

joshblakeley:
Removes reliability on docopt which is currently unsupported.


main.go
```diff
@@ -753,13 +771,16 @@ func setupLogger() {
...
 
-func initialiseSystem(arguments map[string]interface{}) error {
+func initialiseSystem() error {
 
 	// Enable command mode
 	for _, opt := range commandModeOptions {
-		v := arguments[opt]
-		if v != nil && v != false {
-			handleCommandModeArgs(arguments)
+		if opt != false {
```
mvdan: You should deduplicate these two, like `if opt != false || opt != "" {`

main.go
```diff
@@ -73,6 +73,24 @@ var (
 
 	runningTests = false
 
+	help               = kingpin.CommandLine.HelpFlag.Short('h')
+	linter             = kingpin.Flag("lint", "enable linter check").Bool()
```
mvdan: Note that you want `tyk lint`, not `tyk --lint`.

If kingpin doesn't handle this, you should use something like `flag.Args()` (the kingpin version of that).

joshblakeley: Easily fixed I think

main.go
joshblakeley: I did have it like that but it fails go vet as a "suspect or"

joshblakeley: When I run go vet:
main.go:778: suspect or: opt != false || opt != ""
exit status 1

joshblakeley: But obviously it should work fine in this case.

mvdan: You're absolutely right - need to investigate further. This will likely be a go vet bug report :)

mvdan: On second thought, I'm retarded and vet is right.

`opt != false || opt != ""` is always `true`, because `opt == false && opt == ""` can never happen. Not sure what you wanted to do with this logic, but it's different from the old one.

joshblakeley: I've fixed it with some type assertion now

joshblakeley:
Arghhhh why has so much changed?

PR #1263, opened 2017-11-02
Changed cli package to Kingpin

joshblakeley:
docopt is no longer supported so changed to an active package


joshblakeley:
Still fixing up some small parts.

mvdan:
@joshblakeley would be useful for everyone if you asked for reviews once you're happy with the PR and CI is green - otherwise you have to ping us again once everything is good :)

lonelycode:
Question: Why Kingpin? We use cobra in our other cli tools.

mvdan:
They're different libraries at core - kingpin is just a version of `flag` with more features, while `cobra` is an entire framework to write command-line tools and is much, much heavier.

We only use cobra in the cli project that I know, while we use flag packages (mostly docopt) in every other project.

joshblakeley:
@mvdan how opposed are you to having lint as a flag. I've been getting some weird behaviour when trying to use it as an argument.

mvdan:
There must be a way for arguments to still work - otherwise kingpin would be useless. We can look into this tomorrow if you'd like. Making it a flag wouldn't be the end of the world, but I'd rather not accomodate our behavior to the limitations (if any) of a lib.

joshblakeley:
Glad to go through it together, it really should have been an easy modification but it adversely affects the behaviour of the command_mode.go flow

mvdan:
@joshblakeley is this ready? not sure what you did to get the coverage to decrease so much - we could look into it together later.

joshblakeley:
Yep if we can sort that out then it's ready to go.

buger:
@joshblakeley pls rebase and solve conflicts

mvdan:
Still looks OK, and no more conflicts or data races. Ready to test.

buger:
Looks good, will test it during 2.5 integration testing.

PR #1267, opened 2017-11-03
Fix connection leak: make HTTP transport API spec property

buger:
Right now Transport is created for each requests, which means that
connection pooling is not used and on each request it creates new
connection to the upstream server.

This differs from 2.3 behavior, which had global Transport for all
APIs. During the 2.4 refactor it was changed to the behavior described
above.

This change introduces transport caching, so it gets re-used between API
calls. In addition, this is better than 2.3 approach, because each API
can have different transport settings and timeouts, and now each API
gets own transport.

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -135,6 +135,7 @@ type APISpec struct {
 	LastGoodHostList         *apidef.HostList
 	HasRun                   bool
 	ServiceRefreshInProgress bool
+	HTTPTransport            http.RoundTripper
 }
 
 // APIDefinitionLoader will load an Api definition from a storage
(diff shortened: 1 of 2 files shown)


mvdan:
I understand this changes from one new transport per request, to one new transport per spec. This seems fine.

I don't understand how the pooling works and why it requires a common transport, though. Where have you read that? I can see no such mention on `net/http`.

buger:
@mvdan Transport object is responsible for managing all the connections and doing requests. If we talking about pool, you can control it using `MaxIdleConns` and `MaxIdleConnsPerHost` options, which we also do.

mvdan:
Ah yes, it does mention "connection caching" instead of "pooling". As long as this is tested to work fine, works for me :)

mvdan:
@buger are we waiting for QA, or can this be merged?

buger:
Well, we definitely need to QA performance for the GW, but probably for the whole master. So merging now.

Issue #1271, opened 2017-11-07
API loader is fragile when many endpoints are set up on a single domain


lonelycode:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When a hybrid node is reloading, there is a chance that two similar paths that only slightly differ in length will load in a different order.

**What is the expected behavior?**
The endpoints should always load in longest-first order

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. Create Two APIs with *slugs* of user and user-login
2. Differentiate them somehow (maybe a header inject)
3. Send requests to the gateway between hot reloads - the match order will swap

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
Hybrid, potentially others

mvdan:
If this was indeed on stable (since Hybrid runs 2.3.x), my recent sort order changes like 177758fe93fb8bc78ad8ba0632fdb4400beb722a cannot be a cause since they were never backported to stable.

Still looking at this, but at least we can rule that out.

mvdan:
I have also written a test to check that the sorting by path works with tricky listen paths in semi-random initial orders, and it passes fine on master.

ewah:
So quickly I've made a gorilla mux script. 

```
package main

import ("fmt"; "net/http"; "github.com/gorilla/mux")

func helloCatchAll(w http.ResponseWriter, r *http.Request) { fmt.Fprintf(w, "hello CatchAll\n") }
func helloRoot(w http.ResponseWriter, r *http.Request) { fmt.Fprintf(w, "hello Root\n") }
func helloHello(w http.ResponseWriter, r *http.Request) { fmt.Fprintf(w, "hello Hello\n") }
func helloWorld(w http.ResponseWriter, r *http.Request) { fmt.Fprintf(w, "hello World\n") }

func main() {
  r := mux.NewRouter()

  // as per api_loader.go:loadApps around line 555
  // concurrency says these can happen at any order
  s1 := r.Host("{fqdn:abc.xyz}").Subrouter() // for '/hello-world'
  s2 := r.Host("{fqdn:abc.xyz}").Subrouter() // for '/hello'
  s3 := r.Host("{fqdn:abc.xyz}").Subrouter() // for '/'

  // as per api_loader.go:loadApps around line 582
  // these proper happen in descending order
  s1.HandleFunc("/hello-world{rest:.*}", helloWorld)
[...]

PR #1275, opened 2017-11-08
test: add an api loader sort test

mvdan:
Wrote it while investigating #1271. The test passes, so it's unlikely
that the sort itself isn't broken, but this test is useful nevertheless.

It loads a bunch of API definitions with overlapping listen paths in a
semi-random initial order, and then uses them all making sure that the
right API spec responded to the request.

### DIFF
--- a/api_test.go
+++ b/api_test.go
@@ -708,3 +708,40 @@ func TestContextSession(t *testing.T) {
 	}()
 	ctxSetSession(r, nil)
 }
+
+func TestApiLoaderLongestPathFirst(t *testing.T) {
+	ids := []string{"abcde", "abc_x", "a", "abc", "abcd", "ab-xx", "ab"}
+	specs := func() (res []*APISpec) {
+		for _, id := range ids {
+			def := sampleAPI
+			def = strings.Replace(def, `"1"`, `"`+id+`"`, 1)
+			def = strings.Replace(def, `"/sample"`, `"/`+id+`"`, 1)
+			// put the path in the target URL too, to make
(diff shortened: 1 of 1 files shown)


buger:
I think it makes sense add all permutations to this test, using smth like this https://play.golang.org/p/Ulyo1H2Bii

Basically mix them up, in all possible orders. Also, I would suggest adding sub urls, like `/ab/a` `/ab/c` and etc.

buger:
I think it makes sense add all permutations to this test, using smth like this https://play.golang.org/p/Ulyo1H2Bii

Basically mix them up, in all possible orders. Also, I would suggest adding sub urls, like `/ab/a` `/ab/c` and etc.

mvdan:
Doing all possible permutations in every test run is a bad idea IMHO. It limits the number of APIs we can work with, and it means the test will take O(n!). It's simply too overkill and would waste time and CPU.

However, something I could do (which I planned to do, but ran out of time today) is generate a long list and shuffle it. That way, you could get similar results by doing `go test -run=ThisTest -count=1000`, the test itself would still be fast, and Travis would still catch probabilistic issues over time (much like it finds race bugs, even though you can run the test suite many times without running into them).

ewah:
re. my comment in #1271, would you be able to add custom domains into this test?

mvdan:
I should be able to, yes. Thanks for the heads up.

mvdan:
I mostly re-wrote the test and now I'm working on a fix too, so I'm going to open a new PR.

PR #1276, opened 2017-11-08
Added strip auth option and middleware

### LINKED ISSUE #1273: Disable forwarding auth related data to the upstream
Feature

Currently auth related header information is proxied to the upstream.

Auth related headers can be stripped from requests when proxied through the gateway as it presents a possible security issue.

All versions

lonelycode:
Fixes #1273

Adds a new option to the API Definition (this PR modifies the vendored version, actual change to tykcommon to follow) called "strip_auth_data" to the root of the object.

Setting this to true will cause the middleware to:

- Check if there are query string params to remove (if enabled) and remove them
- Check if there is an auth header to remove, and remove it
- If no header name is enforced (e.g. Basic Auth, HMAC or OIDC all use the "Authorization" header), strip that name instead

### DIFF
--- a/mw_strip_auth.go
+++ b/mw_strip_auth.go
@@ -36,7 +36,7 @@ func (a *MWStripAuthData) IsEnabledForSpec() bool {
 func (m *MWStripAuthData) StripAuth(r *http.Request, spec *APISpec) {
 	if spec.APIDefinition.Auth.UseParam {
 		// Remove the query string value
-		copy, err := url.Parse(r.URL.String())
+		cp, err := url.Parse(r.URL.String())
 		if err != nil {
 			log.WithFields(logrus.Fields{
 				"path":   r.URL.Path,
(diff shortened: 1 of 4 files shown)


mvdan:
Did you do this against release-2.3 on purpose? I thought we wouldn't do another 2.3.x release.

mw_strip_auth.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// GetConfig retrieves the configuration from the API config - we user mapstructure for this for simplicity
+func (m *MWStripAuthData) GetConfig() (interface{}, error) {
+	var thisModuleConfig MiddlewareContextVarsConfig
+	return thisModuleConfig, nil
+}
+
+func (a *MWStripAuthData) IsEnabledForSpec() bool {
+	if a.Spec.StripAuthData {
```
MangoMM: `return a.Spec.StripAuthData`

lonelycode:
Yes this is against 2.3 on purpose as it means we can just release it as a patch

buger:
I wonder if it is more convenient to put this code directly to auth key middleware.

mw_strip_auth.go
```diff
@@ -0,0 +1,88 @@
...
+	key := r.Header.Get(n)
+	if key != "" {
+		r.Header.Del(n)
+	}
+}
+
+// ProcessRequest will run any checks on the request on the way through the system, return an error to have the chain fail
+func (m *MWStripAuthData) ProcessRequest(w http.ResponseWriter, r *http.Request, configuration interface{}) (error, int) {
+
+	if !m.Spec.StripAuthData {
```
MangoMM: nit

```
if m.Spec.StripAuthData {
    m.StripAuth(r, mSpec)
}

return nil, 200
```

Also, does ProcessRequest function implement an interface and that's why we are also returning an error?

mvdan: I wouldn't worry about nits at all, as this patch will be completely different once ported to master :)

And yes, this is an interface.

buger:
Looks good. @ilijabojanovic ready for testing.

buger:
@MangoMM will to help with merging to master?

PR #1278, opened 2017-11-09
loader: set up exactly one sub-router per host

### LINKED ISSUE #1271: API loader is fragile when many endpoints are set up on a single domain

mvdan:
Plus a test, and another minor loader fix. See the commit messages for details.

Fixes #1271.

/cc @ewah

### DIFF
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -106,7 +106,6 @@ func createDefinitionFromString(defStr string) *APISpec {
 	loader := APIDefinitionLoader{}
 	def := loader.ParseDefinition(strings.NewReader(defStr))
 	spec := loader.MakeSpec(def)
-	spec.APIDefinition = def
 	return spec
 }
 
--- a/api_loader.go
+++ b/api_loader.go
@@ -552,7 +552,7 @@ func loadApps(specs []*APISpec, muxer *mux.Router) {
 					"api_name": spec.Name,
 					"domain":   spec.Domain,
 				}).Info("Custom Domain set.")
-				subrouter = mainRouter.Host(spec.Domain).Subrouter()
+				subrouter = subrouter.Host(spec.Domain).Subrouter()
 			}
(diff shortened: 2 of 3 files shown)


ewah:
This solves our issues.

Just one note because Host can be domain regex-es does it need to resort based on Domain length to normalize it even more... 

ie. something like
Host("{fqdn:abc.(?:(dev.)?).com}").Subrouter()
Host("{fqdn:abc.(?:.*.?)com}").Subrouter()

The first should always load before the second.

mvdan:
You're right that this fix does not cover regexes. However, as far as I can tell, we don't officially support regexes there - our docs don't mention regexes anywhere.

I don't know how we could possibly support all regexes here. The router doesn't backtrack - as soon as one `Host()` subrouter matches, it doesn't try the other host sub-routers. We could document that we support regexes with the current approach, as long as they never overlap - if they do, stuff will break as before.

ewah:
@mvdan I'd be happy just doing a sort of length on Domain (regex or not)... it might not be 100% accurate -- but it is 100% consistent.

mvdan:
Ah, I see your point about consistency. Yes, at the moment the host subrouter creation order is not deterministic.

Doing it just by length won't do, as different strings with the same length would still be sorted arbitrarily. I'm thinking of doing first length then alphabetical, or just alphabetical.

The only reason to do first length and then alphabetical is that it may help more often than not with regexes.

mvdan:
Done - ended up using both length and chars, in decreasing order.

@buger PTAL

buger:
Nice! Pls create same PR for release-2.3 branch.

PR #1280, opened 2017-11-10
coprocess: fire AuthFailed event when using custom authentication middleware

### LINKED ISSUE #1279: Fire AuthFailed event when using custom auth plugins
Bug

It's not possible to use a custom auth plugin and attach an event listener to the "AuthFailed" event.

It should be possible to trigger this event when using a custom auth plugin.
This scenario doesn't make sense if you're attaching an event handler that's implemented in a plugin, as the plugin already knows about this authentication event. However the user may want to attach `eh_web_hook_handler` or similar (built-in) while using a plugin for authentication.

- Create a plugin that implements an authentication mechanism and also an event handler:
```python
from tyk.decorators import *

@Hook
def MyAuthMiddleware(request, session, metadata, spec):
    print("MyAuthMiddleware is called")

@Event
def MyEventHandler(event, spec):
    print("MyEventHandler is called")
    print(event, spec)
```
- Generate a bundle and configure the API definition to load it and also set an event handler:
```json
{
    "enable_coprocess_auth": true,
    "use_keyless": false,
    "enable_coprocess_auth": true,
    "custom_middleware_bundle": "bundle.zip",
    "event_handlers": {
      "events": {
        "AuthFailure": [
          {
            "handler_name": "cp_dynamic_handler",
            "handler_meta": {
              "name": "MyEventHandler"
            }
          }
        ]
      }
    }
}
```
- When you send a request and the authentication fails, `MyEventHandler` won't be triggered.
- If you change `enable_coprocess_auth` to `false`, `MyEventHandler` will be triggered because the built-in authentication mechanism will be used and `AuthKey` is correctly setup on this one.

Master from what I've tested

matiasinsaurralde:
Fixes #1279.

When the custom authentication middleware doesn't return a session object we already assume it's a bad auth (we return 403 by default), so this enhances the flow, allowing the user to setup event handlers on top of this.

It also assumes that the authentication header name is what the middleware is using to perform the authentication process.

### DIFF
--- a/coprocess.go
+++ b/coprocess.go
@@ -285,6 +285,8 @@ func (m *CoProcessMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Requ
 	if m.Spec.EnableCoProcessAuth && m.HookType == coprocess.HookType_CustomKeyCheck {
 		// The CP middleware didn't setup a session:
 		if returnObject.Session == nil {
+			authHeaderValue := r.Header.Get(m.Spec.Auth.AuthHeaderName)
+			AuthFailed(m, r, authHeaderValue)
 			return errors.New("Key not authorised"), 403
 		}
 


buger:
@ilijabojanovic ready to be test

mvdan:
The issue was still in the backlog, so I've moved it to ready to test.

PR #1288, opened 2017-11-13
port: strip_auth middleware to tip

asoorm:
refs #1273 

TODO:

- [x] Write Tests
- [x] Establish how the mw should react on error `stripFromParams` (Ignore errors)
- [x] confirm where `mwAppendEnabled` should be placed (Same as Martin's example)

### DIFF
--- a/mw_strip_auth.go
+++ b/mw_strip_auth.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"net/http"
+	"net/url"
+
+	"github.com/Sirupsen/logrus"
+)
+
+type StripAuth struct {
+	BaseMiddleware
+}
+
+func (sa *StripAuth) Name() string {
+	return "StripAuth"
+}
+
+func (sa *StripAuth) EnabledForSpec() bool {
+	return sa.Spec.StripAuthData
+}
+
+func (sa *StripAuth) ProcessRequest(w http.ResponseWriter, r *http.Request, _ interface{}) (error, int) {
+
+	config := sa.Spec.Auth
+
+	log.WithFields(logrus.Fields{
+		"prefix": sa.Name(),
+	}).Debugf("sa.Spec.Auth: %+v\n", config)
+
+	if sa.Spec.Auth.UseParam {
+		sa.stripFromParams(r)
+	} else {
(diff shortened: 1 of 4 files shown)


mw_strip_auth.go
```diff
@@ -0,0 +1,85 @@
...
+	return nil, 200
+}
+
+func (sa *StripAuth) stripFromParams(r *http.Request) error {
+
+	config := sa.Spec.Auth
+
+	reqUrlPtr, err := url.Parse(r.URL.String())
+	if err != nil {
+		return err
```
buger: I think it should be totally safe ignore both of this errors in this function. In the first case, you are parsing current URL, which was already validated. In the second case, almost the same URL object, just changed query string a bit.

mw_strip_auth_test.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func TestStripAuth_Name(t *testing.T) {
+
+	sa := StripAuth{}
+
+	if sa.Name() != "StripAuth" {
+		t.Error("Incorrect Name")
+	}
+}
```
mvdan: This is a bit unnecessary.

mw_strip_auth_test.go
```diff
@@ -0,0 +1,131 @@
...
+	if sa.EnabledForSpec() != sa.Spec.StripAuthData {
+		t.Error("EnablesForSpec be true when sa.Spec.StripAuthData is true")
+	}
+
+	sa.Spec.StripAuthData = false
+
+	if sa.EnabledForSpec() != sa.Spec.StripAuthData {
+		t.Error("EnablesForSpec should false when sa.Spec.StripAuthData is false")
+	}
+}
```
mvdan: This one too - you're writing a dozen lines of test for a single line of trivial code.

If you want to be sure that the middleware works, don't unit test every single part of it. Unit test the complex bits like you do below, and then do a full end-to-end test. Should be easy with httptest and loading a custom spec.

buger:
Looks good! Lets test together with UI part.

mvdan:
@asoorm please rebase - we need to merge this in for 2.4

asoorm:
@mvdan Done

PR #1295, opened 2017-11-16
main: fix hot restart regression in 2.4

### LINKED ISSUE #1268: Hot restart via SIGUSR2 broken in 2.4 (regression from 2.3)
Golang 1.9.2
tyk master branch

I send kill -SIGUSR2 xxx to tyk, but tyk don‘t hot restart.’

mvdan:
In 26043b48, some changes were made to allow specifying a different
control API port.

However, this broke the SIGUSR2 hot restart in two ways. First, it added
a second goAgainErr, overwriting the one that we need to use to detect
that the fork happened. This meant that, when the child was forked, the
parent would never exit properly.

Second, it made it impossible for generateListener to use a port 0,
meaning to use a random port. This would result in both parent and child
trying to use the same port before the parent is killed, meaning that
the child would not be able to start up properly.

After these fixes, a hot restart with the default tyk.conf works again.
Writing a Go test for hot restarts is impossible though, so we will
write them separately as python end-to-end tests, covering all
possibilities of listen options combined with SIGUSR2.

Note that we will need further fixes to the goagain library once we
switch to Go 1.9, as it uses reflection and the struct it peeks at
changed in 1.9.

Fixes #1268.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1025,8 +1025,8 @@ func main() {
 		}
 	}
 
-	l, _ := goagain.Listener(onFork)
-	controlListener, goAgainErr := goagain.Listener(onFork)
+	l, goAgainErr := goagain.Listener(onFork)
+	controlListener, _ := goagain.Listener(onFork)
 
 	if err := initialiseSystem(arguments); err != nil {
 		log.WithFields(logrus.Fields{
@@ -1164,9 +1164,6 @@ func start(arguments map[string]interface{}) {
 
 func generateListener(listenPort int) (net.Listener, error) {
 	listenAddress := config.Global.ListenAddress
-	if listenPort == 0 {
-		listenPort = config.Global.ListenPort
-	}
 
 	targetPort := fmt.Sprintf("%s:%d", listenAddress, listenPort)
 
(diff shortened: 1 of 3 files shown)


mvdan:
@jacyo please try this branch if you have a few minutes, to confirm that it fixes the bug for you. Remember that you'll need Go 1.8 for it to work (or 1.9 or later, as long as you fix the reflection line like you did before).

buger:
I pushed my changes. The previous fix not really worked because it always used the random port, and instead it should be using `ListenPort`. I think the real reason behind this issue is that goagain by design do not support multiple listeners, and when we called it twice, it just broke. 

In addition, I applied goagain fix suggested by @jacyo and it should work for both Go 1.9 and Go 1.8 https://github.com/TykTechnologies/goagain/commit/e23361255278b7e2dcdfe6d1407191725340fa00

Note that I could not make control API work if it set on a separate port, after fork, it just do not start litener. I do not fully understand the issue, but it is not the regression since it not worked on 1.3 branch too: control API on separate port is a new feature. 

We need move to something like https://github.com/facebookgo/grace in future.

PR #1298, opened 2017-11-17
Fixed panic when loading spec with Domain on env without custom domain support

buger:
If Domain is non empty and enable_custom_domains is false, API router
becomes nil, and panics.

This was part of recent change to fix order of API loading. Restored
original behavior when default router picked as fallback.

### DIFF
--- a/api_loader.go
+++ b/api_loader.go
@@ -585,6 +585,15 @@ func loadApps(specs []*APISpec, muxer *mux.Router) {
 	for i, spec := range specs {
 		go func(spec *APISpec, i int) {
 			subrouter := hostRouters[spec.Domain]
+			if subrouter == nil {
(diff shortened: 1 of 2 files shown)


api_loader.go
```diff
@@ -585,6 +585,15 @@ func loadApps(specs []*APISpec, muxer *mux.Router) {
 	for i, spec := range specs {
 		go func(spec *APISpec, i int) {
 			subrouter := hostRouters[spec.Domain]
+			if subrouter == nil && spec.Domain != "" {
```
mvdan: I would make this be just `if subrouter == nil {` - otherwise, we could be missing yet another `subrouter == nil` case.

And if we did forget about a separate case, I'd rather have an incorrect warning logged than a crash.

Issue #1203, opened 2017-10-06
Tyk log entry exposes pre-hashed session key value in logs


bitsofinfo:
**Do you want to request a *feature* or report a *bug*?**

feature/bug

**What is the current behavior?**

Certain error conditions yields a log statement like this that exposes the clear (pre hashed) session key to the log file. In our case the session key (part of it) contains sensitive data that should not flow into a log file

```
2017-10-06T21:07:07.724962728Z time="Oct  6 21:07:07" level=info msg="Attempted access to unauthorised API." api_found=false key="orgid-keypart1-keypart2-/my/path-SENSITIVE-PART=" origin=192.168.1.111 path="/my/path"
```

**What is the expected behavior?**

This should be a configurable option to permit omission or obfuscation of the data like this flowing into the log file

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**

latest 1.3.9/2.3.9

PR #1301, opened 2017-11-17
new config var enable_key_logging added, request logging unified

dencoded:
added changes to fix https://github.com/TykTechnologies/tyk/issues/1203

1. added helper function to log request info and have more grain control about what to put in log for request
2. added new config parameter `enable_key_logging` with default value false which will switch on/off logging for tokens/keys

I tried to to catch all cases where we log info about http-request and changed them to use new helper function `getLogEntryForRequest`. It prepares logrus.Entry with passed data but leaves freedom to decide about message log-level to caller. If we go with this approach the assumption is to keep eye on how we log things and use this function to make sure sensitive info won't be logged.

### DIFF
--- a/log_helpers.go
+++ b/log_helpers.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"net/http"
+
+	"github.com/Sirupsen/logrus"
+
+	"github.com/TykTechnologies/tyk/config"
+)
+
+func getLogEntryForRequest(r *http.Request, key string, data map[string]interface{}) *logrus.Entry {
+	// populate http request fields
+	fields := logrus.Fields{
+		"path":   r.URL.Path,
+		"origin": requestIP(r),
+	}
+	// add key to log if configured to do so
+	if key != "" && config.Global.EnableKeyLogging {
+		fields["key"] = key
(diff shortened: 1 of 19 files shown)


buger:
See the error in circleCI it fails on `goimports` check.

log_helpers.go
```diff
@@ -0,0 +1,28 @@
...
+
+func getLogEntryForRequest(r *http.Request, key string, data map[string]interface{}) *logrus.Entry {
+	// populate http request fields
+	fields := logrus.Fields{
+		"path":   r.URL.Path,
+		"origin": requestIP(r),
+	}
+	// add key to log if configured to do so
+	if key != "" && config.Global.EnableKeyLogging {
+		fields["key"] = key
```
buger: Looks good! But I think it makes sense make it obvious to the user that tokes are hidden. Smth like this:
```
if key != "" {
    fields["key"] = key

    if !config.Global.EnableKeyLogging {
      key = "<hidden>"      
    }
}
```

mvdan:
bump @buger

buger:
👌

Issue #1297, opened 2017-11-16
If beginning of a random path of api call matches an existing slug we return 403 (instead of 404.)


letzya:
**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
`http://my-instance.com:8080/api1 works`, but
`http://my-tyk-instance.com:8080/api1ss` (i.e. adding some letters at the end of the slug) generates  **403 forbidden** while putting some string right after the hostname `http://my-tyk-instance.com:8080/ss` just gives `404 page not found`
**What is the expected behavior?**
http://my-tyk-instance.com:8080/api1ss ==> should return 404

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. create api
2. make a call to it, check it's fine
3. make a call with some extra chars at the end.
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.3

mvdan:
I cannot reproduce this. Before the API is created, both without and with extra chars are 404. After the API is created, both are 200.

I even added tests for this, which pass:

```
diff --git a/gateway_test.go b/gateway_test.go
index 752269da..7fe4162d 100644
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -919,8 +919,10 @@ func TestListener(t *testing.T) {
 		{method: "POST", path: "/tyk/apis", body: sampleAPI, adminAuth: true, code: 200},
 		// API definitions not reloaded yet
 		{method: "GET", path: "/sample", code: 404},
+		{method: "GET", path: "/samplefoo", code: 404},
 		{method: "GET", path: "/tyk/reload/?block=true", adminAuth: true, code: 200},
 		{method: "GET", path: "/sample", code: 200},
+		{method: "GET", path: "/samplefoo", code: 200},
 	}
 
 	// have all needed reload ticks ready
```

Please provide an API definition and the `curl` commands to reproduce it. Does it reproduce when the API is keyless?

letzya:
Happens only with this api, will check again.

asoorm:
I cannot reproduce:

Correct
```
$ curl -X GET https://ahmet-1.cloud.tyk.io/api1/ -I
HTTP/1.1 401 Unauthorized
Content-Type: application/json
Date: Mon, 20 Nov 2017 16:13:43 GMT
Content-Length: 46
Connection: keep-alive
```

Not found for all other combinations:
```
[...]

PR #1306, opened 2017-11-21
test: include listen path suffix tests

mvdan:
To be extra sure about issues like #1297.

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -921,6 +921,9 @@ func TestListener(t *testing.T) {
 		{method: "GET", path: "/sample", code: 404},
 		{method: "GET", path: "/tyk/reload/?block=true", adminAuth: true, code: 200},
 		{method: "GET", path: "/sample", code: 200},
+		{method: "GET", path: "/samplefoo", code: 200},
+		{method: "GET", path: "/sample/", code: 200},
+		{method: "GET", path: "/sample/foo", code: 200},
 	}
 
 	// have all needed reload ticks ready


gateway_test.go
```diff
@@ -921,6 +921,9 @@ func TestListener(t *testing.T) {
 		{method: "GET", path: "/sample", code: 404},
 		{method: "GET", path: "/tyk/reload/?block=true", adminAuth: true, code: 200},
 		{method: "GET", path: "/sample", code: 200},
+		{method: "GET", path: "/samplefoo", code: 200},
+		{method: "GET", path: "/sample/", code: 200},
+		{method: "GET", path: "/sample/foo", code: 200},
```
dencoded: somewhat to mention, when it comes to standard string-values I would prefer to use already defined constants vs typing them (source of errors), like for http-verbs defined in `net/http` we have `MethodGet `

mvdan: What constant are you talking about? `"GET"`, `"/sample"`, or `200`?

All three of them are short, so I'm not sure that named constants are necessary. `net/http` itself is full of `"GET"` literals, for example. The constants there are defined for consistency more than for usefulness.

mvdan: Also worth noting that a typo in any of these literals would result in a test failure or panic, so I'm not too worried about the mistakes we could make either.

dencoded: no worries, the test code is totally fine. I meant that for http-verbs like `GET` (and yeah - response codes like `200`) we have predefined constants in `net/http` which we could use - `http.MethodGet` and `http.StatusOK`

Issue #1299, opened 2017-11-17
Tests take way longer and are noisy after the mutual TLS merge


mvdan:
Before:

```
go test ./...
ok      github.com/TykTechnologies/tyk  2.691s
?       github.com/TykTechnologies/tyk/apidef   [no test files]
ok      github.com/TykTechnologies/tyk/apidef/importer  0.043s
ok      github.com/TykTechnologies/tyk/config   0.038s
?       github.com/TykTechnologies/tyk/coprocess        [no test files]
ok      github.com/TykTechnologies/tyk/lint     0.039s
?       github.com/TykTechnologies/tyk/log      [no test files]
?       github.com/TykTechnologies/tyk/storage  [no test files]
?       github.com/TykTechnologies/tyk/user     [no test files]
```

After:

```
$ go test ./...
ok      github.com/TykTechnologies/tyk  7.013s
?       github.com/TykTechnologies/tyk/apidef   [no test files]
ok      github.com/TykTechnologies/tyk/apidef/importer  0.006s
ok      github.com/TykTechnologies/tyk/certs    2.300s
ok      github.com/TykTechnologies/tyk/config   0.014s
?       github.com/TykTechnologies/tyk/coprocess        [no test files]
ok      github.com/TykTechnologies/tyk/lint     0.043s
?       github.com/TykTechnologies/tyk/log      [no test files]
?       github.com/TykTechnologies/tyk/storage  [no test files]
?       github.com/TykTechnologies/tyk/user     [no test files]
```

Also, when testing just the main package, it now shows garbage output:

```
$ go test
2017/11/17 15:38:17 http: TLS handshake error from [::1]:45108: tls: no certificates configured
2017/11/17 15:38:17 http: TLS handshake error from [::1]:59068: tls: client didn't provide a certificate
2017/11/17 15:38:17 http: TLS handshake error from [::1]:59070: tls: failed to verify client's certificate: x509: certificate signed by unknown authority
2017/11/17 15:38:18 http: TLS handshake error from 127.0.0.1:46904: tls: client didn't provide a certificate
2017/11/17 15:38:18 http: TLS handshake error from 127.0.0.1:46914: tls: failed to verify client's certificate: x509: certificate signed by unknown authority
2017/11/17 15:38:19 http: TLS handshake error from 127.0.0.1:46918: tls: failed to verify client's certificate: x509: certificate signed by unknown authority (possibly because of "x5
[...]

PR #1307, opened 2017-11-21
test: discard all std log output too by default

mvdan:
Before:

	$ go test
	2017/11/21 15:58:57 http: TLS handshake error from [::1]:54058: tls: no certificates configured
	2017/11/21 15:58:58 http: TLS handshake error from [::1]:58572: tls: client didn't provide a certificate
	2017/11/21 15:58:58 http: TLS handshake error from [::1]:58574: tls: failed to verify client's certificate: x509: certificate signed by unknown authority
	2017/11/21 15:58:59 http: TLS handshake error from 127.0.0.1:46488: tls: client didn't provide a certificate
	2017/11/21 15:58:59 http: TLS handshake error from 127.0.0.1:46498: tls: failed to verify client's certificate: x509: certificate signed by unknown authority
	2017/11/21 15:59:00 http: TLS handshake error from 127.0.0.1:46502: tls: failed to verify client's certificate: x509: certificate signed by unknown authority (possibly because of "x5
	09: invalid signature: parent certificate cannot sign this kind of certificate" while trying to verify candidate authority certificate "serial:206981011806556900346702653301449444008
	")
	2017/11/21 15:59:00 http: TLS handshake error from 127.0.0.1:50170: tls: failed to verify client's certificate: x509: certificate signed by unknown authority
	PASS
	ok      github.com/TykTechnologies/tyk  6.257s

With this change, it's back to being quiet.

Updates #1299.

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"html/template"
 	"io/ioutil"
+	stdlog "log"
 	"log/syslog"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


buger:
I wonder if test logging like t.Log still works. Sometimes it is useful to add some debugging calls, but do not enable logging.

mvdan:
Yes - it still works, as long as `go test -v` is used (just like before). This is because the testing package does not use the log package.

PR #1304, opened 2017-11-21
added extra condition to allow reload with 1 api

dencoded:
fixes https://github.com/TykTechnologies/tyk/issues/1035

PR #1313, opened 2017-11-24
all: fix numerous global apiSpecs data races

mvdan:
Also make gateway_test.go help the race detector by writing to it in the
background. This made the races show up on every 'go test -race' run on
my laptop, as opposed to just sometimes.

While at it, don't use apiSpecs as a name for local variables, as that
makes the code very confusing.

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -281,17 +281,17 @@ func (a APIDefinitionLoader) FromDashboardService(endpoint, secret string) []*AP
 	}
 
 	// Process
-	var apiSpecs []*APISpec
+	var specs []*APISpec
 	for _, def := range apiDefs {
 		spec := a.MakeSpec(def)
-		apiSpecs = append(apiSpecs, spec)
+		specs = append(specs, spec)
 	}
 
 	// Set the nonce
 	ServiceNonce = list.Nonce
 	log.Debug("Loading APIS Finished: Nonce Set: ", ServiceNonce)
 
-	return apiSpecs
+	return specs
 }
 
 // FromCloud will connect and download ApiDefintions from a Mongo DB instance.
@@ -325,7 +325,7 @@ func (a APIDefinitionLoader) processRPCDefinitions(apiCollection string) []*APIS
 		return nil
 	}
 
-	var apiSpecs []*APISpec
+	var specs []*APISpec
 	for _, def := range apiDefs {
 		def.DecodeFromDB()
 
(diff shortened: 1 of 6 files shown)


cert.go
```diff
@@ -102,6 +102,7 @@ func getTLSConfigForClient(baseConfig *tls.Config, listenPort int) func(hello *t
 			return newConfig, nil
 		}
 
+		apisMu.RLock()
```
buger: Note that it will be called for each incoming connection, and I would rather avoid it. 

Can we lock only writers here, and keep readers not touched?

buger:
@dencoded you spent quite lot of time on it, what do you think?

cert.go
mvdan: Well, this is what happens if you use a global. All the other api operations obtain a read lock. If we only use the lock for writes, it's absolutely useless.

dencoded:
@buger according `getTLSConfigForClient` and access to global `apiSpecs` - these changes look similar to ones I pushed to my branch yesterday https://github.com/TykTechnologies/tyk/pull/1304/commits/1a924dee652cf4e8f88662ccd2edf57957fb4bee (which is still failing). So changes in this brunch look Ok to me.

Speaking in general - I don't like idea about sharing API spec via global, especially, like you said, we we access them with in every incoming request

cert.go
dencoded: @buger @mvdan 
I wander if we can benefit from using `sync.atomic`'s `Value` and `copy-on-write` approach showed here: https://golang.org/pkg/sync/atomic/#example_Value_readMostly ?

Also, kind of related to that comment - I was considering that approach https://golang.org/pkg/sync/atomic/#example_Value_config  for re-loading api definitions and use them to serve requests

api_definition.go
```diff
@@ -281,17 +281,17 @@ func (a APIDefinitionLoader) FromDashboardService(endpoint, secret string) []*AP
 	}
 
 	// Process
-	var apiSpecs []*APISpec
+	var specs []*APISpec
```
dencoded: 👍 I think we these change we eliminate a big confusion as it was not clear that we are "hiding" global apiSpecs by declaring local var

mvdan:
I agree that globals and locks are bad, but this isn't a refactor - it's a race fix. It's the smallest and least intrusive one, because we need this fixed fast and backported to 2.4.

buger:
Got it.

I'm happy to get rid of globals. Let's try to come up with the plan.

cert.go
buger: @dencoded make sense. But in fact, we modify apiSpec records only inside `apiLoader`, so we modify them all at once, and can replace in Value.

PR #1316, opened 2017-11-27
upstream cache control configurable ttl header

### LINKED ISSUE #1194: Extend gateway upstream cache control to allow specifying response header
As a service owner, I should be able to control the caching behaviour of Tyk based on headers that my service sends:

- I should be able to set custom header values instead of the existing built-in ones

Extend the existing `EnableUpstreamCacheControl` option to add a configuration object that lets the header to listen for the TTL value that is currently used. This defaults to our existing custom header if unset (retaining backwards compatibility), and allows the user to specify `Expires` as the header to use in order to offer an MvP of the functionality

asoorm:
resolves #1194

Extends the existing EnableUpstreamCacheControl option to add a
configuration object that lets the header to listen for the TTL value
that is currently used. This defaults to our existing custom header if
unset (retaining backwards compatibility), and allows the user to
specify Expires as the header to use in order to offer an MvP of the
functionality

### DIFF
--- a/mw_redis_cache.go
+++ b/mw_redis_cache.go
@@ -191,8 +191,13 @@ func (m *RedisCacheMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Req
 				log.Warning("Upstream cache action not found, not caching")
 				cacheThisRequest = false
 			}
-			// Do we override TTL?
-			ttl := reqVal.Header.Get(upstreamCacheTTLHeader)
+
+			cacheTTLHeader := upstreamCacheTTLHeader
+			if m.Spec.CacheOptions.CacheControlTTLHeader != "" {
+				cacheTTLHeader = m.Spec.CacheOptions.CacheControlTTLHeader
+			}
+
+			ttl := reqVal.Header.Get(cacheTTLHeader)
 			if ttl != "" {
 				log.Debug("TTL Set upstream")
 				cacheAsInt, err := strconv.Atoi(ttl)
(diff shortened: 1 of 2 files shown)


asoorm:
@buger @mvdan I dont like the name `upstream_cache_control_ttl_header`. What do you think if I rename to `cache_ttl_header_name` or `expires_header_name_override`?

mvdan:
I think that `upstream_`, `_name` and `_override` are all redundant. I'd go for `cache_control_ttl_header`, as it's immediately obvious that it's related to the setting just above it.

asoorm:
@mvdan - I've changed the header name to `cache_control_ttl_header`. Ready for review & merge again.

mw_redis_cache.go
```diff
@@ -17,8 +17,7 @@ import (
 )
 
 const (
-	upstreamCacheHeader    = "x-tyk-cache-action-set"
-	upstreamCacheTTLHeader = "x-tyk-cache-action-set-ttl"
+	upstreamCacheHeader = "x-tyk-cache-action-set"
```
mvdan: Nit, but I'd keep the constants as they were - next to each other. They're very closely related, so it's useful.

asoorm: but is no longer a constant. We set the default to `"x-tyk-cache-action-set"` and change it if header name has been changed.

mvdan: Sure, but you could still have it there under `upstreamCacheTTLHeaderDefault` or something similar.

mvdan:
I'll leave it to @buger to decide whether or not to have QA in this.

PR #1320, opened 2017-11-28
Fixes RPC session being zapped due to struct member rename

excieve:
This is a side effect of 2a9d7d5b6c289ac75d, which renamed `InboundData.SessionState` to `InboundData.Session`. The receiving part didn't recognize a new member and initialised the old `SessionState` to an empty string, effectively killing the org session on the first call to `SetKey`.

### DIFF
--- a/rpc_storage_handler.go
+++ b/rpc_storage_handler.go
@@ -21,12 +21,12 @@ import (
 )
 
 type InboundData struct {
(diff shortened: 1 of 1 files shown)


excieve:
Hmm, tests timeout on 1.9 doesn't seem to be related to this.

PR #1260, opened 2017-11-01
Bundle loader enhancements (WIP)

matiasinsaurralde:
This includes more enhancements like #1229, #1237.
I've modified the dispatcher to load the modules with a namespace, this solves most of the issues with PYTHONPATH and ambiguous function names across different middlewares.

PR #1322, opened 2017-11-29
coprocess_python: fix backport of #1237

matiasinsaurralde:
This enhances the usage of absolute Python module paths.
It also drops the call to "load_middlewares" that occurs during the dispatcher initialization because "load_bundle" is used instead, #1260 implements more enhancements related to this.

### DIFF
--- a/coprocess/python/dispatcher.py
+++ b/coprocess/python/dispatcher.py
@@ -26,7 +26,6 @@ class TykDispatcher:
 
         self.middlewares = []
         self.hook_table = {}
-        self.load_middlewares()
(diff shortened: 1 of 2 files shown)


matiasinsaurralde:
The issue happened because the backport didn't use the full module path (which the previous fix expects to use). I've introduced `module_path` which will make it work again, however colliding function names and related issues aren't handled by this issue and is the reason why I started to work on #1260.
We should look forward to have #1260 fully compatible with all the bundle scenarios and introduce it whenever it's possible.

Issue #607, opened 2017-03-09
Block the use of invalid policies in the Gateway


kohrVid:
**Do you want to request a *feature* or report a *bug*?**
Enhancement/feature

**What is the current behaviour?**
It is currently possible for users to access APIs with invalid policies. The use of invalid policies with JWT can lead to the creation of master keys.

**What is the expected behaviour?**
We would like to add a configuration option that will allow API owners to block requests sent with invalid policies.

**If the current behaviour is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

1. An API is created using the REST API and configured to use a JWT.
2. Valid policies are created using our [documentation](https://tyk.io/docs/tyk-v2-2-documentation-components/tutorial-create-a-policy-ce/)
3. A request is made to the API with the correct JWT and the non-existent policy.

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
All

csyllac:
Hi - to add more detail to this, it is a bug - but when it was described on the forum there was some confusion about the actual steps to cause it.

The actual issue:
Changing policies in the JWT is not properly handled by tyk if the SUB (identity) has already established a valid session.

To reproduce:
- create a policy file with two policies, 'a' and 'b'
- create JWT with sub '1' and policy 'a'
- make valid request using that JWT - which will pass
- create a second JWT with sub '1' and policy 'b'
- make a request, and tyk will still enforce policy 'a', not policy 'b''
- create a third JWT with sub '1' and policy 'c' (which does not exist in the policy file)
- make a request and tyk will still enforce policy 'a' even though 'c' does not exist at all

According to the documentation, " the policy will be applied the first time the token is seen and then modified if the policy changes in the claim".

What actually happens is that tyk keeps applying the policy of the first successful request instead of detecting that the policy has changed and applying the new policy (or failing if the new policy does not exist).

If the first request is made with a JWT that contains a non-existent policy tyk returns 403 correctly.
This was tested with tyk 2.3.3 CE.

lonelycode:
@csyllac Thanks for the STR, it will help us get this issue solved.

buger:
To fix it we need to check on each request if policy id of current active session is the same as policy in JWT

PR #1325, opened 2017-11-29
one more check for policy ID in mw_jwt added

dencoded:
added fix for a bug https://github.com/TykTechnologies/tyk/issues/607 and one more test for this situation


buger:
Nice! 👍

PR #1326, opened 2017-11-29
log: get rid of unnecessary checks

mvdan:
Ranging over a nil and/or empty map is already a no-op.


dencoded:
@mvdan sorry for delay. LGTM postfactum, code reads clearer for sure.

PR #1332, opened 2017-11-30
cache: EnabledForSpec should just use EnableCache

### LINKED ISSUE #1329: Caching: Cache all safe request does not work
BUG
Cache all safe requests" does not work for protected api's
When we enable "Cache all safe requests", we should cache requests and see them in redis
1. Create protected api definition 
2. Generate key for protected api definition
3. Trigger get/post requests to gateway 
Release 2.4

mvdan:
Don't require there to be a cached extended path too. Otherwise, one
can't use any sort of caching on an API without adding cached extended
paths.

At the same time, don't run the middleware if there are any cached
extended paths yet EnableCache is false. The docs tell the user to
enable the cache for the whole API, so that is expected.

Add a test to prove that caching of a simple API works. This test was
failing before, as the API does not have any cached extended paths.

Fixes #1329.

### DIFF
--- a/mw_redis_cache.go
+++ b/mw_redis_cache.go
@@ -37,15 +37,7 @@ func (m *RedisCacheMiddleware) Init() {
 }
 
 func (m *RedisCacheMiddleware) EnabledForSpec() bool {
-	if !m.Spec.CacheOptions.EnableCache {
-		return false
-	}
-	for _, version := range m.Spec.VersionData.Versions {
-		if len(version.ExtendedPaths.Cached) > 0 {
-			return true
-		}
-	}
-	return false
+	return m.Spec.CacheOptions.EnableCache
 }
 
 func (m *RedisCacheMiddleware) CreateCheckSum(req *http.Request, keyName string) string {
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -15,6 +15,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


mw_redis_cache.go
```diff
@@ -37,15 +37,7 @@ func (m *RedisCacheMiddleware) Init() {
 }
 
 func (m *RedisCacheMiddleware) EnabledForSpec() bool {
-	if !m.Spec.CacheOptions.EnableCache {
-		return false
-	}
-	for _, version := range m.Spec.VersionData.Versions {
```
buger: I think it was sort of optimization here, to not run this middleware all the time. 
And it could be fixed without removing this `for` loop and adding an additional check like:
```
if m.Spec.CacheOptions.CacheAllSafeRequests {
   return true
}
```

mvdan: Sure - but with this PR, the middleware does *not* run all the time. It only runs if caching is enabled.

And when it's enabled but only for some cases, the cases that end up not having the caching run the middleware for very little time. They are discarded very early on in ProcessRequest.

I think it is best if we just keep it simple, as we have clearly fucked it up once already. And the even more complex logic would still have edge cases. For example, what if caching is enabled, but `CacheAllSafeRequests` is false and there are no cached extended paths? I'm not that sure that the middleware should be removed entirely in such a case.

Or, to put it another way, what would `EnableCache` even be for, if we didn't rely on it here?

buger: Yeah, I'm ok with the current solution too, it should work well. Just needs to be confirmed.

mvdan: To clarify - I also started tweaking the current heuristic, instead of just reducing it to the one bool field. But after thinking about it a little bit, I realised it was silly and overly complex.

Issue #1035, opened 2017-08-25
Can't deactivate API if it is the only one


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
If there is only 1 active API, and the user tries to deactivate it, the gateway does not apply this changes, and not disable the API. Here are the logs:

```
Aug 25 03:58:46 ip-xxx tyk[9302]: time="Aug 25 03:58:46" level=info msg="Detected 0 APIs"
Aug 25 03:58:46 ip-xxx tyk[9302]: time="Aug 25 03:58:46" level=warning msg="No API Definitions found, not reloading"
```
**What is the expected behavior?**
It should disable API even if it the only one

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
On the empty environment, add new active API, ensure that it proxied. De-activate it in the dashboard. API still proxied.

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.3.8

mvdan:
This is a dumb question, but I assume this doesn't happen if you have 2 but deactivate 1? What if you have 2 and deactivate both?

lonelycode:
If the list from the dashboard returns 0 (it is filtered on the activation flag) then Tyk takes it as a failed get and doesn't unload. So you need to deactivate one at a time.

lonelycode:
If this is closed why is it in the backlog for 2.5?

buger:
Because initially I thought it will be for 2.4.1, but it depends on refactorings I do not want to backport. So moved from 2.4.1 to 2.5.

ilijabojanovic:
Verified

lonelycode:
Re-opening this as we may need a better fix for this issue than what is currently in place.

PR #1334, opened 2017-11-30
Only one api deactivating fix

dencoded:
added changes to fix https://github.com/TykTechnologies/tyk/issues/1035


dencoded:
@buger @mvdan please take a look

buger:
Nice! (I'll squash)

PR #1337, opened 2017-12-01
cache: don't depend on timings for the test

mvdan:
We were seeing failures on CI, like:

	--- FAIL: TestWithCacheAllSafeRequests (0.06s)
		gateway_test.go:1269: wanted GET / to cache, but it didn't

They can be easily reproduced locally if a delay on Redis is forced,
such as by adding a time.Sleep(10 * time.Millisecond) right before the
"go m.CacheStore.SetKey" line.

The reason is that the redis write is asynchronous, so we need to wait
for it to be done before we can do the next request and see whether or
not its response was cached.

Much like with the reload queue, add a channel so that the request
handling goroutine and the test goroutine can communicate and
coordinate.

The tests passed 100 runs with the sleep after the fix, confirming that
the flake should be gone for good.

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -1248,6 +1248,8 @@ func TestWithCacheAllSafeRequests(t *testing.T) {
 		{"GET", "/", true},   // still cached after the POST
 	}
 	highestID := -1
+	cachedRequest = make(chan bool)
+	defer func() { cachedRequest = nil }()
 	for _, tc := range tests {
 		req, err := http.NewRequest(tc.method, baseURL+tc.path, nil)
 		if err != nil {
@@ -1271,5 +1273,6 @@ func TestWithCacheAllSafeRequests(t *testing.T) {
 		if testResp.ID > highestID {
 			highestID = testResp.ID
 		}
+		<-cachedRequest
 	}
 }
--- a/mw_redis_cache.go
+++ b/mw_redis_cache.go
@@ -96,11 +96,23 @@ func (m *RedisCacheMiddleware) decodePayload(payload string) (string, string, er
 	return "", "", errors.New("Decoding failed, array length wrong")
 }
 
+// cachedRequest is for the tests to know if a request was cached,
+// as otherwise it's hard to tell given that the redis write is async.
+var cachedRequest chan bool
+
+func didCache(cached bool) {
+	if cachedRequest != nil {
+		// in a goroutine, to let the response be sent
+		go func() { cachedRequest <- cached }()
+	}
+}
+
 // ProcessRequest will run any checks on the request on the way through the system, return an error to have the chain fail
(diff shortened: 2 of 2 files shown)


buger:
race?

mvdan:
The race seems unrelated - likely another one of the same races that @dencoded was finding.

joshblakeley:
@buger this is a blocker for me at the moment I think for #1340

dencoded:
I think suggested changes solves the problem. We write to cache storage in async-way and if we need to wait for the moment when those changes actually made to cache - this instrumentation will allow this. 

I have a couple of concerns though. It is code needed just to sync tests with actual cache storage implementation which is part of the real app. Another concern - I think it is normal behaviour if we do request assuming it will be cached asynchronously, and then do another one but it is still not in cache.

I's suggest two options (which I believe simpler, maybe too simple):

1. sleep between requests using some jitter and fail after some number of retries (if requests wasn't cached/not cached as expected)
2. or after http-request - read cached key from redis directly as part of test-code to wait until it cached/not cached, to make sure it is there and then test that next request was retrieved from cache (using existing logic around returned request ID)

mvdan:
> sleep between requests using some jitter and fail after some number of retries

That would still leave the test being flakey. What if you wait for a total of 5 seconds, but redis took 6 to take effect?

> read cached key from redis directly as part of test-code to wait until it cached/not cached

This is essentially the same - you still need a timeout to wait for the write to actually happen.

I'm not sure what the problem is with this approach. I can see that it's a bit verbose, but there's no other way to know exactly when the gateway does the redis write (or decides not to do a redis write at all).

dencoded:
The main problem with that approach I see is that we are making code more testable but It brings extra-complexity and chunks of code gateway wouldn't need when running without tests.

If we want to invest some time to make our code more testable I'd vote for:

- don't relay on cache storage availability/presence in tests (at least particularly in this case)
- in test code inject in `RedisCacheMiddleware`'s `CacheStore` the mock instead of real redis
- provide mock's expectations for methods' calls, functions' params and returns for deferent situation to test how gateway interacts with cache storage and reacts to various replies/errors from cache storage

Or if we don't want mess up with this now use sleeps/speak to redis directly leaving the gateway code base unchanged

buger:
Btw, I simplified test a bit https://github.com/TykTechnologies/tyk/commit/a0772e584fd1a02fff426223f0bc2f0a8e8b9f6d

asoorm:
I think @mvdan PR is fine for time being and should be merged, because it solves the problem and unblocks @joshblakeley .
I think we should create new ticket to clean up and implement @dencoded suggestion about mocks so that we don't need to rely on 3rd party service availability & eventual consistency during *unit* tests.

buger:
Fixed tests with adding additional timing, too many discussions for a such small thing :) 

I do not like this idea either, but introducing separate logic, which lives in production code, just to run the tests is a bad smell. And we for sure should solve this different way in future, by using of the suggestions above.

@joshblakeley rebase from master, issue should be fixed

Issue #1283, opened 2017-11-10
Make tyk-python independent from the system Python version


excieve:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
We're dynamically linking against `libpython3`, however our packages are built for Python 3.4, which is an issue outside of a vetted environment (e.g. docker). Even when targeting a specific distro version, system can upgrade a Python minor, which may break ABI (current Ubuntu LTS has 3.5). This implies laying additional burden on infrastructure of the users who use raw packages and prevents us from easily upgrading even our own docker base images.

**What is the expected behavior?**
Statically link Python runtime and bundle its standard library along with our required dependencies so that we wouldn't depend on a specific version of Python being present on the system.

The proposed solution will make our packages fully self-contained in terms of dependencies, make deployments more simple for the users and let us easily manage it without worrying for compatibility. The downsides are:
 1) increasing the size of the package and executable (but it's not by a lot so I wouldn't call it an issue)
 2) having to find a way for plugin developers to add 3rd-party dependencies as it wouldn't be possible to just use system-wide pip for it

The latter could be solved in a user-friendly way with e.g. `tyk-cli` reading the standard `requirements.txt` and installing in the bundle or maybe creating a virtual env for it, either way it's solvable.

Other possible solutions to the issue are:
 1) Setting `Py_LIMITED_API` preprocessor macro to only use a stable subset of Python API that never breaks ABI for Python 3.2+. Unfortunately this doesn't currently work (tested it) due to our reliance on Cython, which does not support it and [doesn't seem to be keen on it any time soon](https://mail.python.org/pipermail/cython-devel/2011-February/000108.html). Is there a way for us to more or less painlessly drop Cython?
 2) Compiling binaries for every Python 3.4+ minor version. This would work but it's quite ugly IMO and would require us to add more binaries with every minor release of Python.

Other points of interest:
 1) Linking `libpython` statically is actually a preferred mode for the interpreter as it can win about 9% in performance according to some [benchmark](https://gist.github.com/methane/4681131).
 2) Although by default cgo links dynamically, it's capable of static linking as well so it doesn't really require any magic.
 3) Self-contained package means we can easily support any Linux distro and any release. It even makes building `tyk-python` for archs other than `x86-64` much more viable.
[...]

PR #1338, opened 2017-12-02
[WIP] Converts Python coprocess glue module to handcrafted C

excieve:
This is work in progress, to be reviewed but not merged yet.

This replaces Cython-generated code in Python coprocess with handcrafted C using only limited Python C API subset so that Py_LIMITED_API preprocessor macro could be set and the resulting binary dynamically linked with whatever version of Python as long as it's 3.2+.

At the same time plugin interface is left compatible by providing a gateway.py module that actually calls the C wrapper functions.

In addition to this, Python code has been partly made PEP-8 conformant.

Refs #1283.


excieve:
@matiasinsaurralde So that's the `Py_LIMITED_API` approach to #1283. I've tested it somewhat and it works! But more testing and cleanup is needed (will do next week). Please take a look.

matiasinsaurralde:
Thanks, looks good, I will give it a try tomorrow

coprocess/python/tyk/gateway_wrapper.c
```diff
@@ -0,0 +1,78 @@
...
+
+	value = TykGetData(key);
+	// TykGetData doesn't currently handle storage errors so let's at least safeguard against null pointer
+	if (value == NULL) {
+		PyErr_SetString(PyExc_ValueError, "Null pointer from TykGetData");
+		return NULL;
+	}
+	ret = Py_BuildValue("s", value);
+	// CGO mallocs it in TykGetData and Py_BuildValue just copies strings, hence it's our responsibility to free it now
+	free(value);
```
excieve: This might be fixing a memleak btw.

excieve:
So, while this works in terms of being compiled for limited API set, there's a problem with properly linking it to `libpython`. PEP-384 recommend linking against `libpython3.so` in this case, which appears to be a tiny library that does runtime dynamic linking to the actual `libpython3.x.so`. There are two problems with this:
 * It appears not every disto packages it, e.g. I couldn't find it on Ubuntu 16.04.
 * It doesn't seem to work with embedding — fails on linking stage, requires symbols from the actual `libpython`. I think it was meant to be used with extensions rather than to invoke the interpreter itself.

We could probably do some dirty tricks with symlinking the actual library and it should work but that won't do well in the packages. I'll see if statically linking concrete `libpython` would still allow us to use system Python's stdlib and `site/dist-packages`.

excieve:
Right, so this won't work well unfortunately. At least I don't see a clean way to dynamically link `libpython3`.

First, `libpython3` wrapper can't be linked from an executable as it does runtime `dlopen` of the concrete `libpython3.x`. This only works with other dynamic libs (e.g Python extensions). It could _possibly_ work if we move the Python coprocess stuff to a separate package an build it with something like `-buildmode=shared`, link it against generic `libpython3` and then link tyk executable to it.

This, however, brings us to the second issue — distros vary in how they package libpython. So far I've seen two ways:
 * RedHat distros (Fedora, CentOS, EL) would properly package the generic `libpython3.so`.
 * Debian-based (including Ubuntu) would not do that and instead only provide concrete `libpython3.5m.so`, etc.

I believe some others might do the `SONAME` trick and let concrete libs be runtime linked by execs compiled against other versions.

Anyway, that's where the plan with linking against the generic lib in some way falls flat. Unless there are any ideas, I'm going to proceed with compiling in `libpython3.x` statically (this works fine) and packaging the stdlib along with it.

The handcrafted C module is still going to be useful I think — it's got a lot less cruft than the Cython thing and I think it might be fixing a memleak that Cython would happily ignore.

buger:
What about 3rd party python libraries. If they are installed via Pip locally, will they be available to our plugins if it statically linked?

excieve:
@buger Our direct dependencies (`protobuf` and `grpcio`) can be packaged straight into the stdlib we'd distribute.

Things get more difficult with packages installed by the users. We'll need to provide a way for them to install into our custom `PYTHONHOME`. There are several ways to do it, e.g:
 * provide a `pip` in the distribution that's configured for it and document it for the users, this is pretty similar to what `virtualenv` does, in fact it might be possible to just use virtualenv for this
 * integrate it into the bundling so that the `tyk-cli` command would read the plugin dependencies metadata and install them directly into the bundle (and working out the `PYTHONPATH` for them) thus allowing various versions of the same dependencies in different bundles

The latter is probably more user-friendly.

excieve:
This, of course, would not work with existing installations. I wonder if there's a way to provide some compatibility plan for the time being. E.g. statically compiling Python 3.4 and defaulting to its system `PYTHONHOME`. Then deprecating this in some later release and compiling for a more recent Python.

buger:
@excieve I am right that changes introduced in this PR still can be useful to us even without static compilation? 

Can it help us increase compatibility with 3.5, and 3.6? 

My main concern with this changes is compatibility issues. And the scope of this tasks becomes so much bigger, including adding own python packaging tools. Maybe we should think about plan B and ensure that our plugins work on most popular python distributions? Since now we have tests for coprocess plugins, it should be possible use CI to catch any inconsistency bugs.

matiasinsaurralde:
@buger That's right, using the limited API is a better choice for compatibility with other versions.

excieve:
@buger Yes, this is still useful — I think it fixes a memleak in the glue module at the very least.

Re compatibility... it's complicated. Compiling for limited API is supposed to make you compatible with any 3.2+ runtime. However, it seems to only have substantial effect when you're making an extension rather than embedding an interpreter as we do (either by dynamically linking libpythonx.y or statically).

In theory, it should still make the executable more compatible across the runtimes but it does not at all mean dynamic linking compatibility because distros can pack libpython is lots of various ways (as I explained in comments above).

Plan B might mean we'd have to link against libpython3.4 for CentOS6/Debian Jessie derived packages and against libpython3.5 for CentOS7/Debian Stretch packages, which currently covers the supported range of distros. This is of course doable and is mostly a building/packaging concern: separate build steps, different packages. Good thing it's just two cases at the moment but we'll need to keep that updated.

excieve:
I also have a terrible hack in mind to experiment with 👿 There's a tool named `patchelf` that can change some records of executables, including so names of dependencies. This can be done during package installation/upgrade by detecting the system's python libs. This way we wouldn't have to recompile/relink for different distros, provided we're compatible and tested.

buger:
I like how Plan B sounds. If we build this separate packages, can we ensure that each system will automatically get own packages? Does package cloud supports it?

buger:
Also can you update this PR to latest master, to resolve conflicts, so we can merge it.

excieve:
@buger 
> If we build this separate packages, can we ensure that each system will automatically get own packages? Does package cloud supports it?

Of course. I'll just need to make changes to the build/push scripts so that they end up in corresponding distro/version repository.

> Also can you update this PR to latest master, to resolve conflicts, so we can merge it.

Sure.

excieve:
I've resolved conflicts and manually tested with a simple plugin.

PR #1340, opened 2017-12-04
Added default version feature for versioned APIs

### LINKED ISSUE #955: Add a way to define a default API version for when none is requested
If an API endpoint is versioned and a request comes through without specifying a version, we return a `Version information not found` error.

I propose adding a new setting alongside `not_versioned`, called `default_version`, that optionally sets the version that will be used if none is requested.

If `default_version` is not set or left as an empty string, the current behavior is kept.

Note that this won't change the behavior if a request asks for a version that doesn't exist - that will still be the same error.

An example config:

```
"version_data": {
    "not_versioned": true,
    "default_version": "v2",
    "versions": {
        "v1": {"name": "v1", ...},
        "v2": {"name": "v2", ...}
    }
}
```

Original thread with the suggestion: https://community.tyk.io/t/is-there-a-way-to-set-up-a-default-version/1792

joshblakeley:
Fixes #955 

If default_version is not set or left as an empty string, the current behaviour is kept and an error is returned if no default version is explicitly requested in a header, url query param, or url path.

Note that this won't change the behavior if a request asks for a version that doesn't exist - that will still be the same error.

An example config:

```
"version_data": {
    "not_versioned": true,
    "default_version": "v2",
    "versions": {
        "v1": {"name": "v1", ...},
        "v2": {"name": "v2", ...}
    }
}
```

### DIFF
--- a/api.go
+++ b/api.go
@@ -1479,3 +1479,10 @@ func ctxGetUrlRewritePath(r *http.Request) string {
 	}
 	return ""
 }
+func ctxGetDefaultVersion(r *http.Request) bool {
+	return r.Context().Value(VersionDefault) != nil
+}
+
+func ctxSetDefaultVersion(r *http.Request) {
+	setCtxValue(r, VersionDefault, true)
+}
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -377,3 +377,34 @@ func (ln *customListener) Accept() (conn io.ReadWriteCloser, clientAddr string, 
 func (ln *customListener) Close() error {
 	return ln.L.Close()
 }
+
+func TestDefaultVersion(t *testing.T) {
+	ts := newTykTestServer()
+	defer ts.Close()
+
+	buildAndLoadAPI(func(spec *APISpec) {
+		v1 := apidef.VersionInfo{Name: "v1"}
+		v1.Name = "v1"
+		v1.Paths.WhiteList = []string{"/foo"}
+
+		v2 := apidef.VersionInfo{Name: "v2"}
+		v2.Paths.WhiteList = []string{"/bar"}
+
+		spec.VersionDefinition.Location = "url-param"
+		spec.VersionDefinition.Key = "v"
+		spec.VersionData.NotVersioned = false
+
+		spec.VersionData.Versions["v1"] = v1
+		spec.VersionData.Versions["v2"] = v2
+		spec.VersionData.DefaultVersion = "v2"
+		spec.Proxy.ListenPath = "/"
+	})
+
+	ts.Run(t, []test.TestCase{
+		{Path: "/foo", Code: 403},      // Not whitelisted for default v2
+		{Path: "/bar", Code: 200},      // Whitelisted for default v2
+		{Path: "/foo?v=v1", Code: 200}, // Allowed for v1
+		{Path: "/bar?v=v1", Code: 403}, // Not allowed for v1
+	}...)
+
+}
--- a/config/config.go
+++ b/config/config.go
@@ -269,6 +269,7 @@ type Config struct {
 	Security                          SecurityConfig                        `json:"security"`
 	EnableKeyLogging                  bool                                  `json:"enable_key_logging"`
 	NewRelic                          NewRelicConfig                        `json:"newrelic"`
+	VersionHeader                     string                                `json:"version_header"`
 }
 
 type CertData struct {
--- a/handler_error.go
+++ b/handler_error.go
@@ -87,6 +87,8 @@ func (e *ErrorHandler) HandleError(w http.ResponseWriter, r *http.Request, errMs
 
 		t := time.Now()
 
+		addVersionHeader(w, r)
+
 		version := e.Spec.getVersionFromRequest(r)
 		if version == "" {
 			version = "Non Versioned"
--- a/handler_success.go
+++ b/handler_success.go
@@ -24,6 +24,7 @@ const (
 	SessionData = iota
 	AuthHeaderValue
 	VersionData
+	VersionDefault
 	OrgSessionContext
 	ContextData
 	RetainHost
@@ -66,6 +67,16 @@ func tagHeaders(r *http.Request, th []string, tags []string) []string {
 	}
 
 	return tags
+}
+
+func addVersionHeader(w http.ResponseWriter, r *http.Request) {
+	if ctxGetDefaultVersion(r) {
+		if vinfo := ctxGetVersionInfo(r); vinfo != nil {
+			if config.Global.VersionHeader != "" {
+				w.Header().Set(config.Global.VersionHeader, vinfo.Name)
+			}
+		}
+	}
 }
 
 func (s *SuccessHandler) RecordHit(r *http.Request, timing int64, code int, requestCopy *http.Request, responseCopy *http.Response) {
@@ -219,6 +230,8 @@ func (s *SuccessHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) *http
 		log.Debug("Upstream Path is: ", r.URL.Path)
 	}
 
+	addVersionHeader(w, r)
+
 	var copiedRequest *http.Request
 	if recordDetail(r) {
 		copiedRequest = copyRequest(r)
@@ -259,6 +272,8 @@ func (s *SuccessHandler) ServeHTTPWithCache(w http.ResponseWriter, r *http.Reque
 	t1 := time.Now()
 	inRes := s.Proxy.ServeHTTPForCache(w, r)
 	t2 := time.Now()
+
+	addVersionHeader(w, r)
 
 	var copiedResponse *http.Response
 	if recordDetail(r) {
--- a/api_definition.go
+++ b/api_definition.go
@@ -1032,8 +1032,13 @@ func (a *APISpec) Version(r *http.Request) (*apidef.VersionInfo, []URLSpec, bool
 			}
 		} else {
 			// Extract Version Info
+			// First checking for if default version is set
 			vname := a.getVersionFromRequest(r)
-			if vname == "" {
+			if vname == "" && a.VersionData.DefaultVersion != "" {
+				vname = a.VersionData.DefaultVersion
+				ctxSetDefaultVersion(r)
+			}
+			if vname == "" && a.VersionData.DefaultVersion == "" {
 				return &version, nil, false, VersionNotFound
 			}
 			// Load Version Data - General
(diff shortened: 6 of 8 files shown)


api.go
```diff
@@ -1460,3 +1460,14 @@ func ctxSetVersionInfo(r *http.Request, v *apidef.VersionInfo) {
 	}
 	setCtxValue(r, VersionData, v)
 }
+
+func ctxGetDefaultVersion(r *http.Request) bool {
+	if r.Context().Value(VersionDefault) != nil {
```
dencoded: semantically I would just do `return r.Context().Value(VersionDefault) != nil`

joshblakeley: Will do. Theres some debugging output i've left in there too so will clean that up too.

handler_error.go
```diff
@@ -87,6 +87,12 @@ func (e *ErrorHandler) HandleError(w http.ResponseWriter, r *http.Request, errMs
 
 		t := time.Now()
 
+		if ctxGetDefaultVersion(r) {
+
+			vinfo := ctxGetVersionInfo(r).Name
+			w.Header().Set("X-Tyk-Default-Version", vinfo)
```
dencoded: I see string value `X-Tyk-Default-Version` repeated several times, might worth to introduce a constant for it

dencoded:
not in this PR but related to added changes - I think this [comment](https://github.com/TykTechnologies/tyk/pull/1340/files#diff-e174c6f6100dfee8c74de8323d464e36R1009) is not up to date as I see from code we support header, url-param and url 1st part

handler_error.go
```diff
@@ -87,6 +87,12 @@ func (e *ErrorHandler) HandleError(w http.ResponseWriter, r *http.Request, errMs
 
 		t := time.Now()
 
+		if ctxGetDefaultVersion(r) {
+
+			vinfo := ctxGetVersionInfo(r).Name
```
dencoded: might worth to check if `ctxGetVersionInfo(r)` returns nil, looking at implementation it can. I see from code that assumption is that we always set VersionInfo in context, if we ever miss it - we hit into nil pointer panic

handler_error.go
```diff
@@ -87,6 +88,12 @@ func (e *ErrorHandler) HandleError(w http.ResponseWriter, r *http.Request, errMs
 
 		t := time.Now()
 
+		if ctxGetDefaultVersion(r) {
+			if vinfo := ctxGetVersionInfo(r); vinfo != nil {
+				w.Header().Set("X-Tyk-Default-Version", vinfo.Name)
```
buger: I assume you forgot replace constant here

joshblakeley: Good spot

handler_error.go
```diff
@@ -17,6 +17,7 @@ const (
 	defaultTemplateName   = "error"
 	defaultTemplateFormat = "json"
 	defaultContentType    = "application/json"
+	defaultVersionHeader  = "X-Tyk-Default-Version"
```
buger: Let's make it configurable, and add this header ONLY if header value is specified. 

Related to our latest tech meeting. Basic idea that we should not expose user infrastructure, in other words that user use Tyk (which can be easily identified by this header name).

api_definition_test.go
```diff
@@ -145,6 +181,17 @@ func TestNotVersioned(t *testing.T) {
 	}
 }
 
+func TestDefaultVersion(t *testing.T) {
```
buger: This can't be really counted as a real test because it does not show that request was actually sent to the right version. Also, it does not go through HTTP pipeline. Also, I see that you copy pasted API definition, and a lot of the stuff in it does not use at all. There is a better way of writing tests for the gateway, see `buildAndLoadAPI` function usage. But basic idea that your test will look like:

```
func TestDefaultVersion(t *testing.T) {
	config.Global.ListenAddress = "127.0.0.1"

	ln, _ := generateListener(0)
	baseURL := "http://" + ln.Addr().String()
	listen(ln, nil, nil)
	defer func() {
		config.Global.ListenAddress = ""
		ln.Close()
	}()

        buildAndLoadAPI(func(spec *APISpec) {
           v1 := apidef.VirtualMeta{Name: "v1"}
           v1.Paths.WhiteList = []string{"/"}

           v2 := apidef.VirtualMeta{Name: "v2"}
           v2.Paths.WhiteList = []string{"/get"}
           
           spec.VersionData.Location = "url-param"
           spec.VersionData.Key = "v"

           spec.VersionData.Versions["v1"] = v1
           spec.VersionData.Versions["v2"] = v2
           spec.VersionData.DefaultVersion = "v2"
        })
        
        testCases := [...]struct {
		url     string
		code int
	}{
		{"/", 404}, // Not whitelisted for default v2
		{"/get", 200},  // Whitelisted for default v2
		{"/?v=v1", 200}, // Allowed for v1
		{"/?v=v1", 404}, // Not allowed for v1
	}

        for _, tc := range testCases {
          resp, err := http.Get(baseURL + tc.url)
          if err !=nil {
            t.Fatal(err, tc)
          }

           if resp.StatusCode != tc.code {
             t.Error(tc.url, resp.StatusCode, tc.code)
           }
        }
}
```

gateway_test.go
```diff
@@ -650,7 +650,7 @@ func TestVersioningRequestFail(t *testing.T) {
 	recorder := httptest.NewRecorder()
 	req := testReq(t, "GET", "/", nil)
 	req.Header.Set("authorization", "zz1234")
-	req.Header.Set("version", "v1")
+	req.Header.Set("version", "v3")
```
buger: Why you set it to v3?

buger:
@joshblakeley bump

gateway_test.go
joshblakeley: I can't actually remember at this point. Will check when retesting after changes.

api_definition_test.go
joshblakeley: Excellent, I was struggling to see how to do a better test like this. Thanks.

handler_error.go
joshblakeley: So should we have options for header, url-param and to disable?

api_definition_test.go
joshblakeley: Im not quite sure what the VirtualMeta has to do with this as thats related to virtual paths?

handler_error.go
joshblakeley: Doing it like this for now so relevant part of apidef would be configurable as:


"version_data": {
    "not_versioned": false,
    "default_version": "v2",
    "versions": {
        "v1": {"name": "v1", ...},
        "v2": {"name": "v2", ...}
    }
    "DefaultLocation": "url-param"
    "DefaultKey": "dv"
}

api_definition_test.go
joshblakeley: Worked it out nvm

handler_error.go
joshblakeley: Got it, that makes much more sense than what i was doing, cheers.

joshblakeley:
@dencoded  @buger There is a race failure in 1.9. I don't want to go sticking mutex in the code for the sake of the test in order to fix so wondering how to go about sorting this?

dencoded:
@joshblakeley I'd do the same we did last time - open issue with data race output and fix it in master, then rebase this branch on master.

Speaking of data races - I think there are still 3-4 places where we have them, it is just hard to catch because of intermittent nature of them.

joshblakeley:
Ok great i'll do it like that then.

joshblakeley:
Currently blocked by #1383

joshblakeley:
Test now follows framework @buger has laid out.

joshblakeley:

handler_error.go
```diff
@@ -87,6 +88,14 @@ func (e *ErrorHandler) HandleError(w http.ResponseWriter, r *http.Request, errMs
 
 		t := time.Now()
 
+		if ctxGetDefaultVersion(r) {
+			if vinfo := ctxGetVersionInfo(r); vinfo != nil {
+				if config.Global.DefaultVersionHeader {
+					w.Header().Set(defaultVersionHeader, vinfo.Name)
```
buger: I assume you need to remove `defaultVersionHeader` constant and replace it with `config.Global.DefaultVersionHeader` in all places.

handler_success.go
```diff
@@ -219,6 +220,14 @@ func (s *SuccessHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) *http
 		log.Debug("Upstream Path is: ", r.URL.Path)
 	}
 
+	if ctxGetDefaultVersion(r) {
```
buger: This code snippet repeated in 3 places, without any changes. I think make sense move it to separate function, and re-use everywhere.

config/config.go
```diff
@@ -262,6 +262,7 @@ type Config struct {
 	LogLevel                          string                                `json:"log_level"`
 	Security                          SecurityConfig                        `json:"security"`
 	EnableKeyLogging                  bool                                  `json:"enable_key_logging"`
+	DefaultVersionHeader              bool                                  `json:"default_version_header"`
```
buger: Since it is not `Default` anymore, and probably make sense rename option to just `VersionHeader`

buger:
Do not forget that this feature should be added to the Dashboard as well, even if field will be hidden in UI.

config/config.go
```diff
@@ -262,6 +262,7 @@ type Config struct {
 	LogLevel                          string                                `json:"log_level"`
 	Security                          SecurityConfig                        `json:"security"`
 	EnableKeyLogging                  bool                                  `json:"enable_key_logging"`
+	VersionHeader                     bool                                  `json:"default_version_header"`
```
buger: Rename JSON field too, and remove `defaultVersionHeader` constant

handler_success.go
```diff
@@ -68,6 +69,16 @@ func tagHeaders(r *http.Request, th []string, tags []string) []string {
 	return tags
 }
 
+func addVersionHeader(w http.ResponseWriter, r *http.Request) {
+	if ctxGetDefaultVersion(r) {
+		if vinfo := ctxGetVersionInfo(r); vinfo != nil {
+			if config.Global.VersionHeader {
+				w.Header().Set("X-Tyk-Default-Version", vinfo.Name)
```
buger: You forgot to update "X-Tyk-Default-Version" to `config.Global.VersionHeader`

joshblakeley: `config.Global.VersionHeader` is just a boolean flag for whether the header should be added or not.

buger: Right, I just thought you initially wanted make it configurable

joshblakeley: I just intended it to be on-or-off depending on whether the user wanted to expose their configuration. If you think having it be a configurable string is better I can make that change easily.

buger: Yeah, let's make it configurable, update bool to string.

Issue #1339, opened 2017-12-04
Websocket does not work


dikderoy:
**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
When opening a ws connection following exception is raised in Tyk Gateway logs:
```
tyk_gateway_1    | 2017/12/04 05:24:32 http: panic serving 172.18.0.18:53856: interface conversion: http.RoundTripper is *main.WSDialer, not *http.Transport
tyk_gateway_1    | goroutine 43779 [running]:
tyk_gateway_1    | net/http.(*conn).serve.func1(0xc4201d5400)
tyk_gateway_1    | 	/usr/local/go/src/net/http/server.go:1721 +0xd0
tyk_gateway_1    | panic(0xcfedc0, 0xc42041ca40)
tyk_gateway_1    | 	/usr/local/go/src/runtime/panic.go:489 +0x2cf
tyk_gateway_1    | main.(*ReverseProxy).WrappedServeHTTP(0xc4203e3360, 0x1330ee0, 0xc42023b260, 0xc420069a00, 0x5a24dc00, 0x0)
tyk_gateway_1    | 	/src/github.com/TykTechnologies/tyk/reverse_proxy.go:524 +0x1d0f
tyk_gateway_1    | main.(*ReverseProxy).ServeHTTP(0xc4203e3360, 0x1330ee0, 0xc42023b260, 0xc420069a00, 0x120)
tyk_gateway_1    | 	/src/github.com/TykTechnologies/tyk/reverse_proxy.go:349 +0x60
tyk_gateway_1    | main.(*SuccessHandler).ServeHTTP(0xc420249e00, 0x1330ee0, 0xc42023b260, 0xc420069a00, 0x31048f40)
tyk_gateway_1    | 	/src/github.com/TykTechnologies/tyk/handler_success.go:218 +0x16e
tyk_gateway_1    | main.(*DummyProxyHandler).ServeHTTP(0xc420249e00, 0x1330ee0, 0xc42023b260, 0xc420069a00)
tyk_gateway_1    | 	/src/github.com/TykTechnologies/tyk/api_loader.go:518 +0x4b
tyk_gateway_1    | main.createMiddleware.func1.1(0x1330ee0, 0xc42023b260, 0xc420069a00)
tyk_gateway_1    | 	/src/github.com/TykTechnologies/tyk/middleware.go:88 +0x7ec
[...]

PR #1342, opened 2017-12-05
type assertion in WrappedServeHTTP fixed

dencoded:
added fix for https://github.com/TykTechnologies/tyk/issues/1339

Unfortunately, I wasn't able to write good unit test because `WSDialer` is doing hijack for connection, so it is not trivial - need to invest more time to research the best approach to test. I was thinking about `ResponseRecorder` with `http.Hijacker` implemented.


reverse_proxy.go
```diff
@@ -518,12 +518,18 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 	// Circuit breaker
 	breakerEnforced, breakerConf := p.CheckCircuitBreakerEnforced(p.TykAPISpec, req)
 
+	// set up TLS certificates for upstream if needed
+	var tlsCertificates []tls.Certificate = nil
```
buger: Do you added `= nil` for expressivity or it is needed somehow?

buger:
It will be perfect you can add simple test covering websockets, because right now we have nothing.

reverse_proxy.go
dencoded: it is not needed as zero value is already nil, I think I added it just looking at " = nil" assignment in  https://github.com/TykTechnologies/tyk/pull/1342/files#diff-da06648202ab00313675c5150a92bffeL524

dencoded:
ok, I will add some simple test for reverse proxy to websocket

dencoded:
@buger I've added test for reverse proxying to websocket upstream as requested, please take a look

PR #1343, opened 2017-12-05
1163 - Json validation

### LINKED ISSUE #1163: Add way to validate inbound JSON objects with JSON Schema

asoorm:
Fixes #1163
Based on #1180

Adds a new JSON Validation middleware that can be configured as follows:

```
"version_data": {
  "not_versioned": true,
  "versions": {
    "default": {
      "name": "default",
      "use_extended_paths": true,
      "extended_paths": {
        "validate_json": [{
          "method": "POST",
          "path": "me",
          "schema": schema,
          "error_response_code": 422 // 422 default however can override.
        }]
      }
    }
  }
},
```

The schema must be a draft v4 JSON Schema spec. The gateway will attempt
to validate the inbound request against it, if fields are failing the
validation process, a detailed error response is provided for the user
to fix their payload.

example schema:

```
{
    "title": "Person",
    "type": "object",
    "properties": {
        "firstName": {
            "type": "string"
        },
        "lastName": {
            "type": "string"
        },
        "age": {
            "description": "Age in years",
            "type": "integer",
            "minimum": 0
        }
    },
    "required": ["firstName", "lastName"]
}
```

This will require a new Dashboard UI to handle input.


api_definition.go
```diff
@@ -699,6 +702,20 @@ func (a APIDefinitionLoader) compileTrackedEndpointPathspathSpec(paths []apidef.
 	return urlSpec
 }
 
+func (a APIDefinitionLoader) compileValidateJSONPathspathSpec(paths []apidef.ValidatePathMeta, stat URLStatus) []URLSpec {
+	urlSpec := []URLSpec{}
```
dencoded: To avoid slice re-allocation I think we could use `urlSpec := make([]URLSpec, len(paths))` and then do `for index, pathVal .. urlSpec[index] = newSpec`

mw_validate_json.go
```diff
@@ -0,0 +1,81 @@
...
+// ProcessRequest will run any checks on the request on the way through the system, return an error to have the chain fail
+func (k *ValidateJSON) ProcessRequest(w http.ResponseWriter, r *http.Request, _ interface{}) (error, int) {
+
+	_, versionPaths, _, _ := k.Spec.Version(r)
+	found, meta := k.Spec.CheckSpecMatchesStatus(r, versionPaths, ValidateJSONRequest)
+	if !found {
+		return nil, 200
+	}
+
+	mmeta := meta.(*apidef.ValidatePathMeta)
```
dencoded: `mmeta` and `meta` - it is probably better to use variables with more different names (than one-letter difference - source of potential bugs)

mw_validate_json.go
```diff
@@ -0,0 +1,81 @@
...
+	if mmeta.ValidateWith64 == "" {
+		return errors.New("no schemas to validate against"), 400
+	}
+
+	rCopy := copyRequest(r)
+	bodyBytes, err := ioutil.ReadAll(rCopy.Body)
+	if err != nil {
+		return err, 400
```
dencoded: I would count it as `500` reply as something might happen on communication level and retry of the same request seems to be a valid situation

mw_validate_json.go
dencoded: also, might worth to do safe type assert `meta, ok := meta.(*apidef.ValidatePathMeta)` to avoid panics. I know it  is unlikely but still..

mw_validate_json.go
```diff
@@ -0,0 +1,81 @@
...
+
+	_, versionPaths, _, _ := k.Spec.Version(r)
+	found, meta := k.Spec.CheckSpecMatchesStatus(r, versionPaths, ValidateJSONRequest)
+	if !found {
+		return nil, 200
+	}
+
+	mmeta := meta.(*apidef.ValidatePathMeta)
+	if mmeta.ValidateWith64 == "" {
+		return errors.New("no schemas to validate against"), 400
```
dencoded: this seems to be not a problem with incoming requests as something is not right with API spec setup on our end

dencoded: I mean it is probably not a `400`

mw_validate_json.go
```diff
@@ -0,0 +1,81 @@
...
+	rCopy := copyRequest(r)
+	bodyBytes, err := ioutil.ReadAll(rCopy.Body)
+	if err != nil {
+		return err, 400
+	}
+	defer rCopy.Body.Close()
+
+	schemaBytes, err := base64.StdEncoding.DecodeString(mmeta.ValidateWith64)
+	if err != nil {
+		return errors.New("unable to base64 decode schema"), 400
```
dencoded: again, I think it could be not a `400` as incoming request seems to be fine

mw_validate_json.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	result, err := k.validate(bodyBytes, schemaBytes)
+	if err != nil {
+		return err, 400
```
dencoded: might worth to log err so we will be able to explain to customer why it was `400`

mw_validate_json.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	if !result.Valid() {
+		errStr := "payload validation failed"
+		for _, desc := range result.Errors() {
+			errStr = fmt.Sprintf("%s: %s", errStr, desc)
```
dencoded: I think `errStr` will contain in this case the latest element from `result.Errors()`, if we want to show all errors we should provide something like comma-separated list

mw_validate_json_test.go
```diff
@@ -0,0 +1,191 @@
...
+		"not_versioned": true,
+		"versions": {
+			"default": {
+				"name": "default",
+				"use_extended_paths": true,
+				"extended_paths": {
+					"validate_json": [{
+						"method": "POST",
+						"path": "me",
+						"validate_with_64": "ew0KICAgICJ0aXRsZSI6ICJQZXJzb24iLA0KICAgICJ0eXBlIjogIm9iamVjdCIsDQogICAgInByb3BlcnRpZXMiOiB7DQogICAgICAgICJmaXJzdE5hbWUiOiB7DQogICAgICAgICAgICAidHlwZSI6ICJzdHJpbmciDQogICAgICAgIH0sDQogICAgICAgICJsYXN0TmFtZSI6IHsNCiAgICAgICAgICAgICJ0eXBlIjogInN0cmluZyINCiAgICAgICAgfSwNCiAgICAgICAgImFnZSI6IHsNCiAgICAgICAgICAgICJkZXNjcmlwdGlvbiI6ICJBZ2UgaW4geWVhcnMiLA0KICAgICAgICAgICAgInR5cGUiOiAiaW50ZWdlciIsDQogICAgICAgICAgICAibWluaW11bSI6IDANCiAgICAgICAgfQ0KICAgIH0sDQogICAgInJlcXVpcmVkIjogWyJmaXJzdE5hbWUiLCAibGFzdE5hbWUiXQ0KfQ=="
```
dencoded: I'd prefer to see in test the actual schema instead of base64, you can encode it when put value here

mw_validate_json.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	rCopy := copyRequest(r)
+	bodyBytes, err := ioutil.ReadAll(rCopy.Body)
+	if err != nil {
+		return err, 400
+	}
+	defer rCopy.Body.Close()
+
+	schemaBytes, err := base64.StdEncoding.DecodeString(mmeta.ValidateWith64)
```
dencoded: might worth to cache `schemaBytes` as this might be too expensive to decode the same bas64 encoded string with every incoming through this middleware request

mw_validate_json.go
asoorm: @dencoded - Agree with status code changes and have implemented these.
The reason I have not logged before returning error is that IMO it's bad practice to log the error before returning. It's up to the caller of `ProcessRequest` to do the logging if it needs to. Open to thoughts on this.

mw_validate_json.go
asoorm: Check tests. Provides a `:` separated list.

mw_validate_json.go
asoorm: @dencoded nice - suggestions on implementation / examples of where done before please?

api_definition.go
asoorm: I like this idea - but I think for consistency, we keep same as rest of `api_definition.go`. Will create separate GH refactor issue for avoiding slice re-allocation.

dencoded: why introduce new code which will be subject of refactoring later? I mean we can refactor existing code as separated task for sure, but new changes - why not to do them better at the very beginning (if we believe they are better)

mw_validate_json.go
```diff
@@ -0,0 +1,82 @@
...
+
+		return errors.New(errStr), vPathMeta.ValidationErrorResponseCode
+	}
+
+	return nil, http.StatusOK
+}
+
+func (k *ValidateJSON) validate(input []byte, schema map[string]interface{}) (*gojsonschema.Result, error) {
+	inputLoader := gojsonschema.NewBytesLoader(input)
+	rulesLoader := gojsonschema.NewGoLoader(schema)
```
dencoded: this is related to loading/decoding of json schema. I feel like validating/parsing and loading of schema is the task for schema-loading process, we don't need to do this when processing every request inside a middleware

mw_validate_json.go
dencoded: I feel like decoding/loading json-schema is an operation which needs to be done when we load/reload API spec definition, so we can put loaded/ready to use json-schema in some additional field and use it

mw_validate_json.go
buger: Since middleware created per API, I think it should be ok to add some field to ValidateJSON middleware and cache this schema.

asoorm:
@buger @dencoded - Are we ok to merge / send to @ilijabojanovic for test? I am struggling with dashboard component testing given that it relies on api definition within this repo.

apidef/api_definitions.go
```diff
@@ -165,6 +165,16 @@ type MethodTransformMeta struct {
 	ToMethod string `bson:"to_method" json:"to_method"`
 }
 
+type ValidatePathMeta struct {
+	Path   string                 `bson:"path" json:"path"`
+	Method string                 `bson:"method" json:"method"`
+	Schema map[string]interface{} `bson:"schema" json:"schema"`
+	// TODO: Implement multi schema support
+	SchemaVersion string `bson:"schema_version" json:"schema_version"`
```
asoorm: Doesn't really do anything yet.

mw_validate_json.go
```diff
@@ -0,0 +1,93 @@
...
+	if !found {
+		return nil, http.StatusOK
+	}
+
+	vPathMeta := meta.(*apidef.ValidatePathMeta)
+	if vPathMeta.Schema == nil {
+		return errors.New("no schemas to validate against"), http.StatusInternalServerError
+	}
+
+	if vPathMeta.SchemaVersion != "" && vPathMeta.SchemaVersion != "draft-v4" {
```
asoorm: Only checks that is Empty string or not `draft-v4`

philsturgeon:
Hey there! I’m really excited about this, but Draft v04 is fairly outdated. Is there a plan to add support for more recent versions of the spec? Seeing as they declare the Schema in the file, it should be easy enough.

asoorm:
We are a bit limited with golang libs http://json-schema.org/implementations.html#validator-go .

I see that there is now a new lib https://github.com/santhosh-tekuri/jsonschema (untested by us) which supports draft6 but nothing yet for draft7. Any thoughts @buger @lonelycode ?

matiasinsaurralde:
There's an ongoing PR that adds support for newer versions: https://github.com/xeipuuv/gojsonschema/pull/180

asoorm:
And to track progress: https://github.com/xeipuuv/gojsonschema/issues/178

lonelycode:
I would suggest we go live with what we do have support for so as not to delay the feature and then update the library for a later release in a different issue, since it is also used elsewhere it will have a wider testing surface so will need more time to integrate.

philsturgeon:
For what it’s worth I’d agree with releasing draft 04 support now and following up with newer later. 😬

--
Phil Sturgeon
@philsturgeon

Issue #187, opened 2016-02-03
Body Transform does not match correctly with URL Rewrite


veger:
Hello,

I am defining an API with a URL rewrite to match the API to an existing one and a Body Transform, in order to match the body response to the expected response of the (new) API.

The URL Rewrite is as follows

```
"url_rewrites": [
    {
    "path": "my_service/id/{id}",
    "method": "GET",
    "match_pattern": "state/id/(\d+)",
    "rewrite_to": "some/path/service?servicename=my_service&id=$1"
    }
],
```

So far so good, the request is properly redirected to the existing API.

Now I want to modify the body of the response (throwing away some 'internal' data):

```
"transform_response": [
    {
    "template_data": {
        "template_mode": "blob",
        "template_source": "some response in Base64 encoding"
    },
    "path":  "my_service/id/{id}",
    "method": "GET"
    }
],
```

Unfortunately, this transformation is not triggered when a request to `my_service/id/{id}` is done. Even though the path is exactly the same as for the URL rewrite.

I discovered that the transform is working when the URL Rewrite is removed. So that made me wonder whether I need to change to path to something the matches the rewritten URL, which seems the case as changing path to `some/path/service{questionmark}servicename=my_service&id={id}` is working!

I suppose this is a bug as such behavior seems weird, as I need a different path than the API defines in order to match the response??

lonelycode:
What version are you running? The transforms happen before rewrites atm:

https://github.com/TykTechnologies/tyk/blob/master/main.go#L641

Which makes this behaviour difficult to explain ;-)

veger:
Tyk gateway (and dashboard) installed as Ubuntu package. Version 1.9.1.1 (and 0.9.7.2 for dashboard)

veger:
The body response transform also happens before the rewrite? As this transform needs to be executed after a response is received, and thus a rewrite is already done..? (I am quite new to golang, so I am not yet able to check the code at such detail level)

lonelycode:
Ah, wait, f course, sorry - I read that as a *request* body transform. 

Right, if the response is being transformed then it will run after the rewrite, which would explain why you need to match a different URL.

I see the issue though, seems a little counter-intuitive to match two different paths in and out.

veger:
Basically, I would expect to define an endpoint and have everything matching that one in a (sort of) session. So independent of modifications to the request, body or other data.

Is it possibility to change/fix Tyk in such a way? Or should I keep using my/the current work around?

[...]

PR #1350, opened 2017-12-08
allow transform response when path is equal to path of url rewrite

dencoded:
added changes to fix https://github.com/TykTechnologies/tyk/issues/187

also added some tests:

1. we use just transform_response
2. we use url_rewrite and transform_response, transform_response has in `path` value matching to URL request was rewritten to
3. we use url_rewrite and transform_response, transform_response has in `path` value equal to `path` specified in url_rewrite


dencoded:
hm race again, I will look into it

buger:
Nice!

PR #1352, opened 2017-12-08
Add NewRelic integration

nezorflame:
This is a port of the NewRelic integration which was originally done by @nebolsin in his fork of the tyk [here](https://github.com/nebolsin/tyk/tree/wtf).

We use this changed version of the gateway in production and so far it works as intended, but it'd be really awesome to have this in original tyk without the need for maintaining our fork.

Please review this and feel free to make/request changes.


buger:
Nice! 

However, i see that it include a lot of refactorings, unrelated to NewRelic. Is it possible to extract only NewRelic part?

nezorflame:
Will work on that.

nezorflame:
@buger still I'd suggest that these refactored parts need to be reviewed as well. Maybe in another pull request?

buger:
I also see that you were trying to solve bugs, caused by router reload logic. I think this is already fixed in our master in a similar manner, so maybe your custom changes will not be needed anymore.

nezorflame:
@buger I see, maybe they won;t be needed anymore then.
Will try only with the NewRelic stuff.

nezorflame:
@buger finished

coprocess.go
```diff
@@ -102,7 +102,10 @@ func (c *CoProcessor) ObjectFromRequest(r *http.Request) *coprocess.Object {
 
 	object.HookType = c.HookType
 
-	object.Metadata = make(map[string]string)
+	object.Metadata = map[string]string{
+		"RequestURI": r.RequestURI,
```
buger: Does it needed for NewRelic integration somehow, or for your internal purpose?

reverse_proxy.go
```diff
@@ -430,6 +431,10 @@ func httpTransport(timeOut int, rw http.ResponseWriter, req *http.Request, p *Re
 		return wsTransport
 	}
 
+	if txn, ok := rw.(newrelic.Transaction); ok {
```
buger: I assume this line should be run only if new relic enabled?

middleware.go
```diff
@@ -53,6 +54,10 @@ func createMiddleware(mw TykMiddleware) func(http.Handler) http.Handler {
 
 	return func(h http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if txn, ok := w.(newrelic.Transaction); ok {
```
buger: Same here, maybe we should run it only if new relic is enabled?

config/config.go
```diff
@@ -170,6 +170,12 @@ type SecurityConfig struct {
 	Certificates                     CertificatesConfig `json:"certificates"`
 }
 
+type NewRelicConfig struct {
+	Enabled    bool   `json:"enabled"`
```
buger: Do you think it make sense get rid of `Enabled` field, and instead just check if AppName is non empty?

nezorflame: OK, can be done.

reverse_proxy.go
nezorflame: Yes, will change it.

middleware.go
nezorflame: Agree.

coprocess.go
nezorflame: AFAIK we use this in our events.
If it's possible, it would be nice to leave this as it is.

nezorflame:
Added requested changes.

nezorflame:
@buger anything else to fix/check?

nezorflame:
@buger anything else to fix/check?

buger:
In general looks fine, and it will go through our internal testing. The only thing that concern me is the `RequestURI` part which is not part of NewRelic change. 

I do understand why you added it, and I'm happy to merge it as a second feature (even in terms of this PR), but first let's clarify few things. Do you use it only for JSVM or for Coprocess plugins? I'm asking because JSVM implementation looks good, and coprocess one is a hack, and kind of require adding the new field to proto buffer object. I will create a separate ticket for adding field to coprocess request object anyway, but first, need to know in what kind of middleware you use it right now.

nezorflame:
@buger for coprocess plugins.
We register hooks, and use RequestURI in a function like this:

```go
func (f Flow) ExtractData(ctx coprocess.Context, object *coprocess.Object) (*coprocess.Object, error) {
...
originalURI := object.Metadata["RequestURI"]

	level.Debug(f.logger).Log(
		"msg", "extracted request information",
		"uri", originalURI,
		"env", env,
		"version", version,
		"phone", phone,
		"action", action,
		"key", token,
		"verify", verify,
	)
...
}

nezorflame:
Any news on this?

buger:
We definitely going include this functionality to 2.5, just need to properly implement RequestURI on coprocess side, but this change will be done by us. Will submit PR soon.

nezorflame:
@buger awesome! Could you please @ me when this PR arrives? We'd want to be ready to implement it as soon as it arrives.

nezorflame:
Fixed bug when web transactions weren't sent into NewRelic

buger:
Merged manually https://github.com/TykTechnologies/tyk/commit/2580fd35b24be63e99433d6f8b777b61aa39a455

Note that RequestURI moved from Metadata, to request object field https://github.com/TykTechnologies/tyk/pull/1391

nebolsin:
@nezorflame thanks for pushing this forward 👍

Issue #1361, opened 2017-12-15
Make RPC pool size configurable


lonelycode:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
The RPC pool size sits at a constant 50, this can quickly overload a server if there are many gateways and few MDCB hosts.

It should be a configurable value instead of a constant.

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
All.

PR #1362, opened 2017-12-15
Made RPC Connection pool size configurable but default to 50

lonelycode:
Addresses #1361 by adding a new Slave option config: `rpc_pool_size` which can be set to anything, but if non-zero will default to 50 (the old hard-coded value).

To test, run gateway in slave mode with this config value set to a different value (e.g. 10) and then check how many open connections there are like this:

`netstat -anp | grep :80 | grep ESTABLISHED | wc -l`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -103,6 +103,7 @@ type SlaveOptionsConfig struct {
 	GroupID                         string `json:"group_id"`
 	CallTimeout                     int    `json:"call_timeout"`
 	PingTimeout                     int    `json:"ping_timeout"`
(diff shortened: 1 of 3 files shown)


lonelycode:
I just tested the change on a test box and with a number of `10` I now get:

```
root@load-hybrid-gw:/opt/tyk-gateway# netstat -anp | grep :9090 | grep ESTABLISHED | wc -l
10
```

buger:
Looks good, will backport to release-1.4 too

PR #1367, opened 2017-12-18
Jq transform no dep

klyr:
Add a [JQ](https://stedolan.github.io/jq/) transformation middleware (and response middleware).

### For the request middlewaire

The JQ filter is given a JSON object of the following form:

    {
      "body": The actual body of the request,
      "_tyk_context": The tyk context as a json object
    }

And must return a JSON object of the following form:

    {
      "body": The transformd Json body,
      "rewrite_headers": A key/value pair object to rewrite new headers,
      "tyk_context": A key/value pair object of variables to add into the tyk context
    }

### For the response middleware

The JQ filter is given a JSON object of the following form:

    {
      "body": The actual body of the request,
      "_tyk_context": The tyk context as a json object,
      "response_headers": A key/value object with the headers of the reponse
    }

And must return a JSON object of the following form:

    {
      "body": The transformd Json body,
      "rewrite_headers": A key/value pair object to rewrite new headers
    }

The JQ C library is needed at Compile and Run-time.


buger:
Nice! Have you tried linking JQ lib statically?

klyr:
Having the following flags seems to work, but I don't find it very clean ...
// #cgo LDFLAGS: /home/XXX/libs/libjq.a /home/XXX/jq/modules/oniguruma/src/.libs/libonig.a -lm

buger:
What about `//#cgo LDFLAGS: -linkmode external -extldflags "-static"` ?

klyr:
No success at the moment, working on it ...

buger:
JQ by itself support compiling fully static binary https://github.com/stedolan/jq/blob/master/Makefile.am

And it feels like passing 2 flags to the linker: `-static-libtool-libs` and `-all-static`

klyr:
The 2 flags you mentioned are libtools flags, their role is to use the '-static' flag and to rewrite libraries names to fullpath.
Moreover, -static tries to link everything statically even libc.

The best I can have in `jq.go` is (only compatible with GNU ld):

    // #cgo LDFLAGS: -l:libjq.a -l:libonig.a -lm

and  compile the whole thing with:

    govendor build -tags 'coprocess python'

And if the jq and oniguruma libraries are not in default path, prepend the command with for example:

    CGO_LDFLAGS="-L /home/.../jq/.libs -L /home/.../ jq/modules/oniguruma/src/.libs"

So at the end, libjq and libonig are linked statically, but libm, as libc, are linked dynamically.

buger:
I think we end up with dynamic linking. Do you think about nice way of capturing errors if jq lib not installed and showing error message to the user?

buger:
Merged by https://github.com/TykTechnologies/tyk/pull/1570

Issue #1375, opened 2017-12-22
Websockets still not working as expected


dikderoy:
**Do you want to request a *feature* or report a *bug*?**
*BUG*
**What is the current behavior?**
I've checked the fix for #1339 (in 2.4.2) - initially websocket connection is working, but as soon as you drop "first" connection, it stops working, and you cannot create new ws-connection without restarting tyk_gateway.

tyk gateway produces 500 Internal Server Error on websocket open attempt with following logs:

```
tyk_gateway_1    | time="Dec 22 15:57:10" level=error msg="Hijack error: http: connection has been hijacked" origin=172.18.0.18 path="/workspaces/default/chat"
tyk_gateway_1    | time="Dec 22 15:57:10" level=error msg="http: proxy error: http: connection has been hijacked" api_id=b7c7ce1ff3f449357eb28312a6fe8fe7 org_id=5a2b3da0bdf125000143add4 server_name="ws_api:8080" user_id="****b567" user_ip=172.18.0.18 user_name="someuser (platform)"
```
**What is the expected behavior?**
Tyk should not give 500 ISE on second and onward attempts to create websocket connection

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1) setup tyk to allow websockets (setup is same as in #1339)
2) use any sample app which accepts websockets behind tyk gateway
3) open a websocket-connnection (it should open just fine)
4) send couple of messages over opened websocket connection
5) close the connection. either from client or server side.
6) try to open new websocket connection. (it fails and gives 500 ISE)

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.4.2, see #1339

dencoded:
hm, I think I know what is going on, I will provide PR shortly

PR #1376, opened 2017-12-22
handling websocket connection upgrade fixed

dencoded:
added changes to fix https://github.com/TykTechnologies/tyk/issues/1375

also added some tests for different ws-connections open/close situations

### DIFF
--- a/reverse_proxy.go
+++ b/reverse_proxy.go
@@ -438,6 +438,10 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 	if p.TykAPISpec.HTTPTransport == nil {
 		_, timeout := p.CheckHardTimeoutEnforced(p.TykAPISpec, req)
 		p.TykAPISpec.HTTPTransport = httpTransport(timeout, rw, req, p)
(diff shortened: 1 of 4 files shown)


reverse_proxy.go
```diff
@@ -438,6 +438,10 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 	if p.TykAPISpec.HTTPTransport == nil {
 		_, timeout := p.CheckHardTimeoutEnforced(p.TykAPISpec, req)
 		p.TykAPISpec.HTTPTransport = httpTransport(timeout, rw, req, p)
+	} else if IsWebsocket(req) { // check if it is an upgrade request to NEW WS-connection
+		// overwrite transport's ResponseWriter from previous upgrade request
+		// as it was already hijacked and now is being used for other connection
+		p.TykAPISpec.HTTPTransport.(*WSDialer).RW = rw
```
dencoded: this is unsafe type assertion but I couldn't imagine situation when it would store something else (it is websocket request and `p.TykAPISpec.HTTPTransport` is already set)

buger:
@dencoded I included websocket endpoint into default test server ^

Issue #1374, opened 2017-12-22
Unrequested URL Encoding while proxying


paolo-rodeghiero:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
When asking an URL that contains parenthesis, which are used by some proprietary backends like SFCC to define arrays of parameters, they get converted to URL encoded characters like %28 and %29 and therefore get interpreted as a single value

e.g.

  GET http://gateway/promotions/(Test_GWP,Test_GWP)   --->  http://upstream/promotions/ %28Test_GWP,Test_GWP%29

Response: promotion id "(Test_GWA,Test_GWB)" not found

**What is the expected behavior?**

The same behaviour if upstream is called directly, without URL encoding the result

e.g.   GET http://gateway/promotions/(Test_GWP,Test_GWP)   --->  http://upstream/promotions/ (Test_GWP,Test_GWP)

Response: the data on promotion Test_GWA and Test_GWB

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

We can't provide as the affected SFCC backend is protected by a non public shareable secret.
If you contact us, we can provide privately

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**

Current version 2.4
No idea on . older versions

buger:
@paolo-rodeghiero try setting the following settings in tyk.conf:

```
"http_server_options": {
...
  "override_defaults": true,
  "skip_url_cleaning": true,
...
}
```

paolo-rodeghiero:
Thanks for your fast reply Leonid.
I tried but no changes  - same behaviour described

buger:
@dencoded any ideas?

dencoded:
@buger I will check

dencoded:
@buger at 1st look I wasn't able to replicate this issue.

also, `"skip_url_cleaning": true` just switches of usage of logic https://golang.org/src/path/path.go?s=2075:2105#L64 which removes from path extra `/`, `.` and `..`, it doesn't do any kind of URL encoding for incoming request or transforming URI segments if I understand right.

I would ask @paolo-rodeghiero for some details:
- full tyk config and API spec for that path - to have some idea what middlewares are involved
- make sure nothing seats in the middle of tyk-gateway and SFCC-upstream

paolo-rodeghiero:
Hello @buger and @dencoded ,
before opening the bug I replicated it in various environment without anything in the middle.
[...]

PR #1382, opened 2018-01-04
new setting proxy.skip_target_path_escaping in api def added

dencoded:
changes to fix https://github.com/TykTechnologies/tyk/issues/1374

as part of reverse proxying we [do escaping when we write actual request to upstream](https://golang.org/src/net/http/request.go?#L520) - we always do `req.URL.RequestURI()` which is doing escaping for path in request for upstream.

We don't have too much choice here. I've noticed that [RequestURI](https://golang.org/src/net/url/url.go?#L983) doesn't do any escaping if URL.Opaque has value, so I build a logic around it and introduced new `skip_target_path_escaping` flag which can be set in `proxy` section of API definition. It is a little bit hacky but I think it will work as URL struct is used here just in context of preparing URI for http request to be sent.

If we go with that approach we will need to document this param with note that customers should use it at their own risk (in case of some specific upstreams like SFCC) as all this RFCs about escaping were done on purpose.

### DIFF
--- a/reverse_proxy.go
+++ b/reverse_proxy.go
@@ -235,7 +235,6 @@ func TykNewSingleHostReverseProxy(target *url.URL, spec *APISpec) *ReverseProxy 
 			req.URL.Scheme = targetToUse.Scheme
 			req.URL.Host = targetToUse.Host
 			req.URL.Path = singleJoiningSlash(targetToUse.Path, req.URL.Path)
-			// force RequestURI to skip escaping if API's proxy is set for this
 		}
 		if !spec.Proxy.PreserveHostHeader {
 			req.Host = targetToUse.Host
@@ -252,7 +251,13 @@ func TykNewSingleHostReverseProxy(target *url.URL, spec *APISpec) *ReverseProxy 
 		}
 
 		if spec.Proxy.SkipTargetPathEscaping {
-			req.URL.Opaque = req.URL.RawPath // if we set opaque here it will force URL.RequestURI to skip escaping
+			// force RequestURI to skip escaping if API's proxy is set for this
+			// if we set opaque here it will force URL.RequestURI to skip escaping
+			if req.URL.RawPath != "" {
+				req.URL.Opaque = singleJoiningSlash(targetToUse.Path, req.URL.RawPath)
+			}
+		} else {
+			req.URL.RawPath = ""
 		}
 	}
 
--- a/reverse_proxy.go
+++ b/reverse_proxy.go
@@ -235,6 +235,10 @@ func TykNewSingleHostReverseProxy(target *url.URL, spec *APISpec) *ReverseProxy 
 			req.URL.Scheme = targetToUse.Scheme
 			req.URL.Host = targetToUse.Host
 			req.URL.Path = singleJoiningSlash(targetToUse.Path, req.URL.Path)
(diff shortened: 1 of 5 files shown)


buger:
@dencoded I added few new tests including failing one. I can't get it work like your tests, with `spec.Proxy.SkipTargetPathEscaping = false`, not sure why. 

Also, I changed Path to RawPath, because otherwise encoded urls get decoded.

reverse_proxy.go
```diff
@@ -253,6 +250,10 @@ func TykNewSingleHostReverseProxy(target *url.URL, spec *APISpec) *ReverseProxy
 			// this line, we would get the net/http default.
 			req.Header.Set("User-Agent", defaultUserAgent)
 		}
+
+		if spec.Proxy.SkipTargetPathEscaping {
+			req.URL.Opaque = req.URL.RawPath // if we set opaque here it will force URL.RequestURI to skip escaping
```
dencoded: I think this change is not gonna work as `req.URL.RawPath` contains at this point full initial request URI - listen path + the rest of the path

lonelycode:
Did you take a look at my branch @dencoded?

reverse_proxy.go
dencoded: another note - `req.URL.RawPath` is going to have query string (if any) `?abc=xyz` and `RequestURI()` is adding query string here to result of escape-call again: https://golang.org/src/net/url/url.go#L994
if I understand right - we will have get passed to upstream result of `req.URL.Opaque + "?" + u.RawQuery` with `?abc=xyz` doubled.

dencoded:
@lonelycode yes, I am/was looking at it. still need to figure out the idea behind it

reverse_proxy.go
buger: From what I understand RawPath do not include query string

dencoded: ok, lets scratch my comment about query string - it is not included to `RawPath`
I still see issue with setting `req.URL.Opaque = req.URL.RawPath` as it contains at this point initial request path (incoming to gateway) and leads to sending it to upstream as is.

the changes https://github.com/TykTechnologies/tyk/commit/a7a2a04107400cc7f1b60fce81b2ae945d11bd04 @lonelycode submitted should address this issue as we remove listen path, and replace in requests URL with object (result of parsing striped URI)

buger: Pls update your PR to use martin changes if you thin they will work.

dencoded: 👍

dencoded:
@buger I've fixed tests you added - there were a little typo in test data. also, I've added some logic to make use of `URL.RawPath`:
- strip listen path from `RawPath` when it is needed as we use its value to direct request to upstream
- some times `URL.RawPath` [can be empty](https://golang.org/src/net/url/url.go#L628) after parsing incoming requests' URL

plus added more tests for combinations of `strip_listen_path` and `skip_target_path_escaping `

so, things left:
- bring back test coverage (it is -0.5% now)
- add special case when one of Path segments has URL like `/abc/http://test.com`, we need somehow to analyze URL, split in segments with respect to possible URLs in them (split by `/` won't work) and apply to each segments [PathEscape](https://golang.org/src/net/url/url.go#L270)

Issue #1219, opened 2017-10-12
pass HTTP method to gRPC plugin request


amitmandke:
**Do you want to request a *feature* or report a *bug*?**
new feature
**What is the current behavior?**
HTTP method (GET, POST etc) is not available in gRPC plugin. 
**What is the expected behavior?**
Make HTTP method available in "MiniRequestObject" for gRPC plugin. Or any rich plugin for that matter.
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
NA (new feature)
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
NA (new feature)

Rouzz:
Hi, I came by the same issue.
Any info on when it will become available ?

My only hope so far is for each method of the API to add some kind of 'X-HTTP-method' header that will allow the middleware to process requests based on HTTP methods.
That's the only way I could think of in the meantime :)

buger:
@matiasinsaurralde so, it looks like we need extend our protobuffers file, and include this field.

buger:
Was implemented in 2.5

PR #1386, opened 2018-01-08
Add HTTP method to middleware request object

matiasinsaurralde:
This covers both coprocess & JSVM, related to #1219.


buger:
I see you added the following comment:
> There's a data structure that's based on this and it's used for Protocol Buffer support, make sure to update it on every change (coprocess/proto/coprocess_session_state.proto).

But what about `coprocess/coprocess_session_state.pb.go`, does it gets updated automatically? 

If so, what commands I should run to do that?

coprocess/bindings/python/coprocess_object_pb2.py
```diff
@@ -277,155 +277,110 @@
...
-    }
-    cardinalities = {
-      'Dispatch': cardinality.Cardinality.UNARY_UNARY,
-      'DispatchEvent': cardinality.Cardinality.UNARY_UNARY,
-    }
-    stub_options = beta_implementations.stub_options(host=host, metadata_transformer=metadata_transformer, request_serializers=request_serializers, response_deserializers=response_deserializers, thread_pool=pool, thread_pool_size=pool_size)
-    return beta_implementations.dynamic_stub(channel, 'coprocess.Dispatcher', cardinalities, options=stub_options)
-except ImportError:
-  pass
+import grpc
```
buger: This stuff is new, did you used some experimental protobuf version to build this?

matiasinsaurralde: The `Dispatcher` class looks mostly the same, the`BetaDispatcher` changed a bit, I will investigate this (shouldn't affect).

buger: Now it required `grpc` packages (previously it dont). 

I would say that it is better to run script with same protobuf version as we officially use now in CI.

matiasinsaurralde:
I've updated my Protocol Buffers setup to make sure that we're using the latest stable version, so the bindings have been generated using `grpcio-1.8.3` (installed via `pip`) and `protobuf-3.5.1`.
`protoc` returns the following:
```
$ protoc --version
libprotoc 3.5.1
```
I have also updated the instructions and the script that we use to generate the bindings to explicitly use Python 3 (previously it could mess things up if you ran the Python generator using "python" instead of "python3").
I will push the second PR that extends our definitions in a few minutes but would like to get this merged first so that we can rebase the other PR.

PR #1391, opened 2018-01-10
Add request URI to middleware request object

matiasinsaurralde:
I've added the field to the request object (and not the CP object) so that it looks like what's implemented in #1352 for JSVM.
Let's merge this after #1386.

Note: the field that's generated by the Go generator would look better as `RequestURI` instead of `RequestUri` but I prefer to keep the generator output as is. Feedback appreciated.


buger:
@matiasinsaurralde  HTTP method merged, ready to be rebased

matiasinsaurralde:
Thanks

matiasinsaurralde:
Updated!

buger:
Merged manually d1e10cfcb290318c617fed827d9c463eb0ae86af

PR #1395, opened 2018-01-11
whitelist for ciphersuites available for TLSconfig

### LINKED ISSUE #1372: Way to specify TLS Ciphers
feature
Right now only default set of ciphers is supported
Security standards vary across companies, and there should be a way to whitelist allowed ciphers
2.4, master

joshblakeley:
User can add values of TLS ciphersuites to array in config, the values of which are defined in the Go TLS lib https://golang.org/pkg/crypto/tls/#Config

Fixes #1372

### DIFF
--- a/cert.go
+++ b/cert.go
@@ -22,6 +22,31 @@ type APIAllCertificates struct {
 	CertIDs []string `json:"certs"`
 }
 
+var cipherSuites = map[string]uint16{
+	"TLS_RSA_WITH_RC4_128_SHA":                0x0005,
+	"TLS_RSA_WITH_3DES_EDE_CBC_SHA":           0x000a,
+	"TLS_RSA_WITH_AES_128_CBC_SHA":            0x002f,
+	"TLS_RSA_WITH_AES_256_CBC_SHA":            0x0035,
+	"TLS_RSA_WITH_AES_128_CBC_SHA256":         0x003c,
+	"TLS_RSA_WITH_AES_128_GCM_SHA256":         0x009c,
+	"TLS_RSA_WITH_AES_256_GCM_SHA384":         0x009d,
+	"TLS_ECDHE_ECDSA_WITH_RC4_128_SHA":        0xc007,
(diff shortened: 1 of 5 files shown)


buger:
Nice! Would love to see a small test. 

The basic idea here is that you specify some ciphersuites on server side, next specify incompatible ones on the http client side, you should get an error. 

If we talking in terms of test framework:

```
config.Global.HttpServerOptions.Ciphers = []string{.....}

ts := newTykTestServer()
defer ts.Close()

buildAndLoadAPI(function(s *APISpec){
   s.Proxy.ListenPath = "/"
})

t.Run("Cipher match", function(t *testing.T) {
  client := &http.Client{&http.Transport{TLSClientConfig: &tls.Config{
     CipherSuites: getCipherAliases([]string{....}} // matching ciphers
  }}}

  // If there is internal TSL error it will fail test
  ts.Run(t, test.TestCase{Path: "/"})
})

t.Run("Cipher not match", function(t *testing.T) {
  client := &http.Client{&http.Transport{TLSClientConfig: &tls.Config{
     CipherSuites: getCipherAliases([]string{....}} // not matching ciphers
  }}}

  ts.Run(t, test.TestCase{Path: "/", ErrorMatch: "tls: some internal error"})
})
```

joshblakeley:
Thanks @buger will see where I get

Have slightly altered some imports in main  too as goimports was being annoying and removing them on save.

buger:
Nice job! I made small refactoring, removing unnecessary logic.

buger:
@joshblakeley the final thing left is to prepare documentation update for this page https://tyk.io/docs/security/tls-and-ssl/

Make a separate PR to tyk-docs repo

joshblakeley:
Will do!

Thanks, i'm very happy with what i've done today.

buger:
Note that I also renamed config variable from `ciphers` to `ssl_ciphers`, to reflect what it is

joshblakeley:
Makes sense, thanks.

joshblakeley:
You killed the linter, i'll fix

PR #1396, opened 2018-01-12
hello endpoint not dependent on there being more than 0 apis at load

### LINKED ISSUE #1393: Health-check (/hello) 404 if no APIs defined
Bug

the `/hello` endpoint does not work if there are no APIs defined

The `/hello` endpoint should be available when the application has successfully started listening

1. Start an instance without any APIs defined
2. Visit `/hello` on the gateway
3. You will get a 404

2.4

joshblakeley:
Really dumb fix for issue where hello endpoint is not loaded in gateway when there are no api's present.

fixes #1393

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -634,6 +634,28 @@ func TestCustomDomain(t *testing.T) {
 	})
 }
 
+func TestHelloHealthcheck(t *testing.T) {
+	ts := newTykTestServer()
+	defer ts.Close()
+
+	t.Run("Without APIs", func(t *testing.T) {
+		ts.Run(t, []test.TestCase{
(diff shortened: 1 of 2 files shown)


buger:
Can you link and describe original issue?

joshblakeley:
See root comment

buger:
I do not think your fix is enough, because `mainRouter` gets overridden on each `doReload`

joshblakeley:
If doReload is called then it falls back to how it behaved before and the endpoint is reinitialised in api_loader I think

joshblakeley:
I just tested and it is always there after adding api and reloading too. So I think this covers the fix.

Issue #1402, opened 2018-01-15
404 errors when accessing API during reload 


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
If on high load you trigger API reload, you may see 404 errors while accessing the api
**What is the expected behavior?**
Reload expected to be "hot"
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
On 100 RPS it becomes noticable
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
Starting from 2.6

PR #1403, opened 2018-01-15
Fix 404 during reloads

buger:
Issue was happening on high load, because mainRouter object was
initialized with the fresh router, before API routes were loaded into it.

Additionally added more mutexes.

Fix https://github.com/TykTechnologies/tyk/issues/1402

### DIFF
--- a/main.go
+++ b/main.go
@@ -126,6 +126,9 @@ func apisByIDLen() int {
 
 // Create all globals and init connection handlers
 func setupGlobals() {
+	reloadMu.Lock()
+	defer reloadMu.Unlock()
+
 	mainRouter = mux.NewRouter()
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -656,21 +659,23 @@ func doReload() {
...
 
 	if config.Global.ControlAPIPort == 0 {
-		loadAPIEndpoints(mainRouter)
+		loadAPIEndpoints(newRouter)
 	}
 
-	loadGlobalApps()
+	loadGlobalApps(newRouter)
 
 	log.WithFields(logrus.Fields{
```
dencoded: it is maybe too early to say in log that API reload complete (until we don't swap `mainRouter` with new value)

PR #1406, opened 2018-01-17
Add DNS ttl tests

buger:
I'm not sure that it should be merged because this test adds 2 seconds to test suite. 
Initially, it was made to replicate reported issue with Tyk not picking up new DNS changes.


buger:
Not needed anymore, confirmed that Go handle DNS as expected

Issue #1413, opened 2018-01-22
Fix `close_connection`


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
Seems like this functionality stopped working at some point
**What is the expected behavior?**
It should disable keep alives, and should start new connection with each request
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
It should be possible to write test server which verifies this functionality, by creating simple pure TCP server with full listen/accept loop. 
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.3 and 2.4

PR #1419, opened 2018-01-23
create config option and test for disablekeepalives

joshblakeley:
Makes `close_connections` option explicitly disable keep-alive connections on network level, instead of relying on server. 

Fix https://github.com/TykTechnologies/tyk/issues/1413


joshblakeley:
@buger me and @asoorm struggling to think through this one

buger:
Fix the tests, seems to be working. 
Also introducing new config variable not needed, since we already have `CloseConnections`.

Issue #1425, opened 2018-01-24
Gateway panics on advanced url rewrite for payload and path part match type


ilijabojanovic:
**Do you want to request a *feature* or report a *bug*?**
BUG
**What is the current behavior?**
Gateway panics for payload and path path when we set full url in rewtite path
**What is the expected behavior?**
No panics in log
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. Import api def
2. Trigger call to /rewrite/get?show_env=1
3. Observe gateway log

Panics log:
```
tyk-gateway_1    | time="Jan 24 12:53:35" level=debug msg="/rewrite/get?show_env=1"
tyk-gateway_1    | time="Jan 24 12:53:35" level=debug msg="Inbound path: /rewrite/get?show_env=1"
tyk-gateway_1    | time="Jan 24 12:53:35" level=debug msg="[STORE] SET Raw key is: 5a68526475798c0312a8468d65dc531ae9f4431092108f6e89d561b4"
tyk-gateway_1    | time="Jan 24 12:53:35" level=debug msg="[STORE] Setting key: apikey-9f3ca68b"
tyk-gateway_1    | 2018/01/24 12:53:35 http: panic serving 172.18.0.1:55398: runtime error: invalid memory address or nil pointer dereference
tyk-gateway_1    | goroutine 202331 [running]:
tyk-gateway_1    | net/http.(*conn).serve.func1(0xc420eb7040)
tyk-gateway_1    | 	/usr/local/go/src/net/http/server.go:1697 +0xd0
tyk-gateway_1    | panic(0xd3b5a0, 0x1647550)
tyk-gateway_1    | 	/usr/local/go/src/runtime/panic.go:491 +0x283
tyk-gateway_1    | regexp.(*Regexp).get(0x0, 0xc42001c000)
tyk-gateway_1    | 	/usr/local/go/src/regexp/regexp.go:211 +0x26
tyk-gateway_1    | regexp.(*Regexp).doExecute(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x2, ...)
tyk-gateway_1    | 	/usr/local/go/src/regexp/exec.go:420 +0x40
tyk-gateway_1    | regexp.(*Regexp).FindString(0x0, 0x0, 0x0, 0x0, 0x0)
tyk-gateway_1    | 	/usr/local/go/src/regexp/regexp.go:750 +0xd0
[...]

PR #1426, opened 2018-01-24
changed InitTriggerRx to traverse/change structure elemt with using  …

dencoded:
fix for https://github.com/TykTechnologies/tyk/issues/1425

with `for _, tr := range path.Triggers ..` and doing write-operations with elements on `tr` - the actual `m.Spec.VersionData.Versions` are never changed as `tr` receives copy of `Triggers` element.

everything which are maps (or already set pointers if any) were working as maps are passed/copied by reference

the change is ugly but I think thats the only way to guarantee that we walk through all structure and change things in place (not copies of things)

### DIFF
--- a/mw_url_rewrite_test.go
+++ b/mw_url_rewrite_test.go
@@ -512,66 +512,98 @@ func TestRewriterTriggers(t *testing.T) {
 }
 
 func TestInitTriggerRx(t *testing.T) {
+	// prepare test data
 	testRewriteMW := &URLRewriteMiddleware{
 		BaseMiddleware: BaseMiddleware{
 			Spec: &APISpec{
-				APIDefinition: &apidef.APIDefinition{
-					VersionData: struct {
-						NotVersioned   bool                          `bson:"not_versioned" json:"not_versioned"`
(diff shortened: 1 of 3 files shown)


mw_url_rewrite_test.go
```diff
@@ -510,3 +510,68 @@ func TestRewriterTriggers(t *testing.T) {
...
+						DefaultVersion: "Default",
+						Versions: map[string]apidef.VersionInfo{
+							"Default": {
+								ExtendedPaths: apidef.ExtendedPathsSet{
+									URLRewrite: []apidef.URLRewriteMeta{
+										{
+											Triggers: []apidef.RoutingTrigger{
+												{
+													Options: apidef.RoutingTriggerOptions{
+														HeaderMatches: map[string]apidef.StringRegexMap{
```
buger: I like indentation level

asoorm: anything we can do about this??

dencoded: I got too excited with my 4K monitor. I will try to reformat it a little bit

PR #1437, opened 2018-01-28
Adds Upstart version detection and separate conf, fixes sysv script, compatible service stopping

excieve:
This mends init scripts incompatibility with Upstart 0.x (e.g. 0.6.5 as found on CentOS6/RHEL6/Amazon Linux, etc.)


excieve:
This is for 2.5 branch only as it diverges from master a bit on the packaging stuff. Will do the same for master a bit later, along with ownership changes.

PR #1443, opened 2018-01-29
Fix StatsD instrumentation

### LINKED ISSUE #1442: StatsD Instrumentation not working
Bug

**What is the current/expected behavior?**
StatsD instrumentation doesn't work, the gateway isn't able to send data to a StatsD server.

To reproduce, setup a [StatsD server](https://github.com/etsy/statsd) and run Tyk with the following settings:
- Set environment variable `TYK_INSTRUMENTATION` to `1`.
- Set a prefix `TYK_GW_STATSDPREFIX`, e.g. `myprefix.tykGWTest`.
- Set `statsd_connection_string` in `tyk.conf` using the address of your StatsD server.

Broken in `master`, `2.4` and newer versions

matiasinsaurralde:
Fixes #1442.

`2.3` used the following code:
https://github.com/TykTechnologies/tyk/blob/release-2.3/instrumentation_statsd_sink.go#L93

### DIFF
--- a/instrumentation_statsd_sink.go
+++ b/instrumentation_statsd_sink.go
@@ -90,7 +90,7 @@ const cmdChanBuffSize = 8192 // random-ass-guess
 const maxUdpBytes = 1440     // 1500(Ethernet MTU) - 60(Max UDP header size
 
 func NewStatsDSink(addr string, options *StatsDSinkOptions) (*StatsDSink, error) {
-	c, err := net.ListenPacket("udp", "127.0.0.1:0")
+	c, err := net.ListenPacket("udp", ":0")
 	if err != nil {
 		return nil, err
 	}


excieve:
I wonder if `0.0.0.0:0` would do the same with IPv4?

matiasinsaurralde:
@excieve right, should we change it? Do we support IPv6? 😄

excieve:
I just recalled why everything has been forced to IPv4: IPv6 being disabled in docker in certain environments. Although, since it works on dashboard and pump I suppose it's fine to keep IPv6 — perhaps those issues have been resolved by now.

excieve:
Oh, and I believe Go 1.9 can dynamically choose v4 when v6 isn't available. Since we're compiling this for 1.9 now it should be fine.

Issue #492, opened 2017-02-09
Enable the version to be stripped from the listen path if it is part of the URL


lonelycode:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
If I version an API and store the data in the url (e.g. `listen_path/v1/this/resource`, then only the `listen_path` portion is stripped.

**What is the expected behavior?**
I should be able to strip the version information from the path as well so that `listen_path/v1` is stripped as a block. 

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
All

asoorm:
- [ ] Gateway Update https://github.com/TykTechnologies/tyk/pull/1542
- [ ] Dashboard API Update
- [ ] Dashboard UI Update
- [ ] Docs

PR #1449, opened 2018-02-01
[WIP] feature: strip path version from proxy upstream

asoorm:
resolves: #492 

https://community.tyk.io/t/api-routing-via-header-option-and-catch-all-routing/2365/2


buger:
Do not forget link ticket in description.

buger:
We need to make this logic conditional to make it backward compatible.

Issue #1438, opened 2018-01-29
Gateway should not start proxying traffic until connectivity established with Redis


asoorm:
The gateway starts proxying traffic, before it has established a connection with Redis.
The gateway should die if it cannot make initial connection with Redis, or at least not start listening until all connections have been fully established and everything is ready.

In the situation where gateway is not able to establish initial connection with Redis - e.g. wrong password or configuration or network problem - then it should never start proxying traffic.

PR #1451, opened 2018-02-02
test redis connection before start serving traffic

dencoded:
added change for https://github.com/TykTechnologies/tyk/issues/1438


storage/redis_cluster.go
```diff
@@ -28,6 +33,55 @@ type RedisCluster struct {
 	IsCache   bool
 }
 
+func clusterConnectionIsOpen(cluster *rediscluster.RedisCluster) bool {
+	// check if at least one of cluster handles has an alive connection
+	for item := range cluster.Handles.Iter() {
+		conn := item.Val.(*rediscluster.RedisHandle).GetRedisConn()
+		isOpen := conn.Err() == nil
+		conn.Close()
```
asoorm: not really sure how this works... but if `conn.Err != nil`, can you safely call `conn.Close`?

dencoded: @asoorm it comes to https://github.com/TykTechnologies/tyk/blob/master/vendor/github.com/lonelycode/redigocluster/rediscluster/redishandle.go#L58
which calls https://github.com/TykTechnologies/tyk/blob/master/vendor/github.com/garyburd/redigo/redis/pool.go#L157 
as they say:
```
// Get gets a connection. The application must close the returned connection.
// This method always returns a valid connection so that applications can defer
// error handling to the first use of the connection. If there is an error
// getting an underlying connection, then the connection Err, Do, Send, Flush
// and Receive methods return that error.
```
so we close it

buger:
Are you sure that you not overcomplicating it?
But even if it is not, why not just call redis Get/Set to some dummy key, and expect to get answer in 1 second or similar?

dencoded:
@buger unfortunately, our `RedisCluster.Connect()` always returns `true`, nevertheless Redis is alive or not, so that made me thinking that we need to go deeper to `rediscluster.RedisCluster` method calls. if you think it is too over complicated, we can change https://github.com/TykTechnologies/tyk/pull/1451/files#diff-0a5b7f37e1eee3672b24bd43a12acee6R50 to use set/get dummy key, I think it should work as `SetKey` and `GetKey` calls will return error

dencoded:
I mean they `SetKey` and `GetKey` should return error (I see some reconnection logic with infinite loop). if they don't we will have to fall back to `rediscluster.RedisCluster` API calls

buger:
Let's keep it like this, but update retry interval to 1 second, with 5 attempts. 

Also cache cluster is an optional, it fallback to default redis. Ensure that this case is handled.

dencoded:
@buger  I was about to change it to use set/get key as you suggested as I realized that `ensureConnection()` is actually not ensuring connection for real and won't block `SetKey` operation.
lets maybe change it so `func IsConnected() bool` will operate only in terms of `type RedisCluster struct` methods?

ok, I will change retries and cache cluster case

buger:
pls fix the CI (gofmt)

buger:
Actually found an issue, it was using wrong uuid library, we use another one in code. Test was failing because this lib was not vendored. Made a fix fa76911

dencoded:
👍 
my bad (autocomplete screwed stuff)

Issue #1435, opened 2018-01-26
Make JSVM middleware "timeout" configurable


bitsofinfo:
https://github.com/TykTechnologies/tyk/blob/3f1fd288633611619311449559a8ce83ff11e9f2/mw_js_plugin.go#L340

Currently is hardwired for 5s

Have needs in development where things are slower to bump this number up

buger:
@LuanvP add a new configuration option to tyk.conf with a sensible name and make sure that it will default to 5 if not set.

LuanvP:
Issues while testing for this was due to this issue https://github.com/TykTechnologies/tyk-analytics/issues/696

VirtualEndpoint middleware was using same JSVM object but not checking the timeout value, unlike DynamicMiddleware.

bitsofinfo:
Verified this works, thanks!

PR #1470, opened 2018-02-12
Configurable jsvm timeout

LuanvP:
Addresses https://github.com/TykTechnologies/tyk/issues/1435

Needs test

Circle CI failing

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -247,7 +247,7 @@ type Config struct {
 	ControlAPIPort                    int                                   `json:"control_api_port"`
 	EnableCustomDomains               bool                                  `json:"enable_custom_domains"`
 	EnableJSVM                        bool                                  `json:"enable_jsvm"`
-	JSVMTimeout						  int									`json:"jsvm_timeout"`
+	JSVMTimeout                       int                                   `json:"jsvm_timeout"`
 	CoProcessOptions                  CoProcessConfig                       `json:"coprocess_options"`
 	HideGeneratorHeader               bool                                  `json:"hide_generator_header"`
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -247,6 +247,7 @@ type Config struct {
 	ControlAPIPort                    int                                   `json:"control_api_port"`
 	EnableCustomDomains               bool                                  `json:"enable_custom_domains"`
 	EnableJSVM                        bool                                  `json:"enable_jsvm"`
+	JSVMTimeout						  int									`json:"jsvm_timeout"`
```
dencoded: how about json-tag for this field? we need some name declared so it will be easier to document. also, formatting - make sure you use `go fmt`

mw_js_plugin.go
```diff
@@ -337,7 +337,12 @@ func (j *JSVM) Init(spec *APISpec) {
 	// Add environment API
 	j.LoadTykJSApi()
 
-	j.Timeout = 5 * time.Second
+	if config.Global.JSVMTimeout <= 0 {
+		j.Timeout = 5 * time.Second
```
dencoded: I'd add default timeout as a constant to avoid adding more "magic" numbers in code

config/config.go
dencoded: sorry, just realized that I had to scroll a lot to the right to see `json:"jsvm_timeout"` :) is it formatting issue?

mw_js_plugin.go
matiasinsaurralde: @dencoded Agree!

config/config.go
buger: @LuanvP you should apply gofmt to fix formatting like this: `gofmt -s -w config/.*`

PR #1471, opened 2018-02-13
batch requests client checks for ssl config and add test

### LINKED ISSUE #1456: TykBatchRequest should obey proxy_ssl_insecure_skip_verify

feature/bug

`TykBatchRequest` can't make calls to self-signed CA or mis-matched hostname backends

```
[JSVM]: Request failed: Get https://mybackend.mydomain.com/whatever: x509: certificate is valid for xxxxx, not mybackend.mydomain.com"
```

There should be an option/param/flag for `TykBatchRequest` OR it should just obey what the gateway's `proxy_ssl_insecure_skip_verify` setting says

2.5.x

@see https://github.com/TykTechnologies/tyk/issues/1006

joshblakeley:
Should fix #1456 

Not sure if using a new client for each request is good practice? But should be ok as we close each request body.

### DIFF
--- a/batch_requests.go
+++ b/batch_requests.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -40,7 +41,16 @@ type BatchRequestHandler struct {
 
 // doRequest will make the same request but return a BatchReplyUnit
 func (b *BatchRequestHandler) doRequest(req *http.Request, relURL string) BatchReplyUnit {
-	resp, err := http.DefaultClient.Do(req)
+	tr := &http.Transport{TLSClientConfig: &tls.Config{}}
+
+	if cert := getUpstreamCertificate(req.Host, b.API); cert != nil {
+		tr.TLSClientConfig.Certificates = []tls.Certificate{*cert}
+	}
+
+	tr.TLSClientConfig.InsecureSkipVerify = config.Global.ProxySSLInsecureSkipVerify
+	client := &http.Client{Transport: tr}
+
+	resp, err := client.Do(req)
 	if err != nil {
 		log.Error("Webhook request failed: ", err)
 		return BatchReplyUnit{}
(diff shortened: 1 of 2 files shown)


joshblakeley:
Any improvements on this @buger ?

batch_requests.go
```diff
@@ -40,7 +41,18 @@ type BatchRequestHandler struct {
 
 // doRequest will make the same request but return a BatchReplyUnit
 func (b *BatchRequestHandler) doRequest(req *http.Request, relURL string) BatchReplyUnit {
-	resp, err := http.DefaultClient.Do(req)
+	tr := &http.Transport{TLSClientConfig: &tls.Config{}}
+	if cert := getUpstreamCertificate(req.Host, b.API); cert != nil {
+		tr.TLSClientConfig.Certificates = []tls.Certificate{*cert}
+	}
+
+	if config.Global.ProxySSLInsecureSkipVerify {
```
buger: Can be replaced by `tr.TLSClientConfig.InsecureSkipVerify = config.Global.ProxySSLInsecureSkipVerify`

buger:
This is not really a fair way to test things like this :) 

The issue initially tells about JSVM api calls. JSVM implementation use `ManualBatchRequest` function instead of `BatchRequest`, and main idea here that JSVM batch function works with absolute urls, while our internal batch API works with relative URLs. 

So the valid way to test this feature would be creating test similar to `TestVirtualEndpoint`, but in this case, you create test https server using `https://golang.org/pkg/net/http/httptest/#Server.StartTLS` and inside virtual endpoint call this endpoint using `TykBatchRequest`. 

Also, note that good test should test both success and failure. So you first run it without with `config.Global.ProxySSLInsecureSkipVerify` false, and next enable it and it should pass. 

Note that functionality itself looks nice, and for sure this maybe function not that big to require full blown test but treat it as an exercise :)

joshblakeley:
Test improved and ready for review @buger

buger:
The test looks good! But you can remove SSL configuration of tyk itself, since it is not used here at all, it will help simplify test a lot.

buger:
@joshblakeley I simplified test as suggested check the diff https://github.com/TykTechnologies/tyk/pull/1471/commits/05b5649760513c80830271bb8bf46cb6990c58cd

Sorry for intervention, just want get it included into patch release

Issue #1460, opened 2018-02-07
Log warning if gateway node_secret is missing


LuanvP:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
There is currently no warning, and the dashboard connection will fail with 403

**What is the expected behavior?**
Add a log warning if the field is blank or missing

ilijabojanovic:
Verified

PR #1473, opened 2018-02-13
Add warning when node_secret is missing (#1460)

matiasinsaurralde:
As described in #1460.


dashboard_register.go
```diff
@@ -74,8 +74,11 @@ func (h *HTTPDashboardHandler) Init() error {
 	h.RegistrationEndpoint = buildConnStr("/register/node")
 	h.DeRegistrationEndpoint = buildConnStr("/system/node")
 	h.HeartBeatEndpoint = buildConnStr("/register/ping")
-
-	h.Secret = config.Global.NodeSecret
+	if h.Secret = config.Global.NodeSecret; h.Secret == "" {
+		log.WithFields(logrus.Fields{
+			"prefix": "main",
+		}).Warning("Node secret is not set, required for dashboard connection")
```
buger: I think it should be Fatal instead

matiasinsaurralde:
@buger Changed to `Fatal`

PR #1474, opened 2018-02-13
Add a mutex to org-channel map in off-process org monitor

excieve:
A concurrent read and write (or just concurrent writes) panic with the map between org IDs and channels for monitoring quota for the org can happen when several requests to APIs belonging to the same organisation occur at the same time and there's no mapping yet. This is pretty easy to reproduce by launching a lot of simultaneus requests right before the gateway starts (e.g. with ab tool).

Since this writes to the map just once per org RWMutex should provide better performance by separating read and write locks.

### DIFF
--- a/mw_organisation_activity.go
+++ b/mw_organisation_activity.go
@@ -9,8 +9,12 @@ import (
 	"github.com/TykTechnologies/tyk/config"
 )
 
-var orgChanMap = make(map[string]chan bool)
+type orgChanMapMu struct {
+	sync.Mutex
+	channels map[string](chan bool)
+}
 
+var orgChanMap = orgChanMapMu{channels: map[string](chan bool){}}
 var orgActiveMap sync.Map
 
 // RateLimitAndQuotaCheck will check the incomming request and key whether it is within it's quota and
@@ -104,13 +108,14 @@ func (k *OrganizationMonitor) SetOrgSentinel(orgChan chan bool, orgId string) {
 }
 
 func (k *OrganizationMonitor) ProcessRequestOffThread(r *http.Request) (error, int) {
-
-	orgChan, ok := orgChanMap[k.Spec.OrgID]
+	orgChanMap.Lock()
+	orgChan, ok := orgChanMap.channels[k.Spec.OrgID]
 	if !ok {
-		orgChanMap[k.Spec.OrgID] = make(chan bool)
-		orgChan = orgChanMap[k.Spec.OrgID]
+		orgChan = make(chan bool)
+		orgChanMap.channels[k.Spec.OrgID] = orgChan
 		go k.SetOrgSentinel(orgChan, k.Spec.OrgID)
 	}
+	orgChanMap.Unlock()
 	active, found := orgActiveMap.Load(k.Spec.OrgID)
 
 	go k.AllowAccessNext(orgChan, r.URL.Path, requestIP(r), r)
--- a/install/data/tyk.with_dash.conf
+++ b/install/data/tyk.with_dash.conf
@@ -60,7 +60,7 @@
   "close_connections": true,
   "enable_non_transactional_rate_limiter": true,
   "enable_sentinel_rate_limiter": false,
-  "experimental_process_org_off_thread": true,
+  "experimental_process_org_off_thread": false,
   "local_session_cache": {
     "disable_cached_session_state": false
   },


mw_organisation_activity.go
```diff
@@ -9,8 +9,12 @@ import (
 	"github.com/TykTechnologies/tyk/config"
 )
 
-var orgChanMap = make(map[string]chan bool)
+type orgChanMapMu struct {
+	sync.RWMutex
+	channels map[string](chan bool)
+}
 
+var orgChanMap = orgChanMapMu{channels: map[string](chan bool){}}
```
excieve: As a note on mutex vs `sync.Map`, although in this case the map is append-only and hence kind of works of the latter, [some benchmarks show](https://medium.com/@deckarep/the-new-kid-in-town-gos-sync-map-de24a6bf7c2c) that it only performs better during very high read contention over 4-8+ real cores. In other cases it significantly looses to the mutex-based maps. At the same time `sync.Map` looses type-safety so I decided against this premature optimisation and I'm not sure it fits `orgActiveMap` either (it's not at all append-only as the sentinel goroutine writes at the same keys all the time).

dencoded: we had a conversation about `sync.Map` a while ago, it was added for good reason but breaks backward comparability with older versions of Go.

according to change: if we wrap mutex and protected map into struct - I think it is a good chance to add to this struct `Get` and `Set` methods (with all lock/defers inside them) to simplify code and prevent (at least logically) from using mutexes/maps directly

excieve:
I just realised there can be another bug: imagine the gateway has just started and there are two simultaneous requests to the same API (or different APIs of the same org) — `A` and `B`:
 1. Both `A` and `B` acquire a read lock and pass the critical section with `ok=false` because the channel doesn't exist yet
 2. Both reach the `!ok` block and e.g. `A` acquires the write lock, while `B` is blocked
 3. `A` creates the channel and writes to the map, launches the sentinel goroutine, then releases the lock
 4. `B` unknowing that the channel (map record and sentinel) already exist does (3) again

Result: there are two channels and sentinels, while only the one created by `B` would be eventually used for reading.

The impact of this doesn't sound significant but still creates an "orphan" goroutine that listens on channel that is written to just once (by `A`). This means we either need the whole read-then-write block to be locked for r/w (which means readers will be blocked on every request) or add another map key existence check inside the write block. I need some advice on this.

mw_organisation_activity.go
excieve: Thanks for the suggestion but I think this needs explicit locking control (see another comment on the PR).

buger:
Looks good, but I would probably simplify it a bit, RW mutex not really used here, function is non CPU sensitive, and replacing with the example below, on top `ProcessRequestOffThread` will do the same:
```
orgChanMap.Lock()
defer orgChanMap.Unlock()
```

buger:
@excieve do you want to implement changes mentioned in your last comment in terms of this PR, or it will wait for another one?

excieve:
@buger Will do in this one.

The thing I'm worried about with a simple r/w mutex is that middlewares are executed on the request goroutine and this particular middleware runs on pretty much any request. While I understand there's nothing computationally heavy going on there (especially in this small block), it means a request to any API will always block all the other requests to any other APIs, which can harm throughput under high load causing contention.

That said, if it's not really a concern for this block or there's a way to mitigate it (e.g. by disabling this middleware with an option — I'm not sure it's really needed on on-prem for instance) I'm happy to go with a simpler solution.

excieve:
In fact, I just recalled there is an option to enable/disable off-thread org processing so that should cover it in case of performance issues with a simple mutex. I'll update this PR accordingly. Also I think we should keep it disabled it by default in our config templates as by default on-prem installations don't even have org sessions and on-thread implementation proceeds immediately in this case unlike the off-thread so I'll make this change too if there are no objections.

excieve:
@buger Updated.

buger:
So, I'm right that this option has an extremely confusing name, and in fact should be named smth "enable_org_sessions"?

Asking that if you going switch it to false by default, we need to ensure that documentation explicitly says that in multi-org environment it should be turned on, right?

excieve:
There are several confusing things. The option itself switches the
on-thread and off-thread org processing. Both update the org session state
if it exists. Also on-thread returns earlier if there is no org session. It
so happens that only the off-thread actually checks the org quota though.
That's why it needs to be enabled on the cloud. I'm not sure why is this
and this needs a closer look in the context of quotas but both
implementations update the session.

19 лют. 2018 р. 18:19 "Leonid Bugaev" <notifications@github.com> пише:

buger:
What if we remove this option at all, and enable off-thread processing by default?

buger:
@excieve bump to my latest comment.

excieve:
@buger In this case I'd argue we should exit from the middleware soon as on-thread one does to avoid unnecessary locking and goroutines. Currently off-thread MW launches a bunch of goroutines even if org has no session, then checks for this in one of these goroutines. So it doesn't avoid all this even if there's no session.

Also differences in code paths between the two implementations need to be properly looked at. E.g. off-thread one doesn't go through the rate limiter, instead directly checking the quota, while the on-thread goes through the rate limiter, ignores its messages, listens for the quota message but actually disables quota check for some reason. Then, there's a separate MW that actually does the rate limiting so there's no real reason to go through it here.

So off-thread one seems to be the way to go indeed but needs some work to make sure we don't cause potential contention without a reason.

buger:
I just realized that real reason behind off-thread calculation is our hybrid layer, since this is where we heavily use org level constraints. Organization monitor uses `orgStorage` storage for dealing with quotas, and in case of RPC layer, this is direct communication with our MDCB instance: which can be quite slow. 

@dencoded pls note that when you implement org level rate limiting, it should not use `orgStorage` and RPC layer, it should be redis layer.

buger:
Will merge it now, since its getting old. Raise new PR if you want additional changes.

PR #1477, opened 2018-02-14
compile tracked endpoints logic change to set empty path

dencoded:
added fix for https://github.com/TykTechnologies/tyk-analytics/issues/574

I think we have to set it to `/` as allowing it to be stored in context as empty string (to avoid [current panic](https://github.com/TykTechnologies/tyk/blob/v2.5.1/api.go#L1443)) would confuse our tracked endpoint logic in places:

https://github.com/TykTechnologies/tyk/blob/v2.5.1/handler_success.go#L142

and

https://github.com/TykTechnologies/tyk/blob/master/handler_error.go#L133

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -693,6 +693,13 @@ func (a APIDefinitionLoader) compileTrackedEndpointPathspathSpec(paths []apidef.
 	for _, stringSpec := range paths {
 		newSpec := URLSpec{}
 		a.generateRegex(stringSpec.Path, &newSpec, stat)
+
+		// set Path if it wasn't set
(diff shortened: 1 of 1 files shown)


buger:
So, just to clarify, have you verified that "Track endpoint" functionality works as expected?

matiasinsaurralde:
I've tried this yesterday and can confirm what's described in the analytics issue.

dencoded:
@buger @matiasinsaurralde while doing testing/debugging I confirmed that in case of empty path for tracked endpoint:

- it doesn't panic anymore
- we send analytics record with `Path: "/"` to storage

PR #1479, opened 2018-02-15
Improve RPC stability in case of master failures

buger:
There are a lot of small changes here:

Previously RPC layer recovery mechanism had own reloading functionality, conflicting with existing one, and we constantly forget to sync it with all the changes. Now it uses standard `doReload` and `syncApis`/`syncPolicies` functions know how to fetch backups if RPC layer is down. Worth noticing that it required refactoring how login functionality works, previously it was fully blocking call. So if I call `doReload` first time, it tries to connect to RPC layer, and if fails, `doReload` is stuck, and we can't even call it one more time, because of the mutexes used inside it. Now `Login` functionality is `async` and if Login failed, it will return error, `syncApis`/`syncPolicies` functions will return nil, and user will see normal reload with 0 apis. After it will trigger goroutine with reconnect functionality, which will load backups and call `doReload` one more time, and continue trying to login. When `Login` succeed, it will call `doReload` one more time, now trying to fetch latest API changes. 

When RPC layer is down, before purging analytics it will now do `Ping` first and only after success will call `GetSetAndDelete`, to get analytics from Redis and remove key. Previously if RPC layer is gone, analytics was discarded on the first call, now there is a `chance` for full recovery since analytic record does not get deleted.

Added Policy backup functionality, so now you can use OpenID, JWT or oAuth apis which depend on policies to create keys in local redis, even if RPC layer is down.

Additionally added protection to avoid infinite growths of analytics data if Pump or RPC layers are down. It is implemented by setting expiration to redis key which contains analytic records, and controlled using new attribute `analytics_config.storage_expiration_time`. The default value is 1 minute. If analytics key was not processed and re-created in given time, it will just expire, and data will be lost. 

Added tests for RPC layer failover and backups.

Tests to do:

### RPC failover
1) Start Hybrid with working RPC layer, so it will do backups
2) Stop RPC layer (or block it on firewall, or change RPC connection string to non-existent addr)
3) Start Hybrid test that it still works
4) Re-enable RPC, it should reconnect and fetch new changes

### Policies backup
Test that you can use oAuth or JWT apis without RPC layer

### Analytics overflow
Disable RPC layer, or Pump in case of on-premise, send lot of traffic to Tyk (maybe with detailed request data enabled), ensure that redis memory not grows infinite, and gets cut according to `analytics_config.storage_expiration_time` option

### DIFF
--- a/rpc_backup_handlers.go
+++ b/rpc_backup_handlers.go
@@ -6,18 +6,18 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"io"
-	"net/http"
 	"strings"
 
 	"github.com/Sirupsen/logrus"
-	"github.com/gorilla/mux"
 
 	"github.com/TykTechnologies/tyk/config"
 	"github.com/TykTechnologies/tyk/storage"
+	"github.com/TykTechnologies/tyk/user"
 )
 
 const RPCKeyPrefix = "rpc:"
-const BackupKeyBase = "node-definition-backup:"
+const BackupApiKeyBase = "node-definition-backup:"
+const BackupPolicyKeyBase = "node-policy-backup:"
 
 func getTagListAsString() string {
 	tagList := ""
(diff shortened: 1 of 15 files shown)


rpc_backup_handlers.go
```diff
@@ -46,91 +72,67 @@ func saveRPCDefinitionsBackup(list string) {
...
-	a := APIDefinitionLoader{}
-	return a.processRPCDefinitions(apiListAsString)
-}
-
-func doLoadWithBackup(specs []*APISpec) {
-
-	log.Warning("[RPC Backup] --> Load Policies too!")
-
-	if len(specs) == 0 {
+	if policies, err := parsePoliciesFromRPC(listAsString); err != nil {
```
matiasinsaurralde: We can simplify this:
```
	if policies, err := parsePoliciesFromRPC(listAsString); err != nil {
		log.WithFields(logrus.Fields{
			"prefix": "policy",
		}).Error("Failed decode: ", err)
		return nil
	} else {
		return policies
	}
```
And use:
```
	if policies, err := parsePoliciesFromRPC(listAsString); err != nil {
		log.WithFields(logrus.Fields{
			"prefix": "policy",
		}).Error("Failed decode: ", err)
		return nil
	}
	return policies
```

ilijabojanovic:
@buger approved for merge

PR #1480, opened 2018-02-19
Update tyk.conf.example

LuanvP:
Remove outdated mongo fields


buger:
Do not forget about https://github.com/TykTechnologies/tyk/tree/master/install/data

PR #1481, opened 2018-02-19
bug: cpuProfile & memProfile dont populate

asoorm:
Current scenario, `start()` is called, but immediately returns. As such, `defer
memProfFile.Close()` and `defer pprof.StopCPUProfile()` get called too
quickly.

This change moves the profiling logic just after `start()`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1091,6 +1091,28 @@ func main() {
 
 	start()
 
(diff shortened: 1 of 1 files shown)


lonelycode:
LGTM

PR #1483, opened 2018-02-20
Update tyk.conf.example

LuanvP:
Update tyk.conf.example based on https://github.com/TykTechnologies/tyk/blob/master/install/data/tyk.self_contained.conf

### DIFF
--- a/install/data/tyk.with_dash.conf
+++ b/install/data/tyk.with_dash.conf
@@ -1,91 +1,226 @@
 {
-  "listen_port": LISTEN_PORT,
-  "node_secret": "352d20ee67be67f6340b4c0605b044b7",
-  "secret": "352d20ee67be67f6340b4c0605b044b7",
-  "template_path": "/opt/tyk-gateway/templates",
-  "use_db_app_configs": true,
-  "db_app_conf_options": {
+    "listen_address": "",
+    "listen_port": LISTEN_PORT,
+    "secret": "352d20ee67be67f6340b4c0605b044b7",
+    "node_secret": "352d20ee67be67f6340b4c0605b044b7",
+    "template_path": "/opt/tyk-gateway/templates",
+    "tyk_js_path": "",
+    "middleware_path": "middleware",
+    "policies": {
+        "policy_source": "service",
+        "policy_connection_string": "",
+        "policy_record_name": "tyk_policies",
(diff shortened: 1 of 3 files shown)


install/data/tyk.with_dash.conf
```diff
@@ -49,7 +49,7 @@
     "analytics_config": {
         "type": "",
         "ignored_ips": [],
-        "enable_detailed_recording": true,
+        "enable_detailed_recording": false,
```
LuanvP: Changed this for performance, do we want to enable this by default?

LuanvP:
Copied all settings into the blank conf file generated at run, to add new fields.

For dashboard version, I turned off detailed recording for performance. Do we want this enabled by default? 

Both also have experimental_process_org_off_thread enabled, do we want this also? Same for both having JSVM enabled by default.

Finally, shall I base the tyk.example.conf on the with_dash or the self_contained example? 

@buger @asoorm @joshblakeley

asoorm:
A few suggestions - and is up for discussion:

1. Rename example files to something like: `example.tyk.ce.conf` and `example.tyk.pro.conf`. Could even add a config for `example.tyk.hybrid.conf`.

2. Remove all configs which do not relate to the deployment. Tyk CE is headless. So we should just delete all dashboard related fields from the config as it does not apply and adds confusion. e.g.

```
...
"use_db_app_configs": false,
"db_app_conf_options": {
  "connection_string": "",
  "node_is_segmented": false,
  "tags": null
},
"disable_dashboard_zeroconf": false,
...
```

3. Ensure sensible defaults, and turn off performance degrading configs in example configurations. But if we want an entire complete example, we could have an `example.tyk.full.conf` where everything would be turned on.

buger:
Files inside data folder not really examples in a sense that you can't just use them, it is templates. That's why they hidden. Moreve over, I would say we should not encourage our users to experiment with config files manually and continue forcing them to use our `setup.sh` script. If they really want custom changes, they should do it line by line, after the main config generated by us.

install/data/tyk.with_dash.conf
matiasinsaurralde: Makes sense

LuanvP:
Closing this and will update the files separately

PR #1490, opened 2018-02-21
Virtual endpoint timeout

LuanvP:
Addresses https://github.com/TykTechnologies/tyk-analytics/issues/696

Code run in VirtualEndpoint now times out appropriately. Checking logic copied from https://github.com/TykTechnologies/tyk/blob/master/mw_js_plugin.go#L153 

VirtualEndpoint middleware is now also run in a goroutine

Added Debug logs for jsvm timeout values

### DIFF
--- a/mw_virtual_endpoint.go
+++ b/mw_virtual_endpoint.go
@@ -159,15 +159,6 @@ func (d *VirtualEndpoint) ServeHTTPForCache(w http.ResponseWriter, r *http.Reque
 	}
 
 	// Run the middleware
-	// vm := d.Spec.JSVM.VM.Copy()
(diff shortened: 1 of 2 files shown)


mw_virtual_endpoint.go
```diff
@@ -153,11 +158,57 @@ func (d *VirtualEndpoint) ServeHTTPForCache(w http.ResponseWriter, r *http.Reque
 		return nil
 	}
 
+	// Run the middleware
+	// vm := d.Spec.JSVM.VM.Copy()
```
buger: Pls remove this debugging

LuanvP: Yup! Forgot to add to commit

PR #1495, opened 2018-02-22
Support B64 Encoded Schema

asoorm:
- feature: #1163 B64 encodes schema for storage. Decodes for sending to gateway & exporting.
- feature: #1163 If $schema supplied, check that is Draft-04
- fix: SchemaVersion unnessesary. Schema itself versioned
- bug: Fix caching so that the schema loads and caches at compile stage
- lint: snake case to camel case

### DIFF
--- a/mw_validate_json.go
+++ b/mw_validate_json.go
@@ -6,14 +6,12 @@ import (
 	"io/ioutil"
 	"net/http"
 
-	"github.com/xeipuuv/gojsonschema"
-
+	"github.com/TykTechnologies/gojsonschema"
 	"github.com/TykTechnologies/tyk/apidef"
 )
 
 type ValidateJSON struct {
 	BaseMiddleware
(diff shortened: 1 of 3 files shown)


mw_validate_json.go
```diff
@@ -44,10 +44,6 @@ func (k *ValidateJSON) ProcessRequest(w http.ResponseWriter, r *http.Request, _
 		return errors.New("no schemas to validate against"), http.StatusInternalServerError
 	}
 
-	if vPathMeta.SchemaVersion != "" && vPathMeta.SchemaVersion != "draft-v4" {
```
buger: Why you removed it?

asoorm: because the schema itself supports versioning. unless I misunderstood? http://json-schema.org/latest/json-schema-core.html#rfc.section.7

PR #1496, opened 2018-02-22
bug: uptime tests respect proxy_ssl_insecure_skip_verify

### LINKED ISSUE #1412: Upstream health checks should respect `proxy_ssl_insecure_skip_verify`
bug
If upstream is secured with self-signed https certificate, all health checks will fail because of insecure certificate error
It should respect `proxy_ssl_insecure_skip_verify` configuration option, and if set, allow self-signed certificates
Configure HTTPS protected upstream, set `proxy_ssl_insecure_skip_verify` to true, and try to configure upstream health checks
all

asoorm:
fixes #1412

If upstream is secured with self-signed https certificate, all health
checks will fail because of insecure certificate error

`WARN host-check-mgr: [HOST CHECKER MANAGER] Host is
DOWN: https://test.server:8000/hello`

`CheckHost` function modified to set the `HostCheckerClient` transport
configuration appropriately. This is suitable for both http & https
hosts.

lint: using http constants & camelCase rather than snake_case vars

### DIFF
--- a/host_checker.go
+++ b/host_checker.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/tls"
 	"math/rand"
 	"net/http"
 	"strings"
@@ -9,6 +10,8 @@ import (
 
 	"github.com/jeffail/tunny"
 	cache "github.com/pmylund/go-cache"
+
+	"github.com/TykTechnologies/tyk/config"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


matiasinsaurralde:
LGTM

buger:
We can do even more, and support Mutual TLS certificates here too. 

See @joshblakeley PR, he implemented similar for BathRequest function https://github.com/TykTechnologies/tyk/pull/1471

buger:
Well it is a bug too, since if your API protected by mutual TLS, and you configured certificate inside tyk, the gateway will proxy request to API without any issues, but uptime check will fail because it ignore certificates. 

So I propose just include into the same PR.

asoorm:
cool - will work on today

buger:
Lets merge it like this, and will add certificate functionality if someone will complain. 

Host checker is deprecated anyway.

Issue #1256, opened 2017-10-31
Improve methods to query hashed keys


nickReyn:
## The problem 
When creating keys with hashing turned on, our users can't query those keys without knowing the original key. Additionally, hash can't be obtained until the key is used, so queries such as the following won't work:
* Show me all keys globally
* Show me all keys we have per API/policy

## Potential solution
- Allow listing hashed keys
- When creating a Key, the hash is returned and the user can choose to record this using their own method. The hash can then be used to query keys usage, without exposing the key itself.

## Community feedback
We are keen to understand who else of the Tyk community has issues querying hashed keys and their suggested solutions! 👍

lonelycode:
> When creating a Key, the hash is returned and the user can choose to record this using their own method. The hash can then be used to query keys usage, without exposing the key itself.

You can enable an event hook on the gateway that can send a webhook or run a plugin whenever a key is created. This will include the *raw original token in the payload*, the user can then use this to store the token in a secure location if they need to. This is already possible, we may want to improve the documentation around it.

nextmetaphor:
When key hashing is enabled, for the [gateway API](https://tyk.io/docs/tyk-rest-api/token-management/), it would be great if...

- `/tyk/keys/create` `[POST]` also returned the key hash

- `/tyk/keys/{access_token}` `[GET]` could also accept the key hash if the `hashed` boolean parameter is set, as is currently supported for the `DELETE` method. This obviously wouldn't return the key value itself, but would be used to examine what access a key has access to, rate limits, policies etc from the hash of the key

- `/tyk/keys/` `[GET]` would return the key hash for all keys

- `/tyk/keys/` `[GET]` with a string `policy_id` parameter would return all hashed keys associated with the specified policy

- `/tyk/keys/` `[GET]` with a string `api_id` parameter would return all hashed keys associated with the specified api

This functionality would also be replicated on the dashboard API.

In essence the above allows an administrator of the gateway, who doesn't know (or need to know) the actual key itself, to ensure that the current keys are auditable, and can be removed if required.

lonelycode:
We've had a discussion internally and are going to adopt parts of the recommendation proposed by @nextmetaphor - in particular the improvements to the existing Tyk API.

With regards to the filtering options, these look like they will be superseded by a different feature which is currently in planning.

nickReyn:
@buger @lghiur can we discuss the UX/UI impact of this please?

nickReyn:
[...]

PR #1499, opened 2018-02-25
create/get keys with hashed keys logic added

dencoded:
added changes for https://github.com/TykTechnologies/tyk/issues/1256

- creating key - `POST /keys/create`, `POST /keys` and `POST /keys/{keyName}` also return field `"key_hash"` if key hashing is switched on
- get all (or per API) keys  - `GET /keys` returns list of key hashes if key hashing is switched on
- get one key info by hash `GET /keys/{keyName}?hashed=true"`
- `DELETE /keys/{keyName}?hashed=true"` - remains the same (delete by hash)

### DIFF
--- a/api.go
+++ b/api.go
@@ -591,7 +591,17 @@ func keyHandler(w http.ResponseWriter, r *http.Request) {
 			obj, code = handleGetDetail(keyName, apiID, isHashed)
 		} else {
 			// Return list of keys
-			if isHashed {
+			if config.Global.HashKeys {
+				// get all keys is disabled by default
+				if !config.Global.EnableHashedKeysListing {
+					doJSONWrite(
+						w,
+						http.StatusNotFound,
+						apiError("Hashed key listing is disabled in config (enable_hashed_keys_listing)"),
+					)
+					return
+				}
+
 				// we don't use filter for hashed keys
 				obj, code = handleGetAllKeys("", apiID)
 			} else {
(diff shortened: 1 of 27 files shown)


api.go
```diff
@@ -620,41 +638,22 @@ func handleUpdateHashedKey(keyName, apiID, policyId string) (interface{}, int) {
...
-
-		return apiError("Unmarshalling failed"), 400
-	}
-
 	// Set the policy
 	sess.LastUpdated = strconv.Itoa(int(time.Now().Unix()))
 	sess.SetPolicies(policyId)
 
-	sessAsJS, err := json.Marshal(sess)
+	err := sessionManager.UpdateSession(keyName, &sess, 0)
```
buger: I noticed you not passing any info that key is hashed here, are you sure it will not do double hashing here?

api.go
```diff
@@ -743,7 +735,7 @@ func handleOrgAddOrUpdate(keyName string, r *http.Request) (interface{}, int) {
 		rawKey := QuotaKeyPrefix + storage.HashKey(keyName)
 
 		// manage quotas separately
-		DefaultQuotaStore.RemoveSession(rawKey)
+		DefaultQuotaStore.Store().DeleteRawKey(rawKey)
```
buger: Why this change?

buger:
Looks good, pls add new config variable (false by default), which enable **listing** of hashed keys. There are case, like our cloud when you have lot of keys, when you want to disable listing, but still allow querying hashed keys.

api.go
dencoded: thats true, looks like I missed it. I was experimenting with `UpdateSession` and then decided to leave it as is because it is used a lot. I will push a fix for it

dencoded:
one more thing for documenting:
- added new config value `"enable_hashed_keys_listing"` with `"false"` by default

this config value controls access to `GET /keys` when keys are hashed.
- so, if `"hash_keys":true` and `"enable_hashed_keys_listing":false` then `GET /keys` will return 404 with explanation why
- if `"hash_keys":true` and `"enable_hashed_keys_listing":true` then `GET /keys` will return 200 with list of hashes

PR #1509, opened 2018-03-03
mw_jwt benchmarks added, existing benchmarks fixed

dencoded:
added more benchmarking instrumentation

- started with `mw_jwt`
- added `utils/ci-benchmark.sh` which just runs benchmarking and has optional parameter - regex for benchmarks to run
- travis will run it but it will just produce numbers - no performance/memory allocation asserts there

Some thoughts about how to use it..

I think if we run it as part of travis we need to do one of those:
- save current branch/commit benchmarking output to `/tmp`
- then run benchmark against other branch (master or latest released?) and save benchmarking output to `/tmp` as well
- compare outputs with using https://godoc.org/golang.org/x/tools/cmd/benchcmp and raise error if its output contains `-` for ops and/or `+` for allocations

or

- accumulate benchmarks runs done by travis in file and store it somewhere (say push/pull from S3)
- analyze file with using https://godoc.org/golang.org/x/perf/cmd/benchstat and assert `delta` column from its output

As another option is to trigger running benchmarks on some etalon EC2 instance which will be the same CPU/memory/OS all the time and accumulating run results there.

Either way - benchmarks which make build to be green/red should be always running in one place on the same configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,3 +38,4 @@ script:
   - go build -tags 'coprocess grpc'
   - ./utils/ci-test.sh
   - if [[ $LATEST_GO ]]; then goveralls -coverprofile=<(gocovmerge *.cov); fi
+  - ./utils/ci-benchmark.sh
--- a/api_test.go
+++ b/api_test.go
@@ -724,13 +724,14 @@ func TestHotReloadMany(t *testing.T) {
 }
 
 func BenchmarkApiReload(b *testing.B) {
+	b.ReportAllocs()
+
 	specs := make([]*APISpec, 1000)
 
 	for i := 0; i < 1000; i++ {
-		s := buildAndLoadAPI(func(spec *APISpec) {
+		specs[i] = buildAndLoadAPI(func(spec *APISpec) {
 			spec.APIID = strconv.Itoa(i + 1)
-		})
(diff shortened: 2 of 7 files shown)


buger:
I would suggest adding custom upstream server, which has some internal timeout for each request.

dencoded:
What the point about custom upstream with timeout for each request?
Adding time out won't guarantee the same latency for every request as there will be still some code in play. I think current one with minimal latency is good one as we wanna benchmark our tyke code like how performant stuff before and after reverse-proxy to upstream.

My 1st idea was to eliminate touching upstream for benchmarks but I decided to leave it as we have some non trivial logic in reverse proxy.

buger:
Yeah, makes total sense. 

My main idea with sleep was that if we have some session level or path level locks, or we have some queues, with added latency and high concurrency such issues can become more visible. But I agree that it is a separate case, and probably can be a separate test.

buger:
I'm going to merge this changes now. @dencoded pls create an issue in product report with next steps you want to implement, and so @asoorm can provide his progress as well.

dencoded:
@buger 👍

dencoded:
@buger @asoorm I've created this ticket https://github.com/TykTechnologies/product/issues/94

PR #1511, opened 2018-03-05
Refactor and speedup url-rewrite

buger:
- Added new `replaceTykVariables` function, which replaces all
`$tyk_context` and `$tyk_meta` variables in string`
- URLRewrite regexp now cached. Previously it was compiling on each
call.

### DIFF
--- a/mw_modify_headers.go
+++ b/mw_modify_headers.go
@@ -1,9 +1,7 @@
 package main
 
 import (
-	"fmt"
 	"net/http"
-	"strings"
 
 	"github.com/TykTechnologies/tyk/apidef"
 )
@@ -12,11 +10,6 @@ import (
 type TransformHeaders struct {
 	BaseMiddleware
 }
-
-const (
-	metaLabel    = "$tyk_meta."
-	contextLabel = "$tyk_context."
-)
 
 func (t *TransformHeaders) Name() string {
 	return "TransformHeaders"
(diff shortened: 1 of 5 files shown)


mw_url_rewrite.go
```diff
@@ -17,17 +17,14 @@ import (
...
-		return "", err
+	if meta.MatchRegexp == nil {
+		return "", fmt.Errorf("URLRewrite regexp error", meta.MatchPattern)
 	}
 	path := r.URL.String()
 	log.Debug("Inbound path: ", path)
 	newpath := path
 
-	result_slice := mp.FindAllStringSubmatch(path, -1)
+	result_slice := meta.MatchRegexp.FindAllStringSubmatch(path, -1)
```
dencoded: `snake_case` for var names looks weird in Go

mw_url_rewrite.go
```diff
@@ -157,51 +153,64 @@ func urlRewrite(meta *apidef.URLRewriteMeta, r *http.Request) (string, error) {
...
+
+func replaceTykVariables(r *http.Request, in string, escape bool) string {
 	contextData := ctxGetData(r)
 
-	dollarMatch := regexp.MustCompile(`\$tyk_context.([A-Za-z0-9_\-\.]+)`)
-	replace_slice := dollarMatch.FindAllStringSubmatch(rewriteToPath, -1)
-	for _, v := range replace_slice {
-		contextKey := strings.Replace(v[0], "$tyk_context.", "", 1)
+	if strings.Contains(in, contextLabel) {
+		replace_slice := dollarMatch.FindAllStringSubmatch(in, -1)
```
dencoded: one more `snake_case` spotted

mw_url_rewrite.go
```diff
@@ -157,51 +153,64 @@ func urlRewrite(meta *apidef.URLRewriteMeta, r *http.Request) (string, error) {
...
-	if session := ctxGetSession(r); session != nil {
+	session := ctxGetSession(r)
+	if session == nil {
+		return in
+	}
 
-		metaDollarMatch := regexp.MustCompile(`\$tyk_meta.(\w+)`)
-		metaReplace_slice := metaDollarMatch.FindAllStringSubmatch(rewriteToPath, -1)
+	if strings.Contains(in, metaLabel) {
+		metaReplace_slice := metaDollarMatch.FindAllStringSubmatch(in, -1)
```
dencoded: and one more snake

mw_url_rewrite.go
```diff
@@ -157,51 +153,64 @@ func urlRewrite(meta *apidef.URLRewriteMeta, r *http.Request) (string, error) {
...
+var metaDollarMatch = regexp.MustCompile(`\$tyk_meta.([A-Za-z0-9_\-\.]+)`)
+
+func replaceTykVariables(r *http.Request, in string, escape bool) string {
 	contextData := ctxGetData(r)
 
-	dollarMatch := regexp.MustCompile(`\$tyk_context.([A-Za-z0-9_\-\.]+)`)
-	replace_slice := dollarMatch.FindAllStringSubmatch(rewriteToPath, -1)
-	for _, v := range replace_slice {
-		contextKey := strings.Replace(v[0], "$tyk_context.", "", 1)
+	if strings.Contains(in, contextLabel) {
```
dencoded: we could eliminate a couple of extra ops if move `contextData := ctxGetData(r)` inside this if

mw_url_rewrite.go
```diff
@@ -157,51 +153,64 @@ func urlRewrite(meta *apidef.URLRewriteMeta, r *http.Request) (string, error) {
...
 	// Meta data from the token
-	if session := ctxGetSession(r); session != nil {
+	session := ctxGetSession(r)
+	if session == nil {
+		return in
+	}
 
-		metaDollarMatch := regexp.MustCompile(`\$tyk_meta.(\w+)`)
-		metaReplace_slice := metaDollarMatch.FindAllStringSubmatch(rewriteToPath, -1)
+	if strings.Contains(in, metaLabel) {
```
dencoded: also could get rid of a couple of ops if we move `if session := ctxGetSession(r); ...` inside this if

mw_modify_headers.go
```diff
@@ -87,7 +38,9 @@ func (t *TransformHeaders) ProcessRequest(w http.ResponseWriter, r *http.Request
 
 	// Add
 	if len(vInfo.GlobalHeaders) > 0 {
```
dencoded: seems like extra check as iteration over empty map will be no operation as well

mw_url_rewrite.go
```diff
@@ -17,17 +17,14 @@ import (
 )
 
 func urlRewrite(meta *apidef.URLRewriteMeta, r *http.Request) (string, error) {
-	// Find all the matching groups:
-	mp, err := regexp.Compile(meta.MatchPattern)
-	if err != nil {
-		log.Debug("Compilation error: ", err)
-		return "", err
+	if meta.MatchRegexp == nil {
+		return "", fmt.Errorf("URLRewrite regexp error", meta.MatchPattern)
```
dencoded: we need to add placeholder into format string - this one was caught by my IDE :)

buger:
@dencoded existing benchmark won't show anything because it does not use URL rewrite plugins. Smth we need to address in future.

PR #1512, opened 2018-03-06
Lint: typo fix & use http constants rather than numbers

asoorm:


handler_success.go
```diff
@@ -204,8 +204,8 @@ func (s *SuccessHandler) RecordHit(r *http.Request, timing int64, code int, requ
 
 func recordDetail(r *http.Request) bool {
 	// Are we even checking?
-	if !config.Global.EnforceOrgDataDeailLogging {
-		return config.Global.AnalyticsConfig.EnableDetailedRecording
+	if !config.Global.EnforceOrgDataDetailLogging {
+		return false
```
buger: Looks like logic changed, are you sure about this change?

asoorm: good spot.

PR #1515, opened 2018-03-06
Fix Modify Headers/URL rewrite

LuanvP:
Fixes https://github.com/TykTechnologies/tyk-analytics/issues/663 and https://github.com/TykTechnologies/tyk/issues/1524

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -911,6 +911,8 @@ func (a *APISpec) CheckSpecMatchesStatus(r *http.Request, rxPaths []URLSpec, mod
 				continue
 			} else if mode == HeaderInjectedResponse && v.InjectHeadersResponse.Path != ctxGetUrlRewritePath(r) {
 				continue
+			} else {
+				continue
 			}
 		}
 		switch v.Status {
--- a/api_definition.go
+++ b/api_definition.go
@@ -402,9 +402,16 @@ func (a APIDefinitionLoader) getPathSpecs(apiVersionDef apidef.VersionInfo) ([]U
 }
 
 func (a APIDefinitionLoader) generateRegex(stringSpec string, newSpec *URLSpec, specType URLStatus) {
+	// if strings.HasPrefix(stringSpec, "/") {
+	// 	stringSpec = stringSpec[1:]
+	// }
+	println("TEST")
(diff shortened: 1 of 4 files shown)


res_handler_header_injector_test.go
```diff
@@ -0,0 +1,62 @@
...
+					"add_headers": {
+						"add-this": "header"
+					},
+					"path": "test-no-slash",
+					"method": "GET",
+					"act_on": false
+				}
+			]`), &v.ExtendedPaths.TransformResponseHeader)
+		})
+	})
```
LuanvP: @buger Am I setting up the middleware correctly here? It seems like it's not getting triggered for either path

json is same as for my test API

buger: Check if json.Unmarshal returns error, maybe it will contain some useful debugging info

ilijabojanovic:
@buger @LuanvP 

I we tested this PR and i still see issues.

If i enable modify headers on:
headers endpoint, add/remove request/response headers working as expected
if i enable modify headers on:
/headers endpoint, add/remove request working fine, but add/remove response headers don't working at all

@buger also url-rewrite + modify headers working with when we add endpoints without backslash 
https://github.com/TykTechnologies/tyk/issues/1524

LuanvP:
@ilijabojanovic both bugs should now be fixed

api_definition.go
```diff
@@ -902,17 +902,31 @@ func (a *APISpec) CheckSpecMatchesStatus(r *http.Request, rxPaths []URLSpec, mod
...
+
 		// only return it it's what we are looking for
 		if !match {
 			// check for special case when using url_rewrites with transform_response
 			// and specifying the same "path" expression
-			if mode != TransformedResponse {
-				continue
-			} else if v.TransformResponseAction.Path != ctxGetUrlRewritePath(r) {
+
+			if mode == TransformedResponse {
```
buger: In this case better use switch:

```
switch mode {
case TransformedResponse: 
...
case InjectHeadersResponse:
....
default:
continue 
}

LuanvP: @buger done

buger:
@ilijabojanovic pls check again

ilijabojanovic:
@buger Verified and approved for merge

PR #1522, opened 2018-03-07
Fix conflicting paths

### LINKED ISSUE #1523: Endpoint confilct with different methods
BUG
When we have two endpoints with different methods, longest endpoint overrides allowed method for shortest endpoint.
Each endpoint should have own permissions and methods
GET /customer-servicing/documents/metadata/{importFileName}
POST /customer-servicing/documents/metadata/purge

GET /customer-servicing/documents/metadata/{importFileName} overrides permissions for POST /customer-servicing/documents/metadata/purge. We can do only GET on /customer-servicing/documents/metadata/purge
All stable versions

buger:
Fixes #1523 

Fix bug when there are 2 paths which conflict by url, but have different HTTP methods


ilijabojanovic:
@buger approved for merge

PR #1528, opened 2018-03-09
lint: general main.go cleanup

asoorm:
General Cleanup

### DIFF
--- a/main.go
+++ b/main.go
@@ -113,6 +113,11 @@ var (
 	}
 )
 
+const (
+	defReadTimeout  = 120 * time.Second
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -49,6 +49,7 @@ import (
 
 var (
 	log                      = logger.Get()
+	l                        = log.WithField("prefix", "main")
```
buger: In general, I like this approach to reduce copy paste, and it should be probably applied to rest of modules in future, pls pls name it smth like `mainLog` or similar.

PR #1533, opened 2018-03-11
rate limiter check on org level added

dencoded:
added changes for https://github.com/TykTechnologies/product/issues/54

### DIFF
--- a/mw_organisation_activity.go
+++ b/mw_organisation_activity.go
@@ -65,7 +65,7 @@ func (k *OrganizationMonitor) ProcessRequestLive(r *http.Request) (error, int) {
 		&session,
 		k.Spec.OrgID,
 		k.Spec.OrgSessionManager.Store(),
-		true,
+		session.Per > 0 && session.Rate > 0,
 		true,
 	)
 
@@ -180,7 +180,7 @@ func (k *OrganizationMonitor) AllowAccessNext(orgChan chan bool, path string, IP
 		&session,
 		k.Spec.OrgID,
 		k.Spec.OrgSessionManager.Store(),
-		true,
+		session.Per > 0 && session.Rate > 0,
 		true,
 	)
 
--- a/mw_organisation_activity.go
+++ b/mw_organisation_activity.go
@@ -125,11 +125,7 @@ func (k *OrganizationMonitor) ProcessRequestLive(r *http.Request) (error, int) {
 func (k *OrganizationMonitor) SetOrgSentinel(orgChan chan bool, orgId string) {
 	for isActive := range orgChan {
 		log.Debug("Chan got:", isActive)
-		o, ok := orgActiveMap.Load(orgId)
-		if ok {
-			o = isActive
-			orgActiveMap.Store(orgId, o)
-		}
+		orgActiveMap.Store(orgId, isActive)
 	}
 }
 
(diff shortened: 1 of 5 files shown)


mw_organisation_activity.go
```diff
@@ -59,29 +59,54 @@ func (k *OrganizationMonitor) ProcessRequestLive(r *http.Request) (error, int) {
 		return errors.New("this organisation access has been disabled, please contact your API administrator"), 403
 	}
 
-	// We found a session, apply the quota limiter
-	reason := k.sessionlimiter.ForwardMessage(&session,
+	// We found a session, apply the quota and rate limiter
+	reason := k.sessionlimiter.ForwardMessage(
+		&session,
 		k.Spec.OrgID,
-		k.Spec.OrgSessionManager.Store(), false, false)
```
dencoded: it was set to `false, false` so `k.sessionlimiter.ForwardMessage` was doing basically nothing when it is not in `ExperimentalProcessOrgOffThread` mode

buger:
In general looks good, needs some tests.

buger:
^ races

dencoded:
@buger some comments to my latest commits:
- quota tests cover case when rate limit was not set and we can still process traffic
- I had to increase `go test` timeout to 40 sec because I added more sleeps in tests
- races where in a couple of spots:
  1. setting request's context in go-routine which was read in middle-wares
  2. we change `global.Config` in tests and read it in code under test at the same time so race was there - this is actually common (but intermittent) problem in other our tests
- fixed one bug in `OrganizationMonitor` mw for off-thread processing mode - `orgActiveMap.Store` was never called

mw_organization_activity_test.go
```diff
@@ -221,17 +207,9 @@ func TestProcessRequestLiveRedisRollingLimiter(t *testing.T) {
...
 	config.Global.EnforceOrgQuotas = true
-	enableRedisRollingLimiter := config.Global.EnableRedisRollingLimiter
 	config.Global.EnableRedisRollingLimiter = true
-	experimentalProcessOrgOffThread := config.Global.ExperimentalProcessOrgOffThread
 	config.Global.ExperimentalProcessOrgOffThread = true
-	defer func() {
-		config.Global.EnforceOrgQuotas = enforceOrgQuotas
-		config.Global.EnableRedisRollingLimiter = enableRedisRollingLimiter
-		config.Global.ExperimentalProcessOrgOffThread = experimentalProcessOrgOffThread
-	}()
```
buger: I think you need to call `defer resetTestConfig()` after each config declaration section, in both tests, similar to how its done in the rest of tests.

dencoded: I've tried different approaches including this one but it introduces data races.
I think the problem is that by the time we change `global.Config` in deferred function - there are still some go-routines running and reading config (like ones launched in `ProcessRequestOffThread()`)

buger: For cases with config races, I introduced new mutex `configMu` you can see its usage.

dencoded: yeah, I saw it, just thinking about how feasible would be to introduce new locks in middleware code just because config is being changed in tests. how often do we expect config to be change during gateway run?

I feel like those locks are extra as we change config structure by changing file and restarting actual GW-process

PR #1534, opened 2018-03-12
requestIP to package, tests, benchmarks & context

asoorm:
If we store IP in context in mw_context_vars, then when we need the IP address again,
we can do so with ~90% performance improvement.

```
BenchmarkRealIP_RemoteAddr-4            10000000               184 ns/op              16 B/op          1 allocs/op
BenchmarkRealIP_ForwardedFor-4          10000000               185 ns/op              16 B/op          1 allocs/op
BenchmarkRealIP_RealIP-4                10000000               139 ns/op              16 B/op          1 allocs/op
BenchmarkRealIP_Context-4               100000000               15.2 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/TykTechnologies/tyk/helpers  7.155s
```

### DIFF
--- a/handler_error.go
+++ b/handler_error.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/TykTechnologies/tyk/config"
+	"github.com/TykTechnologies/tyk/request"
 )
 
 const (
@@ -82,7 +83,7 @@ func (e *ErrorHandler) HandleError(w http.ResponseWriter, r *http.Request, errMs
 	token := ctxGetAuthToken(r)
 	var alias string
 
-	ip := requestIP(r)
+	ip := request.RealIP(r)
 	if config.Global.StoreAnalytics(ip) {
 		t := time.Now()
 
--- a/handler_success.go
+++ b/handler_success.go
@@ -12,6 +12,8 @@ import (
 
 	cache "github.com/pmylund/go-cache"
 
+	"github.com/TykTechnologies/tyk/request"
+
 	"fmt"
 
 	"github.com/TykTechnologies/tyk/config"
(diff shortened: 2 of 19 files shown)


helpers/real_ip.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"net"
+	"net/http"
+	"strings"
+)
+
+func RealIP(r *http.Request) string {
+
+	if contextIp := r.Context().Value("RealIP"); contextIp != nil {
```
matiasinsaurralde: perhaps use `contextIP`?

helpers/real_ip_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+		ip := RealIP(r)
+
+		if ip != test.expected {
+			t.Errorf("\texpected %s got %s", test.expected, ip)
+		}
+	}
+
+	t.Log("Context")
+	r, _ := http.NewRequest("GET", "http://abc.com:8080", nil)
```
matiasinsaurralde: `http.MethodGet`

buger:
Change itself look good! 

The only thing that I would probably move it back from helpers package, in order to avoid creating general purpose packages, like "utils", or "helpers": with time such packages start looking like Frankenstein and would be really hard to manage. 

In addition, I would propose not include benchmarks into the final merge. They were really useful in the context of this PR to see how change affects performance, but inside Tyk itself, probably does not make much sense. And we still can compare our @dencoded benchmarks added earlier, before and after the change. @asoorm it actually will be nice to see how it affected performance, maybe you can post them here as a comment?

buger:
I merged but forgot to wait for benchmarks. @asoorm can you pls put here bench before and after the change.

PR #1542, opened 2018-03-16
feature: strip path version from proxy upstream

asoorm:
resolves: https://github.com/TykTechnologies/tyk/issues/492

https://community.tyk.io/t/api-routing-via-header-option-and-catch-all-routing/2365/2


handler_success.go
```diff
@@ -224,6 +225,17 @@ func recordDetail(r *http.Request) bool {
 // Spec states the path is Ignored
 func (s *SuccessHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) *http.Response {
 	log.Debug("Started proxy")
+
+	versionDef := s.Spec.VersionDefinition
+	if versionDef.Location == "url" && versionDef.StripPath {
```
buger: I do not think you should use string replace the whole URL. Make it work with path only, see `StripListenPath ` few lines below, it does exactly what you want and very similar in its logic. Also do not forget that we have also RawPath, which used for analytics.

asoorm: Thanks, will look into

buger:
@asoorm pls create task for UI support

PR #1543, opened 2018-03-16
Include cookie values in mw context data

asoorm:
#1544 
    
 This PR allows the context variables to access the cookie by it's field
 name for use within middleware.
    
 ```
 $tyk_context.cookies_COOKIEFIELDNAME
 ```

### DIFF
--- a/mw_context_vars.go
+++ b/mw_context_vars.go
@@ -53,7 +53,12 @@ func (m *MiddlewareContextVars) ProcessRequest(w http.ResponseWriter, r *http.Re
 	//Correlation ID
 	contextDataObject["request_id"] = uuid.NewV4().String()
 
+	for _, c := range copiedRequest.Cookies() {
+		name := "cookies_" + strings.Replace(c.Name, "-", "_", -1)
+		contextDataObject[name] = c.Value
+	}
+
 	ctxSetData(r, contextDataObject)
 
-	return nil, 200
+	return nil, http.StatusOK
 }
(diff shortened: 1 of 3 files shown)


buger:
Pls create related ticket, assign it to 2.6 project, and attach "require documentation" label.

mw_context_vars.go
```diff
@@ -53,7 +53,12 @@ func (m *MiddlewareContextVars) ProcessRequest(w http.ResponseWriter, r *http.Re
 	//Correlation ID
 	contextDataObject["request_id"] = uuid.NewV4().String()
 
+	for _, c := range copiedRequest.Cookies() {
+		name := "cookies_" + strings.Replace(c.Name, "-", "_", -1)
+		contextDataObject[name] = string(c.Value)
```
dencoded: I am not sure we need this cast as `c.Value` is already a `string`

mw_context_vars_test.go
```diff
@@ -34,3 +37,37 @@ func TestContextVarsMiddleware(t *testing.T) {
...
+	err, code := (&MiddlewareContextVars{}).ProcessRequest(res, req, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if code != http.StatusOK {
+		t.Fatal(errors.New("non 200 status code"))
+	}
+
+	ctx := req.Context().Value(ContextData)
```
dencoded: I think using `ctxGetData(req)` would simplify code in this test and avoid type asserts in if statements below (hard to read for me)

Issue #1506, opened 2018-03-02
bug: CORS options_passthrough on Non OPTIONS requests


asoorm:
When `Enable CORS: false` and `Options Passthrough: true`. In the scenario where the gateway is short-circuiting the proxy, for example due to failed auth, the response headers do not contain required CORS headers.

It is expected by Browsers that the *SAME* CORS headers returned from an OPTIONS request will be returned on a POST request for example.

Steps to Recreate:

Browser sends preflight `OPTIONS` request to GW, GW transparently reverse proxy preflight request to upstream. CORS headers returned.

```
$ curl -X OPTIONS https://tyk-gateway.dev:8080/httpbin/post -H 'Origin: abc.com' -k -v
> OPTIONS /httpbin/post HTTP/1.1
> Host: tyk-gateway.dev:8080
> User-Agent: curl/7.54.0
> Accept: */*
> Origin: abc.com
>
< HTTP/1.1 200 OK
< Access-Control-Allow-Credentials: true
< Access-Control-Allow-Methods: GET, POST, PUT, DELETE, PATCH, OPTIONS
< Access-Control-Allow-Origin: abc.com
< Access-Control-Max-Age: 3600
< Allow: POST, OPTIONS
< Connection: close
< Content-Length: 0
< Content-Type: text/html; charset=utf-8
< Date: Fri, 02 Mar 2018 13:15:42 GMT
< Server: meinheld/0.6.1
< Via: 1.1 vegur
< X-Powered-By: Flask
< X-Processed-Time: 0
< X-Ratelimit-Limit: 0
< X-Ratelimit-Remaining: 0
< X-Ratelimit-Reset: 0
<
```

Browser sends actual request to Gateway. But with invalid credentials - which means 4xx failure.

```
$  curl -X POST https://tyk-gateway.dev:8080/httpbin/post -H 'Origin: abc.com' -H 'Authorization: Bearer JUNK' -k -v
> POST /httpbin/post HTTP/1.1
> Host: tyk-gateway.dev:8080
> User-Agent: curl/7.54.0
> Accept: */*
> Origin: abc.com
> Authorization: Bearer JUNK
>
< HTTP/1.1 403 Forbidden
< Content-Type: application/json
< Date: Fri, 02 Mar 2018 13:18:32 GMT
< Content-Length: 37
<
{
    "error": "Key not authorized"
}
```

This is because the gateway has short-circuited the connection and never hits the upstream in order to get the CORS headers.

[...]

PR #1550, opened 2018-03-20
[DO NOT MERGE] bug: CORS headers not returned when mw shortcircuits upsteam

asoorm:
#1506

In the current scenario, when the gateway does not handle CORS, and
OptionsPassthrough is set to true, if the tyk middleware shortcircuits
the response, tyk is not able to return the required upstream CORS
headers.

This fixes the issue by detecting the appropriate conditions, and
sending an OPTIONS request via loopback, then appending the full set of
CORS headers to the error response.

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -193,7 +193,9 @@ func loopbackCORS(w http.ResponseWriter, r *http.Request, mw TykMiddleware) {
 	corsRes, err := corsClient.Do(corsRequest)
 	if err != nil {
 		log.WithError(err).Error("Error from upstream OPTIONS request, ignoring and returning default response")
+		return
 	}
+	defer corsRes.Body.Close()
 
 	corsHeaders := map[string]string{}
 	corsHeaders[response.AccessControlAllowOrigin] = corsRes.Header.Get(response.AccessControlAllowOrigin)
--- a/middleware.go
+++ b/middleware.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"crypto/tls"
 	"fmt"
 	"net/http"
+	"net/http/httputil"
 	"strconv"
 	"time"
 
@@ -81,8 +83,13 @@ func createMiddleware(mw TykMiddleware) func(http.Handler) http.Handler {
 				h.ServeHTTP(w, r)
 				return
 			}
+
 			err, errCode := mw.ProcessRequest(w, r, mwConf)
 			if err != nil {
+
+				if errCode == http.StatusForbidden {
+					loopbackCORS(w, r, mw)
+				}
 
 				handler := ErrorHandler{mw.Base()}
 				handler.HandleError(w, r, err.Error(), errCode)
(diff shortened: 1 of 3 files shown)


asoorm:
PR submitted for early feedback please

asoorm:
Present implementation: If a POST request is sent with a body, we get an error. But if body is empty, then it works as expected.

```
2018/03/20 10:44:51 http: multiple response.WriteHeader calls
```

Need to investigate this.

letzya:
Per @asoorm have copied from the main issue:
What about other response codes? they would just be blocked?
For instance:

    When rate limit exceeded
    When global rate limit exceeded 429
    When quota exceeded
    When access denided
    Access to this API has been disallowed
    Key has expired, please renew 401

Sorry, haven't thoroughly checked, this is off the top of my head

middleware.go
```diff
@@ -80,9 +86,14 @@ func createMiddleware(mw TykMiddleware) func(http.Handler) http.Handler {
 				h.ServeHTTP(w, r)
 				return
 			}
+
 			err, errCode := mw.ProcessRequest(w, r, mwConf)
 			if err != nil {
 
+				if errCode == http.StatusForbidden {
```
dencoded: how would we know that this `http.StatusForbidden` caused by one of our middlewares, not from actual upstream?
I think if it is reply from upstream we don't need to do anything as it was resposibility of upstream to set CORS headers and they might already there.

also, how about other HTTP 4XX which can be set by other our middlewares?

asoorm: Because at the point this has been injected - it's from the middleware, and not the upstream. So at this stage, the request has not yet been proxied and we are about to return an error to the user.

middleware.go
```diff
@@ -106,6 +117,98 @@ func createMiddleware(mw TykMiddleware) func(http.Handler) http.Handler {
...
+
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: config.Global.HttpServerOptions.SSLInsecureSkipVerify},
+	}
+
+	corsClient := &http.Client{Transport: tr}
+	corsRes, err := corsClient.Do(corsRequest)
+	if err != nil {
+		log.WithError(err).Error("Error from upstream OPTIONS request, ignoring and returning default response")
+	}
```
dencoded: connection leak, we should do `defer corsRes.Body.Close()`

middleware.go
```diff
@@ -106,6 +117,98 @@ func createMiddleware(mw TykMiddleware) func(http.Handler) http.Handler {
...
+
+	log.Debugf("loopback request dump\n%s", string(dump))
+
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: config.Global.HttpServerOptions.SSLInsecureSkipVerify},
+	}
+
+	corsClient := &http.Client{Transport: tr}
+	corsRes, err := corsClient.Do(corsRequest)
```
dencoded: this needs to have some kind of circuit breaker or at least time out for client to avoid waiting for upstream when it is down (if we don't it could cause bump in gateway latency)

middleware.go
```diff
@@ -106,6 +117,98 @@ func createMiddleware(mw TykMiddleware) func(http.Handler) http.Handler {
...
+		return
+	}
+
+	log.Debugf("loopback request dump\n%s", string(dump))
+
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: config.Global.HttpServerOptions.SSLInsecureSkipVerify},
+	}
+
+	corsClient := &http.Client{Transport: tr}
```
dencoded: I'd add `corsClient ` as field of middleware and change `loopbackCORS` to be method of middleware so we won't re-create `http.Client` within every call of `loopbackCORS` and benefit from keep-alived connection

middleware.go
```diff
@@ -106,6 +117,98 @@ func createMiddleware(mw TykMiddleware) func(http.Handler) http.Handler {
...
+	corsRes, err := corsClient.Do(corsRequest)
+	if err != nil {
+		log.WithError(err).Error("Error from upstream OPTIONS request, ignoring and returning default response")
+	}
+
+	corsHeaders := map[string]string{}
```
dencoded: might worth to cache the reply with using as key `Origin` received from browser to improve latency and avoid bombing upstream with `OPTIONS` for the same origin every time we error

middleware.go
asoorm: So the circuit breaker I am not sure is required. because we are calling Tyk itself but this time, as a pre-flight request.

middleware.go
```diff
@@ -106,6 +117,98 @@ func createMiddleware(mw TykMiddleware) func(http.Handler) http.Handler {
...
+	corsHeaders[response.AccessControlAllowOrigin] = corsRes.Header.Get(response.AccessControlAllowOrigin)
+	corsHeaders[response.AccessControlAllowMethods] = corsRes.Header.Get(response.AccessControlAllowMethods)
+	corsHeaders[response.AccessControlAllowCredentials] = corsRes.Header.Get(response.AccessControlAllowCredentials)
+	corsHeaders[response.AccessControlMaxAge] = corsRes.Header.Get(response.AccessControlMaxAge)
+
+	for k, v := range corsHeaders {
+		if v != "" {
+			w.Header().Set(k, v)
```
dencoded: we need to make sure that response wasn't already sent to browser at this point. setting headers on `ttp.ResponseWriter` when response already sent won't take any affect

middleware.go
dencoded: my understanding is that this depends on order of middlewares. it is not super clear for me that  the request has not yet been proxied as this code will be literally part of every middleware (if I understand right)

middleware.go
dencoded: oh, I though we were going to send OPTIONS to upstream directly. it seems to me that calling GW itself again is an extra step (but it should work). I'd add timeout to request anyways

asoorm: That's why it's called loopback. because the gw calls it's own API. The reasoning for this is because the OPTIONS request needs to go through the exact same MW chain as the original pre-flight request would have done before so that we get the same request to the same upstream endpoint.

e.g. there may be MW in place which re-writes the target.

dencoded: ok, got it. according to original pre-flight request - some browsers don't send pre-flight requests for all ajax calls, so original OPTIONS one might never happened. I don't think it would cause any issues in this case, just wanted to mention it.

dencoded: I am still not sure about this loopback round trip business, as looks like all we want to do is just to hit the code placed a couple of lines above here: https://github.com/TykTechnologies/tyk/pull/1550/files#diff-4c166b743d95629a9ab37eb6fe3df22aR86

why not just to prepare OPTIONS request like we do anyways and response writer (we can probably use httptest recoder) and then call `h.ServeHTTP(optionsRecorder, optionsRequest)` to make it go through all underlaying middle-wares down to upstream?

PR #1555, opened 2018-03-21
PoC - support yaml and json config files

asoorm:
This is just a PoC - Not for merging, and obviously needs to be fully
tested.

json is difficult for OPS teams to work with, especially via the command
line.

without going into too much detail about benefits of yaml, we can also
add comments to configs, or even comment some fields out. even offer
explanations / documentation directly into the config file if necessary.

json is subset of yaml. And as such, with a few simple changes, instead of
unmarshal json, we can unmarshal yaml using the github.com/ghodss/yaml
lib.

if data is json, it will unmarshal into struct. if data is yaml, it will
also unmarshal into a struct.

The benefit of this method, is that we can support both yaml and json
configs without having to worry about adding extra struct tags.

Further - if this works well, we could potentially also support yaml api
definitions, and maybe even yaml swagger specs with little effort.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"regexp"
 
+	"github.com/ghodss/yaml"
 	"github.com/kelseyhightower/envconfig"
 
 	"github.com/TykTechnologies/tyk/apidef"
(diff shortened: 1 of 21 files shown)


config/config.go
```diff
@@ -369,9 +370,12 @@ func Load(paths []string, conf *Config) error {
 		log.Info("Loading default configuration...")
 		return Load([]string{path}, conf)
 	}
-	if err := json.NewDecoder(r).Decode(&conf); err != nil {
+
+	bytes, _ := ioutil.ReadAll(r)
+	if err := yaml.Unmarshal(bytes, &conf); err != nil {
```
buger: As mentioned in slack, I would rather not rely on on 3-rd party library for json parsing, and will run `yaml.Unmarshal` only if string looks like YAML, since it is easy to detect by checking first byte.

dencoded:
I like the idea to support both yaml or json. However, would be interesting to see how our config would look like in yaml just to make sure it won't look too crazy.

asoorm:
closing as abandoned.

Issue #1536, opened 2018-03-14
Session metadata empty with gRPC plugin


robertogyn19:
**Do you want to request a *feature* or report a *bug*?**
Bug.

**What is the current behavior?**
Using the javascript middleware the metadata is populated correctly, however when using gRPC plugin the session metadata is empty.

**What is the expected behavior?**
Expect see api key meta_data in the session object.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

Plugin dispatcher in Java I expect to see the key metadata in the `request.getSession().getMetadata()`, but the map is always empty.

```java
package com.testorg.testplugin;

import coprocess.CoprocessObject;
import coprocess.DispatcherGrpc;

public class PluginDispatcher extends DispatcherGrpc.DispatcherImplBase {

	@Override
	public void dispatch(CoprocessObject.Object request,
	                     io.grpc.stub.StreamObserver<CoprocessObject.Object> responseObserver) {

		System.out.println();
		System.out.println("hook type       : " + request.getHookType());
		System.out.println("hook name       : " + request.getHookName());
		System.out.println("request.metadata: " + request.getMetadataMap());
		System.out.println("session.metadata: " + request.getSession().getMetadata());
		System.out.println();

		responseObserver.onNext(request);
		responseObserver.onCompleted();
	}
}
```

Plugin output:
```shell
Initializing gRPC server.

hook type       : PostKeyAuth
hook name       : MyPostKeyAuthMiddleware
request.metadata: {}
session.metadata: 


hook type       : Post
hook name       : MyPostMiddleware
request.metadata: {}
session.metadata: 
```

manifest.json:
```json
{
  "custom_middleware": {
    "driver": "grpc",
    "post": [
      {
        "name": "MyPostMiddleware",
        "require_session": true
      }
    ],
    "post_key_auth": [
      {
        "name": "MyPostKeyAuthMiddleware",
        "require_session": true
      }
    ]
  }
}
```

API definition:
```json
{
  "id": "5aa913965b5c1a00017a09fa",
  "name": "test",
  "slug": "test",
  "api_id": "e6229c01dff5495e68892f403c8a17da",
  "org_id": "5aa7e68b493bc70001f34ea2",
  "auth": {
    "use_param": false,
    "param_name": "",
    "use_cookie": false,
    "cookie_name": "",
    "auth_header_name": "Authorization",
    "use_certificate": false
  },
  "custom_middleware": {
    "pre": null,
[...]

PR #1577, opened 2018-03-28
coprocess: metadata access on post and post-key auth hooks

matiasinsaurralde:
Following the report on #1536, we use `coprocess.Object` metadata field for our custom authentication mechanism, using implementations like this one:
```go
	object.Session = &coprocess.SessionState{
		Rate: 1000.0,
		Per:  1.0,
	}
	object.Metadata = map[string]string{
		"token":   validKey,
		"metakey": "metavalue",
	}

	return object, nil
```
This works fine for custom authentication or the ID extractor mechanism itself (to extract the key only) however the metadata is ignored when `SessionManager.UpdateSession` is called, this patch fixes this.
Additionally, it makes sure that an encoded session object includes metadata values (the encoded session object is used by post and post-key auth hooks).
The reason why we used `coprocess.Object.Metadata` instead of interacting with the session metadata field directly is that the native structure (not the Protobuf-generated one) uses `interface{}` which was trickier to use with Protobuf in the past (see [here](https://github.com/golang/protobuf/issues/370)). For a better support we might modify the `coprocess.Session.Metadata` field to use the `Protobuf.Any` type.
This patch doesn't break the compatibility with existing custom authentication implementations but adds missing data useful for a few scenarios.

### DIFF
--- a/coprocess.go
+++ b/coprocess.go
@@ -133,9 +133,15 @@ func (c *CoProcessor) ObjectFromRequest(r *http.Request) *coprocess.Object {
 
 	// Encode the session object (if not a pre-process & not a custom key check):
 	if c.HookType != coprocess.HookType_Pre && c.HookType != coprocess.HookType_CustomKeyCheck {
-		session := ctxGetSession(r)
-		if session != nil {
+		if session := ctxGetSession(r); session != nil {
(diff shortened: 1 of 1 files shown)


buger:
Make sense.

Pls update the docs, and inform customer how to use this feature.

PR #1584, opened 2018-03-30
Log fields

asoorm:


buger:
@asoorm pls fix conflicts and can be merged

Issue #834, opened 2017-06-22
Replace rsc/letsencrypt with x/crypto/acme/autocert


mvdan:
As per the warning in https://github.com/rsc/letsencrypt:

> In the long term you should be using https://golang.org/x/crypto/acme/autocert instead of this package. Send improvements there, not here.

Also, the one we use seems to be semi-abandoned, whereas the other is much more active.

mvdan:
Speaking of which, do we have tests of any kind for this? We don't seem to have unit tests in this repo. @buger @ilijabojanovic

mvdan:
Ah, unfortunately the APIs are slightly different. `autocert` does not have Manager methods to marshal and unmarshal the internal state into a string.

mvdan:
I believe the right fix is to supply a `Manager.Cache` backed by Redis, instead of using `Watch` in combination with `Marshal` and `Unmarshal`. To confirm my suspicion, I've sent https://groups.google.com/forum/#!topic/golang-nuts/LHEuEUHWnGI.

In any case, this switch is less trivial than I thought and the library we currently use isn't bringing any trouble. Moving back to 2.5.

dernise:
Bump, the letsencrypt package is now broken and certificate issuance isn't working anymore

PR #1588, opened 2018-03-31
Implemented autocert

maxencehenneron:
This is a work in progress, I updated to autocert as described in this issue: https://github.com/TykTechnologies/tyk/issues/834

A few things: 

- Lets Encrypt disabled TLS-SNI http://www.zdnet.com/article/lets-encrypt-disables-tls-sni-01-validation/ which is what tyk is currently using to generate the certificates
- I had to use the http-01 ACME challenge (see http://letsencrypt.readthedocs.io/en/latest/challenges.html for more informations) which needs to have the port 80 available. 
- I have no idea what the NoticeGatewayLENotification is, it's not used anywhere in the source code, does anyone have any idea what it refers to in the current code base?
- I made a redirection to the https endpoint in case someone tries to access the gateway from the port 80.
- I created a new config variable a user has to set to agree with Letsencrypt's TOS

I'd appreciate to have feedbacks

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -121,6 +121,7 @@ type HttpServerOptionsConfig struct {
 	WriteTimeout           int        `json:"write_timeout"`
 	UseSSL                 bool       `json:"use_ssl"`
 	UseLE_SSL              bool       `json:"use_ssl_le"`
+	AcceptLetsEncryptTOS   bool 	  `json:"accept_lets_encrypt_tos"`
 	SSLInsecureSkipVerify  bool       `json:"ssl_insecure_skip_verify"`
 	EnableWebSockets       bool       `json:"enable_websockets"`
 	Certificates           []CertData `json:"certificates"`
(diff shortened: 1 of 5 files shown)


dernise:
Also, the build seems to fail because I added accept_lets_encrypt_tos, why is the linter not recognizing my newly-created field?

buger:
Hi! Thank you for contribution. 

> I created a new config variable a user has to set to agree with Letsencrypt's TOS

Does this step required?

dernise:
It's a legal requirement. Let's Encrypt wants the user to explicitly agree their TOS. We could override it and force it to be true but it would leave you  legally reliable I guess..

lonelycode:
The notification is used here: https://github.com/TykTechnologies/tyk/blob/master/redis_signals.go#L108

When gateways are load balanced the cluster should be sharing the LE SSL cert to prevent spamming LE, IIRC it stashed the cert in redis and tells the cluster to use that.

Issue #1519, opened 2018-03-07
TykMakeHttpRequest & TykBatchRequest response object inconsistent


bitsofinfo:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

The response object from `TykMakeHttpRequest` and `TykBatchRequest` differ in that `TykMakeHttpRequest`'s response object's headers are in a property called `Headers` but in `TykBatchRequest` the response is called `headers`.  This applies to all properties.

**What is the expected behavior?**

All response object properties should be consistently named 

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**

all

asoorm:
I'm not certain we can change this without breaking backwards compatibility. Any thoughts?

buger:
We can support both formats for some time, and remove the lower case version in 3.0 or similar.

buger:
Released as part of 2.7

PR #1594, opened 2018-04-05
Make TykMakeHttpRequest response fields consistent, add tests

matiasinsaurralde:
Related to #1519.


buger:
@matiasinsaurralde pls prepare tyk-docs PR with this changes. I assume we should convert all our examples to use lowercase?

buger:
@matiasinsaurralde pls fix the test, it is not that obvious

matiasinsaurralde:
Fixed

Issue #1540, opened 2018-03-15
Allow to specify proxy url on API level


buger:
**Do you want to request a *feature* or report a *bug*?**
feature
**What is the current behavior?**
Right now if you need to make tyk use proxy, the only way is to set HTTP(S)_PROXY env variable, and ALL HTTP requests from all APIs will use this proxy. We inherit this feature from Go stack itself.
**What is the expected behavior?**
It should be possible to set specific proxy URL on API level. And wrap standard `http.ProxyFromEnvironment` function, to check for API level variable first. 

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.5

buger:
Note that some of the features may require Go 1.10 https://medium.com/@mlowicki/https-proxies-support-in-go-1-10-b956fb501d6b

PR #1597, opened 2018-04-06
Add support for specifying HTTP Proxy per API

buger:
Added new API definition option: `proxy.transport.proxy_url` to specify custom proxy per API

Fix: https://github.com/TykTechnologies/tyk/issues/1540


asoorm:
Nice

Issue #1590, opened 2018-04-03
Distributed rate limiter connection issues 


ilijabojanovic:
**Do you want to request a *feature* or report a *bug*?**
BUG
**What is the current behavior?**
DRL don't have connection to redis
**What is the expected behavior?**
DRL should not fail during connection to redis
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Start gateway in on-prem mode. Monitor debug log. 
This issue isn't visible always and it's visible only during gateway start.

Here is log during gateway start:
```
[36mtyk-gateway_1          |[0m Trying to compile and run Tyk...
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=debug msg="Enabling debug-level output" 
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=debug msg="Using /develop/confs/tyk.conf for configuration" 
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=debug msg="Setting up analytics DB connection" 
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=debug msg="Connecting to redis cluster" 
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=debug msg="Creating new Redis connection pool" 
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=debug msg="[STORE] Getting WAS: PollerActiveInstanceID" 
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=debug msg="[STORE] Getting: host-checker:PollerActiveInstanceID" 
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=debug msg="Error trying to get value:redigo: nil returned" 
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=debug msg="No Primary instance found, assuming control" 
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=debug msg="[STORE] SET Raw key is: PollerActiveInstanceID" 
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=debug msg="[STORE] Setting key: host-checker:PollerActiveInstanceID" 
[36mtyk-gateway_1          |[0m time="Apr  3 19:45:41" level=info msg="Starting Poller" 
[...]

PR #1599, opened 2018-04-07
start DRL after node registration completed

dencoded:
added fix for https://github.com/TykTechnologies/tyk/issues/1590

so we had this sequence:

1. set global `NodeID` to `solo-...`
2. setup `DRLManager` and start sending notifications
3. `DRLManager` sends notification with NodeID="solo-..."
4. run node registration which sets `NodeID` to some different value
5. DRL's method `NotifyCurrentServerStatus` uses global `NodeID` so starts sending notifications with new NodeID - which confuses underlying DRL which still has "solo-..." as ID for current server, so it treats new NodeID as other server

I've changed startup process - first we do node registration and then when we know that NodeID is "final" we start DRL

### DIFF
--- a/distributed_rate_limiter.go
+++ b/distributed_rate_limiter.go
@@ -91,7 +91,12 @@ func onServerStatusReceivedHandler(payload string) {
 	log.Debug("Received DRL data: ", serverData)
 
 	if DRLManager.Ready {
-		DRLManager.AddOrUpdateServer(serverData)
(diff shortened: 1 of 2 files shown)


lonelycode:
Will this still work without node registration (CE cluster)?

dencoded:
I works fine without node registration as it it skips that step (registering node) and just runs DRL with 1 current node with default NodeID=`solo-...`

Issue #1446, opened 2018-01-30
Advanced URL rewriter should support negative logic


buger:
**Do you want to request a *feature* or report a *bug*?**
feature

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Right now I can define only what should match. Introducing negative logic can simplify boolean logic a lot. I would suggest adding "not_match_rx" field.

LuanvP:
This is mainly finished on this branch: https://github.com/TykTechnologies/tyk/pull/1601

I've updated StringRegexMap to include both a MatchPattern and a NotMatchPattern. The Check method takes both of these into account.

I've added tests to cover negative and mixed-logic cases, there is one test commented out which was behaving strangely and it seemed to be due to some caching behaviour in the way the tests were run (?)

buger:
Closed in favor of https://github.com/TykTechnologies/tyk/pull/2096

ilijabojanovic:
Released in 2.9 release

PR #1601, opened 2018-04-10
URL Rewrites implement negative trigger logic

LuanvP:
To implement https://github.com/TykTechnologies/tyk/issues/1446

### DIFF
--- a/mw_url_rewrite_test.go
+++ b/mw_url_rewrite_test.go
@@ -91,7 +91,7 @@ func TestRewriterTriggers(t *testing.T) {
 			patt := "hello"
 			r.Header.Set("x-test", patt)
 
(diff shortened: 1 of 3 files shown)


mw_url_rewrite_test.go
```diff
@@ -560,6 +560,52 @@ func TestRewriterTriggers(t *testing.T) {
 				r,
 			}
 		},
+		// func() TestDef {
```
LuanvP: @ilijabojanovic I think there's some weird caching behaviour here, if I uncomment this test it fails even though the rewrite logic is triggered.

The rewrite path prints as "/change/to/me/$tyk_context.trigger-0-payload" which comes from the previous test, and the URL stays as "/change/to/me/ignore"

Any thoughts?

PR #1607, opened 2018-04-18
Port process ownership revert and init script fixes

excieve:
Master currently has potentially backwards incompatible behaviour in the Linux init scripts and packages related to change of ownership from root user to "tyk". This reverts master to previous behaviour until backwards compatible upgrade path is introduced.

Additionally it lacks the fix for broken upstart 0.x scripts from 2.5.x releases so that's ported too.


excieve:
@buger Could you please merge this into master and 2.6 branch so that I could build the test packages?

Issue #991, opened 2017-08-07
transport: No gRPC URL is set


bitsofinfo:
I need to use the bundle downloader to pull a some `otto` js middleware and use it in an api.

To do this, one of the options it says I must turn on is `enable_coprocess`

```
     "coprocess_options": {
        "enable_coprocess": true,
        "coprocess_grpc_server": ""
    },
```

I do this and restart the gateway, then I start getting these errors regarding GRPC... I have no need for GRPC

```
2017-08-07T15:27:19.848880480Z 2017/08/07 15:27:19 grpc: addrConn.resetTransport failed to create client transport: connection error: desc = "transport: No gRPC URL is set!"; Reconnecting to { <nil>}
```

bitsofinfo:
can this please be fixed, constant useless noise in logs

bitsofinfo:
not stale

PR #1608, opened 2018-04-22
Enhance gRPC errors

matiasinsaurralde:
Enhancement for #991 
The dialer already does an URL check (parsing) but this is a simpler thing to check during init

### DIFF
--- a/main.go
+++ b/main.go
@@ -195,7 +195,11 @@ func setupGlobals() {
 	}
 
 	if config.Global().CoProcessOptions.EnableCoProcess {
-		CoProcessInit()
+		if err := CoProcessInit(); err != nil {
+			log.WithFields(logrus.Fields{
+				"prefix": "coprocess",
+			}).Error(err)
+		}
 	}
 
 	// Get the notifier ready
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -190,7 +190,11 @@ func setupGlobals() {
 	}
 
 	if config.Global.CoProcessOptions.EnableCoProcess {
-		CoProcessInit()
+		if err := CoProcessInit(); err != nil {
+			log.WithFields(logrus.Fields{
```
asoorm: Can just use `log.WithField("prefix", "coprocess")`

matiasinsaurralde: Updated

matiasinsaurralde:
This is a very small change, to test this, set this into your `tyk.conf` and use the standard Tyk build (the one we get from `go build -tags 'coprocess grpc'`:
```json
     "coprocess_options": {
        "enable_coprocess": true,
        "coprocess_grpc_server": ""
    },
```
Without the patch you should see the recurring dialer errors.

matiasinsaurralde:
Just fixed conflicts on this PR

PR #1613, opened 2018-04-24
feature: Make MaxIdleConns configurable

### LINKED ISSUE #1560: Allow max connection per host be greater then 100
feature
While we allow control `MaxIdleConnsPerHost` using `max_idle_conns_per_host`, the top maximum `MaxIdleConns` is still 100, and not configurable. 

On high traffic environments, especially when upstream not that fast, this value can become a bottleneck. 
`MaxIdleConns` should be configurable via config
2.5

asoorm:
fixes #1560

This PR could be a bit of a problem.

https://golang.org/src/net/http/transport.go#L156
> MaxIdleConns controls the maximum number of idle (keep-alive)
> connections across all hosts. Zero means no limit.

Zero value with this PR will set default MaxIdleConns to 100 - which matches existing behaviour. But that means that impossible to allow unlimited connections.

However without setting sensible default value, zero means that all prev installations will then change current functionality to unlimited idle connections.

Any thoughts @buger ?

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -272,6 +272,7 @@ type Config struct {
 	EnableBundleDownloader            bool                                  `bson:"enable_bundle_downloader" json:"enable_bundle_downloader"`
 	AllowRemoteConfig                 bool                                  `bson:"allow_remote_config" json:"allow_remote_config"`
 	LegacyEnableAllowanceCountdown    bool                                  `bson:"legacy_enable_allowance_countdown" json:"legacy_enable_allowance_countdown"`
+	MaxIdleConns                      int                                   `bson:"max_idle_connections" json:"max_idle_connections"`
 	MaxIdleConnsPerHost               int                                   `bson:"max_idle_connections_per_host" json:"max_idle_connections_per_host"`
 	MaxConnTime                       int64                                 `json:"max_conn_time"`
 	ReloadWaitTime                    int                                   `bson:"reload_wait_time" json:"reload_wait_time"`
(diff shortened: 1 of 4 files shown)


joshblakeley:
Would anyone truly need unlimited connections? They still have the option to pass it a ridiculously large number should they wish to.

buger:
I believe there is some misunderstanding here. I'm not sure that `infinite` applicable to this option. From what I read in Go source before, 0, will mean that Go will set it to 2 - its default value. If you want it unlimited, you need set it just to a very high number.

buger:
^ update on above

I confused it with `MaxIdleConnsPerHost` which is 2 by default. 

And in this case, I probably agree to keep it 0 by default and avoid magic numbers like -1.

dencoded:
my comment about unlimited as default was according to option we change in this PR - `"max_idle_connections"`. I believe we shouldn't do `100` as default as it is almost equal to have it hardcoded to `100` as it was before (unless users change it to big number or -1).

buger:
Yes, pretty much, because it won't change anything. Users still will be capped with `MaxIdleConnsPerHost` 100, which is set in all our default configs.  

So by setting it to 0, you will only allow it to set `MaxIdleConnsPerHost` more then 100.

asoorm:
@buger Removed the -1 / unlimited functionality. 0 means unlimited now.

buger:
@asoorm pls fix CI, from what I see only gofmt is needed

PR #1614, opened 2018-04-25
adding more benchmarks

dencoded:
added more benchmarks in different middle-wares


dencoded:
most of the changes are just reusing existing unit-tests code to run benchmarks

dencoded:
also, added two new profiling flags to tyk:
- `--blockprofile` - when tyk exits it will generate and flush to disk block-profile with name `tyk.blockprof`
- `--mutexprofile` - when tyk exits it will generate and flush to disk mutex-profile with name `tyk.mutexprof`
useful when you don't want to enable `http/pprof` (or tyk is irresponsive to http calls) but want to see what was going on with blocking

dencoded:
ok, now I need to solve conflicts here as well

Issue #929, opened 2017-07-20
feature: (Extension) Expose request context to JSVM/Plugins


bitsofinfo:
Running Tyk on docker, w/ some customer JSVM middleware plugins. I'd like to be able to launch the gateway containers w/ a docker env var such as "-e "MYVAR=MYVAL".... and subsequently would like to be able to access the value of MYVAR from within a middleware JS plugin

https://community.tyk.io/t/access-environment-variable-from-withing-jsvm-middleware/1765

bitsofinfo:
any thoughts on this?

buger:
Assigned milestone. Make sense.

bitsofinfo:
Will this be added at some point?

asoorm:
What are thoughts about setting env var prefixed with `TYK_CONTEXT_`?

We could potentially expose the entire context vars + env vars within Tyk context - and then make accessible via function?

e.g. Rough example `var ctx = TykGetContext();`

```
ctx: {
   env: {
      ENV_KEY1: ENV_VAL1,
      ENV_KEY2: ENV_VAL2
   },
   ... // other context vars
}
```

bitsofinfo:
That would be great, but what about real os level environment vars that are defined (for example) on container start (such as via docker w `-e var1=val1`). Would those be accessible via this method?

asoorm:
My thoughts are that it would work like this:

`docker run -itd --rm --name gateway -e TYK_CONTEXT_VAR1=value1 tykio/tyk-gateway:latest`

bitsofinfo:
But why limit to only those prefixed with `TYK_CONTEXT_` ?

asoorm:
I was thinking in terms of security perspective not to expose everything - as will not always be run inside a docker container.

bitsofinfo:
true, yeah as long as we can set it via an `-e` syntax (and not have to mess w/ tyk config file), then I can live w/ the prefix. Sounds good

buger:
As an alternative, we may just add "context" attribute to request object itself. By doing this way, even gRPC plugins will get support of this feature.

buger:
@matiasinsaurralde I would like to make it in more generalized way:
Environment variables prefixed with TYK_CONTEXT, should be added to the actual `tyk_context` object in mw_context_vars.go, and JSVM and gRPC plugins will be able to access whole `tyk_context` object (and maybe modify it). It will probably mean new field in middleware objects.

buger:
[...]

PR #1616, opened 2018-04-25
Expose prefixed environment variables to JS middleware

matiasinsaurralde:
Discussed in #929.

With this you can access environment variables:
```js
var sampleMiddleware = new TykJS.TykMiddleware.NewMiddleware({});
sampleMiddleware.NewProcessRequest(function(request, session) {
    var myvar = TykJS.Environment["TYK_CONTEXT_MY_VAR"]
    return sampleMiddleware.ReturnData(request, session.meta_data);
});
```
We may optionally strip the "TYK_CONTEXT_" prefix, so that it becomes "MY_VAR". Environment variables that don't start with "TYK_CONTEXT_" will be ignored.
VE should also benefit from this. We avoid having this in the request context.
`getCtxEnv` can be re-used for rich plugins code.

### DIFF
--- a/mw_js_plugin.go
+++ b/mw_js_plugin.go
@@ -310,6 +310,9 @@ const defaultJSVMTimeout = 5
 func (j *JSVM) Init(spec *APISpec) {
 	vm := otto.New()
 
+	// Inject environment variables, this is referenced by TykJS.Environment:
+	vm.Set("env", getCtxEnv())
+
 	// Init TykJS namespace, constructors etc.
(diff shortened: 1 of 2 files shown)


asoorm:
i think would work better if the `TYK_CONTEXT_SOMETHING` env var was set on host OS, then make that env var accessible via `$tyk_context.env_XXX`

buger:
@asoorm see my comment here https://github.com/TykTechnologies/tyk/issues/929

Issue #1662, opened 2018-05-03
profile performance of BenchmarkRewriter


dencoded:
Let’s focus on significant performance improvements, the ones with less than 1%-5% difference can be just statistical noise (consider to increase bench time with flag `-benchtime=…` in case of low throughput).

Need to do profiling of all types (cpu, memory, block and mutex). Provide `benchcmp` results for benchmark output - before and after improvement (or prove that this piece of code cannot be improved in any ways).

matiasinsaurralde:
I've found a related issue here, at some point we refactored the API definition code in order to avoid pointers when loading `URLSpec` data structures, see [here](https://github.com/TykTechnologies/tyk/commit/c2f80ff6bac2f2c7c92c559be3ae02a452a5c882#diff-e174c6f6100dfee8c74de8323d464e36). This assumed all `URLSpec` access was read only, which isn't the case for the optimization that's described here:
While trying to optimize the URL rewriter middleware I noticed it wasn't possible to pre-compile the regular expression and store it as part of `URLRewriteMeta`, in [`mw_url_rewriter.go`](https://github.com/TykTechnologies/tyk/blob/master/mw_url_rewrite.go) we use the following condition:
```go
	if meta.MatchRegexp == nil {
		var err error
		meta.MatchRegexp, err = regexp.Compile(meta.MatchPattern)
		if err != nil {
			return path, fmt.Errorf("URLRewrite regexp error %s", meta.MatchPattern)
		}
	}
```
This is called every single time and the compiled regular expression isn't persisted (as we expect). To fix this it would make sense to use `[]*URLSpec` again and refactor the functions that use it, mainly `APISpec.Version`, `CheckSpecMatchesStatus`, `MakeSpec`. I've did this and the code seems to work as expected, looking for suggestions anyway.

PR #1624, opened 2018-04-30
Use precompiled regex for ID extractor

matiasinsaurralde:
When using regular expressions with the ID extractor, we compile them on every request, following the suggestions from the performance document, this modifies the code in order to compile the expression only once.

To test this, the following manifest can be used:
```json
  "custom_middleware": {
    "auth_check": {
      "name": "MyAuthCheck"
    },
    "id_extractor": {
      "extract_from": "header",
      "extract_with": "regex",
      "extractor_config": {
        "header_name": "Authorization",
        "regex_expression": "prefix(.*)",
        "regex_match_index": 0
      }
    },
    "driver": "grpc"
  }
```

PR #1672, opened 2018-05-04
Access the URL rewriter meta data structure using a pointer

matiasinsaurralde:
Need to do some more tests on this, the idea is to avoid changing other parts of the code like this: https://github.com/TykTechnologies/tyk/commit/c2f80ff6bac2f2c7c92c559be3ae02a452a5c882#diff-e174c6f6100dfee8c74de8323d464e36

With this PR, the regular expressions used in the URL rewriter are compiled only once, following the idea of #1629, #1624. This is part of #1662.

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -105,7 +105,7 @@ type URLSpec struct {
 	InjectHeadersResponse     apidef.HeaderInjectionMeta
 	HardTimeout               apidef.HardTimeoutMeta
 	CircuitBreaker            ExtendedCircuitBreakerMeta
-	URLRewrite                apidef.URLRewriteMeta
+	URLRewrite                *apidef.URLRewriteMeta
 	VirtualPathSpec           apidef.VirtualMeta
 	RequestSize               apidef.RequestSizeMeta
 	MethodTransform           apidef.MethodTransformMeta
@@ -678,7 +678,7 @@ func (a APIDefinitionLoader) compileURLRewritesPathSpec(paths []apidef.URLRewrit
 		newSpec := URLSpec{}
 		a.generateRegex(stringSpec.Path, &newSpec, stat)
 		// Extend with method actions
-		newSpec.URLRewrite = stringSpec
+		newSpec.URLRewrite = &stringSpec
 
 		urlSpec = append(urlSpec, newSpec)
 	}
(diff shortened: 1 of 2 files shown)


matiasinsaurralde:
Any feedback on this is appreciated, not sure it will affect other existing functionality, tests look ok

dencoded:
avoid doing compile is cool. I was also thinking about creating some kind of internal service/wrapper arround `regexp` functions where we can cache results of all match/find/submatch like operations and avoid doing them every time, say some `map[string]bool` where key is just regexp+tested_string

say we have 100 requests per sec of endpoint calls `GET /abcxyz` and it matches regexp `abc.*` - it is not efficient to do 100 `regexp.Match` call per sec knowing the result is the same - it does lots of CPU and memory allocations under the hood. maybe this kind of caching is already done on package `"regexp"` level, needs to be proved 1st if we wanna go with that approach.

api_definition.go
```diff
@@ -990,7 +990,7 @@ func (a *APISpec) CheckSpecMatchesStatus(r *http.Request, rxPaths []URLSpec, mod
 			}
 		case URLRewrite:
 			if r.Method == v.URLRewrite.Method {
-				return true, &v.URLRewrite
+				return true, v.URLRewrite
```
buger: Not sure if I follow, here you actually removed usage of pointer, what I'm missing?

buger: /cc @matiasinsaurralde

matiasinsaurralde: `URLRewrite` was changed, so the `&` isn't required.
Without this patch this is the behavior:
```
Sending request for the first time:
meta.MatchRegexp = 0xc4204ada88
meta.MatchRegexp is nil, calling regexp.Compile
Sending the same request again:
meta.MatchRegexp = 0xc42049ea88
meta.MatchRegexp is nil, calling regexp.Compile
```
Using the patch:
```
Sending request for the first time:
meta.MatchRegexp = 0xc4203a8478
meta.MatchRegexp is nil, calling regexp.Compile
Sending the same request again:
meta.MatchRegexp = 0xc4203a8478
```

matiasinsaurralde:
`benchcmp` output (will refactor the benchmark on a different PR)
```
% benchcmp old new
benchmark               old ns/op     new ns/op     delta
BenchmarkRewriter-8     197470        173320        -12.23%

benchmark               old allocs     new allocs     delta
BenchmarkRewriter-8     501            501            +0.00%

benchmark               old bytes     new bytes     delta
BenchmarkRewriter-8     378877        378876        -0.00%
```

Issue #1623, opened 2018-04-30
Avoid/replace JSON serialization for sessions


matiasinsaurralde:
**Do you want to request a *feature* or report a *bug*?**
Enhancement

**What is the current behavior?**
Currently when a session is created, we serialize the `SessionState` as JSON and store it on Redis. Using an alternative serialization method could improve Redis memory usage and also speed up read operations ([this article](http://adamnengland.com/2012/11/15/redis-performance-does-key-length-matter/) mentions that using smaller keys results in faster reads).
When a session is retrieved from Redis, we deserialize the object and so on.

**What is the expected behavior?**
It should be possible to switch the serialization format and get a better performance, also in combination with a fix for #1622.
I've been running an experiment on replacing JSON with Protocol Buffers (as we already have these definitions in place, we use them for rich plugins). This is how we currently store a session (note the key length):
<img width="886" alt="redis_key_json" src="https://user-images.githubusercontent.com/20110/39423998-ecb366f8-4c74-11e8-83fb-774f999ff55b.png">
This is after switching to PB (key length is around 4 times smaller):
<img width="881" alt="redis_key_pb" src="https://user-images.githubusercontent.com/20110/39424016-fd0565ec-4c74-11e8-9079-af410f301c6d.png">

I ran a few benchmarks disabling the local session cache.
Using PB:
```
% boom -H 'Authorization: 59f358a13565200001116d6f02c5e14d780847ad8e7161f35891aac5' -n 100 -c 2 http://localhost:8080/quickstart/
Summary:
  Total:	0.0408 secs
  Slowest:	0.0046 secs
  Fastest:	0.0005 secs
  Average:	0.0008 secs
  Requests/sec:	2451.8248
  Total data:	600 bytes
  Size/request:	6 bytes

Status code distribution:
  [200]	100 responses

Response time histogram:
  0.001 [1]	|
  0.001 [78]	|∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  0.001 [13]	|∎∎∎∎∎∎
  0.002 [6]	|∎∎∎
  0.002 [0]	|
  0.003 [0]	|
  0.003 [0]	|
  0.003 [0]	|
  0.004 [0]	|
  0.004 [0]	|
  0.005 [2]	|∎

Latency distribution:
  10% in 0.0006 secs
[...]

PR #1674, opened 2018-05-04
Update gRPC, Protocol Buffers and other dependencies

matiasinsaurralde:
Currently we're using the gRPC package that was released in 2016, this package had different enhancements in the past years, see [here](https://github.com/grpc/grpc-go/releases).
These aren't breaking changes and the Protocol Buffers versions is still 3.x. Have tested this with a few gRPC plugin samples we have.

Updates/adds the following dependencies as well (required by the gRPC and Protocol Buffers package)
- golang.org/x/net/context
- golang.org/x/net/http2
- golang.org/x/net/trace


matiasinsaurralde:
Found this: https://grpc.io/2017/08/22/grpc-go-perf-improvements.html

buger:
Can we benchmark this change, do we have gRPC plugin benchmark, and if don't can we add?

matiasinsaurralde:
Sure, did some quick tests with `hey` while testing #1623 using the latest gRPC package and they looked positive. Will work on it.

buger:
Does it mean that we updated dependencies based on their masters? I thought we going to use stable tags?

matiasinsaurralde:
Yes, `master` is stable, we could also use the last release tag

asoorm:
@matiasinsaurralde have you looked at: https://github.com/gogo/protobuf

Issue #1678, opened 2018-05-07
Tyk fails when using global event handlers


matiasinsaurralde:
**Do you want to request a *feature* or report a *bug*?**
After #1592 we use a different method to access/modify the global configuration, global event handlers need to be updated accordingly.

**What is the current behavior?**
The following configuration will make Tyk crash:
```json
{
    "listen_port": 8080,
    "secret": "352d20ee67be67f6340b4c0605b044b7",
    "template_path": "templates",
    "middleware_path": "middleware",
    "db_app_conf_options": {
    },
    "app_path": "apps/",
    "storage": {
        "type": "redis",
        "host": "localhost",
        "port": 6379,
        "database": 0,
        "optimisation_max_idle": 100,
        "optimisation_max_active": 0
    },
      "local_session_cache": {
    "disable_cached_session_state": false
  },
  "event_handlers": {
    "events": {
      "AuthFailure": [
        {
          "handler_name": "eh_web_hook_handler",
          "handler_meta": {
            "method": "POST",
            "target_path": "http://localhost:1111"
          }
        }
      ]
    }
  }
}
```
Output is:
```
% ./tyk --conf=test.conf
[May  7 16:00:29]  INFO webhooks: Loading default template. target=http://localhost:1111
[May  7 16:00:29] ERROR webhooks: Could not load the default template: open default_webhook.json: no such file or directory
[May  7 16:00:29] ERROR Failed to init event handler: open default_webhook.json: no such file or directory
[May  7 16:00:29]  INFO main: PIDFile location set to: /var/run/tyk/tyk-gateway.pid
[May  7 16:00:29] ERROR main: Failed to write PIDFile: error opening pidfile /var/run/tyk/tyk-gateway.pid: open /var/run/tyk/tyk-gateway.pid559671885: permission denied
[May  7 16:00:29]  INFO main: Initialising Tyk REST API Endpoints
[May  7 16:00:29]  WARN File descriptor limit 256 too low for production use. Min 80000 recommended.
	This could have a significant negative impact on performance.
	Please refer to https://tyk.io/docs/deploy-tyk-premise-production/#file-handles for further guidance.
[May  7 16:00:29] ERROR Could not EXPIRE key: dial tcp :0: connect: can't assign requested address
[May  7 16:00:29]  INFO Waiting for Redis connection pools to be ready currRetry=0
[...]

Issue #1269, opened 2017-11-07
Investigate why global event handlers are broken on 2.3 and master


mvdan:
And fix them, adding tests.

matiasinsaurralde:
Interesting, it will be useful for #1260 as I plan to adapt those changes to support event handlers.

mvdan:
@matiasinsaurralde is already working on stuff very close to this (and the handlers seem to work for him), and I hadn't started this, so passing it on.

buger:
Moving to 1.4.1

matiasinsaurralde:
Yes, it’s on my list today

buger:
@matiasinsaurralde bump :)

matiasinsaurralde:
I've been investigating this issue again. I see that we have a `config.Global.EventHandlers` field, this is loaded from `tyk.conf` and has the same type as the event handlers block we use for API definitions (`apidef.EventHandlerMetaConfig`):
```go
	EventHandlers                     apidef.EventHandlerMetaConfig         `json:"event_handlers"`
	EventTriggers                     map[apidef.TykEvent][]TykEventHandler `json:"event_trigers_defunct"`
	PIDFileLocation                   string                                `json:"pid_file_location"`
```
However I don't see any references to `config.Global.EventHandlers`, this means we never load these "global" events.
In the other side, event handlers seem to be working fine when attached to API definitions.
I don't see any reference to "global" event handlers in our  docs, was this feature available in the past? I'm investigating previous versions at the moment to see if we broke it somehow, perhaps after configuration tweaks?

buger:
@matiasinsaurralde indeed weird, try to check 2.4 branch on GitHub. 

If it is not in 2.4, estimate how hard is to add them, and maybe it just make sense remove them from all the configs and docs.

matiasinsaurralde:
I've realized that the original issue was about JSVM event handlers, I can confirm that they stopped working after 2.3.
To reproduce:
- Modify `tyk.conf` with `"enable_jsvm": true`
- Setup a protected API
- Add event handlers to it:
```json
"event_handlers": {
    "events": {
        "AuthFailure": [
                {
                    "handler_name":"eh_dynamic_handler",
                    "handler_meta": {
                        "name": "sampleHandler",
                        "path": "event_handlers/sample_handler.js"
                    }
                }
            ]
    }
}
```
- Create `event_handlers/sample_handler.js` with the following contents:
```js
[...]

PR #1679, opened 2018-05-07
Update global event handlers code

matiasinsaurralde:
Not sure if this feature is documented at all but it was breaking my local installation and this was necessary to keep the compatibility (#1678).
Have also changed the variable names a bit, would like to add tests in a separate PR. This might also be related to #1269.

### DIFF
--- a/event_system.go
+++ b/event_system.go
@@ -185,9 +185,9 @@ func (l *LogMessageEventHandler) HandleEvent(em config.EventMessage) {
 	log.Warning(logMsg)
 }
 
-func InitGenericEventHandlers(theseEvents apidef.EventHandlerMetaConfig) map[apidef.TykEvent][]config.TykEventHandler {
-	actualEventHandlers := make(map[apidef.TykEvent][]config.TykEventHandler)
-	for eventName, eventHandlerConfs := range theseEvents.Events {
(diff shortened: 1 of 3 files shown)


buger:
@dencoded so, do you approve it?

matiasinsaurralde:
I've simplified this and added the patch from #1693, have also found the reason why global event handlers aren't working anymore (#1678).

dencoded:
looks OK but there is some data race in CI

matiasinsaurralde:
@dencoded Thanks, have found the issue after adding the patch from #1693, will work on a fix

matiasinsaurralde:
Have fixed the issue, tweaked the log handler in order to accept an alternative `logrus.Logger`, by default it will take the standard logger, when running tests we use a custom one to capture the output (something we do in the JSVM tests). Basically removed dup code that caused a race when accessing the buffer from this custom logger.
Note that we use the event system in `event_handler_webhooks_test.go`, in this scenario it's ok to keep using the standard logger (so we can't force the custom logger for the whole event system code), the proposed solution also takes care of this.

matiasinsaurralde:
Also, tried to use these [test hooks](https://github.com/sirupsen/logrus#testing) but the current solution looked simpler.

PR #1684, opened 2018-05-08
Adds min-length check, adds more robust hash function, fixes #1681

### LINKED ISSUE #1681: Tyk Gateways Allows Invalid 3 Character Authorization Tokens
This is a bug
Any API using Authorization header with Tyk AuthToken (key) that has an access list (not using Profile) lets all requests through that have 3 charactors or less in the Authorization header

Obviously these are not valid keys and should be blocked.

Create an API, Create a Key for the API, make a request to the API with any 3 character combination. Now try a 4 charactor combination

We saw this in Gateway version 2.3.5 through 2.6.1
api 3 chars
auth 4 chars

lonelycode:
This fixes #1681 by checking min length of token (must be at least 3 or a configurable parameter in the config file)

Adds a longer-term fix by using a psuedo-random suffix for new tokens, if this suffix is found, then Murmur2 is used (64 bit), otherwise, if it is exactly 56 characters (an org ID + UUIDv4) OR is less than 56, will revert to murmur3.

Choice for the hashing algo [was based on this rather amazing write-up](https://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed)

### DIFF
--- a/auth_manager.go
+++ b/auth_manager.go
@@ -190,7 +190,9 @@ type DefaultKeyGenerator struct{}
 func (DefaultKeyGenerator) GenerateAuthKey(orgID string) string {
 	u5 := uuid.NewV4()
 	cleanSting := strings.Replace(u5.String(), "-", "", -1)
-	return orgID + cleanSting
+	withOrg := orgID + cleanSting
+
+	return storage.ModifyFromHashRange(withOrg)
 }
 
 // GenerateHMACSecret is a utility function for generating new auth keys. Returns the storage key name and the actual key
--- a/auth_manager_test.go
+++ b/auth_manager_test.go
@@ -1,0 +1,15 @@
+package main
+
+import (
+	"github.com/TykTechnologies/tyk/storage"
+	"testing"
+)
+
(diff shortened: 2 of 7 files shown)


buger:
Just depending on 56 probably won't work since murmur not used in all cases, as well as orgId. 
We have the following key formats:
- random UUID: fixed length 32
- Basic http auth: variadic length depending on user name
- Keys created based on certificates: fixed length 64

And all the keys above can be created with or without orgID (length 24)

buger:
On the other hand, if we make it work only for regular tokens it will be good for this stage, the rest (certs and basic auth), will use old hashing. So the only thing needs to be fixed is optional orgID, and ensuring that rest of token formats fallback to old hashing.

lonelycode:
I agree - this solution isn't robust enough - I think we may need something much more structured. I'm going to close this PR as I think we can do it better. 

I have opened a new PR that addresses the core issue of the three-char access in certain hash conditions, that should at least speed up patch release, completely changing hash mechanism may be something in 2.8.

PR #1680, opened 2018-05-08
Remove fmt call in the session for each call

joshblakeley:
Fixes #1632 

Running gateway headless with auth-key.

Before change:

```
Showing top 10 nodes out of 182
      flat  flat%   sum%        cum   cum%
     0.04s  0.58%  0.58%      0.04s  0.58%  fmt.(*fmt).fmt_sbx
     0.04s  0.58%  1.16%      0.06s  0.87%  github.com/TykTechnologies/tyk/vendor/github.com/TykTechnologies/redigocluster/rediscluster.ConcurrentMap.Iter.func1.1
     0.02s  0.29%  1.45%      0.04s  0.58%  net/textproto.CanonicalMIMEHeaderKey
     0.02s  0.29%  1.74%      0.02s  0.29%  net/textproto.validHeaderFieldByte (inline)
     0.02s  0.29%  2.03%      0.02s  0.29%  strings.TrimLeftFunc
     0.01s  0.14%  2.17%      0.01s  0.14%  bufio.(*Reader).bufio.reset
     0.01s  0.14%  2.32%      0.01s  0.14%  bufio.(*Writer).WriteString
     0.01s  0.14%  2.46%      0.01s  0.14%  bytes.(*Buffer).grow
     0.01s  0.14%  2.61%      0.01s  0.14%  bytes.(*Buffer).tryGrowByReslice
     0.01s  0.14%  2.75%      0.01s  0.14%  encoding/json.(*arrayEncoder).encode
```

After:

```
Showing top 10 nodes out of 153
      flat  flat%   sum%        cum   cum%
     0.02s   0.3%   0.3%      0.02s   0.3%  bytes.(*Buffer).WriteString
     0.02s   0.3%  0.61%      0.02s   0.3%  context.(*cancelCtx).cancel
     0.02s   0.3%  0.91%      0.03s  0.46%  context.(*valueCtx).Value
     0.02s   0.3%  1.22%      0.02s   0.3%  github.com/TykTechnologies/tyk/config.Global (inline)
     0.02s   0.3%  1.52%      0.20s  3.04%  main.createMiddleware.func1.1
     0.02s   0.3%  1.82%      0.02s   0.3%  net/textproto.CanonicalMIMEHeaderKey
     0.02s   0.3%  2.13%      0.02s   0.3%  sync.(*RWMutex).RLock
     0.01s  0.15%  2.28%      0.01s  0.15%  bytes.(*Buffer).Write
[...]

PR #1687, opened 2018-05-09
Benchmark user session hash function - to help with string casting ticket

joshblakeley:
As suggested by @dencoded in comments of #1680 

I have done this as it's own PR so it can be merged to master separately as it is expansion of benchmarking rather than part of a fix.

### DIFF
--- a/user/session_test.go
+++ b/user/session_test.go
@@ -1,0 +1,23 @@
+package user
+
+import (
+	"testing"
+)
+
+func BenchmarkHash(b *testing.B) {
+	s := SessionState{
+		Allowance:        1000.0,
+		Rate:             1000.0,
+		Per:              1,
+		Expires:          1458669677,
+		QuotaRemaining:   1000,
+		QuotaRenewalRate: 3600,
+	}
+
+	b.ReportAllocs()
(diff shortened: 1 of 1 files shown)


user/session_test.go
```diff
@@ -0,0 +1,11 @@
+package user
+
+import "testing"
+
+var session SessionState
+
+func BenchmarkHash(b *testing.B) {
```
dencoded: need to enable memory allocations with `b.ReportAllocs()`

user/session_test.go
```diff
@@ -0,0 +1,11 @@
+package user
+
+import "testing"
+
+var session SessionState
```
dencoded: might worth to create session with some data instead of zero-value struct

user/session_test.go
```diff
@@ -0,0 +1,11 @@
...
+
+func BenchmarkHash(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		session.Hash()
```
dencoded: need to save result into variable out of scope to avoid compiler optimizations

joshblakeley:
Thanks @dencoded will improve tomorrow

Issue #1551, opened 2018-03-20
Add option to ignore kid header in JWT validation


LuanvP:
**Do you want to request a *feature* or report a *bug*?**
feature, customer request: https://community.tyk.io/t/beginner-tyk-cloud-keycloak-kid-header/2472
**What is the current behavior?**
For JWT validation, "sub" claim will only be used for ID if the "kid" header is empty
https://github.com/TykTechnologies/tyk/blob/master/mw_jwt.go#L106
**What is the expected behavior?**
Add a config option to ignore the "kid" header if present use "sub" header instead. Required when JWT provider already uses kid header for its own purpose

neterium:
Just FYI, Keycloak is an example of provider using a kid header in its generated JWT access tokens

neterium:
Have to agree Buger :).
But if Tyk delegates the JWT generation to a 3rd party IDP (which is good practice I presume), it also has to accept that the kid header is already used to store a 3rd party identifier (which is also good practice then). Definitely a blocking issue for me :(.

buger:
Added new `jwt_skip_kid` API definition boolean field to ignore kid sub.

PR #1688, opened 2018-05-09
Skip kid #1551

letzya:
Made Tyk skip kid in case it's not an empty string but already being used for other purpose, and check the `sub` claim (or any clain that the user set on the api def)
This address #1551

Tests are still missing! creating this PR per @buger request.

TestJWTSessionRSABearerInvalidTwoBears - added test as in 'TODO' - to show that this code needs fixing


mw_jwt.go
```diff
@@ -100,31 +108,41 @@ func (k *JWTMiddleware) getSecretFromURL(url, kid, keyType string) ([]byte, erro
...
+	// If is not supposed to ignore KID - will use this as ID if not empty
+	if !k.Spec.APIDefinition.JWTSkipCheckKidAsId {
+		tykId, idFound = token.Header[KID].(string)
+		if idFound {
+			log.Debug("Found: ", tykId)
+			return tykId, idFound, nil
+		}
+	}
+	// In case KID was empty or was set to ignore KID ==> Will try to get the Id from JWTIdentityBaseField or fallback to SUM
+	tykId, err := k.getUserIdFromClaim(token)
```
letzya: This method is checking if there was a configured claim and if not fall back to  'sub'.

letzya:
@buger - please review. thanks

buger:
I think we need to remove option from config level, for consistency reasons. All our JWT options live in API definition at the moment.

buger:
I think we need to remove option from config level, for consistency reasons. All our JWT options live in API definition at the moment.

buger:
Also you added lot of unnecessary files like mw_jwt_test.go.10may2018_0107 tyk.test and etc. Pls remove them.

mw_jwt.go
```diff
@@ -206,22 +225,41 @@ func (k *JWTMiddleware) getBasePolicyID(token *jwt.Token) (string, bool) {
...
+	}
+	if !found {
+		userId, found = token.Claims.(jwt.MapClaims)[SUB].(string)
+		if !found {
+			message := fmt.Sprint("user id was not found in claims")
+			log.Error(message)
+			return "", errors.New(message)
+		}
+	}
+	log.Debugf("Found User Id: ", userId)
```
asoorm: This looks like a bug: shouldn't it be something like: `log.Debugf("Found User Id: %s", userId)`

mw_jwt.go
```diff
@@ -206,22 +225,41 @@ func (k *JWTMiddleware) getBasePolicyID(token *jwt.Token) (string, bool) {
...
+		userId, found = token.Claims.(jwt.MapClaims)[k.Spec.JWTIdentityBaseField].(string)
+
+		if !found {
+			log.Warning("Base Field not found, falling back to SUB")
+		}
+	}
+	if !found {
+		userId, found = token.Claims.(jwt.MapClaims)[SUB].(string)
+		if !found {
+			message := fmt.Sprint("user id was not found in claims")
```
asoorm: no need to `fmt.Sprint` - just set the msg to a string

mw_jwt.go
buger: good point, either this or replica function to just `log.Debug`

letzya: of course. will check what had happened

mw_jwt.go
letzya: Didnt want to repeat the text (then if you need to change it you do it once. maybe I should have added const.
**const message := fmt.Sprint("user id was not found in claims")**

asoorm: What I meant was, what is wrong with: `message := "user id was not found in claims"`

mw_jwt.go
```diff
@@ -396,17 +435,17 @@ func (k *JWTMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Request, _
...
+				return nil, fmt.Errorf("Unexpected signing method: %v and not HMACSign", token.Header["alg"])
 			}
-		case "rsa":
+		case RSASign:
 			if _, ok := token.Method.(*jwt.SigningMethodRSA); !ok {
-				return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
+				return nil, fmt.Errorf("Unexpected signing method: %v and not RSASign", token.Header["alg"])
 			}
-		case "ecdsa":
+		case ECDSASing:
```
asoorm: Typo? `ECDSASing` -> `ECDSASign`

mw_jwt.go
```diff
@@ -396,17 +435,17 @@ func (k *JWTMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Request, _
...
 			if _, ok := token.Method.(*jwt.SigningMethodECDSA); !ok {
-				return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
+				return nil, fmt.Errorf("Unexpected signing method: %v and not ECDSASing", token.Header["alg"])
```
asoorm: Typo? `ECDSASing` -> `ECDSASign`

mw_jwt.go
letzya: yep! changed it to that. Before I had some variable..

mw_jwt_test.go
```diff
@@ -86,39 +87,84 @@ func createJWTSessionWithRSAWithPolicy(policyID string) *user.SessionState {
...
+
+			if claimName != KID {
+				t.Claims.(jwt.MapClaims)[claimName] = tokenKID
+				t.Header[KID] = "ignore-this-id"
+			} else {
+				t.Header[KID] = tokenKID
+			}
+		})
+	default:
+		panic("Bad signing method")
```
asoorm: don't know specifics of implementation, but should we panic here? any better way of gracefully dealing with issue?

new.txt
```diff
@@ -0,0 +1,7 @@
+time="May 16 13:03:10" level=info msg="Disabled feature"
```
asoorm: dont need to commit these

prepareJWTSessionRSA
```diff
@@ -0,0 +1,20 @@
+func prepareJWTSessionRSA(testName string) (*APISpec, string) {
```
asoorm: these should not be committed to repo

mw_jwt_test.go
letzya: We can but I'll have to add checks in all the tests to verify `spec` and `token`, since there weren't any before and I haven't added it myself.

prepareJWTSessionRSA
letzya: Yes, thanks, buger had mentioned that on the main thread

mw_jwt.go
```diff
@@ -100,31 +108,41 @@ func (k *JWTMiddleware) getSecretFromURL(url, kid, keyType string) ([]byte, erro
...
+		k.Spec.APIDefinition.JWTSkipCheckKidAsId = true
 	}
 
-	if !idFound && token.Claims.(jwt.MapClaims)["sub"] != nil {
-		tykId = token.Claims.(jwt.MapClaims)["sub"].(string)
-		idFound = true
+	// Check which claim is used for the id - kid or sub header
+	// If is not supposed to ignore KID - will use this as ID if not empty
+	if !k.Spec.APIDefinition.JWTSkipCheckKidAsId {
+		tykId, idFound = token.Header[KID].(string)
```
dencoded: better to do shorter `if tykId, idFound = token.Header[KID].(string); idFound {` if we don't need vars after `if`

mw_jwt.go
```diff
@@ -100,31 +108,41 @@ func (k *JWTMiddleware) getSecretFromURL(url, kid, keyType string) ([]byte, erro
 	return nil, errors.New("No matching KID could be found")
 }
 
-func (k *JWTMiddleware) getIdentityFomToken(token *jwt.Token) (string, bool) {
-	// Try using a kid or sub header
+// Try using a kid or configured claim or sub claim
+func (k *JWTMiddleware) getIdentityFomToken(token *jwt.Token) (string, bool, error) {
 	idFound := false
 	var tykId string
```
dencoded: I suppose we don't need these variables any more as "pass-through" logic was removed/changed below

config/config.go
```diff
@@ -287,6 +287,7 @@ type Config struct {
 	VersionHeader                     string                                `json:"version_header"`
 	EnableHashedKeysListing           bool                                  `json:"enable_hashed_keys_listing"`
 	MinTokenLength                    int                                   `json:"min_token_length"`
+	JWTSkipCheckKidAsId               bool                                  `json:"jwt_use_id_from_kid"`
```
dencoded: json tag and struct field names don't match - looks confusing

mw_jwt.go
```diff
@@ -206,22 +225,41 @@ func (k *JWTMiddleware) getBasePolicyID(token *jwt.Token) (string, bool) {
...
 }
 
+func (k *JWTMiddleware) getUserIdFromClaim(token *jwt.Token) (string, error) {
+	var userId string
+	var found bool
+
+	if k.Spec.JWTIdentityBaseField != "" {
+		userId, found = token.Claims.(jwt.MapClaims)[k.Spec.JWTIdentityBaseField].(string)
+
+		if !found {
```
dencoded: we also have situation here  to return right away
`if found {
    return userID, nil
}`

mw_jwt.go
```diff
@@ -100,31 +108,41 @@ func (k *JWTMiddleware) getSecretFromURL(url, kid, keyType string) ([]byte, erro
...
+func (k *JWTMiddleware) getIdentityFomToken(token *jwt.Token) (string, bool, error) {
 	idFound := false
 	var tykId string
-	if token.Header["kid"] != nil {
-		tykId = token.Header["kid"].(string)
-		idFound = true
+
+	// Global config overrides and sets the api def
+	if k.Spec.GlobalConfig.JWTSkipCheckKidAsId {
+		k.Spec.APIDefinition.JWTSkipCheckKidAsId = true
```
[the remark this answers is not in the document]
letzya: Actually seemed trivial to me. But if I want it on all my apis, because my jwts always contain kid (that I'm using for something else or cannot control it) then this way I can make sure it's always ON. I will not need to keep on setting it every time or worse - will forget to do that and then the jwt will fail and I'll have to update all my apis. 
Also the case of using kid for something else would probably be always applied to all my apis since I'm using the same IdP and same rules.

mw_jwt.go
```diff
@@ -206,22 +225,41 @@ func (k *JWTMiddleware) getBasePolicyID(token *jwt.Token) (string, bool) {
 	return "", false
 }
 
+func (k *JWTMiddleware) getUserIdFromClaim(token *jwt.Token) (string, error) {
+	var userId string
+	var found bool
+
+	if k.Spec.JWTIdentityBaseField != "" {
+		userId, found = token.Claims.(jwt.MapClaims)[k.Spec.JWTIdentityBaseField].(string)
```
dencoded: we could reduce number of type asserts storing `token.Claims.(jwt.MapClaims)` in variable and use after

mw_jwt.go
```diff
@@ -206,22 +225,41 @@ func (k *JWTMiddleware) getBasePolicyID(token *jwt.Token) (string, bool) {
...
+
+		if !found {
+			log.Warning("Base Field not found, falling back to SUB")
+		}
+	}
+	if !found {
+		userId, found = token.Claims.(jwt.MapClaims)[SUB].(string)
```
dencoded: just a general comment about readability - i am kind of worried about expressions containing several asserts and access to map key, hard to read.

also, current code should work, but `found` is not like real sign that sub was found in claims - it is more it was found in claims and it was a string

mw_jwt.go
dencoded: I'd still follow `logrus`' convention to do `log.WithField("userId", userId).Debug("Found User Id")` as its API was done specifically to remove this formatted outputs.

mw_jwt.go
```diff
@@ -206,22 +225,41 @@ func (k *JWTMiddleware) getBasePolicyID(token *jwt.Token) (string, bool) {
...
-		log.Warning("Base Field not found, using SUB")
-		var found bool
-		baseFieldData, found = token.Claims.(jwt.MapClaims)["sub"].(string)
-		if !found {
-			log.Error("ID Could not be generated. Failing Request.")
-			k.reportLoginFailure("[NOT FOUND]", r)
-			return errors.New("Key not authorized"), http.StatusForbidden
-		}
 
+	var baseFieldData string
```
dencoded: extra var declaration

mw_jwt.go
dencoded: Si**n**g -> Si**g**n

mw_jwt.go
```diff
@@ -424,7 +463,7 @@ func (k *JWTMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Request, _
 		if k.Spec.JWTSigningMethod == "rsa" {
```
dencoded: you introduced constant `RSASign` could you please switch the rest of code to use it?

mw_jwt.go
```diff
@@ -100,31 +108,41 @@ func (k *JWTMiddleware) getSecretFromURL(url, kid, keyType string) ([]byte, erro
...
+		return "", false, err
 	}
 
-	log.Debug("Found: ", tykId)
-	return tykId, idFound
+	return tykId, true, nil
 }
 
-func (k *JWTMiddleware) getSecret(token *jwt.Token) ([]byte, error) {
+func (k *JWTMiddleware) getSecretToVerifySignature(token *jwt.Token) ([]byte, error) {
```
dencoded: `getSecretFromToken`? it is not necessary to be used to verify signature

mw_jwt.go
```diff
@@ -424,7 +463,7 @@ func (k *JWTMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Request, _
 		if k.Spec.JWTSigningMethod == "rsa" {
 			asRSA, err := jwt.ParseRSAPublicKeyFromPEM(val)
 			if err != nil {
-				log.Error("Failed to deccode JWT to RSA type")
+				log.Error("Failed to decode JWT to RSA type")
```
dencoded: maybe also log error here: `log.WithError(err).Error("Failed to decode JWT to RSA type")` so we have more context in log

mw_jwt.go
```diff
@@ -456,9 +495,8 @@ func (k *JWTMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Request, _
 	if err != nil {
 		logEntry.Error("JWT validation error: ", err)
 		return errors.New("Key not authorized:" + err.Error()), http.StatusForbidden
-	} else {
-		return errors.New("Key not authorized"), http.StatusForbidden
 	}
+	return errors.New("Key not authorized"), 403
```
dencoded: I think it is better to use constants like `http.StatusForbidden` for magic http-codes, easier to read and reduces number of bugs caused by typos

mw_jwt_test.go
```diff
@@ -86,39 +87,84 @@ func createJWTSessionWithRSAWithPolicy(policyID string) *user.SessionState {
 	return session
 }
 
-// JWTSessionHMAC
+type JwtCreator func() *user.SessionState
+
+var createHmacJWKSession = func() *user.SessionState { return createJWTSession() }
```
dencoded: I am confused why do we need this new funcs and not to use just names `createJWTSession` and `createJWTSessionWithRSA` - I think they qualify to be type of `JwtCreator`

mw_jwt_test.go
```diff
@@ -804,8 +909,36 @@ func BenchmarkJWTSessionRSAWithEncodedJWK(b *testing.B) {
...
+
+	//If we skip the check then the Id will be taken from SUB and the call will succeed
+	_, jwtToken := prepareGenericJWTSession(t.Name(), RSASign, "user-id", false, true)
+	authHeaders := map[string]string{"authorization": jwtToken}
+	t.Run("Request with valid JWT/RSA signature/id in user-id claim", func(t *testing.T) {
+		ts.Run(t, test.TestCase{
+			Headers: authHeaders, Code: http.StatusOK,
+		})
+	})
+}
```
dencoded: case when `getUserIdFromClaim` if falling back to SUB is not covered by tests

mw_jwt_test.go
```diff
@@ -86,39 +87,84 @@ func createJWTSessionWithRSAWithPolicy(policyID string) *user.SessionState {
...
+type JwtCreator func() *user.SessionState
+
+var createHmacJWKSession = func() *user.SessionState { return createJWTSession() }
+var createRsaJWKSession = func() *user.SessionState { return createJWTSessionWithRSA() }
+
+func prepareGenericJWTSession(testName string, method string, claimName string, GlobalSkipKidAsId bool, ApiSkipKidAsId bool) (*APISpec, string) {
+
+	tokenKID := testKey(testName, "token")
+
+	globalConf := config.Global()
```
dencoded: since we change config here - we should not forget to do `defer resetTestConfig()` in all tests where this function is being called (to bring config to initial state)

mw_jwt.go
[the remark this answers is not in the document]
dencoded: but it doesn't give you option to make sure it is always OFF for all your apis - thats confusing for me as well

mw_jwt_test.go
letzya: fixed

mw_jwt_test.go
letzya: ok

mw_jwt_test.go
letzya: Added TestJWTRSAIdInClaimsWithBaseField, TestJWTRSAIdInClaimsWithoutBaseField.

letzya:
@dencoded  Have replied all your comment and checked in the changes. I think there were 2 comments I have answered you without making code change.

mw_jwt.go
letzya: ok

mw_jwt.go
letzya: Yes. Have you seen my detailed answer? can't see it here any longer.

mw_jwt.go
letzya: Have fixed it.

letzya:
@dencoded Have replied all your comment and checked in the changes. I think there were 2 comments I have answered you without making code change.

letzya:
Have fully tested the feature.
1. Tested centralised JWT - 
 - With default claim 'sub'  - request pass - GOOD
 - With changing it to 'subject' (anything different from 'sub') - request pass - GOOD
 - Setting 'kid' in the header does not make a difference
2. Individual JWT secrets - 
 - `"jwt_skip_check_kid_as_id": false` 
    - The token (created a key and put my public key in it) in 'kid' - request pass - GOOD
    - The token in 'sub' and 'kid' has another value - request fails - GOOD
 -`"jwt_skip_check_kid_as_id": true`
   -  The token in 'kid' - request fails - GOOD
   -  The token in 'sub' - request pass - GOOD
   -  The token in 'subject', api definition expect 'subject' claim - request pass - GOOD

letzya:
@buger will resolve the conflicts once you know when you can merge.

Issue #1664, opened 2018-05-03
profile performance of BenchmarkVersioning


dencoded:
Let’s focus on significant performance improvements, the ones with less than 1%-5% difference can be just statistical noise (consider to increase bench time with flag `-benchtime=…` in case of low throughput).

Need to do profiling of all types (cpu, memory, block and mutex). Provide `benchcmp` results for benchmark output - before and after improvement (or prove that this piece of code cannot be improved in any ways).

matiasinsaurralde:
In addition to #1664 I've been tracing the versioning-related calls, if you have an API with a few extended paths, these are the main functions that get called over and over, perhaps it's possible to optimize something here (fewer calls? we already keep the versioning data in the request context):
```
- RequestValid is called from github.com/TykTechnologies/tyk.(*VersionCheck).ProcessRequest
- Version is called from github.com/TykTechnologies/tyk.(*APISpec).RequestValid
- URLAllowedAndIgnored is called from github.com/TykTechnologies/tyk.(*APISpec).RequestValid
- Version is called from github.com/TykTechnologies/tyk.(*TransformHeaders).ProcessRequest
- CheckSpecMatchesStatus is called from github.com/TykTechnologies/tyk.(*TransformHeaders).ProcessRequest
- Version is called from github.com/TykTechnologies/tyk.(*URLRewriteMiddleware).ProcessRequest
- CheckSpecMatchesStatus is called from github.com/TykTechnologies/tyk.(*URLRewriteMiddleware).ProcessRequest
- Version is called from github.com/TykTechnologies/tyk.(*ReverseProxy).CheckHeaderInRemoveList
- CheckSpecMatchesStatus is called from github.com/TykTechnologies/tyk.(*ReverseProxy).CheckHeaderInRemoveList
```
Also `reverse_proxy.go` calls `Version` three different times as part of `WrappedServeHTTP`, it should be possible to refactor the helper functions (`CheckHardTimeoutEnforced`, etc.) in order to avoid these extra calls.
I've shared #1677 to include `DoMockReply` in our tests/benchmarks.

PR #1690, opened 2018-05-09
Versioning enhancements

matiasinsaurralde:
For #1664.


buger:
@matiasinsaurralde does it still a WIP? Do you want to include bencmp output?

matiasinsaurralde:
I'm still testing this, have found that my whitelist cache code was making things slower, currently `BenchmarkDefaultVersion` is around 8% faster than `master` but `GetVersionFromRequest` calls can be improved.

buger:
@matiasinsaurralde this PR becomes very old, pls rebase it against latest master and let's merge it.

matiasinsaurralde:
Updated!

buger:
Merged manually

PR #1695, opened 2018-05-11
ID extractor enhancements: extended tests, benchmarks and BodyExtractor fix

matiasinsaurralde:
- I've renamed the previous `coprocess_id_extractor_test.go` in order to allow ID extractor testing on normal builds, this also allows us to benchmark ID extractor without the `cgo` overhead.
To allow this ID extractor references were removed from `coprocess_dummy.go`.
- The ID extractor Python tests will remain the same and run when the build tag is used.
- I've added new tests and benchmarks for every extractor type and source combination.
- Have fixed the `BodyExtractor` as its implementation was missing, this is covered by the new tests too.
- Have removed unused fields from `XPathExtractorConfig`.

This will enhance ID extractor benchmarking for #1629 and #1624.
Other potential enhancements that this allows:
- Supporting ID extractor on JSVM middleware.
- Test/benchmark gRPC ID extractor integration separately.

### DIFF
--- a/coprocess_dummy.go
+++ b/coprocess_dummy.go
@@ -61,5 +61,3 @@ func CoProcessInit() error {
 }
 
 func doCoprocessReload() {}
-
-func newExtractor(referenceSpec *APISpec, mw BaseMiddleware) {}
--- a/coprocess_id_extractor.go
+++ b/coprocess_id_extractor.go
@@ -1,11 +1,10 @@
-// +build coprocess
-
 package main
 
 import (
 	"crypto/md5"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"regexp"
 	"strings"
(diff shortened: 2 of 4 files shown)


buger:
Looks great! 

Can you extend a bit on:
> Supporting ID extractor on JSVM middleware.

Can we already do it with this change, and if so, can you write some docs?

matiasinsaurralde:
@buger Thanks, it's not possible to use with JSVM yet but should be possible adding some of these lines: https://github.com/TykTechnologies/tyk/blob/master/coprocess.go#L238-L256
Would be good to unify this code a bit!

buger:
@matiasinsaurralde can you create task about allowing JSVM id extractor and auth plugins, and propose it for 2.8 (in product ticket)? Thank you!

PR #1696, opened 2018-05-14
bench: caching bcrypt function for auth

### LINKED ISSUE #1622: mw_basic_auth is slow
bug

APIs with basic auth is very slow (with cap at about 10-20 req/sec )

APIs protected by basic auth should be fast

Create API protected by basic auth and try to send some loading >10 reqs/sec

all

screen shot 2018-04-26 at 1 29 59 pm

### LINKED ISSUE #1640: profile performance of BenchmarkBasicAuth
Let’s focus on significant performance improvements, the ones with less than 1%-5% difference can be just statistical noise (consider to increase bench time with flag `-benchtime=…` in case of low throughput).

Need to do profiling of all types (cpu, memory, block and mutex). Provide `benchcmp` results for benchmark output - before and after improvement (or prove that this piece of code cannot be improved in any ways).

asoorm:
Fixes #1622 Fixes #1640

```
BenchmarkBasicAuth_CacheEnabled-4           3000            369157 ns/op           72023 B/op        451 allocs/op
BenchmarkBasicAuth_CacheDisabled-4            20          66711644 ns/op          509487 B/op       9704 allocs/op
```

Current Master:

```
BenchmarkBasicAuth-4          10         135002061 ns/op         1035518 B/op      19855 allocs/op
```

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -329,8 +329,12 @@ type APIDefinition struct {
 		AllowedAuthorizeTypes  []osin.AuthorizeRequestType `bson:"allowed_authorize_types" json:"allowed_authorize_types"`
 		AuthorizeLoginRedirect string                      `bson:"auth_login_redirect" json:"auth_login_redirect"`
 	} `bson:"oauth_meta" json:"oauth_meta"`
-	Auth                          Auth                 `bson:"auth" json:"auth"`
-	UseBasicAuth                  bool                 `bson:"use_basic_auth" json:"use_basic_auth"`
+	Auth         Auth `bson:"auth" json:"auth"`
+	UseBasicAuth bool `bson:"use_basic_auth" json:"use_basic_auth"`
+	BasicAuth    struct {
+		DisableCaching bool `bson:"disable_caching" json:"disable_caching"`
+		CacheTTL       int  `bson:"cache_ttl" json:"cache_ttl"`
+	} `bson:"basic_auth" json:"basic_auth"`
 	UseMutualTLSAuth              bool                 `bson:"use_mutual_tls_auth" json:"use_mutual_tls_auth"`
 	ClientCertificates            []string             `bson:"client_certificates" json:"client_certificates"`
 	UpstreamCertificates          map[string]string    `bson:"upstream_certificates" json:"upstream_certificates"`
--- a/mw_basic_auth.go
+++ b/mw_basic_auth.go
@@ -5,22 +5,30 @@ import (
 	"errors"
 	"net/http"
 	"strings"
+	"time"
 
+	"github.com/Sirupsen/logrus"
+	"github.com/pmylund/go-cache"
+	"github.com/spaolacci/murmur3"
 	"golang.org/x/crypto/bcrypt"
 
 	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/user"
 )
 
+const defaultBasicAuthTTL = time.Duration(60) * time.Second
+
 // BasicAuthKeyIsValid uses a username instead of
 type BasicAuthKeyIsValid struct {
 	BaseMiddleware
+	cache *cache.Cache
 }
 
 func (k *BasicAuthKeyIsValid) Name() string {
 	return "BasicAuthKeyIsValid"
 }
 
+// EnabledForSpec checks if UseBasicAuth is set in the API definition.
 func (k *BasicAuthKeyIsValid) EnabledForSpec() bool {
 	return k.Spec.UseBasicAuth
 }
@@ -75,39 +83,22 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
 	if !keyExists {
 		logEntry.Info("Attempted access with non-existent user.")
 
-		// Fire Authfailed Event
-		AuthFailed(k, r, token)
-
-		// Report in health check
-		reportHealthValue(k.Spec, KeyFailure, "-1")
-
-		return k.requestForBasicAuth(w, "User not authorised")
+		return k.handleAuthFail(w, r, token)
 	}
 
-	// Ensure that the username and password match up
-	var passMatch bool
 	switch session.BasicAuthData.Hash {
 	case user.HashBCrypt:
-		err := bcrypt.CompareHashAndPassword([]byte(session.BasicAuthData.Password), []byte(authValues[1]))
-		if err == nil {
-			passMatch = true
+
+		if err := k.compareHashAndPassword(session.BasicAuthData.Password, authValues[1], logEntry); err != nil {
+			logEntry.Warn("Attempted access with existing user, failed password check.")
+			return k.handleAuthFail(w, r, token)
 		}
 	case user.HashPlainText:
-		if session.BasicAuthData.Password == authValues[1] {
-			passMatch = true
+		if session.BasicAuthData.Password != authValues[1] {
+
+			logEntry.Warn("Attempted access with existing user, failed password check.")
+			return k.handleAuthFail(w, r, token)
 		}
-	}
-
-	if !passMatch {
-		logEntry.Info("Attempted access with existing user but failed password check.")
-
-		// Fire Authfailed Event
-		AuthFailed(k, r, token)
-
-		// Report in health check
-		reportHealthValue(k.Spec, KeyFailure, "-1")
-
-		return k.requestForBasicAuth(w, "User not authorised")
 	}
 
 	// Set session state on context, we will need it later
@@ -116,6 +107,65 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
 		ctxSetSession(r, &session, keyName, false)
 	}
 
-	// Request is valid, carry on
 	return nil, http.StatusOK
+}
+
+func (k *BasicAuthKeyIsValid) handleAuthFail(w http.ResponseWriter, r *http.Request, token string) (error, int) {
+
+	// Fire Authfailed Event
+	AuthFailed(k, r, token)
+
+	// Report in health check
+	reportHealthValue(k.Spec, KeyFailure, "-1")
+
+	return k.requestForBasicAuth(w, "User not authorised")
+}
+
+func (k *BasicAuthKeyIsValid) doBcryptWithCache(cacheDuration time.Duration, hashedPassword []byte, password []byte) error {
+	if err := bcrypt.CompareHashAndPassword(hashedPassword, password); err != nil {
+
+		return err
+	}
+
+	hasher := murmur3.New32()
+	hasher.Write(password)
+	k.cache.Set(string(hashedPassword), string(hasher.Sum(nil)), cacheDuration)
+
+	return nil
+}
+
+func (k *BasicAuthKeyIsValid) compareHashAndPassword(hash string, password string, logEntry *logrus.Entry) error {
+
+	cacheEnabled := !k.Spec.BasicAuth.DisableCaching
+	passwordBytes := []byte(password)
+	hashBytes := []byte(hash)
+
+	if !cacheEnabled {
+
+		logEntry.Debug("cache disabled")
+		return bcrypt.CompareHashAndPassword(hashBytes, passwordBytes)
+	}
+
+	cacheTTL := defaultBasicAuthTTL // set a default TTL, then override based on BasicAuth.CacheTTL
+	if k.Spec.BasicAuth.CacheTTL > 0 {
+		cacheTTL = time.Duration(k.Spec.BasicAuth.CacheTTL) * time.Second
+	}
+
+	cachedPass, inCache := k.cache.Get(hash)
+	if !inCache {
+
+		logEntry.Debug("cache enabled: miss: bcrypt")
+		return k.doBcryptWithCache(cacheTTL, hashBytes, passwordBytes)
+	}
+
+	hasher := murmur3.New32()
+	hasher.Write(passwordBytes)
+	if cachedPass.(string) != string(hasher.Sum(nil)) {
+
+		logEntry.Warn("cache enabled: hit: failed auth: bcrypt")
+		return bcrypt.CompareHashAndPassword(hashBytes, passwordBytes)
+	}
+
+	logEntry.Debug("cache enabled: hit: success")
+	return nil
 }
--- a/api_loader.go
+++ b/api_loader.go
@@ -6,10 +6,12 @@ import (
 	"net/url"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/gorilla/mux"
 	"github.com/justinas/alice"
+	"github.com/pmylund/go-cache"
 
 	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/config"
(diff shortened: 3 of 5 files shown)


mw_basic_auth.go
```diff
@@ -84,30 +91,37 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
-		}
-	}
+		cacheKey := session.BasicAuthData.Password + authValues[1]
+		// check if username/password combination in cache
+		val, isCached := k.cache.Load(cacheKey)
+		if isCached {
 
-	if !passMatch {
-		logEntry.Info("Attempted access with existing user but failed password check.")
+			log.Debug("cache hit: ", cacheKey)
```
letzya: Should we print cashed keys? (even in debug mode)?

asoorm: This is for me to test locally that my cache is working as expected.

mw_basic_auth.go
```diff
@@ -117,6 +111,66 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+
+	return k.requestForBasicAuth(w, "User not authorised")
+}
+
+func (k *BasicAuthKeyIsValid) compareHashAndPassword(hash string, password string, logEntry *logrus.Entry) error {
+
+	cacheEnabled := true
+	cacheDuration := time.Duration(60) * time.Second
+	hasher := murmur3.New32()
+
```
matiasinsaurralde: Perhaps add `passwordBytes := []byte(password)` and avoid subsequent `[]byte()` calls?

mw_basic_auth.go
```diff
@@ -137,24 +139,23 @@ func (k *BasicAuthKeyIsValid) compareHashAndPassword(hash string, password strin
 		}
 
 		if cacheEnabled {
+
 			logEntry.Debug("cache enabled: save")
-			k.cache.Set(hashedPassword, password, cacheDuration)
+			k.cache.Set(hashedPassword, string(hasher.Sum([]byte(password))), cacheDuration)
```
dencoded: I think `hasher.Sum([]byte(password))` will give you the same `password`,
need to be
```
hasher := New32()
hasher.Write(password)
hasher.Sum32()
k.cache.Set(hashedPassword, string(hasher.Sum(nil)), cacheDuration)
```

asoorm: Yeah - I think i messed that up... good spot

apidef/api_definitions.go
```diff
@@ -326,8 +326,12 @@ type APIDefinition struct {
 		AllowedAuthorizeTypes  []osin.AuthorizeRequestType `bson:"allowed_authorize_types" json:"allowed_authorize_types"`
 		AuthorizeLoginRedirect string                      `bson:"auth_login_redirect" json:"auth_login_redirect"`
 	} `bson:"oauth_meta" json:"oauth_meta"`
-	Auth                          Auth                 `bson:"auth" json:"auth"`
-	UseBasicAuth                  bool                 `bson:"use_basic_auth" json:"use_basic_auth"`
+	Auth         Auth `bson:"auth" json:"auth"`
+	UseBasicAuth bool `bson:"use_basic_auth" json:"use_basic_auth"`
+	BasicAuth    struct {
+		DisableCaching bool `bson:"enable_caching" json:"enable_caching"`
```
joshblakeley: Should be `EnableCaching` ?

asoorm: We want caching to be enabled by default - so a user needs to disable caching explicitly

joshblakeley: Actually looking at rest of code i think the bson and json tags need changing to `disable_caching`

mw_basic_auth.go
```diff
@@ -117,6 +111,66 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+		if cacheEnabled {
+
+			logEntry.Debug("cache enabled: save")
+			k.cache.Set(hashedPassword, string(hasher.Sum([]byte(password))), cacheDuration)
+		}
+
+		return nil
+	}
+
+	if !cacheEnabled {
```
dencoded: I think this can be done as earlier return at the beginning of func

mw_basic_auth.go
```diff
@@ -117,6 +111,66 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+	AuthFailed(k, r, token)
+
+	// Report in health check
+	reportHealthValue(k.Spec, KeyFailure, "-1")
+
+	return k.requestForBasicAuth(w, "User not authorised")
+}
+
+func (k *BasicAuthKeyIsValid) compareHashAndPassword(hash string, password string, logEntry *logrus.Entry) error {
+
```
dencoded: as a general comment - from where do these "JS-style" empty lines after func name come? (some lint?)

mw_basic_auth.go
```diff
@@ -117,6 +111,66 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+	cacheDuration := time.Duration(60) * time.Second
+	if k.Spec.BasicAuth.CacheTTL > 0 {
+		cacheDuration = time.Duration(k.Spec.BasicAuth.CacheTTL) * time.Second
+	}
+
+	hasher := murmur3.New32()
+	passwordBytes := []byte(password)
+
+	doBcrypt := func(cacheEnabled bool, cacheDuration time.Duration, hashedPassword string, password []byte) error {
+		if err := bcrypt.CompareHashAndPassword([]byte(hash), password); err != nil {
```
dencoded: one confusing thing - shouldn't this be skipped if `cacheEnabled` is true?

mw_basic_auth.go
```diff
@@ -117,6 +111,66 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+	cacheEnabled := !k.Spec.BasicAuth.DisableCaching
+	cacheDuration := time.Duration(60) * time.Second
+	if k.Spec.BasicAuth.CacheTTL > 0 {
+		cacheDuration = time.Duration(k.Spec.BasicAuth.CacheTTL) * time.Second
+	}
+
+	hasher := murmur3.New32()
+	passwordBytes := []byte(password)
+
+	doBcrypt := func(cacheEnabled bool, cacheDuration time.Duration, hashedPassword string, password []byte) error {
```
dencoded: what the purpose of this function literal? the reason I am asking - I don't see it passed anywhere as param or any closure to outer-scoped variables

mw_basic_auth.go
```diff
@@ -117,6 +111,66 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+
+	// Report in health check
+	reportHealthValue(k.Spec, KeyFailure, "-1")
+
+	return k.requestForBasicAuth(w, "User not authorised")
+}
+
+func (k *BasicAuthKeyIsValid) compareHashAndPassword(hash string, password string, logEntry *logrus.Entry) error {
+
+	cacheEnabled := !k.Spec.BasicAuth.DisableCaching
```
dencoded: we could do bcrypt and return here right away

mw_basic_auth.go
```diff
@@ -117,6 +111,66 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+	cacheDuration := time.Duration(60) * time.Second
+	if k.Spec.BasicAuth.CacheTTL > 0 {
+		cacheDuration = time.Duration(k.Spec.BasicAuth.CacheTTL) * time.Second
+	}
+
+	hasher := murmur3.New32()
+	passwordBytes := []byte(password)
```
dencoded: I think you forgot to calculate hash with:
```
hasher.Write(passwordBytes)
hasher.Sum32()
```
hasher methods are confusing

mw_basic_auth.go
```diff
@@ -117,6 +111,66 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+	}
+
+	cachedPass, inCache := k.cache.Get(hash)
+	if !inCache {
+
+		logEntry.Debug("cache enabled: miss: bcrypt failover")
+		return doBcrypt(cacheEnabled, cacheDuration, hash, passwordBytes)
+	}
+
+	if cachedPass.(string) != string(hasher.Sum([]byte(password))) {
```
dencoded: I think this needs to be `cachedPass.(string) != string(hasher.Sum(nil))`

`hasher.Sum` doesn't calculate hash, it returns current state of hash (which needs to be calculated with method `hasher.Sum` earlier) with passed bytes appended to result

dencoded:
left some comments, also it needs to be rebased on master

mw_basic_auth.go
asoorm: It's been a habit that I've had for many years - I think it looks neater. Is not just js... PEP-8, PSR-2 etc.

Also for golang, because `{` is in-line with function, unlike k&r style, it looks clearer in my opinion.

mw_basic_auth.go
asoorm: we need to call `doBcrypt` 3 times in the function. If you look at previous way was implemented, all of content from the function literal was being repeated many times. I didn't want to create pointer receiver function on the struct, because this is the only place that it is used.

I didn't want to place as separate pointer receiver, as is not used elsewhere.

Also, it makes the business logic bit of the code much easier to read & understand.

mw_basic_auth.go
asoorm: We pretty much do. but if we did this, then hasher & doBcrypt function would need to be declared eariler, and would make no difference?

mw_basic_auth.go
```diff
@@ -117,6 +111,66 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+		if err := bcrypt.CompareHashAndPassword([]byte(hash), password); err != nil {
+
+			logEntry.Debug("bcrypt: failed auth")
+			return err
+		}
+
+		if cacheEnabled {
+
+			logEntry.Debug("cache enabled: save")
+			k.cache.Set(hashedPassword, string(hasher.Sum([]byte(password))), cacheDuration)
```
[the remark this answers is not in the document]
asoorm: Will look into

mw_basic_auth.go
asoorm: Will look into

mw_basic_auth.go
asoorm: nop - because if not in cache, or hash doesn't match from cache, then it fails-over to bcrypt as src of truth. doBcrypt also populates the cache if cache enabled

mw_basic_auth.go
dencoded: well I never seen that in Go code and it is not part of `gofmt`, from other point of view `gofmt` doesn't remove those empty lines

mw_basic_auth.go
```diff
@@ -75,39 +81,22 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
-	// Ensure that the username and password match up
-	var passMatch bool
 	switch session.BasicAuthData.Hash {
 	case user.HashBCrypt:
-		err := bcrypt.CompareHashAndPassword([]byte(session.BasicAuthData.Password), []byte(authValues[1]))
-		if err == nil {
-			passMatch = true
+
+		if err := k.compareHashAndPassword(session.BasicAuthData.Password, authValues[1], logEntry); err != nil {
+			logEntry.Info("Attempted access with existing user, failed password check.")
```
dencoded: maybe `Warning` as it requires special attention

mw_basic_auth.go
```diff
@@ -117,6 +106,67 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+}
+
+func (k *BasicAuthKeyIsValid) doBcryptWithCache(cacheDuration time.Duration, hashedPassword []byte, password []byte) error {
+	if err := bcrypt.CompareHashAndPassword(hashedPassword, password); err != nil {
+
+		return err
+	}
+
+	hasher := murmur3.New32()
+	hasher.Write([]byte(password))
```
dencoded: password is already `[]byte`

mw_basic_auth.go
```diff
@@ -117,6 +106,67 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+	hasher.Sum32()
```
dencoded: we don't need to do `hasher.Sum32()` as it is done inside `hasher.Sum(nil)`

mw_basic_auth.go
```diff
@@ -117,6 +106,67 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+	passwordBytes := []byte(password)
+	hashBytes := []byte(hash)
+
+	if !cacheEnabled {
+
+		logEntry.Debug("cache disabled")
+		return bcrypt.CompareHashAndPassword(hashBytes, passwordBytes)
+	}
+
+	cacheDuration := time.Duration(60) * time.Second // set a default TTL, then override based on BasicAuth.CacheTTL
```
dencoded: this can be done as constant instead of calculating it every time

mw_basic_auth.go
```diff
@@ -117,6 +106,67 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+
+	cachedPass, inCache := k.cache.Get(hash)
+	if !inCache {
+
+		logEntry.Debug("cache enabled: miss: bcrypt failover")
+		return k.doBcryptWithCache(cacheDuration, hashBytes, passwordBytes)
+	}
+
+	hasher := murmur3.New32()
+	hasher.Write([]byte(password))
```
dencoded: supposed to use `passwordBytes`?

mw_basic_auth.go
```diff
@@ -117,6 +106,67 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+	hasher.Sum32()
```
dencoded: we don't need this `hasher.Sum32()` as it is done as part of `hasher.Sum(nil)`

mw_basic_auth.go
```diff
@@ -117,6 +106,67 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
...
+	if cachedPass.(string) != string(hasher.Sum(nil)) {
+
+		logEntry.Debug("cache enabled: hit: invalid: bcrypt failover")
+		return k.doBcryptWithCache(cacheDuration, hashBytes, passwordBytes)
```
dencoded: if it is in cache and doesn't match provided password why not to do just `bcrypt.CompareHashAndPassword(hashBytes, passwordBytes)` directly? (and eliminate one function call)  I mean there is no way `doBcryptWithCache` can put anything in cache at this point

asoorm: If user password changes, then we could get a cache hit - but an invalid password.
We call the `doBrcyptWithCache` function to re-set the cache with latest password.

dencoded: if user password changed wouldn't `compareHashAndPassword` be called with new hash and new password? I mean in case of new password `session.BasicAuthData.Password` will contain new bcrypt-hash and there is no way to get cache hit. I think new password situation won't reach this piece of logic, if I am not missing anything here.

dencoded:
@asoorm bump

asoorm:
```
benchmark                old ns/op     new ns/op     delta
BenchmarkBasicAuth-4     134896502     70577659      -47.68%

benchmark                old allocs     new allocs     delta
BenchmarkBasicAuth-4     19842          10522          -46.97%

benchmark                old bytes     new bytes     delta
BenchmarkBasicAuth-4     1033905       594978        -42.45%
```

dencoded:
@asoorm looks pretty cool

PR #1698, opened 2018-05-15
changed 403 to 401 for unauthorised key in oidc

### LINKED ISSUE #1503: OpenID Connect, invalid/expired token returning 403 instead of 401
bug
When a token is expired, gateway returns 403
Should return 401 as specified here: https://tools.ietf.org/html/rfc6750#section-3.1

Note: A valid token with insufficient privileges should return 403

All

joshblakeley:
Fixes #1503

### DIFF
--- a/mw_openid.go
+++ b/mw_openid.go
@@ -104,7 +104,7 @@ func (k *OpenIDMW) ProcessRequest(w http.ResponseWriter, r *http.Request, _ inte
 	if halt {
 		// Fire Authfailed Event
 		k.reportLoginFailure("[JWT]", r)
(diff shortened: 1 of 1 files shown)


joshblakeley:
The change here is very naive. 
1. This could break back compatibility of someone relies on the outputted code.
2. Could make it configurable - but is this too small to bother?
3. Although all these returns say unauthorised, which would warrant a 401 according to the OID RFC, i'm not completely sure how they mirror the underlying OIC libs errors.

buger:
Looks good, but let's make this change for 2.7 instead

PR #1710, opened 2018-05-18
bug: EOF when performing response body transform

asoorm:
fixes: #1729

Description of bug:

1. Create an API with a mock XML response e.g. `/api-a/mock`.
2. Create an API which reverse proxies to `api-a` called `abi-b`.
3. perform a response body transform on `api-b/mock` endpoint. e.g.
convert to json.

Bug:

```
ERROR outbound-transform: Error unmarshalling XML:
xml.Decoder.Token() - XML syntax error on line 167: unexpected EOF
api_id=90a1e135513946cb437b5e5111422227 path=mock
server_name=https://tyk-gateway.dev:8080/api-a/
```

This occurs because the body looks like it has already been read.

Fix: reset the buffer by taking a copy of the http.Response.

### DIFF
--- a/res_handler_transform.go
+++ b/res_handler_transform.go
@@ -27,6 +27,7 @@ func (h *ResponseTransformMiddleware) Init(c interface{}, spec *APISpec) error {
 }
 
 func respBodyReader(req *http.Request, resp *http.Response) io.ReadCloser {
+
 	if req.Header.Get("Accept-Encoding") == "" {
 		return resp.Body
 	}
@@ -64,6 +65,14 @@ func compressBuffer(in bytes.Buffer, encoding string) (out bytes.Buffer) {
 }
 
 func (h *ResponseTransformMiddleware) HandleResponse(rw http.ResponseWriter, res *http.Response, req *http.Request, ses *user.SessionState) error {
+
+	logger := log.WithFields(logrus.Fields{
+		"prefix":      "outbound-transform",
+		"server_name": h.Spec.Proxy.TargetURL,
+		"api_id":      h.Spec.APIID,
+		"path":        req.URL.Path,
+	})
+
 	_, versionPaths, _, _ := h.Spec.Version(req)
 	found, meta := h.Spec.CheckSpecMatchesStatus(req, versionPaths, TransformedResponse)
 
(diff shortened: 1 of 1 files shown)


asoorm:
I'm not 100% happy with this fix. I don't think it's the right fix because it looks wrong. but any advice on the right fix, or if you are happy with this PR would be much appreciated.

buger:
@asoorm so the main question is where body gets read twice, right?

asoorm:
@buger I don't think that the body gets read twice. If it was, then the `copyResponse` fix simply would not work. example:

Code:

```
package main

import (
	"bytes"
	"io"
	"log"
	"net/http"
	"os"
)

func main() {

	res, err := http.Get("http://httpbin.org/xml")
	if err != nil {
		log.Fatalf("unable to create new request: %v", err)
	}

	log.Println("oldRes - START")
	io.Copy(os.Stdout, res.Body)
	log.Println("oldRes - END")

	newRes := copyResponse(res)

	log.Println("newRes - START")
	io.Copy(os.Stdout, newRes.Body)
	log.Println("newRes - END")
}

func copyResponse(r *http.Response) *http.Response {
	r2 := *r
	if r.Body != nil {
		r.Body, r2.Body = copyBody(r.Body)
	}
	return &r2
}

func copyBody(body io.ReadCloser) (b1, b2 io.ReadCloser) {
	if nc, ok := body.(nopCloser); ok {
		buf := *(nc.Reader.(*bytes.Buffer))
		return body, nopCloser{&buf}
	}
	defer body.Close()

	var buf1, buf2 bytes.Buffer
	io.Copy(&buf1, body)
	buf2 = buf1
	return nopCloser{&buf1}, nopCloser{&buf2}
}

// nopCloser is just like ioutil's, but here to let us fetch the
// underlying io.Reader.
type nopCloser struct {
	io.Reader
}

func (nopCloser) Close() error { return nil }
```

Output:

```
$ go19 run main.go 
2018/05/18 13:37:39 oldRes - START
<?xml version='1.0' encoding='us-ascii'?>

<!--  A SAMPLE set of slides  -->

<slideshow 
    title="Sample Slide Show"
    date="Date of publication"
    author="Yours Truly"
    >

    <!-- TITLE SLIDE -->
    <slide type="all">
      <title>Wake up to WonderWidgets!</title>
    </slide>

    <!-- OVERVIEW -->
    <slide type="all">
        <title>Overview</title>
        <item>Why <em>WonderWidgets</em> are great</item>
        <item/>
        <item>Who <em>buys</em> WonderWidgets</item>
    </slide>

</slideshow>2018/05/18 13:37:39 oldRes - END
2018/05/18 13:37:39 newRes - START
2018/05/18 13:37:39 newRes - END
```

buger:
I'm not sure if fully understand why this fix works. Can you explain?

asoorm:
To be honest, I don't understand why the fix works either. Need help please.

dencoded:
@buger @asoorm I've been looking into this issue as well. I found the problem is that xml parser can't work with `ReadCloser` when it actually is `net/http`'s `bodyEOFSignal` - it is always producing `EOF` unless it is already seen.

Also, I've noticed that our  `VersionCheck` MW for mocked endpoints producing Chunked reply without providing finalizing chunk properly, so connection stays open and xml parser bumps into EOF at some point. As potential fix we could add final chunk in response:
```
diff --git a/mw_version_check.go b/mw_version_check.go
index f444d59..6dc791f 100644
--- a/mw_version_check.go
+++ b/mw_version_check.go
@@ -26,7 +26,7 @@ func (v *VersionCheck) Name() string {
 func (v *VersionCheck) DoMockReply(w http.ResponseWriter, meta interface{}) {
        // Reply with some alternate data
        emeta := meta.(*apidef.EndpointMethodMeta)
-       responseMessage := []byte(emeta.Data)
+       responseMessage := []byte(emeta.Data + "\r\n0\r\n\r\n")
        for header, value := range emeta.Headers {
                w.Header().Add(header, value)
        }
```

As another fix - in `func (h *ResponseTransformMiddleware) HandleResponse` we can read whole body 1st and then parse like we do in `TransformMiddleware`:
```
respBody, err := ioutil.ReadAll(r.Body)
...
defer respBody.Close()
...
bodyData, err = mxj.NewMapXml(body)
...
```

buger:
Point about chunk encoding is interesting, but this is TCP level, and Go should handle it by itself. More over chunked encoding turned on optionally, depending on body size. 

Fix with reading whole body first make sense.

dencoded:
also we should probably re-vendor github.com/clbanning/mxj as it contains some optimizations and fixes (our version is about one year old)

asoorm:
@dencoded thanks for help on this. PR modified to use `ioutil.ReadAll` rather than `copyResponse`. Ready for re-review.

buger:
@asoorm looks good, but are you sure that JSON part does not affected? Maybe make sense `readall` body for all cases including JSON?

Also do you think we should apply same logic to request body transformation?

asoorm:
@buger JSON is not affected. It looks like the problem is with `mxj.NewMapXmlReader`, because it cannot work with `ReadCloser`. That' why I only read the bytes for XML rather than everything.

@dencoded looked at possibility of upgrading `mxj`, but there are quite a few changes to the lib, and so for purposes of this PR - this is the backwards compatible workaround until we have opportunity to consider replacing/upgrading mxj.

PR #1726, opened 2018-05-22
Handles Dashboard API failures, fixes #1725

### LINKED ISSUE #1725: A failed API response from the dashboard can stop gateway routing correctly
bug

When a gateway pulls definitions form the dashboard and the request fails for some reason, and there is an existing API spec list in memory, it will replace that list instead of retaining the pre-failed version.

On a login or request failure to the dashboard API, the gateway should retain the last-good API set to ensure availability of services

1. Set up a gateway with a dashboard, with 1 API
2. Ensure the gateway loads the API
3. Mock the API Definition dashboard endpoint with a fake one and return a malformed JSON message
4. The gateway will throw an error and then all APIs that were running will be unloaded

v.2.6.x, master

lonelycode:
Have added a test that simulates a running gateway with 1 definition, and then that same gateway pulling a malformed definition

The gateway will default to what is already loaded.

This introduces the potential for gateways to get out-of-step, but ensures existing services stay loaded until there has been a successful reload.

Perhaps a future improvement would be a signal to the dashboard notifying which gateways have successfully reloaded

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -20,6 +20,7 @@ import (
 
 	"sync"
 
+	"fmt"
 	"github.com/TykTechnologies/gojsonschema"
 	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/config"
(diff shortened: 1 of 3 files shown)


lonelycode:
Forgot to add ticket number in message, fixes #1725

buger:
Looks good. Will fix CI error (seems like its mine).

Issue #1724, opened 2018-05-22
Non-UTF8 in request bodies when using rich plugins


matiasinsaurralde:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When using rich plugins, we currently copy the request body and convert it to string, this breaks the PB decoding in Python when sending an attachment.
This is related to #351 and #778.

**What is the expected behavior?**
We need to make sure that this type of request body works fine.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** See #351 but use it with Python plugins.

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?** All

matiasinsaurralde:
Have been investigating this and seems the issue occurs with PB so it's also present on gRPC plugins, this is what an object looks like in the file attachment scenario:
```
[...]

PR #1727, opened 2018-05-23
coprocess: return error when proto.Marshal fails

matiasinsaurralde:
This is already present for gRPC but will enhance error handling when using Python plugins, related to #1724


matiasinsaurralde:
Have updated this with additional error checks, also modified some of the `C.free` calls

PR #1728, opened 2018-05-23
[WIP] gRPC tests and benchmarks

matiasinsaurralde:
This is part of #1674, currently `benchcmp` looks like this:
I might drop the `TestCase` usage for benchmarks as it might affect the results, the dispatcher seems to perform a bit faster (are the allocations made by the test case code? still investigating this):
```
% benchcmp old new
benchmark                                            old ns/op     new ns/op     delta
BenchmarkGRPCDispatch/Pre_Hook_with_SetHeaders-8     558806        526713        -5.74%

benchmark                                            old allocs     new allocs     delta
BenchmarkGRPCDispatch/Pre_Hook_with_SetHeaders-8     669            803            +20.03%

benchmark                                            old bytes     new bytes     delta
BenchmarkGRPCDispatch/Pre_Hook_with_SetHeaders-8     84621         89969         +6.32%
```
I will add more scenarios and also an API with multiple hooks.

### DIFF
--- a/coprocess_grpc_test.go
+++ b/coprocess_grpc_test.go
@@ -1,0 +1,153 @@
+// +build coprocess
+// +build grpc
+
+package main
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"testing"
+
+	"context"
+
+	"google.golang.org/grpc"
+
+	"github.com/TykTechnologies/tyk/apidef"
+	"github.com/TykTechnologies/tyk/config"
+	"github.com/TykTechnologies/tyk/coprocess"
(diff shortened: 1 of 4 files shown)


matiasinsaurralde:
Have removed `TestCase` (now using `http.Get`/`http.Post` directly)  from the benchmarks:
```
goos: darwin
goarch: amd64
pkg: github.com/TykTechnologies/tyk
BenchmarkGRPCDispatch/Pre_Hook_with_SetHeaders-8         	   50000	     20316 ns/op	    2474 B/op	      43 allocs/op
```

buger:
Allocations are quite small anyway. Looks good.

Issue #1733, opened 2018-05-25
bug: request & Response body transform cannot handle json arrays


asoorm:
Create a json response body transform. Use it to filter an array.

Because the gateway unmarshals `bodyData` into `map[string]interface{}`,
this means that it is unable to unmarshal an array.

```
[May 25 19:49:13] ERROR outbound-transform: Error unmarshalling JSON:
json: cannot unmarshal array into Go value of type map[string]interface
{} api_id=62498653ade2405b64537d94d2666dc0 path=example-array
server_name=https://my-upstream.com/
```

This fix resolves the response body transform by unmarshaling into an
`interface{}`.

We need to find a suitable solution for request body transforms. This is
currently not possible due to:

```
if tmeta.TemplateData.EnableSession {
	session := ctxGetSession(r)
	bodyData["_tyk_meta"] = session.MetaData
}

if contextVars {
	bodyData["_tyk_context"] = ctxGetData(r)
}
```

buger:
Seems so

buger:
Released as part of 2.7

PR #1734, opened 2018-05-25
[WIP] Allow json response transforms to handle array responses

asoorm:
#1733

Create a json response body transform. Use it to filter an array.

Because the gateway unmarshals `bodyData` into `map[string]interface{}`,
this means that it is unable to unmarshal an array.

```
[May 25 19:49:13] ERROR outbound-transform: Error unmarshalling JSON:
json: cannot unmarshal array into Go value of type map[string]interface
{} api_id=62498653ade2405b64537d94d2666dc0 path=example-array
server_name=https://my-upstream.com/
```

This fix resolves the response body transform by unmarshaling into an
`interface{}`.

We need to find a suitable solution for request body transforms. This is
currently not possible due to:

```
if tmeta.TemplateData.EnableSession {
	session := ctxGetSession(r)
	bodyData["_tyk_meta"] = session.MetaData
}

if contextVars {
	bodyData["_tyk_context"] = ctxGetData(r)
}
```

### DIFF
--- a/res_handler_transform.go
+++ b/res_handler_transform.go
@@ -76,7 +76,7 @@ func (h *ResponseTransformMiddleware) HandleResponse(rw http.ResponseWriter, res
 	defer respBody.Close()
 
 	// Put into an interface:
-	var bodyData map[string]interface{}
+	var bodyData interface{}
 	switch tmeta.TemplateData.Input {
 	case apidef.RequestXML:
 		mxj.XmlCharsetReader = WrappedCharsetReader


buger:
I think we need to move from this approach to providing context and meta via template functions. 
And as migration level still support old context and meta approach, but only if bodyData is a hash, e.g. to type checking.

buger:
Well, actually in this exact case your fix will be enough since response transform does not expose cotext or meta. Using functions also probably won't work, since they are request dependant, and to implement such functions we need to generate templates on each request, which will be slow.

I like the idea of special array case.

@dencoded `.index` function not really required, but it can be really handy if you want to get the nested key, or for example key with non-standard characters. And it works with maps too.

Lets do so, will merge this specific fix, and will think a bit more how to properly fix arrays in request body transforms.

PR #1736, opened 2018-05-26
Enhance Python errors

matiasinsaurralde:
This PR introduces some logging enhancements:
- When `tyk-python` is executed with `enable_coprocess` but `python_path_prefix` the following warning is logged:
```
WARN coprocess: Python path prefix isn't set (check "python_path_prefix" in tyk.conf)
```
- When a bundle is loaded and it fails during a request we throw an error:
```
@Hook
def MyHook(request, session, metadata, spec):
  inexistentfunction()
```
This will log:
```
ERROR python: Hook 'MyHook' returned an error: name 'inexistentfunction' is not defined
```
- When an inexistent hook is used in the manifest file and a HTTP request comes:
```
ERROR python: Hook is not defined: MyHook
```
- When the Python code contains a syntax error during initialization:
```
ERROR python: Middleware initialization error: invalid syntax (middleware.py, line 1)
```
- Under other scenarios where the hook table is empty (introduced by @buger):
```
ERROR python: No hooks defined for API: quickstart
```
- When the `cgo` call fails for any reason (in addition to #1727)
```
ERROR coprocess: Dispatch error
```
- When using debug mode, successfully loaded hooks are logged:
```
python: Loading hook 'MyHook' (quickstart_bff3f7c0a59f177778910dfe0c03224b)
```

### DIFF
--- a/coprocess.go
+++ b/coprocess.go
@@ -266,6 +266,9 @@ func (m *CoProcessMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Requ
 
 	returnObject, err := coProcessor.Dispatch(object)
 	if err != nil {
+		log.WithFields(logrus.Fields{
+			"prefix": "coprocess",
+		}).WithError(err).Error("Dispatch error")
 		if m.HookType == coprocess.HookType_CustomKeyCheck {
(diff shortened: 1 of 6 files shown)


buger:
👌

PR #1741, opened 2018-05-30
Add Configured leeway to JWT time validation claims in case of a clock skew

letzya:
[This fix ](https://github.com/TykTechnologies/tyk/commit/abb1b350b7dfdce52901e6e3f730a7fd899fd3f0
)helps to avoid jwt failure but can risk us since it won't validate the time-related claims at all (using disable config fields in api def). 

I have added a field that gives leeway ,in seconds, in case there is a clock skew times between the signing server (Idp for instance) and the verifying server, i.e. Tyk.

Have added tests for both type of fields - disable time claims validation and added leeway to time claims validation.


buger:
@letzya since your JWT disable validation fields not released yet, and do the similar thing but worse, what do you think about removing them at all?

letzya:
@buger It potentially could be useful for troubleshooting (when you can't change your jwt).

buger:
Yes, but your options replace mine. If you set very high clock skew it is same as disabling validation.

letzya:
Ok, will remove it.

letzya:
@buger Removed.

letzya:
@buger Important PR for a client.

letzya:
@buger please review. thanks

buger:
Looks good 👍

Will go to 2.7.2

PR #1743, opened 2018-05-30
Fix whitelists failing when modify headers on same path and add test

### LINKED ISSUE #1732: Whitelists let through unwhitelisted methods when used with other middleware

Bug

1. Create api with whitelisted GET endpoint
2. Send request with PUT and it will be rejected by Tyk
3. Add modify headers middleware to  same endpoint
4. Send PUT request again and it will hit the upstream.

Whitelisted endpoints should only allow through the specified http method

See above.

2.6.1

joshblakeley:
Fixes #1732 

This may need some changing as might have been quite naive with the approach - but this solution fixes the issue described and doesnt break all our other tests. Although there may be other scenarios this change could affect.


joshblakeley:
Closing in favour of #1749

PR #1744, opened 2018-05-30
switch rate limiter order so that keys are rate limited first

asoorm:
This ensures that if a token goes over it's rate-limit, it does not
count towards the global / api-level rate-limit.


buger:
Preferably

PR #1746, opened 2018-05-31
add tyk_context to response body transform

asoorm:
When `EnableContextVars: true`, request context data is added to
`bodyData["_tyk_context"]` for use within response body transform
templates.
    
Similarly, we have also included `bodyData["_tyk_meta"]`.

### DIFF
--- a/res_handler_transform.go
+++ b/res_handler_transform.go
@@ -102,7 +102,7 @@ func (h *ResponseTransformMiddleware) HandleResponse(rw http.ResponseWriter, res
 		if err != nil {
 			logger.WithError(err).Error("Error unmarshalling XML")
 		}
-	default:
+	default: // apidef.RequestJSON
 		var tempBody interface{}
 		if err := json.NewDecoder(respBody).Decode(&tempBody); err != nil {
 			logger.WithError(err).Error("Error unmarshalling JSON")
@@ -114,6 +114,15 @@ func (h *ResponseTransformMiddleware) HandleResponse(rw http.ResponseWriter, res
 		case map[string]interface{}:
 			bodyData = tempBody.(map[string]interface{})
 		}
+	}
+
+	if h.Spec.EnableContextVars {
+		bodyData["_tyk_context"] = ctxGetData(req)
+	}
+
+	if tmeta.TemplateData.EnableSession {
+		session := ctxGetSession(req)
+		bodyData["_tyk_meta"] = session.MetaData
 	}
 
 	// Apply to template
--- a/mw_transform.go
+++ b/mw_transform.go
@@ -87,7 +87,6 @@ func transformBody(r *http.Request, tmeta *TransformSpec, contextVars bool) erro
 			case map[string]interface{}:
 				bodyData = tempBody.(map[string]interface{})
 			}
-
 		default:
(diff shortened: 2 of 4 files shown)


api_loader.go
```diff
@@ -406,8 +406,8 @@ func processSpec(spec *APISpec, apisByListen map[string]int,
 		mwAppendEnabled(&chainArray, &StripAuth{baseMid})
 		mwAppendEnabled(&chainArray, &KeyExpired{baseMid})
 		mwAppendEnabled(&chainArray, &AccessRightsCheck{baseMid})
-		mwAppendEnabled(&chainArray, &RateLimitForAPI{BaseMiddleware: baseMid})
 		mwAppendEnabled(&chainArray, &RateLimitAndQuotaCheck{baseMid})
+		mwAppendEnabled(&chainArray, &RateLimitForAPI{BaseMiddleware: baseMid})
```
letzya: Is this related? or just testing?

asoorm: not related - was a checkout of another branch

res_handler_transform.go
```diff
@@ -103,11 +103,23 @@ func (h *ResponseTransformMiddleware) HandleResponse(rw http.ResponseWriter, res
...
+		case []interface{}:
+			logger.Debug("got array: []interface{}")
+			bodyData["array"] = tempBody
+		case map[string]interface{}:
+			logger.Debug("got object: map[string]interface{}")
+			bodyData = tempBody.(map[string]interface{})
+		}
 	}
 
+	bodyData["_tyk_context"] = ctxGetData(req)
```
buger: Why only context and not meta?

res_handler_transform.go
```diff
@@ -103,11 +103,23 @@ func (h *ResponseTransformMiddleware) HandleResponse(rw http.ResponseWriter, res
...
+		var tempBody interface{}
+		if err := json.NewDecoder(respBody).Decode(&tempBody); err != nil {
 			logger.WithError(err).Error("Error unmarshalling JSON")
 		}
+
+		switch tempBody.(type) {
+		case []interface{}:
+			logger.Debug("got array: []interface{}")
+			bodyData["array"] = tempBody
+		case map[string]interface{}:
```
buger: Do not think you can do this. Type checking for `map[string]interface{}` does not work. You should use `reflect.TypeOf(a).Kind() == reflect.Map` and `reflect.TypeOf(a).Kind() == reflect.Array`

api_loader.go
buger: Can you put link to related PR, so we can merge it, and you will rebase on master.

res_handler_transform.go
dencoded: I think it should be OK as json un-marshalling does `map[string]interface{}` or `[]interface{}` if not strict type provided.

It can also be done using type assertion:
```
if tmpVal, ok := tempBody.([]interface{}); ok {
		fmt.Println("got array: []interface{}", tmpVal)
	} else if tmpVal, ok := tempBody.(map[string]interface{}); ok {
		fmt.Println("got object: map[string]interface{}", tmpVal)
	} else {
		fmt.Println("Got something else", tmpVal)
	}
```

also, I think any of those three have about the same performance

PR #1747, opened 2018-05-31
Fix session metadata conversions

matiasinsaurralde:


buger:
Can you add a bit more implementation details here? 

Does it cross compatible?

matiasinsaurralde:
Previously we had a `metadata` field in the `coprocess.Object` data structure, this was a`map[string]string`, the flow was like this:
- Request arrives, if there's metadata present we add it to `coprocess.Object.Metadata` (and try to stringify all the values).
- The hook gets called and metadata is accessed through `coprocess.Object`.
Using `coprocess.Object` was a workaround as we weren't sure about the metadata usage in this context.

This PR modifies `coprocess.SessionState` to use a `map[string]string` metadata field, this makes more sense (our original data structure uses `map[string]interface{}`).
If the request contains metadata, we copy all the elements and stringify all the values, this is fine for common structures like `{"my_field:" "my_value", "another_field": 10}`, in case a nested value is detected, it gets stored as a JSON-encoded string. Example:
```json
{"my_field": "my_value", "dashboard_data": {"developer_id":123}}
```
Accessing `session.my_field` will return `my_value` (raw string).
Accessing `session.dashboard_data` will return `{"developer_id":123}` (JSON-encoded string).

The session object access was wrapped in the our Python bindings and I've adjusted them so that it doesn't break the current usage.
For gRPC users that need/use metadata, they would have to adjust their code to access `session.Metadata` instead of `object.Metadata`.

matiasinsaurralde:
I've found a fix to keep compatibility with non-Python plugins, Python is already compatible (as mentioned above).

buger:
@matiasinsaurralde can you pls explain your last fix?

matiasinsaurralde:
Previously I've mentioned that users would need to change their middleware to use `session.Metadata` instead of `object.Metadata` (the latter was created as a workaround and `session.Metadata` was only a string (you will find this in the Proto definition `string metadata`). The last commit allows code that still uses `object.Metadata` to work.
The whole PR drops the usage of `object.Metadata` and uses `session.Metadata` in the correct way, additionally dealing with the issue we've found with nested data structures.

buger:
Looks good! Since this stuff is sensitive, pls add tests using python, which test this functionality. 

Thank you.

matiasinsaurralde:
Sure, will extend the gRPC tests for this in particular, Python tests already cover this (they were broken before the compatibility fix).

buger:
If they were broken, why test was not failing? Maybe test need to be fixed?

matiasinsaurralde:
Python tests were failing before, that's the reason for the compatibility commit (the user isn't required to change his code and we handle this transparently). gRPC isn't failing because we don't have any test using metadata in there, needs to be extended with the old approach and the new one, both auth hooks.

buger:
Ah, I see. Can we then extend this test with nested values?

matiasinsaurralde:
Yes, we could add some test values similar to the ones set by the dashboard

PR #1749, opened 2018-06-01
Fix whitelists failing when modify headers on same path V2

### LINKED ISSUE #1732: Whitelists let through unwhitelisted methods when used with other middleware

buger:
Our rxPaths always built with the same order. First comes Whitelist, Ignore, and Blacking, the rest of plugins after.

So, if after we checked for whitelisting, we did not return early from the cycle, we can safely tell that white-listing failed.

Fixes #1732

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -870,6 +870,7 @@ func (a *APISpec) URLAllowedAndIgnored(r *http.Request, rxPaths []URLSpec, white
 		if !v.Spec.MatchString(strings.ToLower(r.URL.Path)) {
 			continue
 		}
+
 		if v.MethodActions != nil {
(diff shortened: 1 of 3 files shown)


joshblakeley:
CI will fix itself if you repush i think

buger:
I just check, and the second clause still needed, in cases when URL has no any matches, and it basically ignores `rxPaths` cycle.

PR #1750, opened 2018-06-01
Benchmark default version profiling

dencoded:
added changes for https://github.com/TykTechnologies/tyk/issues/1635

When I first looked at BenchmarkDefaultVersion I've noticed that we spend some significant time in ErrorHandler and SuccessHandler when analytics is enabled. So I started looking into it and that PR includes changes for analytics as well:

- we did lots of slice re-allocations for `record.Tags`, I've added helper function which calculates capacity based on spec and configuration
- don't spawn go-routine for `analytics.RecordHit`, write to buffered channel instead which is consumed by pool of workers
- added new method to redis storage `AppendToSetPipelined(key string, values []string)` to use pipelined operation in analytics pool worker
- pool workers maintain buffer of consumed messages, that buffer is sent to redis when it is full or when there were no new messages for some time but buffer holds some unsent messages
- the same config value `pool_size` is used to specify number of pool workers
- also there is new config value `records_buffer_size` which specifies buffer size for pool workers (size of each pipeline operation). also, buffered channel depends on this value as well and calculated as `pool_size * records_buffer_size`. The default value for `records_buffer_size` is `1000`, also if value from config is lower it will be forced to `1000`
- we stop analytics worker poll gracefully when tyk is about to exit

Also there some optimizations specifically for `VersionCheck` middle-ware:

- parse version `expires` field at API load time, not at request processing
- some small clean ups which potentially improve performance and readability of code

benchcmp vs master:
```
benchmark                     old ns/op     new ns/op     delta
BenchmarkDefaultVersion-8     1120529       1023211       -8.69%

benchmark                     old allocs     new allocs     delta
BenchmarkDefaultVersion-8     1487           1226           -17.55%

benchmark                     old bytes     new bytes     delta
BenchmarkDefaultVersion-8     213900        190022        -11.16%
```

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -65,7 +65,6 @@ const analyticsKeyName = "tyk-system-analytics"
 
 const (
 	minRecordsBufferSize       = 1000
-	minRecordsWriteBatchSize   = 200
 	recordsBufferFlushInterval = 200 * time.Millisecond
 )
 
@@ -225,8 +224,8 @@ func (r *RedisAnalyticsHandler) recordWorker() {
 	defer r.poolWg.Done()
 
 	// this is buffer to send one pipelined command to redis
-	// use minRecordsWriteBatchSize as cap to reduce slice re-allocations
-	recordsBuffer := make([]string, 0, minRecordsWriteBatchSize)
+	// use r.recordsBufferSize as cap to reduce slice re-allocations
+	recordsBuffer := make([]string, 0, r.recordsBufferSize)
 
 	// read records from channel and process
 	for {
@@ -274,7 +273,7 @@ func (r *RedisAnalyticsHandler) recordWorker() {
 			}
 
 			// identify that buffer is ready to be sent
-			readyToSend = uint64(len(recordsBuffer)) == minRecordsWriteBatchSize
+			readyToSend = uint64(len(recordsBuffer)) == r.recordsBufferSize
 
 		case <-time.After(recordsBufferFlushInterval):
 			// nothing was received for that period of time
@@ -285,7 +284,7 @@ func (r *RedisAnalyticsHandler) recordWorker() {
 		// send data to Redis and reset buffer
 		if readyToSend && len(recordsBuffer) > 0 {
 			r.Store.AppendToSetPipelined(analyticsKeyName, recordsBuffer)
-			recordsBuffer = make([]string, 0, minRecordsWriteBatchSize)
+			recordsBuffer = make([]string, 0, r.recordsBufferSize)
 		}
 	}
 }
--- a/analytics.go
+++ b/analytics.go
@@ -4,9 +4,10 @@ import (
 	"fmt"
 	"net"
 	"regexp"
+	"sync"
+	"sync/atomic"
 	"time"
 
-	"github.com/jeffail/tunny"
 	"github.com/oschwald/maxminddb-golang"
 	"gopkg.in/vmihailenco/msgpack.v2"
 
@@ -62,10 +63,13 @@ type GeoData struct {
 
 const analyticsKeyName = "tyk-system-analytics"
 
+const (
+	minRecordsBufferSize       = 1000
+	minRecordsWriteBatchSize   = 200
+	recordsBufferFlushInterval = 200 * time.Millisecond
+)
+
 func (a *AnalyticsRecord) GetGeo(ipStr string) {
-	if !config.Global().AnalyticsConfig.EnableGeoIP {
-		return
-	}
 	// Not great, tightly coupled
 	if analytics.GeoIPDB == nil {
 		return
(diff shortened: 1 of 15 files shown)


buger:
What if we have some moderate load, like 5 RPS, which does not hit timeout, appends to the buffer, and waits till it will fill. How big analytics delay can be in the worst situation?

analytics.go
```diff
@@ -165,53 +171,120 @@ func (r *RedisAnalyticsHandler) Init() {
...
 		ps = 50
 	}
+	log.WithField("ps", ps).Debug("Analytics pool workers number")
 
-	r.AnalyticsPool, err = tunny.CreatePoolGeneric(ps).Open()
-	if err != nil {
-		log.Error("Failed to init analytics pool")
+	r.recordsBufferSize = r.globalConf.AnalyticsConfig.RecordsBufferSize
+	if r.recordsBufferSize < minRecordsBufferSize {
+		r.recordsBufferSize = minRecordsBufferSize // force it to this value
```
buger: I think it is a bit unfair to call this variable `recordsBufferSize`, because its not total value, actual buffer will be `pool * buffer_size`. It feels like doing `recordsBufferSize = recordsBufferSize / pool_size` will be more fair.

analytics.go
```diff
@@ -165,53 +171,120 @@ func (r *RedisAnalyticsHandler) Init() {
...
-			// Extend tag list to include this data so wecan segment by node if necessary
-			record.Tags = append(record.Tags, "tyk-hybrid-rpc")
-		}
+	log.WithField("recordsBufferSize", r.recordsBufferSize).Debug("Analytics records buffer size")
 
-		if config.Global().DBAppConfOptions.NodeIsSegmented {
-			// Extend tag list to include this data so wecan segment by node if necessary
-			record.Tags = append(record.Tags, config.Global().DBAppConfOptions.Tags...)
-		}
+	chanSize := uint64(ps) * r.recordsBufferSize
```
buger: We need to be careful here, in cases when detailed recording is turned on payload size can up to few mb for each record. Just want to ensure that such big buffer won't cause memory issues in edge cases.

analytics.go
```diff
@@ -165,53 +171,120 @@ func (r *RedisAnalyticsHandler) Init() {
...
+			record.Tags = append(record.Tags, "api-"+record.APIID)
+
+			if encoded, err := msgpack.Marshal(record); err != nil {
+				log.WithError(err).Error("Error encoding analytics data")
+			} else {
+				recordsBuffer = append(recordsBuffer, string(encoded))
+			}
+
+			// identify that buffer is ready to be sent
+			readyToSend = uint64(len(recordsBuffer)) == r.recordsBufferSize
```
buger: Les say we have 50 workers, and 1000 buffer per worker and each receives traffic which has around 200ms around requests (but a bit lower), and do not hit timeout, mentioned above. Time to fill the buffer can be up to ((1000 * 50) * 0.2)/60 ~ 166 minutes. Correct me if I wrong. 

If this true, I suggest putting one more top timeout to this select, which will force writing analytics to redis every second.

gateway_test.go
```diff
@@ -592,6 +592,8 @@ func TestAnalytics(t *testing.T) {
 	})
 
 	// Cleanup before test
+	// let records to to be sent
+	time.Sleep(250 * time.Millisecond)
```
buger: make sense replace with `recordsBufferFlushInterval + 50`?

analytics.go
dencoded: Thats true, current design assumes high traffic. but with low throughput like you said one request every 199ms - data will stale here for long time. We can add for that special case logic to send data to redis anyways if last time we sent data was more than 1 sec ago.

analytics.go
dencoded: we are passing pointers to channel, once record sent GC can pick and free it. also, I made that default buffer size too big with assumption that pool workers will process faster than sender sends to channel. if we receive much more and block on writing to channel I think think effect will be close to situation when we we try to schedule lots of go-routines. I was thinking about adaptable worker pool size, i.e. if sender can't send to channel for too long (which means channel is not read too fast) it could ask analytics to increase pool of workers. this sounds interesting but we have reidis on other end and redis connection pool with fixed size - it can get slow just because of redis (thats why I started thinking about sending records via pipelines). Also, according to redis - starting from 2.4 we can provide to RPUSH more then one value, we can benefit from that as well.

analytics.go
dencoded: I am bad in naming things, will come up with better name

gateway_test.go
dencoded: 👍

analytics.go
buger: Or you can leave it as it is, like "buffer_size", and per pool buffer will be: BufferSize / pool_size. 

IMHO, it is the most logical here. You separate control concurrency via pool size and total buffer size. Just set it to bigger value if you want, like 10000. 

For sure if we will be able to auto-grow this buffer, and we do not need to expose this option at all, it will be even more cooler, but only if implementation does not take too much time.

matiasinsaurralde:
Looks good and agree with Leon's comments, will rebase #1690 with this after the merge to see if more improvements show up

Issue #1751, opened 2018-06-02
Request body transform not working if input is JSON


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
All body transforms fail with `json: Unmarshal(non-pointer map[string]interface {})` error
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Create request body transform with JSON input
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
Bug introduced in 2.6

PR #1752, opened 2018-06-02
Fix body transform when input is JSON

buger:
`json.Unmarshal` needs pointer to map

Fix https://github.com/TykTechnologies/tyk/issues/1751


buger:
Apparently our tests was covering only XML -> JSON transform.

Issue #1694, opened 2018-05-10
Proposal: Replace hash function and make it possible to extend


lonelycode:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
The current hash function i prone to collisions, even though it is very fast.

**What is the expected behavior?**
We should:

1. Replace the hash function with Murmur2 (64-bit)
2. Ensure a way to transition from legacy tokens to new ones to ensure smooth upgrade
3. Make the token format extensible

### Proposal: New hash function: Murmur2

According to [this write up](https://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed) Murmur3 is extremely fast because it operates on 4-byte chunks, however this also increases the chance of collisions and leads to the odd behaviour of being able to use a valid token + 3 arbitrary characters to gain API access when using standard bearer tokens.

the write up also concludes that while FNV-1a is probably the most popular fast hash to use because of good distribution, the author goes on to state tat Murmur2 is not only faster, but provides lower / equal collision protection.

I would also suggest we move from a 32-bit has to a 64-bit hash, as that decreases chances even further. The `github.com/aviddiviner/go-murmur` library implements [MurmurHash64A](https://github.com/aviddiviner/go-murmur/blob/master/murmur2.go#L67) and also [shows solid benchmarks](https://github.com/aviddiviner/go-murmur/blob/master/README.md).

Just to ensure 64a is in fact a variant of murmur2, ere's what wikipedia has to say about it:

> MurmurHash2
> The older MurmurHash2[8] yields a 32-bit or 64-bit value. Slower versions of MurmurHash2 are available for big-endian and aligned-only machines. The MurmurHash2A variant adds the Merkle–Damgård construction so that it can be called incrementally. **There are two variants which generate 64-bit values; MurmurHash64A, which is optimized for 64-bit processors**, and MurmurHash64B, for 32-bit ones. MurmurHash2-160 generates the 160-bit hash, and MurmurHash1 is obsolete.

(emphasis mine)

Based on the above, I would propose that we move to `MurmurHash64A` for our default hashing function.

### Proposal: Make our tokens easier to extend and index

Currently tokens generated by the dashboard are of the format: `OrgID+UUIDv4` (the UUID has it's dashes removed). These are provided in a raw format to the user.

[...]

PR #1753, opened 2018-06-02
New token generation, support for custom hashing algorithms

buger:
Added new scheme of generating tokens, now it is JSON base64 objects, similar to how JWT works, which allows us add any meta information to the tokens.

Example of new token:
`eyJvcmciOiIiLCJpZCI6IjAxZGIwN2Q5NWQ0MDRjYjM5ODg3ZjUwN2ZmNTg0OGUzIiwiaCI6Im11cm11cjEyOCJ9`. After based64 decoding it turns to: `{"org":"","id":"01db07d95d404cb39887f507ff5848e3","h":"murmur128”}`.

Current token has 3 fields at the moment:
- “org” - organization id
- “id” - token id, as we had before. For regular tokens random uuid, but for example for JWTs or OpenID based on sub.
- “h” - token hashing algorithm

At the moment following algorithms are supported: `murmur32`, `murmur64`, `murmur128`, `sha256`.

You can set hashing algorithm to use via new config variable: `hash_key_function`.

If config value not set, it generates legacy tokens like we had before. We may consider switching it to `murmur64` by default in next releases, once feature will stabilize.

You can change hashing function dynamically, and have tokens with multiple hashing algos, including handling of legacy tokens. So if you turn on new key support, you old legacy tokens will continue working. Same if you decided to switch different algorithm or turn this off.

Fix https://github.com/TykTechnologies/tyk/issues/1694


buger:
@lonelycode Worth noticing that I forked https://github.com/spaolacci/murmur3 to our org, and update d murmur32 hashing implementation to the version we had vendored. So all other algos like murmur128 got all the latest code and fixes, and murmur32 legacy support. `github.com/aviddiviner/go-murmur` mentioned in your proposal is older, support only murmur2 (and what we have already murmur3), and do not support 128 bit function.

lonelycode:
Awesome! Is there a way we can check the append-three character issue has been solved?

buger:
@lonelycode Yes. Added tests showing both bug and fix itself, in various combinations: https://github.com/TykTechnologies/tyk/pull/1753/commits/3d25da10d6ebfd3ee557661cc41a6a7a0108222b#diff-092a80b1b0fbc9a08fef31ca557e51a2R20

api.go
```diff
@@ -235,7 +235,7 @@ func handleAddOrUpdate(keyName string, r *http.Request) (interface{}, int) {
 		// Only if it's NEW
 		switch r.Method {
 		case "POST":
-			keyName = newSession.OrgID + keyName
+			keyName = generateToken(newSession.OrgID, strings.TrimPrefix(keyName, newSession.OrgID))
```
dencoded: how about hiding TrimPrefix-logic inside `generateToken` helper? so calling part won't have to remember to do this

auth_manager.go
```diff
@@ -211,11 +211,23 @@ func (b *DefaultSessionManager) Sessions(filter string) []string {
 
 type DefaultKeyGenerator struct{}
 
+func generateToken(orgID, keyID string) string {
+	token, err := storage.GenerateToken(orgID, keyID, config.Global().HashKeyFunction)
+
+	if err != nil {
+		log.WithFields(logrus.Fields{
```
dencoded: we could also do `.WithError(err).WithFields` so error will appear as special error-field in log entry

storage/storage.go
```diff
@@ -46,8 +53,78 @@ type Handler interface {
 	RemoveSortedSetRange(string, string, string) error
 }
 
+const defaultHashAlgorithm = "murmur64"
+
+// If hashing algorithm is empty, use legacy key generation
+func GenerateToken(orgID, keyID, hashAlgorithm string) (string, error) {
+	if keyID == "" {
+		keyID = strings.Replace(uuid.NewV4().String(), "-", "", -1)
```
dencoded: it is not efficient to get canonical string rep of UUID and then remove dashes right away using string search/replace operation, maybe it is OK as long as we don't issue tons of tokens per sec

storage/storage.go
```diff
@@ -46,8 +53,78 @@ type Handler interface {
...
+	}
+
+	if hashAlgorithm != "" {
+		_, err := hashFunction(hashAlgorithm)
```
dencoded: I think it is a little bit of extra work here - we are allocating an instance of hash (which is runtime sys-call and takes extra memory for a bit) just to test if token needs to have a default hash algo, then we ignore allocated space which give some extra work for GC. I am not sure compiler can optimize this.

also I like more readable form `if _, err := hashFunction(hashAlgorithm); err != nil`

storage/storage.go
```diff
@@ -46,8 +53,78 @@ type Handler interface {
...
+		if err != nil {
+			hashAlgorithm = defaultHashAlgorithm
+		}
+
+		jsonToken := fmt.Sprintf(`{"org":"%s","id":"%s","h":"%s"}`, orgID, keyID, hashAlgorithm)
```
dencoded: might worth to use string concatenation instead of `Sprintf` here

storage/storage.go
```diff
@@ -46,8 +53,78 @@ type Handler interface {
...
+	case "", "murmur32":
+		return murmur3.New32(), nil
+	default:
+		return murmur3.New32(), fmt.Errorf("Unknown key hash function: %s. Falling back to murmur32.", algorithm)
+	}
+}
+
 func HashStr(in string) string {
-	h := murmur3.New32()
+	h, _ := hashFunction(TokenHashAlgo(in))
```
dencoded: method `HashStr ` is heavily used when key hashing enabled, correct me if I am wrong - we always do `base64.StdEncoding.EncodeToString` when we generateToken from key, org ID and then have to do `base64.StdEncoding.DecodeString` right away to get from storage or update from storage. this base64 encoding/decoding seems to be extra work to me

storage/storage.go
dencoded: also, we could introduce token as not just string but 1st class citizen in our code base as a struct, so we could put here in its fields hash algo name, Hasher func for later use in case we need to hash it (and do hash only once if it is not laready done), json and and base64 encoded  representation  of this token as fields as well

buger:
Frankly, I would not care at all about key creation performance, since it does not need to be fast. 
But note about HashStr every time using b64 is true, however, I'm not sure what we can do about it. I already added check which checks for legacy tokens without b64 decoding, but for new keys it probably will be required. Enabling caching for tokens seems like over-engineering here. b64 in general very fast, and we avoid full JSON parsing too. Let me add some benchmark results here, on how it affects basic request flow.

buger:
@dencoded fixed code styling issues. Regarding performance, as mentioned above, I would prefer not spending too much time on key creation process (if it looks good from code point of view), since this is quite rarely used functionality.

buger:
That's true but frankly, I can't see a more efficient way to encode custom metadata to token itself. I'm open to suggestions.

buger:
@dencoded change I made in https://github.com/TykTechnologies/tyk/pull/1770 should resolve the situation with key hashing called multiple times.

PR #1802, opened 2018-06-28
add support for array based claims

### LINKED ISSUE #1801: use array jwt claim in header
Seems like a bug

I have jwt with a claim which is an array of strings.  I would like to put those strings into a request header.  I am a _go_ newbie, but judging from [the code](https://github.com/TykTechnologies/tyk/blob/e887423ba9276d3c499a3fd4447062f3c758606f/mw_url_rewrite.go#L229) I would have thought that it would give me a comma-separated list.  However, it doesn’t work.  I get instead an empty header, with the following error in the logs:
```
[Jun 28 22:38:04] ERROR Context variable type is not supported: []interface {}
```

a comma-separated list in the new request header

1. put the following in api definition:
```
        "global_headers": {
          "X-Mids": "$tyk_context.jwt_claims_idlist"
        },
```
2. use a jwt with `idlist` claim which is an array of strings

tried with latest docker image and locally-built v2.6.2

asoorm:
Fixes #1801
    
What about strings with commas in already? should the 
comma be stripped also? If so, then what character 
should that `,` char be replaced with?

### DIFF
--- a/mw_url_rewrite.go
+++ b/mw_url_rewrite.go
@@ -243,7 +243,7 @@ func valToStr(v interface{}) string {
 		tmpSlice := make([]string, 0, len(x))
 		for _, val := range x {
 			if rec := valToStr(val); rec != "" {
-				tmpSlice = append(tmpSlice, rec)
+				tmpSlice = append(tmpSlice, url.QueryEscape(rec))
 			}
 		}
 		s = strings.Join(tmpSlice, ",")
--- a/mw_url_rewrite.go
+++ b/mw_url_rewrite.go
@@ -239,6 +239,14 @@ func valToStr(v interface{}) string {
 			}
 			i++
 		}
+	case []interface{}:
+		tmpSlice := make([]string, 0, len(x))
+		for _, val := range x {
+			if rec := valToStr(val); rec != "" {
+				tmpSlice = append(tmpSlice, rec)
+			}
+		}
+		s = strings.Join(tmpSlice, ",")
 	default:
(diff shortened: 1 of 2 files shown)


altaurog:
@asoorm thanks!  This works well for me.  Perhaps json encoding any non-string values would be a good general solution which would eliminate such edge cases.

mw_url_rewrite.go
```diff
@@ -239,6 +239,14 @@ func valToStr(v interface{}) string {
 			}
 			i++
 		}
+	case []interface{}:
+		var tmpSlice []string
+		for _, val := range v.([]interface{}) {
```
dencoded: looks like an extra type assertion here, better to use `range x`

mw_url_rewrite.go
```diff
@@ -239,6 +239,14 @@ func valToStr(v interface{}) string {
 			}
 			i++
 		}
+	case []interface{}:
+		var tmpSlice []string
```
dencoded: would be better to do `tmpSlice := make([]string, 0, len(x))` to avoid slice re-allocations

asoorm:
@dencoded PTAL

dencoded:
@asoorm looks good to me but CI errored

asoorm:
@dencoded This is separate issue. https://github.com/TykTechnologies/tyk/issues/1820

altaurog:
This is not in version with v2.7.7 release tag.  Is that not the latest (and greatest)?

Issue #1775, opened 2018-06-13
Custom authorization with python plugin doesn't work


ilijabojanovic:
**Do you want to request a *feature* or report a *bug*?**
BUG
**What is the current behavior?**
Custom authorization with python plugin doesn't work
**What is the expected behavior?**
Custom authorization with python plugin should work if we send valid auth header
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

Here is testing definition:
```
{
    "id": "5b1fc34bc000f00018c2ff62",
    "name": "basic",
    "slug": "basic",
    "api_id": "5d67ad64a0fb49d57ee3b605da309d95",
    "org_id": "5b1e7820c000f000173578c2",
    "use_keyless": false,
    "use_oauth2": false,
    "use_openid": false,
    "openid_options": {
        "providers": [],
        "segregate_by_client": false
    },
    "oauth_meta": {
        "allowed_access_types": [],
        "allowed_authorize_types": [],
        "auth_login_redirect": ""
    },
    "auth": {
        "use_param": false,
        "param_name": "",
        "use_cookie": false,
        "cookie_name": "",
        "auth_header_name": "",
        "use_certificate": false
    },
    "use_basic_auth": false,
    "use_mutual_tls_auth": false,
    "client_certificates": [],
    "upstream_certificates": {},
    "pinned_public_keys": {},
    "enable_jwt": false,
    "use_standard_auth": false,
    "enable_coprocess_auth": true,
    "jwt_signing_method": "",
    "jwt_source": "",
    "jwt_identity_base_field": "",
    "jwt_client_base_field": "",
    "jwt_policy_field_name": "",
    "notifications": {
        "shared_secret": "",
        "oauth_on_keychange_url": ""
    },
    "enable_signature_checking": false,
    "hmac_allowed_clock_skew": -1,
    "base_identity_provided_by": "",
    "definition": {
        "location": "header",
        "key": "x-api-version"
    },
    "version_data": {
        "not_versioned": true,
        "default_version": "",
        "versions": {
            "Default": {
                "name": "Default",
                "expires": "",
                "paths": {
                    "ignored": [],
                    "white_list": [],
                    "black_list": []
                },
                "use_extended_paths": true,
                "extended_paths": {},
                "global_headers": {},
                "global_headers_remove": [],
                "global_size_limit": 0,
                "override_target": ""
            }
        }
    },
    "uptime_tests": {
        "check_list": [],
        "config": {
[...]

PR #1803, opened 2018-06-29
Protobuf fix

matiasinsaurralde:
For #1775

### DIFF
--- a/coprocess/python/tyk/middleware.py
+++ b/coprocess/python/tyk/middleware.py
@@ -78,9 +78,7 @@ class TykMiddleware:
             handler(object, object.spec)
             return
         elif handler.arg_count == 4:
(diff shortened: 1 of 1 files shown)


buger:
Why this fix the issue? `MergeFrom` was doing nothing?

matiasinsaurralde:
Seems that `MergeFrom` behaves differently across versions and there's a sort of incompatibility between the version we use to generate the bindings and the runtime version (for tests, CI, etc.), in some versions we would get `no attribute MergeFrom`, etc. I've found that it's simpler to pass it over to the hooks in this way, with the expected behavior.
Tests work fine locally (`go test -tags 'coprocess python`), the reason for the CI fail looks related to the gRPC rollback?

Issue #1771, opened 2018-06-09
WebSocket and HTTP on the same API Definition


dikderoy:
**Do you want to request a *feature* or report a *bug*?**
BUG
**What is the current behavior?**
It is impossible to use websocket and plain-HTTP on the same endpoint (or two endpoints with shared path prefix) defined in single API Definition

**What is the expected behavior?**
Full support of WebSocket and plain-HTTP on per-request/per-protocol basis (I understand what you are caching upstream connections, but this leads to described problem)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

API definition is set to forward any traffic under `/path/{rest:*}` to `http://service_endpoint:8080/`
and for example:
- `/path/socket` is WS endpoint
- `/path/request` is plain HTTP endpoint

then after establishing WS connection on `/path/socket` plain-http requests to `/path/request` will fail (no response).

Following requests to websocket endpoint with WS protocol also fail **until Tyk Gateway is restarted.**

Following appears in log:

```
tyk_gateway_1      | 2018/06/09 19:49:50 http: panic serving 172.19.0.21:54130: interface conversion: http.RoundTripper is *main.WSDialer, not *http.Transport
tyk_gateway_1      | goroutine 3655 [running]:
tyk_gateway_1      | net/http.(*conn).serve.func1(0xc4201f3040)
tyk_gateway_1      | 	/usr/local/go/src/net/http/server.go:1697 +0xd0
tyk_gateway_1      | panic(0xd7d8c0, 0xc4206b0640)
tyk_gateway_1      | 	/usr/local/go/src/runtime/panic.go:491 +0x283
tyk_gateway_1      | main.(*ReverseProxy).WrappedServeHTTP(0xc4202f5d60, 0x14cee40, 0xc42043b500, 0xc4205c7200, 0x28881600, 0x0)
tyk_gateway_1      | 	/src/github.com/TykTechnologies/tyk/reverse_proxy.go:591 +0x2008
tyk_gateway_1      | main.(*ReverseProxy).ServeHTTP(0xc4202f5d60, 0x14cee40, 0xc42043b500, 0xc4205c7200, 0x9)
tyk_gateway_1      | 	/src/github.com/TykTechnologies/tyk/reverse_proxy.go:362 +0x60
[...]

PR #1817, opened 2018-07-13
allow api spec to keep both transports to http and ws upstreams

dencoded:
added changes for https://github.com/TykTechnologies/tyk/issues/1771


dencoded:
@buger bump

PR #1819, opened 2018-07-13
Strip version when listen path is stripped

### LINKED ISSUE #1818: Version not stripped from listen path when added as part of url

Bug

When listen path is stripped the version is not also stripped

Version information will be stripped from listen path

2.6 and 2.7

joshblakeley:
Fixes #1818 

Previously the statement would never evaluate to true so version never stripped from path.

### DIFF
--- a/handler_success.go
+++ b/handler_success.go
@@ -245,7 +245,8 @@ func (s *SuccessHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) *http
 	defer s.Base().UpdateRequestSession(r)
 
 	versionDef := s.Spec.VersionDefinition
-	if !s.Spec.VersionData.NotVersioned && versionDef.Location == "url" && versionDef.StripPath {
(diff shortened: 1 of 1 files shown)


buger:
@asoorm pls verify

joshblakeley:
https://github.com/TykTechnologies/tyk/pull/1542

This was fixed here however the issue is that currently there is no UI element to enable `versionDefinition.StripPath` to be true and is also undocumented so user has no way of intuitively making the feature work.

joshblakeley:
Happy to close this but it would be great to get the UI done/ some interim documentation of the feature as it stands.

asoorm:
@buger have been waiting for UI component in order to prepare documentation. At the moment, this feature should be available via dashboard `raw api definition` and setting `strip_path` to true inside `api_definition.definition`.

```
"definition": {
      "location": "url",
      "key": "version",
      "strip_path": true
    }
```

PR #1849, opened 2018-07-31
Check if claims expiry longer than session

### LINKED ISSUE #1805: bug: Policy update does not propagate token expiry

possible bug

I create a policy with an expiry.
I create a JWT which points (in claims) to the policy.

The token works within the expiry
The token stops working when expired (gateway side) - we now have conflicts between JWT expiry and Policy Expiry.

Because the token is expired (gateway side) we are unable to re-instate access to the API - a new JWT must be created - however this is not possible via the dev portal, because according to the portal, the user already has access to this subscription.

In JWT mode, the key should either inherit the JWT token `exp` claim, and fallback to policy if not present

All

joshblakeley:
Fixes #1805 

At the moment only does this for when JWT is paired to policy.

In case of directly being tied to internal token i.e. with kid header the internal token expiry takes precedence

### DIFF
--- a/mw_jwt.go
+++ b/mw_jwt.go
@@ -284,6 +284,12 @@ func (k *JWTMiddleware) processCentralisedJWT(r *http.Request, token *jwt.Token)
 			log.Error("Could not find a valid policy to apply to this token!")
 			return errors.New("Key not authorized: no matching policy"), http.StatusForbidden
 		}
+		//override session expiry with JWT
+		if f, ok := claims["exp"].(float64); ok {
+			if int64(f) != session.Expires {
+				session.Expires = int64(f)
+			}
+		}
 
 		session = newSession
 		session.MetaData = map[string]interface{}{"TykJWTSessionID": sessionID}
@@ -334,6 +340,13 @@ func (k *JWTMiddleware) processCentralisedJWT(r *http.Request, token *jwt.Token)
 				k.reportLoginFailure(baseFieldData, r)
 				log.WithError(err).Error("Could not apply new policy from JWT to session")
 				return errors.New("Key not authorized: could not apply new policy"), http.StatusForbidden
+			}
+
+			//override session expiry with JWT
+			if f, ok := claims["exp"].(float64); ok {
+				if int64(f) != session.Expires {
+					session.Expires = int64(f)
+				}
 			}
 
 			go SessionCache.Set(session.KeyHash(), session, cache.DefaultExpiration)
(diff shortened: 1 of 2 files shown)


apidef/api_definitions.go
```diff
@@ -351,6 +351,7 @@ type APIDefinition struct {
 	JWTDisableExpiresAtValidation bool                 `bson:"jwt_disable_expires_at_validation" json:"jwt_disable_expires_at_validation"`
 	JWTDisableNotBeforeValidation bool                 `bson:"jwt_disable_not_before_validation" json:"jwt_disable_not_before_validation"`
 	JWTSkipKid                    bool                 `bson:"jwt_skip_kid" json:"jwt_skip_kid"`
+	JWTExpiryOverridesPolicy      bool                 `bson:"jwt_expiry_overrides_policy" json:"jwt_expiry_overrides_policy"`
```
asoorm: @buger @josephearl 

Should this be configurable? If JWT contains an expiry - surely we should just use the JWT's expiry without looking at the Policy expiry?

mw_jwt.go
```diff
@@ -284,6 +284,12 @@ func (k *JWTMiddleware) processCentralisedJWT(r *http.Request, token *jwt.Token)
 			log.Error("Could not find a valid policy to apply to this token!")
 			return errors.New("Key not authorized: no matching policy"), http.StatusForbidden
 		}
+		//override session expiry with JWT if it is longer lived
+		if k.Spec.JWTExpiryOverridesPolicy {
+			if int64(claims["exp"].(float64)) > session.Expires {
```
asoorm: I think - because JWT is not generated by Tyk, rather by a 3rd party - that the type assertion here should check validity otherwise the gateway will panic.

apidef/api_definitions.go
joshblakeley: The thinking with configurability is that it wont disrupt existing flows for users with JWT.

asoorm: Agree with that - but the logic is currently bugged IMO - this is supposed to resolve the issue, not give user option to resolve the issue.

asoorm: That's why I tagged @buger for his opinion

mw_jwt.go
buger: ^ yes, make sense to do right type checking, who knows which garbage can be inside this record.

apidef/api_definitions.go
buger: Probably agree with @asoorm. In this case we solving a bug.

mw_jwt.go
```diff
@@ -335,6 +341,12 @@ func (k *JWTMiddleware) processCentralisedJWT(r *http.Request, token *jwt.Token)
 				log.WithError(err).Error("Could not apply new policy from JWT to session")
 				return errors.New("Key not authorized: could not apply new policy"), http.StatusForbidden
 			}
+			//override session expiry with JWT if it is longer lived
+			if k.Spec.JWTExpiryOverridesPolicy {
+				if int64(claims["exp"].(float64)) > session.Expires {
```
buger: Why this check is needed, and why not to always set expiration form JWT token if it is not zero?

buger:
At the moment you do this check only when the session does not exist, and just created. 

We also should handle the case when JWT expiration was extended for existing session (this is a valid case). So basically if original expiration != new expiration, override it.

buger:
Also pls rebase from latest master, and modify existing tests to test new behavior.

PR #1850, opened 2018-08-01
Enable inialisation of tls client for dash comms

### LINKED ISSUE #1806: Gateway can't connect to dashboard using self-signed certificate
bug
If dashboard deployed using TLS with self-signed cert, gateway will be unable connect to the dashboard with an error saying that certificate verification failed. It works only if dashboard certificate manually added to OS trusted storage.
It should be possible specify an option to skip SSL verification for GW->Dash communication.

2.7 and earlier

joshblakeley:
Fixes #1806

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -254,9 +254,7 @@ func (a APIDefinitionLoader) FromDashboardService(endpoint, secret string) ([]*A
 
 	newRequest.Header.Set("x-tyk-nonce", ServiceNonce)
 
-	c := &http.Client{
-		Timeout: 120 * time.Second,
-	}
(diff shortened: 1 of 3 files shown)


buger:
Sorry for the late review. I think you overdid it a bit. 

Passing `NameToCertificate` and `Certificates` to TLS client is too much. `NameToCertificate` for example used only by TLS servers, and passing `Certificates` for client, used only in mutual TLS flow. 

In this case we need only option to ignore certificate verification for the dashboard connections. Using existing config as you did should be fine.

Moving to in-progress for now.

buger:
👍 

Will be part of 2.7.2

PR #1857, opened 2018-08-03
add leading backslash to path if needed

dencoded:
added changes to fix https://github.com/TykTechnologies/tyk-analytics-ui/issues/514


buger:
Will be part of 2.7.2

PR #1859, opened 2018-08-06
Add Looping support

buger:
Looping means calling itself without additional http request. So you can do an advanced URL rewrite, and for example redirect it to another endpoint, by using URL of the following scheme: `tyk://self/path`.

As url you can read it as: `tyk` scheme, `self` domain (current API), and `/path` is a path. You can also provide `?method`  the argument to specify HTTP method name, like GET or POST, for example: `tyk://self/path?method=POST`. By default, it matches the same HTTP method as the original request.

Additionally added looping support for VirtualEndpoints. To do so you just need to return HTTP redirect in response: 30X redirect, with Location header containing loop HTTP scheme. Using this technique you can turn VirtualEndpoint into advanced URL rewriter with any logic you want. The same idea will be implemented for the rest of the plugins as well. 

Regarding flow itself, during looping, request retains all its metadata, query params, form values and etc. 

You can chain loops, and turn Tyk into a state machine. Worth noticing that it has protection over too deep stack, and limited by 5 loops in the same request.

Related ticket https://github.com/TykTechnologies/product/issues/86
## Example

The easiest example to follow will be SOAP usage: you receive payloads to the single endpoint, but you want to apply specific transforms, like: different upstreams for different endpoints, or converting XML to JSON, and forwarding to the REST endpoint, and etc. 

On this general SOAP processing endpoint you define advanced URL rewrite which, based on body content, loops user to another Tyk endpoint, for example: if body match `action=“getPrices”`, redirect to `tyk://self/get_prices?method=GET`. 

After you define internal endpoints, for example `get_prices`, and add endpoint specific plugins there.


asoorm:
This looks awesome -  I'm just thinking about edge-case. Many users don't use listen-path, but rather have each API listen on custom domain.

So rather than `curl http://gateway:8080/httpbin/get`, they would do `curl -H 'Host: httpbin' http://gateway:8080/get`

If the gateway / api-definition is configured in this way, how would our looping logic support this?

api_loader.go
```diff
@@ -474,12 +474,48 @@ func processSpec(spec *APISpec, apisByListen map[string]int,
 	return &chainDef
 }
 
+// Check for recursion
+const LoopLevelLimit = 5
```
asoorm: This could be a default, and allow override via config file / env var?

buger:
Currently, there is no any query string collision since during looping there is no any query param transfer, original request query params not get modified. I thought about adding support of passing custom query params or headers via url with some prefixed variables, but it can be done in future.

buger:
Loop limit now configurable via "loop_limit" query param variable:`tyk://self/path?loop_limit=2`. It can be set only one-time per request and can be set only for loop url.

buger:
@asoorm 
> Many users don't use listen-path, but rather have each API listen on custom domain.
> So rather than curl http://gateway:8080/httpbin/get, they would do curl -H 'Host: httpbin' http://gateway:8080/get

Looping at works in context of current API, it does not matter if API bound to domain or some listen path prefix, when you set url to `tyk://self/get`, it means find `/get` endpoint in current API, not emulate http call to gateway and actually do `/get` HTTP call.

buger:
Since there are no more comments merged and went to testing. Will add more features once it will be actually used and we will understand full scope.

PR #1861, opened 2018-08-07
Handling propagation of keys to slaves on creation and update events in RPC mode

joshblakeley:
Part of https://github.com/TykTechnologies/tyk-sink/issues/26

Currently in RPC we simply flush cache for all key events that are handled and populate local Redis on first use of a key.

This change now means that on key creation and update events the Redis cache in the slave gateway will have the key prepopulated in it's storage before any calls are made.

### DIFF
--- a/api.go
+++ b/api.go
@@ -376,19 +376,19 @@ func handleGetAllKeys(filter, apiID string) (interface{}, int) {
 	return sessionsObj, http.StatusOK
 }
 
-func handleAddKey(keyName, sessionString, apiID string) {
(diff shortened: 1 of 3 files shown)


buger:
Do you think we can achieve this functionality without MDCB changes? What if for created keys we just issue "update" events? And instead of introducing special events for "created" keys, we just force fetching key after all events, even "delete" event?

joshblakeley:
That does make sense actually - and means we could scale back the changes here and simplify. I'll try it and see.

PR #1863, opened 2018-08-07
Handle adding to local cache for key events

joshblakeley:
Fixes #TykTechnologies/tyk-sink#26
Currently in RPC we simply flush cache for all key events that are handled and populate local Redis on first use of a key.

This change now means that on key creation and update events the Redis cache in the slave gateway will have the key prepopulated in it's storage before any calls are made.


joshblakeley:
@buger this is the implementation without mdcb changes

rpc_storage_handler.go
```diff
@@ -999,24 +999,67 @@ func (r *RPCStorageHandler) CheckForKeyspaceChanges(orgId string) {
...
+				r.getSessionAndCreate(splitKeys[0])
 			}
 		} else {
 			log.Info("--> removing cached key: ", key)
 			handleDeleteKey(key, "-1")
+			SessionCache.Delete(key)
+			RPCGlobalCache.Delete(r.KeyPrefix + key)
+			r.getSessionAndCreate(splitKeys[0])
 		}
-
```
buger: Not sure why you moved code from here. Do not think duplication here is needed. 
If `len(splitKeys) == 0` then `key` is equal to  `splitKeys[0]`.

buger:
Lets try to shrink it even more :)

I do not think that `handleAddKey` and `getKeyForCreation` even needed here. 

Point is that your goal is to warm up the cache. If we look at the pipeline, the main function which query the key is `CheckSessionAndIdentityForValidKey`, which actually does all needed lookups and update cache. The only thing you need to do, is to add flag to this function, so it will accept hashed keys, and will not do double hashing. IMHO it should work.

buger:
I actually realized that it could be not that simple, since `CheckSessionAndIdentityForValidKey` is quite bound to request flow, so probably your approach makes sense.

rpc_storage_handler.go
```diff
@@ -999,24 +999,67 @@ func (r *RPCStorageHandler) CheckForKeyspaceChanges(orgId string) {
...
+	}
+	if config.Global().SlaveOptions.EnableRPCCache {
+		// Cache key
+		RPCGlobalCache.Set(newKeyName, value, cache.DefaultExpiration)
+	}
+	//return hash key without prefix so it doesnt get double prefixed in redis
+	return value.(string), newKeyName[7:], nil
+}
+
+func (r *RPCStorageHandler) getSessionAndCreate(keyName string) {
```
buger: Do not think it should be a part of `RPCStorageHandler` class. Leave it as separate helper function.

rpc_storage_handler.go
joshblakeley: I did move some but then stuck it back again so i've just messed that up a bit - i'll fix.

rpc_storage_handler.go
joshblakeley: Makes sense, yep.

rpc_storage_handler.go
```diff
@@ -999,24 +999,67 @@ func (r *RPCStorageHandler) CheckForKeyspaceChanges(orgId string) {
 	}
 }
 
+func (r *RPCStorageHandler) getKeyForCreation(keyName string) (string, string, error) {
```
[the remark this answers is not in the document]
joshblakeley: We can do that since GetRawKey is just sitting there doing nothing at the moment - however this function is more specific in that it will always look for keys with the apikey prefix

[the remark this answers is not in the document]
buger: Lets then generalize it, we will for sure need `GetRawKey` in future.

rpc_storage_handler.go
```diff
@@ -422,8 +422,30 @@ func (r *RPCStorageHandler) GetKey(keyName string) (string, error) {
 }
 
 func (r *RPCStorageHandler) GetRawKey(keyName string) (string, error) {
```
buger: @joshblakeley do you think we can now rewrite `GetKey` function to use `GetRawKey` internally?

buger:
@joshblakeley made a small change to rewrite GetKey use GetRawKey, also noticed that latter one did you used caching (fixed).

PR #1864, opened 2018-08-09
Add delay to fix test failing in CI all the time

joshblakeley:
Added a 100 millisecond delay to let the sentinel token be set in redis for each token in order to stop test failing.


buger:
👍

PR #1874, opened 2018-08-20
Fix quota calculation for Key API

buger:
Fixes:
https://github.com/TykTechnologies/tyk/issues/1879

It was not using "quota-" prefix.


ilijabojanovic:
Verified @buger please merge on master and release-2.7

PR #1875, opened 2018-08-21
Remove build tags and use dynamic Python loader

matiasinsaurralde:
This solves #1283 (as discussed build tags in this scenario are no longer needed).

- Most of CPython glue code was moved and simplified into its own package (`dlpython` directory).
- The dynamic Python loader uses `python-config` to find available Python versions in the system, the latest version is selected by default.
- If the user wants to override a specific version, it's possible to set `python_version` under `coprocess_options`, specifying `3.7` for example.
- By default, when using `go build`, the resulting binary will support Python and gRPC. Python will be loaded only when it's available in the system and a Python plugin is used.
- The Lua build has its own build tag: `go build -tags 'lua'`.
- It's possible to have multiple APIs using different types of plugins at the same time, e.g. API 1 with a gRPC plugin and API 2 with Python.


matiasinsaurralde:
I have refactored this in order to include `dlpython` in the gateway repo.
Also:
- `dlpython` logging was integrated with Tyk logger.
- Tests were adjusted on `dlpython`.`testVersion` in `dlpython/main_test.go` needs to be set to an appropriate version in order to run `dlpython` tests.
- I have tried to come up with a default Python version load approach but seems that this can be problematic, one of the reasons is that `python-config` somestimes refer to older Python versions, the second reason is that if we load `python-config` we don't really know which version we're loading (we only guess this from the filename).
- We should suggest users to pick a specific version of Python or ship with a default one (every distro might have a different one), for example `3.7`.

matiasinsaurralde:
Updates:
- Tests are passing: `go test -timeout 30s github.com/TykTechnologies/tyk/coprocess/python` and `go test -timeout 30s github.com/TykTechnologies/tyk/coprocess/grpc`.
- The dispatcher logic was refactored in order to allow multiple dispatchers in the same process.
- gRPC and Python can work as part of the same binary, no build tags are required, just `go build`. Also removed build tags from these tests.
- Updated schema for the linter to accept `python_version`.

Missing items to fix today:
- Fix/remove `'coprocess'` tests.
- Fix Lua build and make it work under its own build tag.

asoorm:
> * refactored

Does this mean that a gRPC plugin and a python plugin can co-exist in the same gateway? e.g. one API definition can target a gRPC plugin, and another can target Python?

matiasinsaurralde:
> > * The dispatcher logic was refactored in order to allow multiple dispatchers in the same process.
> 
> Does this mean that a gRPC plugin and a python plugin can co-exist in the same gateway? e.g. one API definition can target a gRPC plugin, and another can target Python?

Yes!

Issue #1783, opened 2018-06-19
Allow merging policies with multiple access rights partitions


asoorm:
Currently it is not possible to mix policies with overlaping access partitions (access rights, rate limits and quotas). 

End goal is ability to specify different rate limits for different access rights. 
It should be possible to define different rate limits and quotas for different APIs or for example different paths. So both of the following scenarious should be possible: `ApiA - 100rps, ApiB - 1000rps` and `ApiA /path1 - 100rps, ApiA /path2 - 200rps` 

In order to impement it with minimal changes, and keeping basic architecture, I propose the following:  instead of extending policy object with access rights array support, I propose instead merge multiple policies. It actually can be not really merging, but more a traversing thought the list of policies, to find matching rule. In order to separately track rate limits and quotas, key names used to track it should include matched policy id. 

To make it backward compatible, so users will not loose rate limits and quotas data after the upgrade, we can add check that if multiple access rights defined in multiple policies, turn on new keys format, otherwise use old ones. 

Additionally, for better policy organization, and in oder to make it work for JWT and OpenID based APIs (when you can specify only one policy) I propose to implement "meta" policies, when you can create special policy which consist of multiple policies (similar to keys, matter of adding `apply_policies` field), and if key created with such policy, its `apply_policies` fields gets populated with "meta" policy `apply_policies` field. No need to support deeply nested scenarious, only 1 level.

lonelycode:
To confirm, this is per-token, per-api?

asoorm:
@lonelycode - In other words - Apply rate limits per API, per policy. So I am user X, with Policy Y. Policy Y states that API(1) has 10 rps and API(2) has 20 rps. And that is applied at the token level. 

So instead of creating a token bucket to catch all, you create a token+policy+api bucket combo and feed that instead of feeding a general bucket.

letzya:
So this will be defined on policy def, right?

buger:
@asoorm I updated title and description

dencoded:
Just some thoughts to clarify what needs to be done. We need to be able to specify `rate limit` and `quotas` on three levels:

1. Per Policy (we have that piece now)
2. Per API included in policy's access rights
3. Per path for any API from policy's access rights list

As I understand that would be on a single policy level without having to create several policies and attaching them to a key.

Another thing - we will likely need some validation rules in place, i.e.:
- rate limits on API level can't be bigger than rate limit on policy level (if specified)
[...]

PR #1883, opened 2018-08-22
per API partitions option added

dencoded:
Added changes for https://github.com/TykTechnologies/tyk/issues/1783

While looking at this feature I tried to follow ideas:
- legacy should continue working without any tricks
- we should be able to use the same concept of partitioned policies but with one more edge case when we can specify quota and rate limit in more than one partitioned policy if they have different APIs in ACL
- we should be able specify quota and rate limit on API level (ACL item) explicitly when we create policy 

The idea is to introduce new option "Per API" in partitions section when we create policy. If this option is unchecked - everything should work as before.

Also, there is one condition - that new option "Per API" comes to play only when another partition option "Enforce access control" (aka ACL) is checked as this is what tells our logic to merge access rights from different policies into one array in key session when several (partitioned) policies attached to key. Imho it is weird, maybe can be fixed on UI level.

Also, I've introduced quota and rate limit on API level (field `limit`) in `AccessDefinition` because:
- we need to be able to specify quota and rate limit per ACL item when we create policy
- we need to store quota and rate limit per ACL policy even if it was not specified per ACL item but just on policy level - so all ACL items kind of inherit quota and rate from policy and overwrite it if it was specified explicitly while policy was created
- when quota and rate on ACL item level inherited from policy `AccessDefinition.Limit` gets new field `SetByPolicy` to true - for informative/debug purposes so we know from where that limit on API level has come from

If we have limits on API level we add API ID to storage keys (leaky bucket and other rate limits keys or keys to store quota for key or org session).

If we don't have limit specified on API level quota and rate limit should work as they were working before that change.

if we try to assign several policies with the same API in ACL it should error.

Please note - I didn't include unit tests intentionally (however I've got some) because:
- I rewrote this logic several times while exploring different options to implement this feature so unit test were through away thing every time
- I am kind of got stuck with the way we test rate limits and quotas (as I receive false positives time to time)

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -204,6 +204,9 @@ func (t BaseMiddleware) UpdateRequestSession(r *http.Request) bool {
 // will overwrite the session state to use the policy values.
 func (t BaseMiddleware) ApplyPolicies(key string, session *user.SessionState) error {
 	rights := session.AccessRights
+	if rights == nil {
+		rights = make(map[string]user.AccessDefinition)
+	}
 	tags := make(map[string]bool)
 	didQuota, didRateLimit, didACL := false, false, false
 	didPerAPI := make(map[string]bool)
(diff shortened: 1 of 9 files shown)


dencoded:
@buger 
I've added changes to make "Per API" independent from other partition checkboxes (no acl partition required like it was mentioned before). Also, "Per API" can be set only without partition checkboxes or partition checkboxes can be set if "Per API" is unset.

Switching on "Per API" allows user to create policy with quota/rate limits on API (acl-item) level. It is possible to specify quota/rate limit for API explicitly or leave it blank and it will be populated from policy fields.

It is possible to assign several policies for key with quota/rate limit on API level only if all policies have "Per API" set to true. Also it is not allowed to assign several policies to key where some have "Per API" set and some have partitions set.

Another constraint - if you assign several policies to a key any API can be in only one policy. Having two policies with the same API in ACL is not allowed.

When "Per API" is not set to true - we have the legacy logic in play for policies or partitioned policies assigned to key.

All those points should be covered by unit tests.

PR #1886, opened 2018-08-23
Fix Python memory issues

matiasinsaurralde:
Related to #1228.

Added Python refcount macros. Modified the C.free calls to avoid any GC issues.

Python_DispatchHook uses memcpy to avoid accessing the internal buffer of the resulting PyObject.

### DIFF
--- a/coprocess/lua/binding.h
+++ b/coprocess/lua/binding.h
@@ -8,7 +8,7 @@
 
 static void LuaInit();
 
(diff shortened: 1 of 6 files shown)


matiasinsaurralde:
Tests are ok now!

buger:
@matiasinsaurralde I see a TODO statement in comments, are you going to do it in terms of this PR?

PR #1906, opened 2018-09-11
Remove forward slash from path join

### LINKED ISSUE #1894: Adding more than one Python file to manifest causes middleware initialisation error
bug
Adding more than one Python file to the manifest.json will cause the error:

“Middleware initialization error: [Errno 2] No such file or directory: ‘/middleware.py’”.

Middleware should initialise normally.

Adding more than one Python file to the manifest.json is sufficient to trigger.

The bug is due to:

https://github.com/TykTechnologies/tyk/blob/dfd77e8e1c5bb271527a7654409027c94bea0eb1/coprocess/python/tyk/middleware.py#L31

This should be:

self.mw_path = os.path.join(self.module_path, "middleware.py")

Tried on 2.7.0 and 2.7.1, affected both.

joshblakeley:
Fixes #1894

### DIFF
--- a/coprocess/python/tyk/middleware.py
+++ b/coprocess/python/tyk/middleware.py
@@ -28,7 +28,7 @@ class TykMiddleware:
         self.module_path = os.path.join(self.bundle_root_path, filepath)
         self.parse_manifest()
 
-        self.mw_path = os.path.join(self.module_path, "/middleware.py")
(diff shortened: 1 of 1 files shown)


ilijabojanovic:
@buger approved for merge, please merge in to master and release-2.7 branch

PR #1909, opened 2018-09-12
Check for policy change and apply to access rights each request

### LINKED ISSUE #1737: Policy access rights do not apply to OIDC clients
a bug

1. When configuring OpenId Connect Authentication for an API, I add an issuer, a client ID and apply a existing policy with access rights to it. Everything is fine, I can reach the API.
2. Then I replace the policy applied to the client with another one. The new selected policy has access rights for another api which has a totally different path. 

I can still reach the API. I shouldn't.

When replacing the policy applied to the client (step 2), I shouldn't be able to reach the API. I should have a 403 forbidden response.

See _What is the current behavior?_

Gateway 2.6.1, Pump 0.5.2, Dashboard 1.6.1
Same issue with Gateway 2.5.1, Pump 0.5.1, Dashboard 1.5.1

joshblakeley:
Fixes #1737


ilijabojanovic:
@buger verified. please merge in to release-2.7 and master

Issue #1761, opened 2018-06-07
In-memory and distributed quota proposal


lonelycode:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Requests with quotas hit redis on each request to incremeent their respective quotas, this is wasteful.

**What is the expected behavior?**
We should reduce the number of round-trips to redis

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
All

## Proposal:

As proposed by @dencoded, to reduce the number of calls to redis, we could make the calls periodic and change how we store quotas. 

I was thinking we could use a redis hash which we update with a redis transaction (multipls `HINCR` calls wrapped in a transaction or a pipeline - there are issues here with Cluster).

To make the updates work across a gateway cluster, we increment a server-based `total` with a gateway based `local` count, and then after that transaction we pull the new totals back into our local quota map and base quota violations on the `total+local` count, resetting the local count to 0 as the total count should be the "integrated" value from the entire cluster.

I've created an [in-memory chnnel-based quota counter](https://github.com/TykTechnologies/inmem-quota) (that should be thread-save), that can also handle snapshots and merges.

The approach leaves a  margin for error as the counts are integrated at a periodic interval, but that *should* be ok so long as they are small. 

With this approach:

- There are no writes to redis per requst, so redis traffic is no longer a function of actual traffic, redis writes will be at specific intervals (e.g. every 5s)
- Quotas are calculated and enforced in-memory per-gateway and are updated with channels so should add minimal latency
- This mechanism can be extended to MDCB, and  allow for quotas to be merged across clusters, solving the global quota problem.

The quota lib also has a Clean() method, that will purge keys over a certain "last-update age" from the list, that could be used to clean the master server list too (or in the case of MDCB it could run the Clean function on behalf of the clients)

Opinions please...

dencoded:
I've been looking into this problem as well and come up with some thoughts, mostly matching what was mentioned by @lonelycode. However, I thought there should be simpler solution, probably I am missing something while saying that solution is simpler.

Some summary about things we have at the moment.

We are limiting number of requests on two levels:
1. Organization session quota
2. API key (aka key session) quota

Both levels use the same implementation from `SessionLimiter` method `RedisQuotaExceeded`

We use Redis as a centralized storage for all quota-counters so in case of several gateways serving traffic quota will be counted correctly (unless Redis is available).

Problem.

[...]

PR #1916, opened 2018-09-21
Chunked quota limiter

dencoded:
added changes for https://github.com/TykTechnologies/tyk/issues/1761

Current implementation includes:

1. Key quota can be stored in local Redis or in Master DC if slave options are enabled - in both cases acts as central storage for quotas all gateways use - it is possible to have just on local gateway cluster sharing the same quotas in Redis or have several gateway groups in Slave DCs sharing quotas stored in Master DC over RPC calls
2. When we serve traffic with quota limiter we ask central storage for chunk of quota (based on new store method `IncrementByWithExpire` which is `INCRBY` + `EXPIRE` or `TTL` combination on Redis side), that chunk has ttl equal to quota renewal rate and then we decrement that quota chunk for all subsequent requests while serving traffic. The size of chunk is configurable.
3. When current chunk is done we ask for another one - thats how we don't speak to Redis within every request. The bonus is that chunks are issued for gateways based on their demands - how much traffic they serve
4. If gateway's traffic dropped - active chunk will be returned after some configurable timeout.
5. If current chunk is done and there are no more available (quota exceeded but renewal is not happened yet) the gateway which faces that situation issues special chunk refund  notification to other gateways in that group with hope that next try to get chunk will be successful
6. Any gateway can receive chunk refund notification from others in group and refund part of current chunk to central storage (percentage of that refund is configurable). This happens only if given gateway is the slowest one in group (it uses DRL notifications to understand that).

All this applies to how we calculate quotas per key basis - chunks are issued per keys, chunk expire or refund as well happens per key.

Issues I found during manual testing:

Current implementation `IncrementByWithExpire` can increment to value much higher than actual quota max for the given key, it is Ok for callers as they receive new value and ttl every time back and understand that situation (quota is exceeded, no more chunks available), but refund process will be screwed up - when we refund part of chunk it will be decremented from quota max + some extra.
As a solution we can supply `IncrementByWithExpire` with max value and inside we can simulate conditional update like as sequence of:
1. get current value
2. check if incrementing it by chunk size won't jump over passed max value
3. if it jumps over max value just set it to max value to identify that it is exceeded and make happy refunders - it will be correctly decremented from max and never jumps over max value in future
4. if current value + chunk size are still under max value just add it
5. all this needs to be done atomically
For atomicity we can do with `IncrementByWithExpire` two things:
- implement with optimistic locking using check-and-set i.e. Redis combination of `WATCH + GET + MULTI + SETEX/INCRBY + EXPIRE + EXEC` and then `TTL` to get current remain of key's ttl so this chunk will have it
- or Lua scripting

Another issue - this change needs to respect recent changes connected with quota limits on policy ACL API level. It can be easily done as it is just different key format picking quota max and renewal from acl item instead of key.

I will also add some sequence diagrams here.

### DIFF
--- a/cli/lint/schema.go
+++ b/cli/lint/schema.go
@@ -1,0 +1,778 @@
+package lint
+
+const confSchema = `{
+"$schema": "http://json-schema.org/draft-04/schema#",
+"type": "object",
+"additionalProperties": false,
+"definitions": {
+	"StorageOptions": {
+		"type": ["object", "null"],
+		"additionalProperties": false,
+		"properties": {
+			"database": {
+				"type": "integer"
+			},
+			"enable_cluster": {
+				"type": "boolean"
+			},
+			"use_ssl":{
+				"type": "boolean"
+			},
+			"ssl_insecure_skip_verify":{
+				"type": "boolean"
+			},
+			"host": {
+				"type": "string",
+				"format": "host-no-port"
+			},
+			"hosts": {
+				"type": ["array", "null"]
+			},
+			"optimisation_max_active": {
+				"type": "integer"
+			},
+			"optimisation_max_idle": {
+				"type": "integer"
+			},
+			"password": {
+				"type": "string"
+			},
+			"port": {
+				"type": "integer"
+			},
+			"type": {
+				"type": "string",
+				"enum": ["", "redis"]
+			},
+			"username": {
+				"type": "string"
+			}
+		}
+	}
+},
+"properties": {
+	"allow_insecure_configs": {
+		"type": "boolean"
+	},
+	"allow_master_keys": {
+		"type": "boolean"
+	},
+	"allow_remote_config": {
+		"type": "boolean"
+	},
+	"analytics_config": {
+		"type": ["object", "null"],
+		"additionalProperties": false,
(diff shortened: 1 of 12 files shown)


dencoded:
Here are some sequence diagrams to show channel communication flow.

API traffic request processing, getting new quota chunk:
request processing getting new quota chunk

Refund processing, send refund request from gateway A:
refund processing send refund request on gateway a

Refund processing, handle refund request on some gateway B:
refund processing handle refund request on some gateway b

dql.go
```diff
@@ -0,0 +1,268 @@
...
+		store:              store,
+		chunkedQuotaConfig: chunkedQuotaConfig,
+		chunkReturnTimeout: time.Duration(chunkedQuotaConfig.ChunkReturnTimeout) * time.Second,
+		channels:           make(map[string]*quotaChannels),
+	}
+}
+
+func (d *DQL) getQuotaChannels(key string) *quotaChannels {
+	d.channelsMu.Lock()
+	d.channelsMu.Unlock()
```
matiasinsaurralde: Should it be `defer d.channelsMu.Unlock()`?

PR #1918, opened 2018-09-24
Prepend apiid to path 

joshblakeley:
Fixes #860

This make endpoints in aggregate become namespaced by apiid and path so endpoint popularity graphs will be fixed.

i.e. When record has `trackPath=true` record.path becomes apiid/path

First path element before forward slash will always be apiid in aggregate collections.

### DIFF
--- a/analytics.go
+++ b/analytics.go
@@ -277,6 +277,10 @@ func (r *RedisAnalyticsHandler) recordWorker() {
 			if !strings.HasPrefix(record.RawPath, "/") {
 				record.RawPath = "/" + record.RawPath
 			}
+			//if tracking path prepend apiid for grouping in aggregate queries
+			if record.TrackPath {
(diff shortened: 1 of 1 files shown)


joshblakeley:
@buger confirmed as enough for frontend to have working endpoint popularity functions - can you plz review

joshblakeley:
We are back to just apiid again as using api name wouldnt be good due to it being possible to change over time - meaning records could become out of date with old names.

buger:
@joshblakeley I just realized that probably better place for this code can be actually a Pump, and only mongo pumps. It should be more backward compatible, since if person already use ES pump or similar, they setup will be broken with this code.

joshblakeley:
That sounds good to me. Not sure why we didnt think about that at this point. I'll close this and open pump PR.

PR #1919, opened 2018-09-24
If no key in keymeta found write as empty string

### LINKED ISSUE #1860: During rewrites with dynamic values, like metadata or context, if key not found, searched string does not get replaced with empty value
bug
If I inject some header attribute to url, which depends on token meta, and token meta not found, it will leave in the header string like: `$tyk_meta.key_with_empty_value`. 
If key not found, it should just replace with an empty value.

joshblakeley:
fixes #1860


buger:
Also should work with context variables. Also pls check if it used in other places like body transforms.

joshblakeley:
Just in url rewrites and headers - will add change for context

joshblakeley:
context and meta now left empty if no key match

PR #1922, opened 2018-09-27
moved RPC to rpc package

dencoded:
Added changes to move RPC stack into new `rpc` package which will be reused in context of https://github.com/TykTechnologies/tyk-pump/issues/80

There is some callback business and new function setupRPC() to glue `rpc` package with globals in tyk.

### DIFF
--- a/api_definition.go
+++ b/api_definition.go
@@ -17,6 +17,8 @@ import (
 	"text/template"
 	"time"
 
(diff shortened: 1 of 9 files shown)


dencoded:
cancelled by https://github.com/TykTechnologies/tyk/pull/1923

PR #1923, opened 2018-09-28
move RPC transport and analytics purger into package

dencoded:
Another potentially better approach for https://github.com/TykTechnologies/tyk-pump/issues/80
This approach allows to move some re-usable part of RPC without changing RPC server side.

if it works then it cancels https://github.com/TykTechnologies/tyk/pull/1922 and https://github.com/TykTechnologies/tyk-sink/pull/51

The idea is to do `rpc.Connect(...)` where we supply connection info, a couple of call backs for emergency mode and map of rpc funcs definitions to add to RPC client dispatcher.

Then we can create `rpc.Purger` with all needed fields and run `PurgeLoop` with some ticker.


buger:
👌

joshblakeley:
@dencoded Change here was causing panic in purger in slaved gateways due to reflect bug deep in msgpack. Described at bottom of https://github.com/TykTechnologies/tyk-sink/issues/26

Fixed by copying struct into rpc package and instantiating directly instead of using cast into interface{} 

See https://github.com/TykTechnologies/tyk/pull/1974 for details

We should condider moving analytics.go into its own package also i think.

dencoded:
@joshblakeley moving analytics.go sounds good to me but it is not easy (tried it as part of this change and it didn't work at 1st)

PR #1936, opened 2018-10-11
coprocess: fix double free on dispatch errors (#1933)

matiasinsaurralde:
After debugging this with the bundle that's provided in #1933, I've found that the object is already freed in [coprocess_python.go](https://github.com/TykTechnologies/tyk/blob/master/coprocess_python.go#L140), when a dispatch error occurs.
Dispatch errors are rare, in the scenario described in #1933 it occurs because the dependencies were missing during the test.


ilijabojanovic:
@buger approved for merge

PR #1937, opened 2018-10-15
vendor: remove net/context

matiasinsaurralde:
Also fixes the errors during tests:
`go test -tags 'coprocess grpc'`

From the official source code [here](https://github.com/golang/net/blob/master/context/context.go#L8-L9):
```
As of Go 1.7 this package is available in the standard library under the name context.  https://golang.org/pkg/context.
```


buger:
Closing in favor of https://github.com/TykTechnologies/tyk/pull/1954

Issue #1944, opened 2018-10-24
Whitelist and blacklist regexps are too broad


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
Whitelist `GET /accounts/{accountId}`, and `GET /accounts/{accountId}/paymentPlans` will be whitelisted too
**What is the expected behavior?**
Everything different then `/accounts/{accountId}` should be rejected
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
See above
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.7

buger:
Issues happen because during checking blacklist/whitelist endpoints we replace `{...}` parts with `(?.*)` https://github.com/TykTechnologies/tyk/blob/master/api_definition.go#L446, which is not matched with what `gorilla/mux` router has - match till first `/` character: `([^/]*)`. So basically regexp becomes too broad. 

The way to overcome the issue at the moment is to blacklist all the "nested" paths, in this exact example, blacklist `GET /accounts/{accountId}/paymentPlans`

ilijabojanovic:
@buger issue is still visible n release-2.7 branch.

Here is testing api-def
```
{
    "id": "5c2df1a7cce9f30cdfa34418",
    "name": "basic",
    "slug": "basic",
    "api_id": "c914dc985bdf45524e37972409f3dead",
    "org_id": "5c2df0c0cce9f30c9930598b",
    "use_keyless": false,
    "use_oauth2": false,
    "use_openid": false,
    "openid_options": {
        "providers": [],
        "segregate_by_client": false
    },
    "oauth_meta": {
        "allowed_access_types": [],
        "allowed_authorize_types": [],
        "auth_login_redirect": ""
    },
    "auth": {
        "use_param": false,
        "param_name": "",
        "use_cookie": false,
        "cookie_name": "",
        "auth_header_name": "Authorization",
        "use_certificate": false
    },
    "use_basic_auth": false,
    "basic_auth": {
        "disable_caching": false,
        "cache_ttl": 0
    },
    "use_mutual_tls_auth": false,
    "client_certificates": [],
    "upstream_certificates": {},
    "pinned_public_keys": {},
    "enable_jwt": false,
    "use_standard_auth": true,
    "enable_coprocess_auth": false,
    "jwt_signing_method": "",
    "jwt_source": "",
    "jwt_identity_base_field": "",
[...]

PR #1956, opened 2018-10-31
Fix whitelist/blacklist regex too broad

Gnouc:
Fixes https://github.com/TykTechnologies/tyk/issues/1944

### DIFF
--- a/api_definition_test.go
+++ b/api_definition_test.go
@@ -74,8 +74,8 @@ func TestURLRewrites(t *testing.T) {
 		})
 
 		ts.Run(t, []test.TestCase{
-			{Path: "/rewrite1?show_env=1", Code: 200, BodyMatch: `"URI":"/get?show_env=2"`},
-			{Path: "/rewrite", Code: 200, BodyMatch: `"URI":"/get?just_rewrite"`},
+			{Path: "/rewrite1?show_env=1", Code: http.StatusOK, BodyMatch: `"URI":"/get?show_env=2"`},
+			{Path: "/rewrite", Code: http.StatusOK, BodyMatch: `"URI":"/get?just_rewrite"`},
 		}...)
 	})
 }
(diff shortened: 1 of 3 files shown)


Gnouc:
cc @buger

buger:
Thank you for the PR! 

I think we can make it a bit smaller, like this:

```
apiLangIDsRegex := regexp.MustCompile(`{([^}]*)}`)
asRegexStr := apiLangIDsRegex.ReplaceAllString(stringSpec, `([^/]*)`)
```

Basically removing logic of detection end of string. And at the same time modifying tests to do that:
```
v.Paths.WhiteList = []string{"/foo/{fooId}$", "/foo/{fooId}/bar/{barId}$"}
```

Basic idea here that here at Tyk we have backward compatibility policy, and this regexp was here for a few years, so there is chance that some of our customers relied on its "confusing" behavior. Thankfuly it was possible to fix it using "$" at the end of string, to explicitly set end of the string. In case of #1944  bug this trick was not working for urls containing {...}. Goal of this PR in this case can be making {...} groups support addign "$" to the end of URL.

Hope it makes sense!

api_definition_test.go
```diff
@@ -133,6 +133,27 @@ func TestWhitelist(t *testing.T) {
...
+			updateAPIVersion(spec, "v1", func(v *apidef.VersionInfo) {
+				v.Paths.WhiteList = []string{"/foo/{fooId}", "/foo/{fooId}/bar/{barId}"}
+				v.UseExtendedPaths = false
+			})
+
+			spec.Proxy.ListenPath = "/"
+		})
+
+		ts.Run(t, []test.TestCase{
+			{Path: "/foo", Code: 403},
```
asoorm: would be nice to use http status code constants.

Gnouc:
@buger @asoorm Updated the PR, please take a look.

Gnouc:
@buger @asoorm any updating?

PR #1962, opened 2018-11-02
coprocess: fix tests to match the new dispatcher arguments

matiasinsaurralde:
The dispatcher arguments changed a bit in #1886 so I've updated the tests to cover this.
I've found that we don't actually call `go build -test 'coprocess'` in our current CI configuration file (we do run Python and gRPC tests) but the fix is useful for development tasks.


matiasinsaurralde:
I've fixed the CI.

PR #1965, opened 2018-11-04
Cautious reload

### LINKED ISSUE #1964: Potential condition where a failed API load wipes all loaded APIs out
bug 

When a reload occurs, and there are 0 APIs returned, but there is a list of APIs loaded, it will remove all APIs. This is fine if it was intentional, but if it is an erroneous load then it could potentially cause all services to be downed.

The history of this issue is that we had in place a protection against a nil API list load, but the fix was then amended with the constraint to not be enabled when there are already APIs loaded, which undoes the reason for the original fix :-/

When 0 APIs are returned, do not reload. And find a different fix for #1035 

TBC

2.5+

lonelycode:
Fixes #1964 partially by removing the loaded API constraint

### DIFF
--- a/main.go
+++ b/main.go
@@ -609,8 +609,9 @@ func doReload() {
 	// load the specs
 	count := syncAPISpecs()
 	// skip re-loading only if dashboard service reported 0 APIs
-	// and current registry had 0 APIs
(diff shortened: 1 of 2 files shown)


lonelycode:
Dropping this in favour of #1966

PR #1969, opened 2018-11-06
Enable logstash hook to handle reconnections

### LINKED ISSUE #1968: Logstash hook for logs cannot handle reconnections

bug

If logstash is reconfigured or times out the gateway needs to be restarted to reconnect the hook.

The hook should attempt to reconnect when connection is broken.

2.x

joshblakeley:
Fixes #1968

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,6 +32,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/justinas/alice"
 	"github.com/lonelycode/osin"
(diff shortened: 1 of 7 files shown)


buger:
Can you add a bit more details, since this is a quite non-standard solution. 
Does it means that built-in logstash driver does not handle re-connections?

joshblakeley:
Exactly see issue here:

https://github.com/bshuster-repo/logrus-logstash-hook/issues/48

Issue #1772, opened 2018-06-12
Access certificate data inside plugins


buger:
**Do you want to request a *feature* or report a *bug*?**
feature
**What is the current behavior?**
At the moment Plugin can access only certificate fingerprint. In order to implement custom certificate authentification it is not enough, and having access to certificate DN is required. DN itself looks like ` CN=client1,OU=MyClients,O=MongoDB-Cluster,L=Austin,ST=TX,C=US` which is basically representation of certificate Subject fields.
**What is the expected behavior?**
We can add new plugin field, called `CertificateDN` which will expose this information.

buger:
Should be streightforward in Go like this:

```
cert, err := x509.ParseCertificate(d)
fmt.Printf("%+v\n", cert.Subject.ToRDNSequence())
```

buger:
Should be great if we can get access to DN inside advanced url rewriter too.

maxtarasuik:
Looks awesome, I suspect this will be of more use when/if Mutual TLS works with a CA rather than individual certificates.
That way you can Authenticate with a CA, and Authorize users using this in a plugin (did I get your terminology right ?)

I should add I don't think you can Authenticate with only a DN, as anyone could sign a cert with "CN=client1,OU=MyClients,O=MongoDB-Cluster,L=Austin,ST=TX,C=US". 

 👍

buger:
This functionality should extend request object to provide 2 things:
1) raw certificate information (can be accessed via req.TLS.PeerCertificates[...].Raw), so the plugin can do parsing and validation by itself
2) certificate meta (we already have helpers to extract it, see `CertificateMeta` in code).

Note that `req.TLS.PeerCertificates[...]` is an array, so both fields mentioned above also should be an arrays. 

For GRPC compatibility, the second field with meta can be a list of JSON strings.

andyo-tyk:
I'm closing this because, whereas the PR linked above was never merged into a release, we hae subsequently developed support for Go Plugins; these have access to the request object and certificate

PR #1970, opened 2018-11-07
Allow certificate data access in plugins

matiasinsaurralde:
For #1772, still need to refactor certificate meta to see how it can be added to the request object.
This allows raw certificate data access.


buger:
You have type "bytes" for this record, but in fact this field can contain multiple certificates. Should not it be array of bytes instead? 

As for Certificate Meta, maybe just serialize to JSON? Also to avoid overhead, I suggest adding new manifest variable `require_client_certificate`, similar to `require_session`, and only if its true, it will load this variables.

coprocess_grpc_test.go
```diff
@@ -204,26 +208,28 @@ func loadTestGRPCAPIs() {
 	})
 }
 
-func startTykWithGRPC() (*tykTestServer, *grpc.Server) {
+func startTykWithGRPC(grpcPort int) (*tykTestServer, *grpc.Server) {
 	// Setup the gRPC server:
-	listener, _ := net.Listen("tcp", grpcListenAddr)
+	listener, _ := net.Listen("tcp", fmt.Sprintf(":%d", grpcPort))
```
lsjurczak: It's better to use `strconv.Itoa`. Sprintf will use the package reflect, and it will allocate one more object.

buger:
@matiasinsaurralde bump

matiasinsaurralde:
Updated with above changes and tests!

buger:
Nice! Do you think we can make it available to JSVM plugins too?

marksou:
I'm assuming the docs will need updating?

buger:
yes, I added "needs documenting" label for the task. 
It will be released as part of 2.8.1 and 2.7.5

matiasinsaurralde:
I will rebase this (we updated the PB bindings recently) and add JSVM support, should be possible!

PR #1972, opened 2018-11-08
Optionally don't strip trailing slash when proxying only root target url

### LINKED ISSUE #1538: Root listen path 302 redirect
bug

After upgrading from tyk gateway 1.3.4 to 1.5.3, my existing apis 302 requests from the root of the listen path to the path specified in my target url:

listen path = '/example_api/'
target url = 'http://api.example.com/api/'
```
curl http://example.com/example_api/ -I
HTTP/1.1 302 Found
Content-Type: text/plain; charset=utf-8
Date: Wed, 14 Mar 2018 17:29:44 GMT
Location: /api/
X-Ratelimit-Limit: -1
X-Ratelimit-Remaining: -1
X-Ratelimit-Reset: 1507675133
Connection: keep-alive
```

I should be able to do a GET on the root of my listen path and be returned whatever my app is serving there.

In my experience, any API I have set up is no longer accessible through tyk at the root of my listen path.

It used to work as expected in 1.3.4, but after upgrading to 1.5.3 I see this issue

joshblakeley:
Fixes #1538 

Uses new api definition config option `proxy.disable_strip_prefix` 

It only takes effect in one case: when there is no path to append onto the target url and when the config option is `true`

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -377,6 +377,7 @@ type APIDefinition struct {
 		PreserveHostHeader          bool                          `bson:"preserve_host_header" json:"preserve_host_header"`
 		ListenPath                  string                        `bson:"listen_path" json:"listen_path"`
 		TargetURL                   string                        `bson:"target_url" json:"target_url"`
+		DisableStripPrefix          bool                          `bson:"disable_strip_prefix" json:"disable_strip_prefix"`
 		StripListenPath             bool                          `bson:"strip_listen_path" json:"strip_listen_path"`
 		EnableLoadBalancing         bool                          `bson:"enable_load_balancing" json:"enable_load_balancing"`
 		Targets                     []string                      `bson:"target_list" json:"target_list"`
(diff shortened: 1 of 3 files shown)


reverse_proxy.go
```diff
@@ -234,9 +234,11 @@ func TykNewSingleHostReverseProxy(target *url.URL, spec *APISpec) *ReverseProxy
 		if targetToUse == target {
 			req.URL.Scheme = targetToUse.Scheme
 			req.URL.Host = targetToUse.Host
-			req.URL.Path = singleJoiningSlash(targetToUse.Path, req.URL.Path)
+			fmt.Println(spec.Proxy.TargetURLDontStripSlash)
```
buger: Remove debugging info here, and few lines below.

apidef/api_definitions.go
```diff
@@ -377,6 +377,7 @@ type APIDefinition struct {
 		PreserveHostHeader          bool                          `bson:"preserve_host_header" json:"preserve_host_header"`
 		ListenPath                  string                        `bson:"listen_path" json:"listen_path"`
 		TargetURL                   string                        `bson:"target_url" json:"target_url"`
+		TargetURLDontStripSlash     bool                          `bson:"target_url_dont_strip_slash" json:"target_url_dont_strip_slash"`
```
buger: `dont` usually not used in naming variables. 
Also, since this is already inside the proxy section, you probably can name it: `disable_strip_prefix` ?

joshblakeley: i like that better, yes - i was unsure on naming

Issue #1833, opened 2018-07-20
There is no way to access regexp group inside URL rewriter


buger:
**Do you want to request a *feature* or report a *bug*?**
bug/feature
**What is the current behavior?**
Advanced URL rewriter allows you to define regexp matches against values, but at the moment you get only whole "matched" value, and if you are using groups, they basically useless.
**What is the expected behavior?**
At the moment Tyk define special `$tyk_context.trigger-{n}-{name}-{i}` https://tyk.io/docs/transform-traffic/url-rewriting/ context value which holds trigger matched string. I propose extend it, and define additional level, which allows you specify group index too `$tyk_context.trigger-{n}-{name}-{i}-{group}`

This is especially needed for handling SOAP based protocols, then you will be able to define body regexp like: `<wsdlsoap:address location="https://example.com/([^"]+)"``, and inside rewrite_to use it like: `$tyk_context.trigger-0-payload-1`, which will give us automatic mapping between SOAP actions and Tyk endpoints. 

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.7

dskoval:
Interested to work on this
I propose following as implementation:
- `$tyk_context.trigger-{n}-payload` - correspond to first match (as current implementation)
- `$tyk_context.trigger-{n}-payload-{i}` - i'th regexp match (i starts from 0)
- `$tyk_context.trigger-{n}-payload-{i}-{g}` - g'th capturing group for i'th regexp match (g starts from 0)

buger:
https://github.com/TykTechnologies/tyk/pull/1973 was merged to `release-2.7` branch
<details><summary></summary>Created via API</details>

buger:
https://github.com/TykTechnologies/tyk/pull/1973 was merged to `release-2.8` branch
<details><summary></summary>Created via API</details>

PR #1973, opened 2018-11-08
Add access to regexp capturing group Inside URL rewriter

dskoval:
- add additional level  for $tyc_context trigger keys
- $tyk_context.trigger-{n}-{name}-{i}-{groupNum}  holds group value

Fixes https://github.com/TykTechnologies/tyk/issues/1833

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -540,6 +540,14 @@ func (s *StringRegexMap) Check(value string) string {
 	return s.matchRegex.FindString(value)
 }
 
+func (s *StringRegexMap) FindStringSubmatch(value string) []string {
+	return s.matchRegex.FindStringSubmatch(value)
+}
+
+func (s *StringRegexMap) FindAllStringSubmatch(value string, n int) [][]string {
+	return s.matchRegex.FindAllStringSubmatch(value, n)
+}
+
 func (s *StringRegexMap) Init() error {
 	var err error
 	if s.matchRegex, err = regexp.Compile(s.MatchPattern); err != nil {
--- a/mw_url_rewrite.go
+++ b/mw_url_rewrite.go
@@ -16,8 +16,10 @@ import (
 )
 
 const (
-	metaLabel    = "$tyk_meta."
-	contextLabel = "$tyk_context."
+	metaLabel        = "$tyk_meta."
+	contextLabel     = "$tyk_context."
+	triggerKeyPrefix = "trigger"
+	triggerKeySep    = "-"
 )
 
 var dollarMatch = regexp.MustCompile(`\$\d+`)
@@ -178,7 +180,7 @@ func replaceTykVariables(r *http.Request, in string, escape bool) string {
 
 		replaceGroups := contextMatch.FindAllStringSubmatch(in, -1)
 		for _, v := range replaceGroups {
-			contextKey := strings.Replace(v[0], "$tyk_context.", "", 1)
+			contextKey := strings.Replace(v[0], contextLabel, "", 1)
 
 			if val, ok := contextData[contextKey]; ok {
 				valStr := valToStr(val)
@@ -202,7 +204,7 @@ func replaceTykVariables(r *http.Request, in string, escape bool) string {
 
 		replaceGroups := metaMatch.FindAllStringSubmatch(in, -1)
 		for _, v := range replaceGroups {
-			contextKey := strings.Replace(v[0], "$tyk_meta.", "", 1)
+			contextKey := strings.Replace(v[0], metaLabel, "", 1)
 
 			val, ok := session.MetaData[contextKey]
 			if ok {
@@ -362,10 +364,13 @@ func checkHeaderTrigger(r *http.Request, options map[string]apidef.StringRegexMa
 		vals, ok := r.Header[mhCN]
 		if ok {
 			for i, v := range vals {
-				b := mr.Check(v)
-				if len(b) > 0 {
-					kn := "trigger-" + strconv.Itoa(triggernum) + "-" + mhCN + "-" + strconv.Itoa(i)
-					contextData[kn] = b
+				match := mr.FindStringSubmatch(v)
+				if len(match) > 0 {
+					kn := buildTriggerKey(triggernum, mhCN, i)
+					contextData[kn] = match[0]
+
+					addGroupsToContextData(&contextData, kn, match[1:])
+
 					fCount++
 				}
 			}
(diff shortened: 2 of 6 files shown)


dskoval:
@buger 
will also try to remove some code duplication in rewrite middleware

buger:
Nice one! 

@dencoded can you pls review?

apidef/api_definitions.go
```diff
@@ -540,6 +540,14 @@ func (s *StringRegexMap) Check(value string) string {
 	return s.matchRegex.FindString(value)
 }
 
+func (s *StringRegexMap) FindStringSubmatch(value string) []string {
+	return s.matchRegex.FindStringSubmatch(value)
```
dencoded: could you also add caching logic for `FindStringSubmatch` in `tyk/regexp`?

apidef/api_definitions.go
```diff
@@ -540,6 +540,14 @@ func (s *StringRegexMap) Check(value string) string {
...
+}
+
+func (s *StringRegexMap) FindAllStringSubmatch(value string, n int) [][]string {
+	return s.matchRegex.FindAllStringSubmatch(value, n)
```
dencoded: the same thing for `FindAllStringSubmatch ` we need to add caching logic for this function in `tyk/regexp` so we won't spend time on RX with the same params

dskoval: Looks like it's already done
https://github.com/TykTechnologies/tyk/blob/23e56e71ddfa5beb0f50a2aebe32e2ebccb3e68e/regexp/regexp.go#L402

dencoded: ah, right. this one is already cached

apidef/api_definitions.go
dskoval: Please, check https://github.com/TykTechnologies/tyk/pull/1973/commits/990474b4369a223c92c547d2fbc3d4d22fa58eee - Add caching logic for FindStringSubmatch in tyk/regexp

buger:
Thank you a lot for all changes! 

@dencoded does it looks good to you?

dskoval:
Hi, at this very moment i'm removing repeating/unused code in cached regexp implementation.
At bare minimum i still have to compare benchmarks between mine and original implementation.
If it will look promising, should i add it to this PR, create issue for it or just create new PR?
Asking because i'm new to GitHub and doesn't want to create mess

buger:
Better new PR, to separate feature thing from refactoring itself. Thank you!

mw_url_rewrite.go
```diff
@@ -362,10 +364,13 @@ func checkHeaderTrigger(r *http.Request, options map[string]apidef.StringRegexMa
...
-				b := mr.Check(v)
-				if len(b) > 0 {
-					kn := "trigger-" + strconv.Itoa(triggernum) + "-" + mhCN + "-" + strconv.Itoa(i)
-					contextData[kn] = b
+				match := mr.FindStringSubmatch(v)
+				if len(match) > 0 {
+					kn := buildTriggerKey(triggernum, mhCN, i)
+					contextData[kn] = match[0]
+
+					addGroupsToContextData(&contextData, kn, match[1:])
```
dencoded: `match[1:]` looks a little suspicious - what if regex didn't have any groups? in worst case I would expect panic with index bound of range, in best case - call to function with no-op inside as `match[1:]` would give empty slice

mw_url_rewrite.go
```diff
@@ -392,11 +397,13 @@ func checkQueryString(r *http.Request, options map[string]apidef.StringRegexMap,
 		vals, ok := qvals[mv]
 		if ok {
 			for i, v := range vals {
-				b := mr.Check(v)
-				if len(b) > 0 {
-					kn := "trigger-" + strconv.Itoa(triggernum) + "-" + mv + "-" + strconv.Itoa(i)
+				match := mr.FindStringSubmatch(v)
```
dencoded: I've noticed logic like:
```
match := mr.FindStringSubmatch(v)
if len(match) > 0 {
	kn := buildTriggerKey(triggernum, mv, i)
	contextData[kn] = match[0]
	addGroupsToContextData(&contextData, kn, match[1:])
```
repeated in several places, might worth to move it to function if possible

mw_url_rewrite.go
```diff
@@ -479,13 +490,39 @@ func checkPayload(r *http.Request, options apidef.StringRegexMap, triggernum int
...
+	if len(indices) > 0 {
+		for _, index := range indices {
+			parts = append(parts, strconv.Itoa(index))
+		}
+	}
+
+	return strings.Join(parts, triggerKeySep)
+}
+
+func addGroupsToContextData(cd *map[string]interface{}, keyPrefix string, groups []string) {
```
dencoded: you don't need to pass map as pointer as maps (and slices) are passed by reference

dencoded:
@buger @dskoval left some comments which I believe need to be addressed

apidef/api_definitions.go
dencoded: 👍

mw_url_rewrite.go
dskoval: All expressions `match[1:]` placed inside of `if len(match) > 0 { .... }` blocks. 
If regexp doesn't have any groups it will result in `match` having only one item and `match[1:]` would be an empty slice with no-op call to function `addGroupsToContextData`
https://play.golang.org/p/-2KdJvgW8D9

@dencoded 
i'll wrap calls `addGroupsToContextData(&contextData, kn, match[1:])` into `if len(match) > 1 { .... }` - it doesn't affect functionality but increase readability

mw_url_rewrite.go
dskoval: Hmm, for me this one looks suspicious.
Intent of `addGroupsToContextData` is to modify (add items to) passed map.
As far as i know, in golang everything is value and passed by value (copied). Slices can look like passed by reference, because internally they holds pointer to baking array, first item index, length and capacity.
It can feels like they passed by reference (in case of modifying existing values), but in case of adding items they not. What's why `append()` returns updated slice.
https://play.golang.org/p/arTxwX1w9mf

I've checked same approach with map and it works, probably because in essence map is just a pointer to underlying structure - i don't know.

Will remove pointers to map as requested

dskoval:
@buger @dencoded implemented requested changes in last commit
https://github.com/TykTechnologies/tyk/pull/1973/commits/68e92f49beb9ccc4378d4fad492b1f680e601ae4

mw_url_rewrite.go
dencoded: this is tricky one - slices are different because they have `len` and `cap` in addition to pointer to underlaying array with elements, so when we pass it to function it receives new "descriptor" of slice but original pointer to elements.

if we do append inside function we have to return slice and re-assign result on calling side, otherwise calling part will hold slice-descriptor with appended elements but old `len` and `cap` (thats why Println outputs only one element). you can find more info about that situation here https://golang.org/doc/effective_go.html#slices

Changing elements would work if your slice is pre-allocated, i.e.:
https://play.golang.org/p/ZoK7WLz_Xtw

That's actually how all functions like `func Read(slice []interface{}) (int, error)` reading data into passed slice without any explicit pointers.

But my comment was about maps, they work differently.

Issue #1989, opened 2018-11-24
During rewrites with dynamic values, if key not found, whole searched string get replaced with empty value


dskoval:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
If I inject some variable (`$tyk_meta.key` or `$tyk_context.key`) and corresponding value not found in metadata (context) whole `rewrite_to` value replaced with empty string. As result, downstream call would be made to target_url specified in proxy section.
**What is the expected behavior?**
Replace only corresponding variable with empty string

Related to #1860 

Lines causing such behavior:
https://github.com/TykTechnologies/tyk/blob/d31e82a37f31c287859898c131ac0d2200632f03/mw_url_rewrite.go#L192-L194
https://github.com/TykTechnologies/tyk/blob/d31e82a37f31c287859898c131ac0d2200632f03/mw_url_rewrite.go#L217-L219
should be replaced with `in = strings.Replace(in, v[0], "", -1)`

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
master

dskoval:
One more note:
rewriter also replaces `$tyk_meta.### ` with corresponding values from session metadata. If session not present it will return url with `$tyk_meta.### ` variables in it. I think in such case all `$tyk_meta.### ` should be replaced with empty strings too.

buger:
https://github.com/TykTechnologies/tyk/pull/1993 was merged to `release-2.7` branch
<details><summary></summary>Created via API</details>

buger:
https://github.com/TykTechnologies/tyk/pull/1993 was merged to `release-2.8` branch
<details><summary></summary>Created via API</details>

PR #1993, opened 2018-11-26
Fix url rewrite issue #1989

dskoval:
- if varible not found, replace only varible with empty string
- if sessin not found, replace session vars with empty string

fix https://github.com/TykTechnologies/tyk/issues/1989

### DIFF
--- a/mw_url_rewrite.go
+++ b/mw_url_rewrite.go
@@ -177,49 +177,38 @@ func urlRewrite(meta *apidef.URLRewriteMeta, r *http.Request) (string, error) {
 func replaceTykVariables(r *http.Request, in string, escape bool) string {
 	if strings.Contains(in, contextLabel) {
 		contextData := ctxGetData(r)
-
-		replaceGroups := contextMatch.FindAllStringSubmatch(in, -1)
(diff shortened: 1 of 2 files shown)


buger:
Thank you! Can you a bit explain why `FindAllStringSubmatch` was replaced by `FindAllString` ?

dskoval:
Sure. FindAllStringSubmatch returns slice of string slices `[][]string`. In every subslice first element is match and rest of elements are capturing group matches. In current implementation
https://github.com/TykTechnologies/tyk/blob/5e01175e7bf9fc340b970dd236b15d28418c8072/mw_url_rewrite.go#L181-L191
only first element (`v[0]`, lines 183 and 191) which is actual match is used, capturing groups simply not needed. 
Because of that i replaced FindAllStringSubmatch with FindAllString which returns only matches.

Though on other hand call to FindAllStringSubmatch could be left as-is and later returned capturing group could be used as key without call to https://github.com/TykTechnologies/tyk/blob/5e01175e7bf9fc340b970dd236b15d28418c8072/mw_url_rewrite.go#L183

buger:
Makes sense!

PR #1999, opened 2018-11-27
Optimize regexp key generation

### LINKED ISSUE #1998: Cached regexp performance improvement
Enhancement

Current cached regexp implementation make at least one mem alloc when building cache key. 
It is possible to create reusable keybuilder, similar to bytes.Buffer and strings.Builder and reuse it from sync.Pool. This will reduce allocations required for key building to zero in case of item lookup and to exactly one alloc in case of storing item in cache.
This improvement will affect url rewrite middleware and other parts which use cached regexp.

Related to #1662 

master

dskoval:
This pull request will introduce reusable keybuilder which reduce allocations in cached regexp
Resolve #1998 
Related to #1662

### DIFF
--- a/mw_url_rewrite_test.go
+++ b/mw_url_rewrite_test.go
@@ -58,15 +58,38 @@ var testRewriterData = []struct {
 	},
 }
 
+type testRewriterCase struct {
+	name     string
+	meta     *apidef.URLRewriteMeta
+	reqMaker func() *http.Request
+	want     string
+}
+
+func prepareRewriterCases() []testRewriterCase {
+	tcs := make([]testRewriterCase, len(testRewriterData))
+	for i, td := range testRewriterData {
+		reqTarget := td.in
+		tcs[i] = testRewriterCase{
+			name: td.name,
+			meta: &apidef.URLRewriteMeta{
+				MatchPattern: td.pattern,
+				RewriteTo:    td.to,
+			},
+			reqMaker: func() *http.Request {
+				return httptest.NewRequest("GET", reqTarget, nil)
+			},
+			want: td.want,
+		}
+	}
+	return tcs
+}
+
 func TestRewriter(t *testing.T) {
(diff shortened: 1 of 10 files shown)


dskoval:
Benchmarks above compare results after rewriter benchmarks fix in https://github.com/TykTechnologies/tyk/pull/1999/commits/af94f1219b07458d9d7ce7a3efd15cb56e79637f with results after optimization in https://github.com/TykTechnologies/tyk/pull/1999/commits/acb6061d459358f04c2b46cc9aeb0c78e6ea6855

dskoval:
PS It is possible shave off some nanoseconds by explicitly putting keybuilder back to pool before each return clause instead of deferred call pool.Put(). This will make regexps benchmark CPU performance on par with previous implementation and rewriter CPU performance even 1-2% better. But on other hand it will make code more verbose. So i'm not sure is it worth to do or not.

buger:
Using `unsafe` package is a bit tricky and dangerous. Do tests cover UTF-8/16 cases?

dskoval:
Agree, because of that unsafe method named UnsafeKey() and provided with comments on usage.

Could you explain a bit your concerns about UTF-8/16 cases, please?
In Go strings are UTF-8 encoded. There is separate package https://golang.org/pkg/unicode/utf16/ to deal with UTF-16 characters (convert 2 bytes to rune and vice versa).
If you mean handling multi-byte characters, i didn't foresee any issues with it (byte slices and strings are pretty same under the hood)

dskoval:
If you are unsure about how it will work with multibyte characters - let me know. I'll simply add/change tests to include them, it would not take much time

buger:
This will be great. This part of the code very senstive.

dskoval:
@buger added multi-byte characters in keyBuilder tests arguments, please check last commit

buger:
Well, seems like you removed all blockers. 

Nice work 👍

Issue #1996, opened 2018-11-27
Global cache is broken


ilijabojanovic:
**Do you want to request a *feature* or report a *bug*?**
BUG
**What is the current behavior?**
Error on console for cached content
**What is the expected behavior?**
We should not have error and gw should return cached body
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. Create api definition 
2. On advanced tab enable Cache all safe requests
3. Trigger several calls to gw

Actual:
```
[Nov 27 10:35:40] ERROR proxy: http: proxy error during body copy: unexpected EOF api_id=a5105834b7c241947b7956b23b6393e7 org_id=5bd82c299b7d7c0873b203f1
[Nov 27 10:35:41]  WARN Incorrect key expiry setting detected, correcting
[Nov 27 10:35:41] ERROR proxy: http: proxy error during body copy: unexpected EOF api_id=a5105834b7c241947b7956b23b6393e7 org_id=5bd82c299b7d7c0873b203f1
[Nov 27 10:35:42]  WARN Incorrect key expiry setting detected, correcting
[Nov 27 10:35:42] ERROR proxy: http: proxy error during body copy: unexpected EOF api_id=a5105834b7c241947b7956b23b6393e7 org_id=5bd82c299b7d7c0873b203f1
```
Api definition for testing:
```
{
    "id": "5bec76cfcce9f328a2fb6107",
    "name": "load",
    "slug": "load",
    "api_id": "a5105834b7c241947b7956b23b6393e7",
    "org_id": "5bd82c299b7d7c0873b203f1",
    "use_keyless": false,
    "use_oauth2": false,
    "use_openid": false,
    "openid_options": {
        "providers": [],
        "segregate_by_client": false
    },
    "oauth_meta": {
        "allowed_access_types": [],
        "allowed_authorize_types": [],
        "auth_login_redirect": ""
    },
    "auth": {
        "use_param": true,
        "param_name": "auth",
        "use_cookie": false,
        "cookie_name": "",
        "auth_header_name": "Authorization",
        "use_certificate": false
    },
    "use_basic_auth": false,
    "basic_auth": {
        "disable_caching": false,
        "cache_ttl": 0
    },
    "use_mutual_tls_auth": false,
    "client_certificates": [],
    "upstream_certificates": {},
    "pinned_public_keys": {},
    "enable_jwt": false,
    "use_standard_auth": true,
    "enable_coprocess_auth": false,
[...]

PR #2000, opened 2018-11-28
analytics RecordHit fixed

dencoded:
added changes for https://github.com/TykTechnologies/tyk/issues/1996

Thus situation was happening when detailed analytics recording (copy of actual response in analytics record) is enabled.

The code in RecordHit was screwing up copyResponse.Body underlaying buffer by reseting its length to 0. So item put into cache had only headers without body and it was causing that error `http: proxy error during body copy: unexpected EOF` as we use reverse-proxy code to instantiate response from cache item value.

### DIFF
--- a/gateway_test.go
+++ b/gateway_test.go
@@ -646,6 +646,101 @@ func TestAnalytics(t *testing.T) {
 			t.Error("Analytics record do not match", record)
 		}
 	})
+
+	t.Run("Detailed analytics", func(t *testing.T) {
+		defer resetTestConfig()
+		globalConf := config.Global()
+		globalConf.AnalyticsConfig.EnableDetailedRecording = true
+		config.SetGlobal(globalConf)
+
+		buildAndLoadAPI(func(spec *APISpec) {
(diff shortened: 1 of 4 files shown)


dencoded:
Agree that it is tricky code, unfortunately I couldn't find better way to do a fix as the place when source Body buffer gets reset to len 0 is inside Go's standard library code.

According to empty empty `RawResponse` - I think this is because we don't pass response copy for cache hits here https://github.com/TykTechnologies/tyk/blob/master/mw_redis_cache.go#L266
I didn't change that peace and it looks like a legacy thing.

dencoded:
@buger after looking again at code history I am pretty sure we have introduced problem with EOF from cache and empty RawResponse here: https://github.com/TykTechnologies/tyk/pull/1778

so might worth to revise that merged PR and come up with better fix

dencoded:
I will look into that PR changes with potentially another fix

dencoded:
@buger PTAL

buger:
Nice! So much clearer.

PR #2005, opened 2018-11-30
Fix quota calculated twice

dskoval:
don't take into account "loop" requests when incrementing quota counter

fix https://github.com/TykTechnologies/tyk/issues/2001

### DIFF
--- a/session_manager.go
+++ b/session_manager.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/TykTechnologies/leakybucket"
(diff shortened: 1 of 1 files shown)


buger:
Replaced by https://github.com/TykTechnologies/tyk/pull/2014

It is a bit simpler, but explicitly in all middlewares which accountable for RL and Quotas

dskoval:
@buger  Indeed, it's simpler, but looks like you missed some "cosmetic" stuff. Your fix doesn't return correct X-Ratelimit-Remaining header. Check GIF below, I've recorded it with your fix and api setup from #2001 
What's exactly why my fix a bit more complicated

peek 2018-12-06 15-43

buger:
Thats, interesting, thank you for pointing to this! Will check.

dskoval:
BTW, i checked code one more time and it looks like all middlewares delegate rate limit check to `SessionLimiter`
https://github.com/TykTechnologies/tyk/blob/f8f6306e88f8db3320056ef206869a4babd95b35/mw_api_rate_limit.go#L71-L73
https://github.com/TykTechnologies/tyk/blob/f8f6306e88f8db3320056ef206869a4babd95b35/mw_rate_limiting.go#L72-L74
https://github.com/TykTechnologies/tyk/blob/f8f6306e88f8db3320056ef206869a4babd95b35/mw_organisation_activity.go#L87-L92
`ProcessRequestOffThread` and `ProcessRequestLive` in turn also call `sessionLimiter.ForwardMessage`
So IMO it's `sessionLimiter` responsibility to handle quota checks (and is it loop or not) if you want to achieve consistent behavior among all rate limiting middlewares

buger:
Frankly, I'm a bit puzzled by your code, because you only read the counter, and do not update it. Not sure how it can fix this bug.

Issue #2016, opened 2018-12-06
Tyk crashes when running xml body transform on non xml response


andrewscode:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When using body transform for XML responses,  Tyk crashes when posting to an endpoint which returns a blank response instead of xml
(Technically it crashes if valid xml is not returned back)

**What is the expected behavior?**
Tyk should return the usual "Error proxying request"

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. Create API with Body Transform on the Response and set type to XML
2. Provide anything in the template  (non-empty)
3. Send to an endpoint that returns a non 200 status code (or simply non xml) .

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.7.4

andrewscode:
Stack trace
[tyk-stacktrace.txt](https://github.com/TykTechnologies/tyk/files/2653985/tyk-stacktrace.txt)

matiasinsaurralde:
Just a note, to reproduce this context vars should be enabled in the API settings.

buger:
Should be fixed by https://github.com/TykTechnologies/tyk/pull/2070

andrewscode:
This is still crashing with the latest release.

andrewscode:
It's important to note (after reading #2070) that the final endpoint isn't returning an empty body.  It's returning non-xml

buger:
@andrewscode re-opening the issue. Can you pls share the full Tyk stack trace?

andrewscode:
Sure,  it's relatively the same as the previous stack trace
[stack2.txt](https://github.com/TykTechnologies/tyk/files/2895548/stack2.txt)

ilijabojanovic:
@buger please merge this in to release-2.8

PR #2017, opened 2018-12-06
Handle nil map in body transform

matiasinsaurralde:
This is a potential fix for #2016.

This fix makes sure that we don't write to the `nil` map that's initialized by `mxj.NewMapXml` when passing an invalid XML input.

We don't throw any HTTP errors from this middleware, if we decide to do that we should modify the lines that handle the request body.


buger:
Should be fixed by https://github.com/TykTechnologies/tyk/pull/2070, but via another approach, by properly handling empty bodies.

PR #2018, opened 2018-12-07
special case for basic auth added

dencoded:
Fix https://github.com/TykTechnologies/tyk-analytics/issues/976 https://github.com/TykTechnologies/tyk/issues/2038

Added new attributes `org_id` and `username` (boolean) attributes, which should be passed together when you query for username.

### DIFF
--- a/api.go
+++ b/api.go
@@ -657,23 +657,10 @@ func keyHandler(w http.ResponseWriter, r *http.Request) {
 	apiID := r.URL.Query().Get("api_id")
 	isHashed := r.URL.Query().Get("hashed") != ""
 	isUserName := r.URL.Query().Get("username") == "true"
+	orgID := r.URL.Query().Get("org_id")
 
 	// check if passed key is user name and convert it to real key with respect to current hashing algorithm
 	if r.Method != http.MethodPost && isUserName {
-		orgID := "default"
-		// check if we have real orgID
-		if !strings.HasPrefix(keyName, "default") && len(keyName) > 24 {
-			orgID = keyName[:24]
-		}
-		// check if organization ID is real
-		if spec := getSpecForOrg(orgID); spec == nil {
-			doJSONWrite(
-				w,
-				http.StatusNotFound,
-				apiError("Key not found"),
-			)
-			return
-		}
 		keyName = generateToken(orgID, keyName)
 	}
 
--- a/api.go
+++ b/api.go
@@ -324,6 +324,10 @@ func handleGetDetail(sessionKey, apiID string, byHash bool) (interface{}, int) {
 	spec := getApiSpec(apiID)
 	if spec != nil {
 		sessionManager = spec.SessionManager
+		// special case for basic auth with custom hashing algo
+		if spec.UseBasicAuth && strings.HasPrefix(sessionKey, spec.OrgID) {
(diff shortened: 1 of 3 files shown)


buger:
I get the idea, but I'm afraid this will not work. When dashboard sends the request to gateway API, it does not set the right API id.

dencoded:
it worked for me in dashboard - tried to find two different BA-keys (created for different BA apis).

the problem here that our gateway logic uses API-id just to get session storage, then it can find any key as they are prefixed with `"apikey-"` (no API id in prefix). I think dashboard picks latest added API ID to do call and relays on that weird legacy behaviour.

So nothing really changed except it works with hashing algo. There is one problem though - if passed API ID is not BA - it is not gonna work. This condition https://github.com/TykTechnologies/tyk/pull/2018/files#diff-651a84f8ad6a38c1f7ccdf63f2692410R328 makes sense only dashboard is sending right API ID, maybe we should remove `spec.UseBasicAuth` part from there.

The question is how would dashboard know the right API ID? Looks like asking for right API ID int this case is something extra as gateway's API already support key retrieval without API ID - in endpoint `GET /tyk/keys/{keyName}?api_id={apiID}` query string param can be omitted and then our logic falls back to `FallbackKeySesionManager ` https://github.com/TykTechnologies/tyk/pull/2018/files#diff-651a84f8ad6a38c1f7ccdf63f2692410R323

buger:
That's why I was thinking about extending Tyk Keys API, and for example, pass query attribute "?username=true" to the endpoint, and if it set, it will run logic similar to what you do in PR.

Dashboard can't know right APIID in advance.

dencoded:
makes sense to me. just to mention  - we also need OrgID to generate right token and get key from storage. I guess the assumption with this approach would be that dashboard would be still passing key as `OrgID + username` and `?username=true`, then we can extract OrgID from that value (it has fixed length) and generate right storage key

buger:
@dencoded makes sense to me

dencoded:
@buger I've added boolean parameter `username`  to endpoints:
- `GET /tyk/keys/{keyName}`
- `PUT /tyk/keys/{keyName}`
- `DELETE /tyk/keys/{keyName}`

`username=true` triggers to form correct key token with respect to current hash algo

`POST /tyk/keys` stays the same as we have everything to generate right key in request payload

api.go
```diff
@@ -656,6 +656,26 @@ func keyHandler(w http.ResponseWriter, r *http.Request) {
 	keyName := mux.Vars(r)["keyName"]
 	apiID := r.URL.Query().Get("api_id")
 	isHashed := r.URL.Query().Get("hashed") != ""
+	isUserName := r.URL.Query().Get("username") == "true"
+
+	// check if passed key is user name and convert it to real key with respect to current hashing algorithm
+	if r.Method != http.MethodPost && isUserName {
```
buger: I do understand why it was done like this, however, I feel bad while looking at it.

I propose to make it simple, and in such case, just require passing one more attribute "org_id".

dencoded: do you mean more attribute "org_id" in addition to "username=true" ?

buger:
Exactly, instead orgid extraction logic you implemented. Also note that gateway api is an admin one, and we trust user input, so check if orgid valid also redundant.

Issue #1834, opened 2018-07-20
Mapping JWT scopes to policies


buger:
**Do you want to request a *feature* or report a *bug*?**
feature
**What is the current behavior?**
At the moment Tyk ignores `scopes` attached to oAuth/JWT tokens, and it just forces you to use the specified fixed policy id for all keys.  Moreover, at the moment access_token data not used at all.
**What is the expected behavior?**
It should be possible to specify "scope" -> "policy" mapping in API definition, and if done so, Tyk should call oAuth [token introspection endpoint](https://www.oauth.com/oauth2-servers/token-introspection-endpoint/) to get scopes info about the token scopes, and use policy merging technique described in this task https://github.com/TykTechnologies/tyk/issues/1783#issuecomment-406580854

So you can define smth like: "users:read" -> policy1, "users:write" -> policy2, and etc.

Note that it should be possible to dynamically revoke user scopes from IDP side, and tyk should see this changes. Doing it in secure and performant way, described in oAuth.com link above:

> One way to mitigate this problem is for consumers to never cache the value beyond the expiration time of the token, which would have been returned in the “exp” parameter of the introspection response.

letzya:
Maybe we can hold off on integrating with `/introspect` since this endpoint is important to verify if an access_token has been revoked or not when we want to be extra careful and not just trust the `exp` claim. It is also useful to get scopes when the access_token is opaque but for now, many IdPs support JWT access_token so it might be enough for releasing this feature and the extra security of `/introspect` can be achieved with a plugin.

ilijabojanovic:
We have several issues related to this feature, i will link new issues to this task

PR #2019, opened 2018-12-07
Move OpenID JWT scope settings to root level

buger:
The main intent that it probably will be too complex, especially from UI point of view. Lets start with defining same mapping for all clients. As a bonus, all fields shared with JWT middleware.

Also, JWT middleware scope name made configurable using `JWTScopeClaimName`

Plus empty map check now use: `len` instead of nil, to support empty maps.

Updates https://github.com/TykTechnologies/tyk/pull/1943 and https://github.com/TykTechnologies/tyk/issues/1834

### DIFF
--- a/mw_openid.go
+++ b/mw_openid.go
@@ -120,7 +120,7 @@ func (k *OpenIDMW) ProcessRequest(w http.ResponseWriter, r *http.Request, _ inte
 		return errors.New("Key not authorised"), http.StatusUnauthorized
 	}
 
-	providerConf, ok := k.providerConfigs[iss.(string)]
+	_, ok := k.providerConfigs[iss.(string)]
 	if !ok {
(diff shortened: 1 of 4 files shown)


buger:
@dencoded pls check

mw_openid.go
```diff
@@ -120,15 +120,15 @@ func (k *OpenIDMW) ProcessRequest(w http.ResponseWriter, r *http.Request, _ inte
 		return errors.New("Key not authorised"), http.StatusUnauthorized
 	}
 
-	providerConf, ok := k.providerConfigs[iss.(string)]
+	_, ok := k.providerConfigs[iss.(string)]
```
dencoded: I think this one is not needed

buger: True. Just made the change.

PR #2023, opened 2018-12-13
 Fix Schema to make it backward compatible

### LINKED ISSUE #2024: API definition schema is not backward compatible
bug
After the moving API definition schema to Tyk repo https://github.com/TykTechnologies/tyk-analytics/pull/979, creating API with null values for complex fields, like arrays or objects, stopped working. Now it throws that "null" value is permitted, which broken compatibility with previous versions.
It should allow null values for all the fields except required ones.
Try creating API using Dashboard API, where API definition does not specify all possible fields, for example, https://github.com/TykTechnologies/tyk-develop-env/blob/master/confs/api/keyless.json
master

furkansenharputlu:
This PR fixes the API definition Schema to make it backward compatible. An API definition adding request will work when `null` values exist in a definition.
Fixes #2024

### DIFF
--- a/apidef/schema.go
+++ b/apidef/schema.go
@@ -1,7 +1,7 @@
 package apidef
 
 const Schema = `{
-    "type": "object",
+    "type": ["object", "null"],
     "$schema": "http://json-schema.org/draft-04/schema",
     "id": "http://jsonschema.net",
     "additionalProperties": false,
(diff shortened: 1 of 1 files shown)


buger:
Makes sense!

buger:
Pls re-vendor it to tyk-analytics

PR #2027, opened 2018-12-18
Update Go bindings to use "context"

matiasinsaurralde:
Previous bindings used `golang.org/x/net/context`.


matiasinsaurralde:
@buger Fixed 👍

Issue #2030, opened 2018-12-19
Dynamically specify TLS certificates for APIs


buger:
**Do you want to request a *feature* or report a *bug*?**
feature
**What is the current behavior?**
If your API has a custom domain, it will require you to modify tyk.conf to include TLS certificate details and reload the process. 
**What is the expected behavior?**
It will be really cool if we can dynamically specify certificate ID (from our cert storage) inside API definition, and it will be automatically recognized by Gateway
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
master

buger:
https://github.com/TykTechnologies/tyk/pull/2031 was merged to `release-2.8` branch
<details><summary></summary>Created via API</details>

buger:
https://github.com/TykTechnologies/tyk/pull/2031 was merged to `release-2.7` branch
<details><summary></summary>Created via API</details>

PR #2031, opened 2018-12-19
Add support for dynamic API certificates

buger:
Added new API definition field: `certificates`, which is a string array.

You just need to specify a list of Certificate IDs, and Tyk will dynamically load this certificates, without process restart. 

Fix https://github.com/TykTechnologies/tyk/issues/2030

### DIFF
--- a/cert.go
+++ b/cert.go
@@ -257,6 +257,25 @@ func getTLSConfigForClient(baseConfig *tls.Config, listenPort int) func(hello *t
 		}
 
 		apisMu.RLock()
+
+		// Dynamically add API specific certificates
+		for _, spec := range apiSpecs {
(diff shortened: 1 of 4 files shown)


cert.go
```diff
@@ -257,6 +257,25 @@ func getTLSConfigForClient(baseConfig *tls.Config, listenPort int) func(hello *t
 		}
 
 		apisMu.RLock()
+
```
letzya: Since this code handles client certificates for mutual tls and server certificates - Can we add comments for it. Something along these lines:
261: // Dynamically add API specific certificates for the server authentication
278: // Add certificate for Client authentication for the relevant domain
250: // Add certificate for Client authentication for ControlAPI

PR #2040, opened 2018-12-25
[WIP]feature/issue-495-Cache_DNS

evityuk:
Issue -  #495 
Added a global dns cache in order to reduce the number of request to gateway's local dns server and appropriate gateway config section
```
...
"dns_cache": {
    "enabled": true | false, //Turned off by default
    "ttl": [-1;max(int64)], //Time in seconds before the record will be removed from cache
    "multiple_ips_handle_strategy": "pick_first" | "random" | "no_cache" //A strategy, which will be used when dns query will reply with more than 1 ip address per single host.
},
...
```
This feature is turned off by default, in order to turn it on set 
dns.enable_cache in tyk.conf to `true` and set appropriate `ttl` & `multiple_ips_handle_strategy` values.

What was changed?
 - Change reverse_proxy.go httpTransport in order to lookup dns records from cache before resolving the dns record ips - https://github.com/TykTechnologies/tyk/blob/61476838d04be19bd8b7df85aafbe0f89983e7ef/reverse_proxy.go#L307
 - Add configuration with schema
 - Add a DnsCacheStorage wrap around(github.com/pmylund/go-cache) - https://github.com/TykTechnologies/tyk/blob/feature/issue-495-Cache_DNS/dns_cache/storage.go#L19
 - Add unit tests for reverse_proxy, config linter, dns cache storage structs
 - Make dns mock server reusable between tests and update it to dynamically extend existing configurations

How to debug?
 - Set logging to at least `debug` level and with enabled caching look for log entries with `dns-cache` prefix.

Docs:
==================
https://tyk.io/docs/configure/tyk-gateway-configuration-options/
dns_cache
This section enables the global configuration of the expireable dns records caching for gateway API endpoints. By design caching affects only http(s), ws(s) protocols apis and doesn't affect any plugin/middleware dns queries.

dns_cache.enabled
Valid values: true, false
Default value: false
Setting this value to `true` will enable caching of dns queries responses used for API endpoint's host names.
By default caching is disabled.

dns_cache.ttl
Valid range: [-1;18446744073709551615] 
Units: seconds
Default value: `0`
This setting allows to specify duration in seconds before the record will be removed from cache after being added to it on first dns query resolution of API endpoints. 
Setting ttl to `-1` prevents record from being expired and removed from cache on next check interval. 

dns_cache.multiple_ips_handle_strategy
Valid values:  "pick_first" | "random" | "no_cache"
Default value: `no_cache`
A strategy, which will be used when dns query will reply with more than 1 ip address per single host. 
As dns query response ip addresses can have changing order depending on dns server balancing strategy(eg: round robin, geographically dependent origin-ip ordering, etc) this option allows to not to limit connection to first host within cached response list or prevent response caching.
By default `no_cache` strategy is enabled. Setting this value to `pick_first` will instruct gateway to connect to *first* ip in returned ips list and **cache** the response. 
Setting this value to `random` will instruct gateway to connect to *random* ip in returned ips list and **cache** the response. 
Setting this value to `no_cache` will instruct gateway to connect to *first* ip in returned ips list and fetch each addresses list without caching on each API endpoint dns query. 

https://tyk.io/docs/configure/gateway-env-variables/

| JSON Value               | Environment Variable Name      |
|--------------------------|--------------------------------|
| dns_cache.enabled        | TYK_GW_DNSCACHE_ENABLED       |
| dns_cache.ttl            | TYK_GW_DNSCACHE_TTL           |
| dns_cache.multiple_ips_handle_strategy            | TYK_GW_DNSCACHE_MULTIPLEIPSHANDLESTRATEGY           |

==================

Team dev docs:

| dns dependent entity  | use caching? |
|-----------------------|--------------|
| http(s) api endpoint  | T            |
| ws(s) api endpoint    | T            |
| grpc plugin           | F            |
| python plugin         | F            |
| LuaJIT plugin         | F            |
| javascript middleware | F            |


config/config.go
```diff
@@ -334,6 +341,7 @@ var Default = Config{
 	AnalyticsConfig: AnalyticsConfigConfig{
 		IgnoredIPs: make([]string, 0),
 	},
+	Dns:DnsConfig{EnableCaching:false, TTL: 3600 * 1000, CheckInterval: 3600 * 1000},
```
joshblakeley: This line isnt neccessary if it's off by default

evityuk: 29984086768d9d417ec85294c7d9ec0bfcaeed24
fixed

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
+package dns_cache
```
dencoded: this package naming encourages to use snake case in the client code and considered to be not the best, more info here: https://golang.org/doc/effective_go.html#package-names

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+	InitDNSCaching(ttl, checkInterval time.Duration)
+	WrapDialer(dialer *net.Dialer) dialContextFunc
+}
+
+type DnsCacheManager struct {
+	DnsCache *DnsCacheStorage
+}
+
+func NewDnsCacheManager() *DnsCacheManager {
+	return &DnsCacheManager{nil}
```
dencoded: not clear why we do not prepare here usable `DnsCacheManager` instantiating `DnsCache` field instead of nil, that I guess purpose of these constructor-like functions

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+
+type dialContextFunc func(ctx context.Context, network, address string) (net.Conn, error)
+
+type IDnsCacheManager interface {
+	InitDNSCaching(ttl, checkInterval time.Duration)
+	WrapDialer(dialer *net.Dialer) dialContextFunc
+}
+
+type DnsCacheManager struct {
+	DnsCache *DnsCacheStorage
```
dencoded: might need to move it to private field as looks like it is pure implementation

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+func (m *DnsCacheManager) doCachedDial(d *net.Dialer, ctx context.Context, network, address string) (net.Conn, error) {
+	if m.DnsCache == nil {
+		return d.DialContext(ctx, network, address)
+	}
+
+	separator := strings.LastIndex(address, ":")
+	ips, err := m.DnsCache.FetchItem(address[:separator])
+
+	if err != nil {
+		logger.Infof("doCachedDial error: %v. network=%v, address=%v", err.Error(), network, address)
```
dencoded: I think we miss here `return nil, err` as looks like that would be actual DNS-error (if any)

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+		return m.doCachedDial(dialer, ctx, network, address)
+	}
+}
+
+func (m *DnsCacheManager) doCachedDial(d *net.Dialer, ctx context.Context, network, address string) (net.Conn, error) {
+	if m.DnsCache == nil {
+		return d.DialContext(ctx, network, address)
+	}
+
+	separator := strings.LastIndex(address, ":")
```
dencoded: just thinking about readability - we can split address with using `net.SplitHostPort` and use host and port vars for logic below

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (m *DnsCacheManager) WrapDialer(dialer *net.Dialer) dialContextFunc {
+	return func(ctx context.Context, network, address string) (net.Conn, error) {
+		return m.doCachedDial(dialer, ctx, network, address)
+	}
+}
+
+func (m *DnsCacheManager) doCachedDial(d *net.Dialer, ctx context.Context, network, address string) (net.Conn, error) {
+	if m.DnsCache == nil {
```
dencoded: might worth to add one more short path to check if address already contains IP address (in some cases upstream might be specified by IP for host name) and return right away `return d.DialContext(ctx, network, address)`

This would make it slightly faster and we won't cache items like key,value->IP,IP

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	separator := strings.LastIndex(address, ":")
+	ips, err := m.DnsCache.FetchItem(address[:separator])
+
+	if err != nil {
+		logger.Infof("doCachedDial error: %v. network=%v, address=%v", err.Error(), network, address)
+	}
+
+	return d.DialContext(ctx, network, ips[0]+address[separator:])
```
dencoded: if we always use `ips[0]` why to store list of IPs in cache? (instead of 1st value)

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+}
+
+func (m *DnsCacheManager) doCachedDial(d *net.Dialer, ctx context.Context, network, address string) (net.Conn, error) {
+	if m.DnsCache == nil {
+		return d.DialContext(ctx, network, address)
+	}
+
+	separator := strings.LastIndex(address, ":")
+	ips, err := m.DnsCache.FetchItem(address[:separator])
```
dencoded: seems like we break here idea behind DNS load balancing (multiple IPs). if we reach this function - when new connection is about to do dial-call - that means we need to hit actual DNS because it will give us IP list with different order so connection will be load-balanced to another IP behind that domain name. Probably the right solution would be - don't cache DNS calls if resolver returns list of IPs per domain name

dns_cache/storage.go
```diff
@@ -0,0 +1,81 @@
...
+		return DnsCacheItem{}, false
+	}
+	return item.(DnsCacheItem), found
+}
+
+func (dc *DnsCacheStorage) FetchItem(key string) ([]string, error) {
+	item, ok := dc.cache.Get(key)
+	if ok {
+		result, _ := item.(DnsCacheItem)
+		logger.Debugf("Dns record was populated from cache: key=%q, addrs=%q", key, result.addrs)
```
dencoded: following logrus-convention it might better to do (so debug message context will appear as message fields):
```
logger.WithFields(logrus.Fields{
  "key": key,
  "addrs": result.addrs,
}.Debug("Dns record was populated from cache")
```

dns_cache/storage.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"net"
+	"reflect"
+	"time"
+
+	cache "github.com/pmylund/go-cache"
+)
+
+type DnsCacheItem struct {
```
dencoded: a little confusing the purpose of this struct, would be equal to use just `type DnsCacheItem []string` or store in cache just `string` (see my comment about always using 1st element from addrs list above)

dns_cache/storage.go
```diff
@@ -0,0 +1,81 @@
...
+	addrs []string
+}
+
+func (item *DnsCacheItem) IsEqualsTo(addrs []string) bool {
```
dencoded: generally we don't add to main code things like this - code pieces needed to run unit-tests

dns_cache/storage.go
```diff
@@ -0,0 +1,81 @@
...
+func (dc *DnsCacheStorage) Set(key string, addrs []string) {
+	logger.Debugf("Adding dns record to cache: key=%q, addrs=%q", key, addrs)
+	dc.cache.Set(key, DnsCacheItem{addrs}, cache.DefaultExpiration)
+}
+
+func (dc *DnsCacheStorage) Clear() {
+	dc.cache.Flush()
+}
+
+func (dc *DnsCacheStorage) resolveDNSRecord(host string) ([]string, error) {
```
dencoded: what is purpose of this function? it looks like it is just alias for `net.LookupHost`

dns_cache/storage.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func NewDnsCacheStorage(expiration, checkInterval time.Duration) *DnsCacheStorage {
+	storage := &DnsCacheStorage{cache.New(expiration, checkInterval)}
+	return storage
+}
+
+//Return map of non expired dns cache items
+func (dc *DnsCacheStorage) Items(includeExpired bool) map[string]DnsCacheItem {
+	var nonExpiredItems map[string]cache.Item = dc.cache.Items()
```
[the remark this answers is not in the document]
dencoded: also, `map[string]cache.Item` type can be omitted as we assign value in var declaration

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/TykTechnologies/tyk/log"
+)
+
+var (
+	logger = log.Get().WithField("prefix", "dns-cache")
+)
+
+type dialContextFunc func(ctx context.Context, network, address string) (net.Conn, error)
+
+type IDnsCacheManager interface {
```
dencoded: comment is missing explaining purpose of interface

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+	InitDNSCaching(ttl, checkInterval time.Duration)
+	WrapDialer(dialer *net.Dialer) dialContextFunc
+}
+
+type DnsCacheManager struct {
```
dencoded: comment is missing explaining purpose of type

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+	DnsCache *DnsCacheStorage
+}
+
+func NewDnsCacheManager() *DnsCacheManager {
```
dencoded: comment is missing explaining purpose of func

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+	return &DnsCacheManager{nil}
+}
+
+func (m *DnsCacheManager) WrapDialer(dialer *net.Dialer) dialContextFunc {
```
dencoded: comment is missing explaining purpose of func

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+	ips, err := m.DnsCache.FetchItem(address[:separator])
+
+	if err != nil {
+		logger.Infof("doCachedDial error: %v. network=%v, address=%v", err.Error(), network, address)
+	}
+
+	return d.DialContext(ctx, network, ips[0]+address[separator:])
+}
+
+func (m *DnsCacheManager) InitDNSCaching(ttl, checkInterval time.Duration) {
```
dencoded: comment is missing explaining purpose of func

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+	if m.DnsCache == nil {
+		logger.Infof("Initialized dns cache with ttl=%s, duration=%s", ttl, checkInterval)
+		m.DnsCache = NewDnsCacheStorage(ttl, checkInterval)
+	}
+}
+
+func (m *DnsCacheManager) DisposeCache() {
```
dencoded: comment is missing explaining purpose of func

dns_cache/manager.go
```diff
@@ -0,0 +1,62 @@
...
+	m.DnsCache.Clear()
+	m.DnsCache = nil
```
dencoded: this would also stop DNS caching, just wanted to clarify is this function to clear cache or clear and stop caching? maybe just lack of Godoc coments

dencoded:
also, as a general note - could you squash all that long list of commits?

api_loader.go
```diff
@@ -35,16 +35,16 @@ type ChainObject struct {
 	Subrouter      *mux.Router
 }
 
-func prepareStorage() (storage.RedisCluster, storage.RedisCluster, storage.RedisCluster, *RPCStorageHandler, *RPCStorageHandler) {
-	redisStore := storage.RedisCluster{KeyPrefix: "apikey-", HashKeys: config.Global().HashKeys}
-	redisOrgStore := storage.RedisCluster{KeyPrefix: "orgkey."}
-	healthStore := storage.RedisCluster{KeyPrefix: "apihealth."}
-	rpcAuthStore := RPCStorageHandler{KeyPrefix: "apikey-", HashKeys: config.Global().HashKeys}
-	rpcOrgStore := RPCStorageHandler{KeyPrefix: "orgkey."}
+func prepareStorage() (*storage.RedisCluster, *storage.RedisCluster, *storage.RedisCluster, *RPCStorageHandler, *RPCStorageHandler) {
```
dencoded: need explanation - why pointers are eliminated here?

config/config.go
```diff
@@ -211,6 +217,7 @@ type Config struct {
 	EnableAnalytics                   bool                                  `json:"enable_analytics"`
 	AnalyticsConfig                   AnalyticsConfigConfig                 `json:"analytics_config"`
 	HealthCheck                       HealthCheckConfig                     `json:"health_check"`
+	Dns                               DnsConfig                             `json:"dns"`
```
dencoded: might worth to rename config key to `dns_cache` as dns is more general term

storage/redis_cluster.go
```diff
@@ -209,11 +220,15 @@ func (r RedisCluster) ensureConnection() {
...
 
 // GetKey will retrieve a key from the database
-func (r RedisCluster) GetKey(keyName string) (string, error) {
+func (r *RedisCluster) GetKey(keyName string) (string, error) {
 	r.ensureConnection()
 	log.Debug("[STORE] Getting WAS: ", keyName)
 	log.Debug("[STORE] Getting: ", r.fixKey(keyName))
+	currentMu := r.getSingletonMU()
+	currentMu.RLock() //Lock rediscluster.RedisCluster.SingleRedisMode from writes
+	defer currentMu.RUnlock()
```
dencoded: as a general note - locking mutex, even read-locking with `RLock`, do some network I/O which might take ages and then unlock mutex - I feel like it is not right, this is not what mutexes for

storage/redis_cluster.go
```diff
@@ -209,11 +220,15 @@ func (r RedisCluster) ensureConnection() {
...
 	value, err := redis.String(r.singleton().Do("GET", r.fixKey(keyName)))
+
```
dencoded: this looks confusing - how about all other methods where we do `r.singleton().Do`, there is no any mutex there

storage/redis_cluster.go
```diff
@@ -490,18 +505,23 @@ func (r RedisCluster) DeleteKeys(keys []string) bool {
 
 // StartPubSubHandler will listen for a signal and run the callback for
 // every subscription and message event.
-func (r RedisCluster) StartPubSubHandler(channel string, callback func(interface{})) error {
-	if r.singleton() == nil {
+func (r *RedisCluster) StartPubSubHandler(channel string, callback func(interface{})) error {
```
dencoded: if I understand right we have only one call to this function - when tyk process gets started. There would be only one `currentMu.Lock()` ever, not clear why do we need it at this place. I'd look into other approach tbh

evityuk:
> also, as a general note - could you squash all that long list of commits?
It can be merged by github `PR` merge&stash functionality, which merges all commits into single one. 
Should I merge it manually?

dnscache/manager.go
```diff
@@ -0,0 +1,118 @@
...
+		return nil, err
+	}
+
+	if ip := net.ParseIP(host); ip != nil {
+		return safeDial(address, "")
+	}
+
+	ips, err := m.cacheStorage.FetchItem(host)
+	if err != nil {
+		logger.WithFields(logrus.Fields{
```
dencoded: better to use `logger.WithError(err).WithFields.(..`

dnscache/storage.go
```diff
@@ -0,0 +1,94 @@
...
+	"reflect"
+	"time"
+
+	"fmt"
+
+	"github.com/pmylund/go-cache"
+)
+
+// DnsCacheItem represents single record in cache
+type DnsCacheItem struct {
```
dencoded: a little confusing the purpose of this struct, would be equal to use just `type DnsCacheItem []string` or store in cache just `string`

dnscache/storage.go
```diff
@@ -0,0 +1,94 @@
...
+	addrs []string
+}
+
+func (item *DnsCacheItem) IsEqualsTo(addrs []string) bool {
```
dencoded: don't add to main code things like this - code pieces needed to run unit-tests

dnscache/storage.go
```diff
@@ -0,0 +1,94 @@
...
+
+// FetchItem returns list of ips from cache or resolves them and add to cache
+func (dc *DnsCacheStorage) FetchItem(hostName string) ([]string, error) {
+	if hostName == "" {
+		return nil, fmt.Errorf("hostName can't be empty. hostName=%v", hostName)
+	}
+
+	item, ok := dc.Get(hostName)
+	if ok {
+		logger.Debugf("Dns record was populated from cache: hostName=%q, addrs=%q", hostName, item.addrs)
```
dencoded: following logrus-convention it might better to do (so debug message context will appear as message fields):

```
logger.WithFields(logrus.Fields{
  "hostName": hostName,
  "addrs": item.addrs,
}.Debug("Dns record was populated from cache")
```

dnscache/storage.go
```diff
@@ -0,0 +1,94 @@
...
+	logger.Debugf("Adding dns record to cache: key=%q, addrs=%q", key, addrs)
+	dc.cache.Set(key, DnsCacheItem{addrs}, cache.DefaultExpiration)
+}
+
+// Clear deletes all records from cache
+func (dc *DnsCacheStorage) Clear() {
+	dc.cache.Flush()
+}
+
+func (dc *DnsCacheStorage) resolveDNSRecord(host string) ([]string, error) {
```
dencoded: what is purpose of this function? it looks like it is just alias for `net.LookupHost`

config/config.go
```diff
@@ -334,6 +341,11 @@ var Default = Config{
 	AnalyticsConfig: AnalyticsConfigConfig{
 		IgnoredIPs: make([]string, 0),
 	},
+	DnsCache: DnsCacheConfig{
+		Enabled:       false,
+		TTL:           3600,
```
dencoded: better to move 3600 and 60 to constants

reverse_proxy.go
```diff
@@ -302,18 +302,23 @@ type ReverseProxy struct {
...
+	timeout := 30
+	if dialerTimeout > 0 {
+		log.Debug("Setting timeout for outbound request to: ", dialerTimeout)
+		timeout = dialerTimeout
+	}
+
 	return &http.Transport{
-		DialContext: (&net.Dialer{
-			Timeout:   30 * time.Second,
+		DialContext: dnsCacheManager.WrapDialer(&net.Dialer{
```
dencoded: seems like in situation when we know for sure that DNS cache is disabled (in config) we don't want to wrap Dialer as it introduces extra closure and some logic inside dnsCacheManager to be run

dnscache/manager.go
```diff
@@ -0,0 +1,118 @@
...
+			"network": network,
+			"address": address,
+		}).Errorf("doCachedDial cachedStorage.FetchItem error: %v. ips=%v", err.Error(), ips)
+
+		return safeDial(ips[0]+":"+port, "")
```
dencoded: looks like `FetchItem` returns error when `net.LookupHost` failed, also `ips` here will be nil.
I think we have valid reason to do `return nil, err`

dnscache/storage.go
dencoded: actually just `string` might not work in case we want to know that this domain has several IPs and build some logic on top of it, i.e. always do real DNS request foor those hostnames in cache

config/config.go
```diff
@@ -81,6 +81,12 @@ type HealthCheckConfig struct {
 	HealthCheckValueTimeout int64 `json:"health_check_value_timeouts"`
 }
 
+type DnsCacheConfig struct {
+	Enabled       bool  `json:"enabled"`
+	TTL           int64 `json:"ttl"`
+	CheckInterval int64 `json:"check_interval"`
```
dencoded: Looks like this value regulates how often in-app cache will be wiping out expired items (for memory usage purposes). Why do we need to care about it on tyk.conf level?

test/dns.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	if domainsErrorMap != nil {
+		dnsMux.domainsToErrors = domainsErrorMap
+	}
+
+	mockServer.Handler = dnsMux
+
+	go func() {
+		err := mockServer.ActivateAndServe()
```
dencoded: might simplify `startResultChannel <- mockServer.ActivateAndServe()`

test/dns.go
```diff
@@ -0,0 +1,198 @@
...
+		if err != nil {
+			startResultChannel <- err
+			return
+		}
+		startResultChannel <- nil
+	}()
+	select {
```
dencoded: might simplify here without `select`: `err := <-startResultChannel`

dencoded:
@evityuk I've left some new comments about logic which I believe need to be addressed and a couple of code styling

dnscache/manager.go
```diff
@@ -0,0 +1,118 @@
...
+	host, port, err := net.SplitHostPort(address)
+	if err != nil {
+		return nil, err
+	}
+
+	if ip := net.ParseIP(host); ip != nil {
+		return safeDial(address, "")
+	}
+
+	ips, err := m.cacheStorage.FetchItem(host)
```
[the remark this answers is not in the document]
evityuk: Why? 
We'll break the balancer logic in that case. 

In our case we'll always get a cached list and the connection will be established with the 1st list item/ip until cache for that record will expire and next dns server query return **probably** list with newly ordered elements.

config/config.go
evityuk: fixed in 7f332e6a861e73fc8eb846e0e9aadd5f96fc724e

dnscache/manager.go
evityuk: fixed in 7f332e6a861e73fc8eb846e0e9aadd5f96fc724e

dnscache/manager.go
evityuk: @dencoded How about to allow to fail current dialer DialContext and simply to delegate the dialing to d.DialContext, which can eventually be successful. 
Add it in in 7f332e6a861e73fc8eb846e0e9aadd5f96fc724e

dnscache/storage.go
evityuk: fixed in 7f332e6a861e73fc8eb846e0e9aadd5f96fc724e

dnscache/storage.go
evityuk: >it looks like it is just alias for `net.LookupHost`
Yes, exactly. But in general it isn't and it's designed so. 
Why? 
Because of good-design principles(intention to separate business logic to separate logical methods) and because this method do some io(eg: connect to dns server and exchange messages with it), which should be mocked within unit tests. 
And due to the fact that currently unit tests aren't written well(by written well I mean that each "method-under-the-test" mocks it's dependencies using dependecyInjection pattern or mocking library and doesn't have a separate dns server for that mock) and someday they will, I left this method for this purpose(with respect to KISS principle).
Does that answer your question in case if your concern is design-redundancy?

test/dns.go
evityuk: fixed in [7f332e6](https://github.com/TykTechnologies/tyk/commit/7f332e6a861e73fc8eb846e0e9aadd5f96fc724e)

test/dns.go
evityuk: fixed in [7f332e6](https://github.com/TykTechnologies/tyk/commit/7f332e6a861e73fc8eb846e0e9aadd5f96fc724e)

config/config.go
dencoded: yes, I get that `CheckInterval` is added here in config because hard-coding doesn't allow it to be changed.
my question was why we want to have it configured. if you look up code by `cache.New` you will see that we use in-app caching a lot and none of those have cleanup interval as config value.
Customers will be asking why do they need to configure this, I don't see any reasonable answer tbh because it is something which controls in-app memory management and doesn't influence items expiration. That might introduce some noise in support.

dnscache/manager.go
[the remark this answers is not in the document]
dencoded: `in our case we'll always get a cached list and the connection will be established with the 1st list item/ip until cache for that record will expire` - that is my concern. With this implementation - no matter how DNS reply shuffles IPs we will be always bombing the last know/cached IP until it expires in cache. Why `**probably** list with reordered elements`. I think it is not probably, it is more likely as several A-records in DNS reply is the implementation of round robin on DNS level

dnscache/storage.go
dencoded: `net.LookupHost` direct call looks OK to me, it does what it says in name and nothing else. as to `resolveDNSRecord` - my 1st though was that it might do some logic inside connected with different types of DNS records (as it suggest in name) but it doesn't do anything specific to DNS-record. I would be ok to leave it here as is  but this is just extra function call with extra memory allocation. Thus code will likely be changed to direct call to `net.LookupHost` with next performance optimization.

reverse_proxy.go
dencoded: I wasn't able to find those cases in code (when cache will be re-enabled). Looks like we enable DNS cache at gateway process start up time and it comes from tyk.conf which gets read again only when we start (or restart) process.

We've spent a while recently on performance optimization. This wrapper is adding extra ops and allocations every time we connect/reconnect to upstream and I think we can easily eliminate it and use standard `net.Dialer` when DNS cache is disabled.

dnscache/manager.go
dencoded: that would be doubled call to lookup hostname:
1st inside `FetchItem` and failed (because `FetchItem` returned error)
2nd inside `safeDial` because standard `DialContext` will try to resolve the same address `FetchItem` was trying and failed

dencoded:
@evityuk replied to comments, might need to have inputs from other reviewers

dnscache/manager.go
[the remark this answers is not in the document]
evityuk: So, the suggestion is: 
a) to use cached ip only if there' exist 1 item for CNAME record and for `>1' case - fallback to native dial using `net.Dialer#DialContext(...address)`, which will [dialSerial](https://golang.org/src/net/dial.go#L406)
b)add a tyk's round robin on cached ips if we want to archieve a balancing on cached records(and therefor not to hit 1st only from list)
c) none of the above?

dnscache/manager.go
evityuk: Yes, 2 calls are by design in order to provide additional `HA` of dns lookup as here we aren't limited by any timeouts(anyway if we're than that will be a rare case)

dencoded: if you think about HA - that might not be the best place to do it, here we have some high level logic. Why not to do N retires in `resolveDNSRecord` say with some jitter then? (it could be a good reason to do it as a function)

dnscache/manager.go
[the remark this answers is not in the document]
dencoded: not sure I follow here, what do you mean by `CNAME`? when we resolve hostname to addr we receive back IP, we don't know was it CNAME or A

config/config.go
evityuk: There's a doc for this value, which we can improve or add more information in order to explain the purpose of this parameter:
```
dns_cache.check_interval
Valid range: [-1;18446744073709551615]
Units: seconds
Default value: `0`
This setting allows to specify interval in seconds of expired records purging.
Setting this value to `-1` prevents checks from being executed and therefore dns cache records from being removed.
```

But in general, I don't have a clear vision regarding your remarks?As managing the inter

evityuk: There's a doc for this value, which we can improve or add more information in order to explain the purpose of this parameter:

```
dns_cache.check_interval
Valid range: [-1;18446744073709551615]
Units: seconds
Default value: `0`
This setting allows to specify interval in seconds of expired records purging.
Setting this value to `-1` prevents checks from being executed and therefore dns cache records from being removed.
```

But as exposing this value to configuration can be harmful, I've removed it from there and from docs. 
related commit - ed314a2894a5617e72fdd80971a45565dec91ec4

reverse_proxy.go
evityuk: So are you suggesting to remove this extension point(by this I mean the ability to detect cache on/off in runtime) in order to make the code performant?

dnscache/manager.go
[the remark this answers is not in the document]
evityuk: But this `PR` is created to fix caching for *some* servers, which doesn't support keep-alive or *do support it*(@buger FYI) and close/decide to close it when *tyk is initiating such connection* or *after it sends* response to tyk. 

Here's the original issue for details: 
`In the case when upstream does not support keep-alive connections, DNS query will be made on each HTTP request, which makes the huge load on DNS server.`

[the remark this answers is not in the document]
evityuk: >not sure I follow here, what do you mean by `CNAME`? when we resolve hostname to addr we receive back IP, we don't know was it CNAME or A
Sorry, by this 
>to use cached ip only if there' exist 1 item for CNAME record

I meant "to use cached ip only if there exist 1 ip after recursive dns resolution chain finishes".

dnscache/storage.go
buger: Agree. Main code should not contain helpers used for tests. Needs be moved to test suite.

dnscache/manager.go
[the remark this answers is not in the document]
buger: Using all IPs from the record is the safest approach. I vote to just fetch random IP record instead of the first one, which is kinda simplified round robin.

reverse_proxy.go
buger: I do not care that much about performance here. What I do care about, is that when we wrap dialer like this, we made all our users use this new code (which for sure has bug somewhere, like with all new code). So if we wrap dialer only when DNS turned on, this potential bugs will happen only for users who enabled DNS caching.

dnscache/storage.go
buger: Let's keep it like this for now.

dnscache/manager.go
```diff
@@ -96,6 +105,21 @@ func (m *DnsCacheManager) doCachedDial(d *net.Dialer, ctx context.Context, netwo
 		return safeDial(address, "")
 	}
 
+	if m.strategy == config.NoCacheStrategy {
+		if len(ips) > 1 {
+			m.cacheStorage.Delete(host)
+			return safeDial(ips[0]+":"+port, "")
```
dencoded: what is the difference between `random` and `pick_first` then?

buger: You probably meant what is the difference between "pick first" and "no_cache". 
In think idea here, that since `m.cacheStorage.Delete(host)` is run, it forces "cache" wrapper to run DNS every time. Additionally "pick first" is a default Go DNS strategy too. In other words, we first cache and then disregard cache immediately.

dnscache/manager.go
```diff
@@ -96,6 +105,21 @@ func (m *DnsCacheManager) doCachedDial(d *net.Dialer, ctx context.Context, netwo
...
+		}
+	}
+
+	if m.strategy == config.RandomStrategy {
+		if len(ips) > 1 {
+			rand.Seed(int64(len(ips)))
```
dencoded: seeding it every time with `len(ips)` would make `ips[rand.Intn(len(ips))]` to return the same IP every time.
also, `rand.Seed` this is something you want to do only once as it is global seed on `rand` package level.

dnscache/manager.go
dencoded: ah right, I meant difference between "pick_first" and "no_cache".

I see here that our cache logic is still influences the choice of which IP to pick. Go DNS strategy depends on env var `GODEBUG=netdns=`. There probably won't be any Go resolver in place.

buger: True, but the alternative, in this case, is do a second DNS call, and run `safeDial(host, "")` 😛

dencoded: I think I need to clarify:

I mean if we leave multiple IPs entries in cache - it will do real resolve only once to put it in cache then if item from cache is multi-IPs we do `return safeDial(address, "")` - that would guarantee that `go` or `cgo` (whatever set to be used) is making choice which IP to pick.

also, in this case we caching, but caching (with TTL) only that fact that given host has multiple IPs

dnscache/manager.go
evityuk: yeah, I've fixed it in this commit - 2dae7488979dcd54c6cd38f7fd26037d990905af .
Now rand is lazyly initialized with the first access to randomStrategy handler. Current seed should be enought for the require level of randomness.

dnscache/manager.go
```diff
@@ -0,0 +1,162 @@
...
+			return safeDial(ip+":"+port, host)
+		}
+		return safeDial(ips[0]+":"+port, host)
+	}
+
+	return safeDial(ips[0]+":"+port, host)
+}
+
+func (m *DnsCacheManager) getRandomIp(ips []string) (string, error) {
+	if m.strategy == config.RandomStrategy {
```
dencoded: sorry for being picky, somewhat about code styling - I am a big fan of early returns like:
```
func (m *DnsCacheManager) getRandomIp(ips []string) (string, error) {
	if m.strategy != config.RandomStrategy {
            return ....
         }

        // other logic
        .....
}
```
also testing `m.strategy` for  in func `getRandomIp` and calling this func only when `m.strategy == config.RandomStrategy` looks la little awkward to me, but not big deal.

dnscache/manager.go
dencoded: @buger tbh I am OK with this implementation

just wanted to mention that it is no_cache but still `pick_first` at the same time, discarding system dealer logic for multiple IPs. if we are Ok with this let's leave it.

dnscache/manager.go
evityuk: >I am a big fan of early returns like

+1, :) 

Left a check in order to prevent memory allocation for cases, when other strategies will be involved. 
With requirements change it can removed though. 
Commit - b1aae61dc18bc7ec25ad0946c5adc341811ffb9d

dnscache/manager.go
evityuk: @dencoded I think you missed some point:
>that would guarantee that `go` or `cgo` (whatever set to be used) is making choice which
It won't, because it will have a list with 1 ip and this is the only ip which will be used for connection establishment(in other words - no additional dns request would take place). See [sources](https://golang.org/src/net/lookup.go#L198): LookupIPAddr is called before [connecting](https://golang.org/src/net/dial.go#L402)

@buger FYI

dencoded: @evityuk my point was that for multiple IPs we need to call `net.Dial` passing `address` (not IP) and it will do `LookupIPAddr` inside applying its system/default logic
instead of:
doing call `net.Dial` with 1st IP picked by us and eliminating any logic inside standard `net.Dial`

PR #2044, opened 2018-12-31
Support multiple signature validation modes

asoorm:
#2045

Allows optional addition of `validate_signature` field to
`apidefintion.auth` object.

Currently supports mashery signature validation modes `MasherySha256`
and `MasheryMd5`.

In order to test, add something like the following to auth field of api definition. Declarative syntax, in that signature validation logic is only enabled based upon the presence of the validate_signature object.

```json
"validate_signature": {
  "mode": "MasherySha256",
  "signature_header_key": "x-Signature",
  "session_meta_key": "secret",
  "allowed_clock_skew": 30,
  "error_code": 403,
  "error_message": "your signature is invalid"
}
```

Ensure to add a shared secret to the token meta.

Benchmarks:

```
BenchmarkMasherySha256Sum_Hash-4          500000              2094 ns/op
208 B/op          4 allocs/op
```


apidef/api_definitions.go
```diff
@@ -435,12 +435,22 @@ type APIDefinition struct {
...
+	UseCookie         bool               `mapstructure:"use_cookie" bson:"use_cookie" json:"use_cookie"`
+	CookieName        string             `mapstructure:"cookie_name" bson:"cookie_name" json:"cookie_name"`
+	AuthHeaderName    string             `mapstructure:"auth_header_name" bson:"auth_header_name" json:"auth_header_name"`
+	UseCertificate    bool               `mapstructure:"use_certificate" bson:"use_certificate" json:"use_certificate"`
+	ValidateSignature *ValidateSignature `mapstructure:"validate_signature" bson:"validate_signature" json:"validate_signature,omitempty"`
+}
+
+type ValidateSignature struct {
+	Mode             string `mapstructure:"mode" bson:"mode" json:"mode"`
+	SignatureKey     string `mapstructure:"signature_key" bson:"signature_key" json:"signature_key"`
```
letzya: Do` signature_header_key` and `signature_key` refer to the same config field?

signature_validator/validate_test.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		if err == nil && s.Error != nil {
+			t.Errorf("expected error (%s), got success", s.Error.Error())
+			t.FailNow()
+		}
+	}
+}
+
+func TestValidateSignature_Validate(t *testing.T) {
```
letzya: Dont we need an end-to-end test with session check as well? The code that pulls the shared secret from the `session_meta_key` should be tested.

Issue #2036, opened 2018-12-21
Host checks does not work if we have 2 hybrid gateways


ilijabojanovic:
**Do you want to request a *feature* or report a *bug*?**
BUG
**What is the current behavior?**
When we have 2 slaved gateways on same redis, host check does not work on second gateway. If 1st gateway detect that some host is down, that info will be not propagated to second gateway.
**What is the expected behavior?**
If host under check is down, all gateway should exclude that host from routing list
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. Start 2 hybrid gateways
2. Create api definition with load balancing and enable host checks 

Gateway config for host checks:
```json
"uptime_tests": {
        "disable": false,
        "config": {
            "failure_trigger_sample_size": 2,
            "time_wait": 10,
            "checker_pool_size": 50,
            "enable_uptime_analytics": false
        }
```
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
Stable and Master

ahmedkaballo:
hello 
we contacted the  support team and they provided us with test license for two week for testing to nodes architecture but they didn't provide us with a documentation of how to add another gateway.
so how can i add two gateways??
should i just configure the second gateway to reach the db and configure a load balancer with floating ip for the two gateways?

ilijabojanovic:
Verified

buger:
https://github.com/TykTechnologies/tyk/pull/2049 was merged to `%!s(MISSING)` branch
> Created via API

buger:
https://github.com/TykTechnologies/tyk/pull/2049 was merged to `release-2.7` branch
> Created via API

buger:
https://github.com/TykTechnologies/tyk/pull/2049 was merged to `release-2.8` branch
> Created via API

buger:
<small>Test small text</small>

buger:
https://github.com/TykTechnologies/tyk/pull/2049 was merged to `release-2.7` branch
<details><summary></summary>Created via API</details>

buger:
https://github.com/TykTechnologies/tyk/pull/2049 was merged to `release-2.8` branch
<details><summary></summary>Created via API</details>

PR #2049, opened 2019-01-08
Fix host checker

buger:
Current implementation has multiple flaws. 
At the moment first HostDown event happens after a number of specified tries, and if the host is still down, the next HostDown event will happen after the same number of tries. So if `time_wait` is 10s, and `failure_trigger_sample_size` 2, it means 20 seconds between events. At the same time we set redis expiration key for "host down" key, to `time_wait` + 1, which means key will be expired 9 seconds before second HostDown event. This PR fix it by setting expiration time to `time_wait` * `failure_trigger_sample_size`

Another issue is how host considered to be up. At the moment on first succesful attempt we just enabled the host, however if upstream is unstable, having single healthy attempt does not mean that it is recovered. This change makes Host UP logic work exactly like Host Down and now consider number of tries. 

Fix https://github.com/TykTechnologies/tyk/issues/2036

### DIFF
--- a/host_checker.go
+++ b/host_checker.go
@@ -112,8 +112,21 @@ func (h *HostUptimeChecker) HostReporter() {
 		case okHost := <-h.okChan:
 			// Clear host from unhealthylist if it exists
 			if h.unHealthyList[okHost.CheckURL] {
-				h.upCallback(okHost)
-				delete(h.unHealthyList, okHost.CheckURL)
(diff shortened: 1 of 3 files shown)


letzya:
@buger Could be nice to add here response time of that sample against an expected_resp_time, then we can raise an event of degradation in service if it's longer and back-in-service it's <= expected_resp_time.

PR #2058, opened 2019-01-14
Preparation for log-language

asoorm:
`Sirupsen` changed github handle to `sirupsen`.

We have a very old version of logrus. In order to update logrus, had to change and update a lot of vendored packages


buger:
@asoorm this PR conflicts a lot with recently merged PR by @furkansenharputlu, which mostly do the same stuff. I think you have to re-do your change with replacing `github.com/Sirupsen/logrus`

Issue #2061, opened 2019-01-16
Can't transform GET request to POST request with body


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
Let's say we have a GET endpoint which should call POST service, with some payload. 
In the context of plugins, it means that we need to add 2 plugins:
1) Method transform: from GET to POST
2) Body transform: from empty to smth else

**What is the expected behavior?**
It fails on body transform level
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
See above
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.7

letzya:
https://github.com/TykTechnologies/tyk/issues/2063

buger:
https://github.com/TykTechnologies/tyk/pull/2062 was merged to `release-2.8` branch
<details><summary></summary>Created via API</details>

buger:
https://github.com/TykTechnologies/tyk/pull/2062 was merged to `release-2.7` branch
<details><summary></summary>Created via API</details>

PR #2062, opened 2019-01-16
Make transform work with empty bodies

buger:
Fix https://github.com/TykTechnologies/tyk/issues/2061


ilijabojanovic:
@buger approved for merge

Issue #2064, opened 2019-01-16
Multiple plugins stop working when URL rewrite plugin changed url


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
If URL rewriter changed the path, let's say from "/get" to "/ip", the following plugins stop working:
- Method transform
- Request size limit
- Virtual Endpoint
**What is the expected behavior?**
Plugins should work
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
See above
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.7

buger:
https://github.com/TykTechnologies/tyk/pull/2065 was merged to `release-2.8` branch
<details><summary></summary>Created via API</details>

buger:
https://github.com/TykTechnologies/tyk/pull/2065 was merged to `release-2.7` branch
<details><summary></summary>Created via API</details>

PR #2065, opened 2019-01-16
Fix multiple plugins stopped working after URL rewrite

buger:
It just checkes for original URL in call cases. 

Fix https://github.com/TykTechnologies/tyk/issues/2064


ilijabojanovic:
@buger approved for merge

Issue #2067, opened 2019-01-17
Listen path name conflicts with endpoint names


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
If you set up a listen path which looks like `/post-with-get` and one of your API endpoints defined plugins for the path which include the same prefix as listen path `/post`, plugins applied to this endpoint, will be applied to ALL urls. 

**What is the expected behavior?**
Listen path should not interfere with endpoint paths

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
see above
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.7

buger:
https://github.com/TykTechnologies/tyk/pull/2068 was merged to `release-2.8` branch
<details><summary></summary>Created via API</details>

buger:
https://github.com/TykTechnologies/tyk/pull/2068 was merged to `release-2.7` branch
<details><summary></summary>Created via API</details>

PR #2068, opened 2019-01-17
Exclude listen path prefix from url when matching plugins

buger:
Issue was happening because ListenPath was included into request URL

Fix https://github.com/TykTechnologies/tyk/issues/2067


ilijabojanovic:
@buger verified and approved for merge

Issue #553, opened 2017-02-21
Extend CoProcess plugins to support Go 1.8 plugins


lonelycode:
**Do you want to request a *feature* or report a *bug*?**
Feature 

**What is the current behavior?**
Tyk CP plugins do not support go plugins

**What is the expected behavior?**
We should support Go plugins

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
All.

Checklist:
- [x] Gateway implementation
- [x] Build env
- [ ] Docs

mvdan:
Do we have any sort of coprocess plugins in Go right now, or is it all Python/Lua/js?

matiasinsaurralde:
@mvdan At the moment you can write a plugin using any language that supports Protocol Buffers / gRPC, including Go, pretty similar to [this](https://github.com/hashicorp/go-plugin#architecture).

mvdan:
Also worth noting that we can't add this for another 5 months at least, until 1.9 is out and we can drop 1.7 support.

And support for them in 1.8 is only available in Linux. But since we don't plan on extending support to *BSD, Mac or Windows anytime soon, that should not be a problem.

buger:
Need to check if it is already done and check status https://github.com/TykTechnologies/tyk-plugin-demo-golang

mvdan:
Different things as far as I can see - both would run Go code, but that repo is using grpc, while this is about native Go plugins.

matiasinsaurralde:
@mvdan is right, the above sample uses the current plugin architecture, so it's a gRPC server written in Go that implements the dispatcher service.
I did an experiment with actual Go plugins months ago, but we were discussing the idea of using them as part of the middleware chain directly (not using/loading them using our current architecture: bundles, etc.).

buger:
@dencoded transferring to you.

General idea is to come up with a proper spec for this task, gather all requirements, and do the proper PoC. @asoorm work should be a good start https://github.com/TykTechnologies/tyk/pull/1868 (feel free to use same branch if you want)

I think maybe we could extend direct middleware interface here, with `ProcessRequest` and etc. 

Also think about developer experience. It would probably mean creating a separate docker image which is used to build tyk at our CI and for building plugins themselves (since they are need to be build at the same env). 

[...]

PR #2077, opened 2019-01-21
PoC: Go-plugins support added

dencoded:
added initial changes for https://github.com/TykTechnologies/tyk/issues/553

- provides low level access to plugin supplying actual `http.ResponseWriter`, `*http.Request`
- assumes that plugin optionally sends reply to `http.ResponseWriter` - in this case all tyk's MW-chain stops and reply returned to client
- assumes that plugin always wire reply to `http.ResponseWriter` in case of error
- recovers with 500 error reply in case plugin did panic 
- our logic around custom middleware was used
- we specify all settings on API spec level in `custom_middleware` field, i.e.:
```
"custom_middleware": {
      "pre": [],
      "post": [
        {
          "name": "MyProcessRequest",
          "path": "/abc/xyz.so",
          "require_session": true
        }
      ],
      "post_key_auth": [],
      "auth_check": {},
      "driver": "goplugin",
      ...
    }
```
- added new plugin driver `"goplugin"`
- supports `"pre"`, `"post"`, `"post_key_auth"` and `"auth_check"` custom middlewares as described in https://tyk.io/docs/customise-tyk/plugins/rich-plugins/rich-plugins-work/#coprocess-dispatcher-hooks 
- .so file can also be packaged as a bundle and loaded/deployed using our logic for `custom_middleware_bundle`


buger:
As discussed previously, let's make default plugin interface just be as simple as `func(ResponseWriter, *Request)`, and things like getting and setting session, will be done using request context, exactly like our internal code does. In this case Auth middleware will be exactly same as Pre, just with setSession call: in this case the only difference will be only supporting [ID extractor](https://tyk.io/docs/customise-tyk/plugins/rich-plugins/id-extractor/)

dencoded:
@buger PTAL

- I've changed Go-plugin signature to `func(http.ResponseWriter, *http.Request)` as suggested. I wanted to reuse standard type `http.HandlerFunc` but type casting from loaded symbol didn't like it so ended up just with `func(http.ResponseWriter, *http.Request)`
- added new package `"github.com/TykTechnologies/tyk/ctx"` to be used by Go-plugin developers (for auth) with exported functions as for now:
  - `func GetAuthToken(r *http.Request) string`
  - `func GetSession(r *http.Request) *user.SessionState`
  - `func SetSession(r *http.Request, s *user.SessionState, token string, scheduleUpdate bool)` (go plugin MW forces to schedule session update if Go-plugin called `SetSession` but with scheduleUpdate=false)

so the simple Go-plugin with hit counters could be:
```
package main

import (
	"encoding/json"
	"net/http"
	"sync"

	"github.com/TykTechnologies/tyk/ctx"
	"github.com/TykTechnologies/tyk/log"
	"github.com/TykTechnologies/tyk/user"
)

var logger = log.Get()

// plugin exported functionality
func MyProcessRequest(rw http.ResponseWriter, r *http.Request) {
	endPoint := r.Method + " " + r.URL.Path
	logger.Info("Custom middleware, new hit:", endPoint)

	hitCounter := recordHit(endPoint)
	logger.Debug("New hit counter value:", hitCounter)

	if hitCounter > 100 {
		logger.Warning("Hit counter to high")
	}

	reply := myReply{
		Session:    ctx.GetSession(r),
		Endpoint:   endPoint,
		HitCounter: hitCounter,
	}

	jsonData, err := json.Marshal(reply)
	if err != nil {
		logger.Error(err.Error())
		rw.WriteHeader(http.StatusInternalServerError)
		return
	}

	rw.Header().Set("Content-Type", "application/json")
	rw.WriteHeader(http.StatusOK)
	rw.Write(jsonData)
}

// called once plugin is loaded, this is where we put all initialization work for plugin
// i.e. setting exported functions, setting up connection pool to storage and etc.
func init() {
	hitCounter = make(map[string]uint64)
}

// plugin internal state and implementation
var (
	hitCounter   map[string]uint64
	hitCounterMu sync.Mutex
)

func recordHit(endpoint string) uint64 {
	hitCounterMu.Lock()
	defer hitCounterMu.Unlock()
	hitCounter[endpoint]++
	return hitCounter[endpoint]
}

type myReply struct {
	Session    *user.SessionState `json:"session"`
	Endpoint   string             `json:"endpoint"`
	HitCounter uint64             `json:"hit_counter"`
}
```

Also, I will rebase/resolve conflicts with another commit push.

buger:
Love it!
Does specifying the direct path to file inside `spec.custom_middleware` will work too?

mw_go_plugin.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	// call Go-plugin function
+	m.handler(rw, r)
+
+	// check if we need to schedule session update in case session was updated by Go-plugin
+	// but update wasn't scheduled
+	if prevMD5Hash != "" {
+		if session := ctxGetSession(r); session != nil && prevMD5Hash != session.MD5Hash() {
+			ctxScheduleSessionUpdate(r)
```
buger: Why not to trust what passed to `ctxSetSession` ?

buger:
I think it makes sense add unit tests as well. In this case we can dynamically generate `so` files inside `ci-tests.sh`, and conditionally run tests only if specific tag is passed, or add some test helper which dynamicall compile `Go` code specific in some string, so flow will be similar to https://github.com/TykTechnologies/tyk/blob/master/coprocess_id_extractor_python_test.go#L57

mw_go_plugin.go
dencoded: plugin code might do call `ctx.SetSession` with passing `scheduleUpdate=false` so we are forcing it there. we can remove this logic assuming developer of plugin will do update session properly

dencoded:
@buger PTAL

buger:
Like it!

Happy to merge.

The next step is documentation and build pipeline (e.g. instructions on how to build binaries compatible with official builds).

dencoded:
I've started working on docs changes as well. Docs wait for build pipe line (we need to document how to build plugins). Yes, build pipeline is a next priority.

buger:
Updated to latest master and made small change to run tests only if "goplugin" tag is passed (enabled on CI), because otherwise just running `go test` will always fail.

buger:
@dencoded I think my change found a potential issue: when Go compiled with `coprocess` tag (our default distribution), it throws: `time="Jun  2 17:11:01" level=error msg="CP Driver not supported: goplugin" `

dencoded:
@buger I think this error message should be Ok as it means that `CoProcessMiddleware` won't be used for Go-plugin which looks Ok to me

Issue #2080, opened 2019-01-23
Handlers for API are registered concurrently


komalsukhani:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
I ran into read-write race condition after running code against go race detector. After digging through code I realized that APIs are registered concurrently. But gorilla/mux doesn't support concurrent registering of handlers. I have confirmed it by raising an issue on gorilla/mux [#441](https://github.com/gorilla/mux/issues/441#issue-402104437)

**What is the expected behavior?**

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

Here is output of go run -race

```console
WARNING: DATA RACE
Read at 0x00c0004322a8 by goroutine 40:
  runtime.growslice()
      /usr/local/go/src/runtime/slice.go:98 +0x0
  github.com/TykTechnologies/tyk/vendor/github.com/gorilla/mux.(*Router).HandleFunc()
      /home/komal/GoWorkspace/src/github.com/TykTechnologies/tyk/vendor/github.com/gorilla/mux/mux.go:256 +0x37c
  main.addBatchEndpoint()
      /home/komal/GoWorkspace/src/github.com/TykTechnologies/tyk/main.go:457 +0x1dd
  main.processSpec()
      /home/komal/GoWorkspace/src/github.com/TykTechnologies/tyk/api_loader.go:217 +0x8097
  main.loadApps.func3()
      /home/komal/GoWorkspace/src/github.com/TykTechnologies/tyk/api_loader.go:602 +0x3de

Previous write at 0x00c0004322a8 by goroutine 98:
  github.com/TykTechnologies/tyk/vendor/github.com/gorilla/mux.(*Router).HandleFunc()
      /home/komal/GoWorkspace/src/github.com/TykTechnologies/tyk/vendor/github.com/gorilla/mux/mux.go:256 +0x212
  main.addBatchEndpoint()
      /home/komal/GoWorkspace/src/github.com/TykTechnologies/tyk/main.go:457 +0x1dd
  main.processSpec()
      /home/komal/GoWorkspace/src/github.com/TykTechnologies/tyk/api_loader.go:217 +0x8097
  main.loadApps.func3()
      /home/komal/GoWorkspace/src/github.com/TykTechnologies/tyk/api_loader.go:602 +0x3de

Goroutine 40 (running) created at:
  main.loadApps()
[...]

PR #2082, opened 2019-01-23
Fix API loader race

buger:
Remove using goroutine when processing specs.
API loader gets called very rare and does not need to be parallelized.

Fix https://github.com/TykTechnologies/tyk/issues/2080

### DIFF
--- a/api_loader.go
+++ b/api_loader.go
@@ -546,8 +546,6 @@ func loadApps(specs []*APISpec, muxer *mux.Router) {
 		return len(specs[i].Proxy.ListenPath) > len(specs[j].Proxy.ListenPath)
 	})
 
-	chainChannel := make(chan *ChainObject)
-
 	// Create a new handler for each API spec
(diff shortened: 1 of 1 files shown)


api_loader.go
```diff
@@ -584,32 +582,23 @@ func loadApps(specs []*APISpec, muxer *mux.Router) {
...
+			}).Warning("Trying to load API with Domain when custom domains are disabled.")
+			subrouter = muxer
+		}
 
-			chainObj.Index = i
-			chainChannel <- chainObj
-			apisMu.Lock()
-			spec.middlewareChain = chainObj
-			apisMu.Unlock()
-		}(spec, i)
```
dencoded: just something draw my attention - here we possibly had the same spec passed to go-routine because for-cycle vars are being shared between iterations

buger: Maybe, but in general, it is very unsafe thing if you run on the multi-core machine and it schedules these goroutines in parallel.

PR #2085, opened 2019-01-24
Remove unwanted mutex

komalsukhani:
As goroutines are removed, mutex can be removed as well.


komalsukhani:
Oh. I thought mutex was used just because multiple goroutines were accessing it.

Issue #2093, opened 2019-01-30
Make sure non-expired tokens issued for oauth-client cannot be used after oath-client is deleted


dencoded:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
We delete oauth client but leave issued token 

**What is the expected behavior?**
We need to make non-expired tokens issued for that oauth-client cannot be used after oath-client is deleted

This needs to be done in scope of https://github.com/TykTechnologies/product/issues/19

ilijabojanovic:
Verified

PR #2095, opened 2019-02-01
oauth-client check added in Oauth2KeyExists MW

dencoded:
Added changes https://github.com/TykTechnologies/tyk/issues/2093

- check if oauth client is present in Redis
- trying to avoid hitting Redis every time:
  - for deleted oauth client there will be only one hit, then result cached in memory cache
  - for not deleted oauth client result is cached for 1 sec

Also, was exploring option just deleting all tokens at the time of oauth-client deletion but that might not work as sync call in case of big amount of issued tokens.

### DIFF
--- a/handler_success.go
+++ b/handler_success.go
@@ -40,8 +40,16 @@ const (
 	Trace
 )
 
-var SessionCache = cache.New(10*time.Second, 5*time.Second)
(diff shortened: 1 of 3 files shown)


buger:
Makes sense

buger:
Deleting all tokens is not an options, since their analytics data potentially can be useful. And we already have options to control expiration of oauth tokens.

PR #2096, opened 2019-02-01
 Add NotMatchPattern to advanced url rewriting

### LINKED ISSUE #1446: Advanced URL rewriter should support negative logic

furkansenharputlu:
Fixes #1446

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -135,6 +135,7 @@ type CircuitBreakerMeta struct {
 
 type StringRegexMap struct {
 	MatchPattern string `bson:"match_rx" json:"match_rx"`
+	Reverse      bool   `bson:"reverse" json:"reverse"`
 	matchRegex   *regexp.Regexp
 }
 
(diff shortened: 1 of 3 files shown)


apidef/api_definitions.go
```diff
@@ -543,25 +544,48 @@ func (a *APIDefinition) DecodeFromDB() {
...
+		return
+	}
+
 	return s.matchRegex.FindString(value)
 }
 
-func (s *StringRegexMap) FindStringSubmatch(value string) []string {
-	return s.matchRegex.FindStringSubmatch(value)
+func (s *StringRegexMap) FindStringSubmatch(value string) (matched bool, match []string) {
+	if s.matchRegex == nil {
```
buger: I think this clause, and in `Check` function not needed anymore.

furkansenharputlu: If `s.matchRegex, err = regexp.Compile(s.MatchPattern);` fails, it will be `nil` so doesn't it panic ?

Issue #2094, opened 2019-01-30
Add new endpoint to edit oauth-client


dencoded:
**Do you want to request a *feature* or report a *bug*?**
Feature
**What is the current behavior?**
It is not possible to edit oauth-client info at the moment.
**What is the expected behavior?**
We need to add new endpoint to edit oauth-client. Also,  add new field to oauth-client to store "title" , allow to edit "title", "redirect_uri" and "policy_id" fields. Also, we need to decide what to do with issued tokens after policy_id changed as it might provide different access. Probably just delete non expired tokens issued for that oauth-client.

This is in scope of https://github.com/TykTechnologies/product/issues/19

ilijabojanovic:
Verified

PR #2099, opened 2019-02-05
store user data in oauth-client and in sessions for issued tokens

dencoded:
added changes for https://github.com/TykTechnologies/tyk-analytics/issues/1043 and https://github.com/TykTechnologies/tyk-analytics/pull/1058

- added logic to populate field `meta_data` in newly created OAuth-client, this data might contain anything. I.e. dashboard passes here key-request fields, developer-fields and developer ID - as part of oauth-app key-request approval flow so we can easily get oauth-app owner (portal developer)
- when new oauth-token gets issued and oauth-client has `meta_data` set  - we assign it to key's `session.MetaData` so session will have developer data. Also, we can report more detailed statistics about oauth (i.e. add tag with developer ID)
- endpoint to create OAuth-client can accept new field `meta_data` where we can put some data related to this new oauth-client
- also get oath-client and list of oauth-clients will have this new field `meta_data` in response

### DIFF
--- a/api.go
+++ b/api.go
@@ -1171,10 +1171,11 @@ func createKeyHandler(w http.ResponseWriter, r *http.Request) {
 type NewClientRequest struct {
 	ClientID          string      `json:"client_id"`
 	ClientRedirectURI string      `json:"redirect_uri"`
-	APIID             string      `json:"api_id"`
-	PolicyID          string      `json:"policy_id"`
+	APIID             string      `json:"api_id,omitempty"`
+	PolicyID          string      `json:"policy_id,omitempty"`
 	ClientSecret      string      `json:"secret"`
 	MetaData          interface{} `json:"meta_data"`
+	Description       string      `json:"description"`
 }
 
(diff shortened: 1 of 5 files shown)


api.go
```diff
@@ -1169,11 +1169,12 @@ func createKeyHandler(w http.ResponseWriter, r *http.Request) {
...
-	ClientRedirectURI string `json:"redirect_uri"`
-	APIID             string `json:"api_id"`
-	PolicyID          string `json:"policy_id"`
-	ClientSecret      string `json:"secret"`
+	ClientID          string      `json:"client_id"`
+	ClientRedirectURI string      `json:"redirect_uri"`
+	APIID             string      `json:"api_id"`
+	PolicyID          string      `json:"policy_id"`
+	ClientSecret      string      `json:"secret"`
+	UserData          interface{} `json:"user_data"`
```
buger: Lets rename to `meta_data` like for keys

dencoded:
also added more changes for https://github.com/TykTechnologies/tyk/issues/2094

Issue #2105, opened 2019-02-11
Pass though gRPC proxying not working


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
While gRPC main transport is HTTP/2 and Tyk supports it, gRPC proxying does not work.
**What is the expected behavior?**
It should proxy without any issues. Looks like Tyk stripping "Te" trailer which is required for gRPC.
See https://github.com/golang/go/issues/21096 for more details.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
It's broken
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.7.x

PR #2106, opened 2019-02-11
Fix gRPC proxying

buger:
Pass "Te" trailer. See https://github.com/golang/go/issues/21096

Should fix https://github.com/TykTechnologies/tyk/issues/2105


buger:
Replaced by https://github.com/TykTechnologies/tyk/pull/2119

PR #2109, opened 2019-02-11
Add pagination when listing Oauth clients tokens

adelowo:
I have added an optional query parameter called `page` to the endpoint `/oauth/clients/{apiID}/{keyName}/tokens` . To prevent breaking backwards compatibility, tokens 
will only be paginated if they contain a `page` query item

Part of https://github.com/TykTechnologies/tyk-analytics/issues/1101


oauth_manager.go
```diff
@@ -509,16 +513,42 @@ func (r *RedisOsinStorageInterface) GetClientTokens(id string) ([]OAuthClientTok
 		go r.store.RemoveSortedSetRange(key, "-inf", cleanupStartScore)
 	}
 
+	itemsPerPage := 2
```
adelowo: Not sure of how this will fit in with the config system. But I have it here right now so I could add tests and all of that.

If I am guided on how this will be added to the config or how it should be specified, I'd happily rework this PR

api.go
```diff
@@ -1365,7 +1365,7 @@ func updateOauthClient(keyName, apiID string, r *http.Request) (interface{}, int
 
 	// invalidate tokens if we had a new policy
 	if prevPolicy := client.GetPolicyID(); prevPolicy != "" && prevPolicy != updatedClient.PolicyID {
-		tokenList, err := apiSpec.OAuthManager.OsinServer.Storage.GetClientTokens(updatedClient.ClientID)
+		tokenList, _, err := apiSpec.OAuthManager.OsinServer.Storage.GetClientTokens(updatedClient.ClientID, 1)
```
adelowo: https://github.com/TykTechnologies/tyk/blob/c0dab5d1a023df456372b45795bd8936d6f369ba/api.go#L1366-L1369 was added in https://github.com/TykTechnologies/tyk/pull/2099 so might want to rethink by adding a new method that allows fetching all tokens still, or would passing a `-1` as the page parameter suffix - ideally no?

buger:
Thank you! 

Going to review soon.

oauth_manager.go
adelowo: Ended up going with 
https://github.com/TykTechnologies/tyk/pull/2109/commits/e3d35f385c4cac231e739958344553574a93fbb2 ..

api.go
dencoded: why is it `GetClientTokens(updatedClient.ClientID, 1)` - that would return only 1st page of tokens to be wiped away. however, we still need to cleanup all tokens per client in case of policy change

api.go
```diff
@@ -1548,14 +1548,26 @@ func oAuthClientTokensHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	page := 1
```
dencoded: Here we are changing logic behind already public endpoint https://tyk.io/docs/tyk-rest-api/oauth-key-management/#retrieve-all-current-tokens-for-specified-oauth-client which I believe some of Tyk users already started using.

I think it might be better:
1. to provide a way here to retrieve all tokens and optional pagination so we won't break existing functionality
2. to add new endpoint for paginated tokens as clients will likely need pagination info to build a ruler

I'd vote for 2nd as it provides backward compatibility and gives more tooling to build pagination UI

api.go
```diff
@@ -1548,14 +1548,26 @@ func oAuthClientTokensHandler(w http.ResponseWriter, r *http.Request) {
...
 	// get tokens from redis
-	// TODO: add pagination
-	tokens, err := apiSpec.OAuthManager.OsinServer.Storage.GetClientTokens(keyName)
+	tokens, totalPages, err := apiSpec.OAuthManager.OsinServer.Storage.GetClientTokens(keyName, page)
 	if err != nil {
 		doJSONWrite(w, http.StatusInternalServerError, apiError("Get client tokens failed"))
 		return
 	}
 
+	_ = totalPages
```
dencoded: if we decide to go with new format for paginated reply - might worth to send back pagination info suitable for UIs, i.e.:
```
{
  "pagination": {
     "page_num": 1,
     "page_total": 129,
     "page_size": 10
   },
   "tokens": [
      ...
    ]
}
```

config/config.go
```diff
@@ -265,6 +265,7 @@ type Config struct {
 	OauthTokenExpire                  int32                                 `json:"oauth_token_expire"`
 	OauthTokenExpiredRetainPeriod     int32                                 `json:"oauth_token_expired_retain_period"`
 	OauthRedirectUriSeparator         string                                `json:"oauth_redirect_uri_separator"`
+	PaginationItemsPerPage            int                                   `json:"pagination_items_per_page"`
```
dencoded: if it is a global param for all paginated data  - name is OK, if it is specific to oauth tokens pagination I'd prefer something like `oauth_tokens_per_page`

main.go
```diff
@@ -222,6 +222,11 @@ func setupGlobals() {
 	if config.Global().NewRelic.AppName != "" {
 		NewRelicApplication = SetupNewRelic()
 	}
+
+	if globalConfig := config.Global(); globalConfig.PaginationItemsPerPage <= 0 {
+		globalConfig.PaginationItemsPerPage = 10
```
dencoded: I think it is better move default value `10` to constant, to eliminate "magic" numbers in code base

oauth_manager.go
```diff
@@ -384,13 +385,12 @@ type ExtendedOsinStorageInterface interface {
 	// Custom getter to handle prefixing issues in Redis
 	GetClientNoPrefix(id string) (osin.Client, error)
 
+	GetClientTokens(id string, page int) ([]OAuthClientToken, float64, error)
```
dencoded: I'd prefer new method leaving existing one as is retrieving all tokens from redis, something like `GetClientTokensPaginated`

oauth_manager.go
```diff
@@ -554,16 +557,42 @@ func (r *RedisOsinStorageInterface) GetClientTokens(id string) ([]OAuthClientTok
...
+		}
+
+		if startIdx > n {
+			startIdx = n
+		}
+	}
+
+	totalPages := math.Ceil(float64(len(tokens)) / float64(itemsPerPage))
+
+	tokens = tokens[startIdx:endIdx]
```
dencoded: I am fine with this implementation but in case when we have many tokens  - I do have concerns because here we still pull over network all elements (note how we do `"+inf"`)

Might worth to consider taking advantage of sorted set - being able to move "window" easily using items score. Say we do `r.store.GetSortedSetRange(key, startScore, endScore)` where by default `endScore` is `startScore` plus some range like one day and then we go like this until we fill the page of required size.

We could also count number of elements in sorted set with `ZCOUNT initial_start_score +inf` to get number of possible pages

In that case client would need to receive `next_start_with` and provide it for next page (instead of page number), `next_start_with` is just some value with encoded in it `startScore` + `token` which are last used `endScore` + `token` on previous page. If we do so we will know from where to move "window" in our sorted set for next page.

It might sound not trivial.. just thinking about reducing network I/O and offloading Redis

api.go
adelowo: That is obviously wrong which is why I left the comment about a potential new method so all tokens can be wiped out. Just like the `GetClientTokensPaginated` you described in https://github.com/TykTechnologies/tyk/pull/2109#discussion_r257043315

config/config.go
adelowo: I didn't find anything related to pagination so I thought a ___global config value___ would be it. Fine to change it to the specified name if OAuth tokens are the only place pagination will be needed.

oauth_manager.go
adelowo: I went with `GetPaginatedClientTokens`.

api.go
adelowo: I'd go with the 2nd option too. I'd update this PR to reflect that

adelowo: Kindly take a look at https://github.com/TykTechnologies/tyk/pull/2109/commits/6b91b049204211be7df6505b4924ed57859d0525

buger:
I removed the pagination control option altogether, setting the default to 100.

PR #2110, opened 2019-02-12
Add HTTP/2 support option and add test

furkansenharputlu:
This PR adds HTTP/2 support option to Tyk. 

For example:
There are an HTTP/2 downstream and an HTTP/2 upstream. Tyk will be able to work in that case.

This PR adds two options by default `true` (in default configs, so existing customers not affected):
- `enable_http2` for downstream-Tyk HTTP/2 connection
- `proxy_enable_http2` for Tyk-upstream HTTP/2 connection

Fixes https://github.com/TykTechnologies/product/issues/98

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,6 +19,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/net/http2"
+
 	newrelic "github.com/newrelic/go-agent"
 
 	"github.com/TykTechnologies/tyk/checkup"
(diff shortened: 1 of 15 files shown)


main.go
```diff
@@ -1145,6 +1147,10 @@ func generateListener(listenPort int) (net.Listener, error) {
 			CipherSuites:       getCipherAliases(httpServerOptions.Ciphers),
 		}
 
+		if httpServerOptions.EnableHttp2 {
+			tlsConfig.NextProtos = append(tlsConfig.NextProtos, http2.NextProtoTLS)
```
joshblakeley: I would use `tlsConfig.NextProtos = append(tlsConfig.NextProtos, []string{"h2"},` to avoid the extra import in main package.

Issue #2108, opened 2019-02-11
Deleting key with non existing ID returns success


ilijabojanovic:
**Do you want to request a *feature* or report a *bug*?**
BUG
**What is the current behavior?**
I can pass any Key ID during deletion, and it always tells success
**What is the expected behavior?**
It should fail
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
See above
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.7.X

PR #2111, opened 2019-02-12
Fix validation of key removal

buger:
`RemoveSession` now return boolean and in case of failure it gives an error

Additionally fixed hanging DNS test.
Also speed up tests by making timeout value be a float, e.g. smaller

Fix https://github.com/TykTechnologies/tyk/issues/2108

### DIFF
--- a/api.go
+++ b/api.go
@@ -475,7 +475,7 @@ func handleDeleteKey(keyName, apiID string) (interface{}, int) {
 		}
 		apisMu.RUnlock()
 
-		if removed {
+		if !removed {
 			log.WithFields(logrus.Fields{
 				"prefix": "api",
 				"key":    obfuscateKey(keyName),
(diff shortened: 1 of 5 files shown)


ilijabojanovic:
@buger approved for merge

api.go
```diff
@@ -466,12 +466,24 @@ func handleDeleteKey(keyName, apiID string) (interface{}, int) {
...
 		for _, spec := range apisByID {
-			spec.SessionManager.RemoveSession(keyName, false)
+			if spec.SessionManager.RemoveSession(keyName, false) {
+				removed = true
+			}
 			spec.SessionManager.ResetQuota(keyName, &user.SessionState{}, false)
 		}
 		apisMu.RUnlock()
 
+		if removed {
```
letzya: @buger

buger: fixed

Issue #1838, opened 2018-07-24
Have caching optionally take into account POST bodies


bitsofinfo:
**Do you want to request a *feature* or report a *bug*?**

feature

**What is the current behavior?**

Caching cannot take into account POST bodies

**What is the expected behavior?**

Caching should have an option to hash POST bodies and use that as part of cache key

Again an option, default behavior leave as is, but per api-def, a flag to permit this.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

We have requests where the URI/querystring does not change but the contents of the POST body do vary (not huge bodies, but small ones) where the params are in the body, not the URI.

We cannot cache this data.

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**

2.6.2+

buger:
Right now it works like his:
1) Only safe methods can be cached
2) If you checked only "Enable caching", then endpoint will be cached only if you set "Cache" plugin to endpoint
3) If you check "Cache all safe requests" then it will cache all safe request by default, without needing step 2)

With this new functionality, if you set "Cache" plugin for POST endpoint, it will automatically understand it and will cache POST by the whole body.
Cache plugin for POST/PUT requests will have additional "cache key regexp" field, where you can optionally specify regexp to extract cache key from the body.

Note that even if you enable "Cache all safe requests", then POST request will not be automatically cached. You always had to explicitly set "Cache" plugin for them.

Assigning @komalsukhani for this ticket ^

bitsofinfo:
I'm confused, so this is already functionality that is enabled? or no (per the assignment)

komalsukhani:
@bitsofinfo It is under development.

bitsofinfo:
not stale

bitsofinfo:
not stale @buger

PR #2114, opened 2019-02-13
Feature/1838 cache post requests

komalsukhani:
Fix issue #1838 

A new field `advance_cache_config` is added in API definition. It has following format:
```json
"advance_cache_config": [
				{
					"method":"POST",
					"path":"addBooks",
					"cache_key_regex": "pattern"
				}
			]
```
 If user wants to enable caching of POST for specific path, an entry should be added in `advance_cache_config`. 

`cache_key_regex` enables user to specify regexp to extract cache key from the body. If it set to "", entire post body will be used to generate cache key. `method` field enables caching of request for specific HTTP method. `cache_key_regex` will not be used for safe methods(GET/HEADERS/OPTIONS). It's value is ignored.


buger:
Thank you! Going to review soon.

mw_redis_cache.go
```diff
@@ -41,13 +44,42 @@ func (m *RedisCacheMiddleware) EnabledForSpec() bool {
 	return m.Spec.CacheOptions.EnableCache
 }
 
-func (m *RedisCacheMiddleware) CreateCheckSum(req *http.Request, keyName string) string {
+func (m *RedisCacheMiddleware) CreateCheckSum(req *http.Request, keyName string, regex string) (string, error) {
 	h := md5.New()
 	io.WriteString(h, req.Method)
 	io.WriteString(h, "-")
 	io.WriteString(h, req.URL.String())
+	if req.Method == "POST" {
```
asoorm: I would prefer to use package contstant `http.MethodPost`

mw_redis_cache.go
```diff
@@ -41,13 +44,42 @@ func (m *RedisCacheMiddleware) EnabledForSpec() bool {
...
+		if req.Body != nil {
+			bodyBytes, err := ioutil.ReadAll(req.Body)
+			defer req.Body.Close()
```
asoorm: `defer req.Body.Close()` should be called after `if err != nil {...}`

komalsukhani: Oh yeah. I will fix it.

mw_redis_cache.go
komalsukhani: ok

gateway_test.go
```diff
@@ -1032,6 +1032,39 @@ func TestCacheAllSafeRequests(t *testing.T) {
...
+                                ]`), &v.ExtendedPaths.AdvanceCacheConfig)
+		})
+
+		spec.Proxy.ListenPath = "/"
+	})
+
+	headerCache := map[string]string{"x-tyk-cached-response": "1"}
+
+	ts.Run(t, []test.TestCase{
+		{Method: "POST", Path: "/", HeadersNotMatch: headerCache, Delay: 10 * time.Millisecond},
```
joshblakeley: In this unit test are you actually sending a body with the post request?

joshblakeley: We should probably have a case testing the regex key against specific body.

komalsukhani: Fixed test cases.

PR #2118, opened 2019-02-14
add dns caching support for wss:tls connection

evityuk:
- Add dns caching for wss proxying by using raw tls updage from native cacheable net.Conn
- Use more "collision prone" ips in dns caching tests in reverse_proxy_test.go


buger:
Closed for now, but will keep the branch

PR #2122, opened 2019-02-15
adding policy ID and developer ID to analytics tags

dencoded:
added changes required for https://github.com/TykTechnologies/tyk-analytics/issues/1046


dencoded:
@buger actually these changes are not required for TykTechnologies/tyk-analytics#1046

However, I believe these changes extend analytics for future use.

Also, I've added change to set session alias to developer email who owns oauth-client app (as we do it for regular API keys)

PR #2124, opened 2019-02-18
Add mutual, basic and token-based authentication tests for gRPC

furkansenharputlu:
This PR adds the following tests for gRPC:
- Mutual Authentication
- Basic Authentication
- Token Based Authentication

extends #2119


gateway/server.go
```diff
@@ -887,14 +887,6 @@ func initialiseSystem() error {
 		mainLog.Error("Failed to write PIDFile: ", err)
 	}
 
-	if globalConf.UseDBAppConfigs && globalConf.Policies.PolicySource != config.DefaultDashPolicySource {
```
buger: Does not feels like part of this PR.

PR #2126, opened 2019-02-19
Swagger documentation

adelowo:
See https://github.com/TykTechnologies/product/issues/88

- [x] Definition of swagger models and responses
- [x] CI build step to validate the schema


adelowo:
Right now, the spec will be validated in CI. But in the long run, we might want to push the file 
directly to GitHub from Travis but for now, `.utils/ci-swagger.sh` can be run and the changes in the
`gateway-swagger.yaml` can be committed alongside other change - CI will validate it again though.

swagger.go
```diff
@@ -0,0 +1,21 @@
+package main
+
+import (
+	"github.com/TykTechnologies/tyk/apidef"
+	"github.com/TykTechnologies/tyk/user"
+)
+
+// parameterBodies
+// swagger:response parameterBodies
+type swaggerParameterBodies struct {
```
buger: What it is? and why it is needed?

buger:
Nice!

As mentioned previously our swagger consist of 2 parts, manual part and auto-generated part.
So next step here, will be taking swagger.yaml from here `https://github.com/TykTechnologies/tyk-docs/blob/db26f221108604da80d7cdf4a5469b83e722c1c7/tyk-docs/static/others/gateway-swagger.yaml` and replacing its `components` object, part, with auto-generated stuff from your script: so we will have manually defined endpoints and automatic model definitions.

swagger.go
adelowo: Without this, `go-swagger` won't generate definitions for `apiStatusMessage` and the rest. Thus the swagger spec will be invalid as `"$ref": "#/definitions/APIDefinition"` for example won't work. 

It is more like a way to actually trick `go-swagger` to generating them as definitions in the output file

adelowo:
I might need to make a few updates to this later on today

buger:
The final change is to move the final swagger file from this PR to here https://github.com/TykTechnologies/tyk-docs/pull/560/files, and modify the script, so it updates definitions section.

PR #2128, opened 2019-02-21
Respect tags coming from key level

### LINKED ISSUE #1671: Adding tag on key level is not saved on the key object
bug
Adding a key is not reflected in the key object 

To be shown in a similar way to the policy tag (or remove this option)
image

1. Create an api
2. Create a key - addthe api to the key and add tag to the key
==> you'll see only the hashed tag i.e. key-{hash}

another scenario:
1. Create an api
2. Create policy and add the api and add tag
2. Create a key - add the policy to the key and add tag to the key
==> you'll see only the policy's tag

Latests was checked on the cloud

furkansenharputlu:
When a policy with a tag is applied to a key, it doesn't respect key's tag. This PR fixes the bug if in FE,
https://github.com/TykTechnologies/tyk-analytics-ui/issues/1082 should be fixed for this PR to work properly.

Fixes #1671

### DIFF
--- a/middleware.go
+++ b/middleware.go
@@ -400,7 +400,6 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 
 	// set tags
 	if len(tags) > 0 {
-		session.Tags = make([]string, 0, len(tags))
 		for tag := range tags {
 			session.Tags = append(session.Tags, tag)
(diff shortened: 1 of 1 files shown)


buger:
Merged manually 5aa1466b

PR #2134, opened 2019-02-24
Notify dashboard whenever monitor for a key quota limit kicks in

adelowo:
Per discussion in https://github.com/TykTechnologies/tyk-analytics/issues/1013

~~Will send the PR for the dashboard in the appropriate repository~~ See https://github.com/TykTechnologies/tyk-analytics/pull/1186

### DIFF
--- a/gateway/dashboard_register.go
+++ b/gateway/dashboard_register.go
@@ -92,7 +92,7 @@ func (h *HTTPDashboardHandler) Init() error {
 	return nil
 }
 
-// NotifyDashboardOfKeyQuotaTrigger acts as a form of event which informs the
+// NotifyDashboardOfEvent acts as a form of event which informs the
 // dashboard of a key which has reached a certain usage quota
 func (h *HTTPDashboardHandler) NotifyDashboardOfEvent(event interface{}) error {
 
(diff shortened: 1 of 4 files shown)


buger:
I think we need it more general purpose, to accept multiple event types.

lghiur:
Can we integrate this somehow with analyitics? So we can maybe highlight it on the charts?

buger:
So far I do not know how to do that. It will require some separate "events" collection, which can be mapped to existing analytics data.

buger:
@lghiur it should for sure be logged to Audit. So when we integrate Audit to the dashboard and expose this thing to UI by default, events should be available there as well.

buger:
@lghiur https://github.com/TykTechnologies/tyk-analytics/issues/344

adelowo:
> I think we need it more general purpose, to accept multiple event types.

Done in https://github.com/TykTechnologies/tyk/pull/2134/commits/f458e8bc0148e7ac0d767473ac268657add280da

adelowo:
I figured out that this webhook only returned the trigger limit not the usage of the API key, I have updated it to also contain usage percentage as it makes much more sense if the dashboard sends an email with the actual usage percentage

buger:
Pls rebase and can be merged

Issue #760, opened 2017-05-23
Specify rate limit weights using upstream headers


buger:
**Do you want to request a *feature* or report a *bug*?**
feature
**What is the current behavior?**
By default, each request counted as 1 rate limit point. Sometimes there are endpoints which consume a different amount of resources, and it makes sense to apply different weight for them when calculating rate limits. 

This is an essential feature for granular rate limiting for GraphQL and similar APIs: where the complexity of response heavily depends on user search query.

**What is the expected behavior?**
A server can respond with: `X-Rate-Limit-Weight` header, which specifies the weight of current request.

lonelycode:
This should be configurable as a header so it works with different things

Also maybe we should ensure that rate limits and quotas can use arbitrary metrics first (generalised metric counter), then we can easily plug new ones in

buger:
@lonelycode can you add some examples on "generalised metric counter"? Not sure if I fully understand the concept.

lonelycode:
So if a user wants to do "100 records per second", how do you define a record if a request could return 1 or 50 objects in a list? And that data might be in XML or JSON? 

So you'd want something that can take an arbitrary object, parse it, make some basic assumptions and then count the number of records - if it's generic enough it can fit any use case

Another one we get a lot is KB or MB per second or total - so based on content-size. But that's not generalisable I think.

buger:
@joshblakeley moving to 2.7

This is definitely should be done in a more generalized way. Its configuration file can look smth like:
"weight": {
   "extract_from": "header",
   "extract_from_key": "X-Rate-Limit-Weight"
}

In addition to "header", possible values can be "response_size", or "response_time". 

Additionally, I'm not sure which object should own these settings. It really looks like Key and Policy attributes. And since we planning [rate limits per path](https://github.com/TykTechnologies/tyk/issues/1108), this weighting can be extended to paths as well.

It feels like can be a big theme for a separate release.

joshblakeley:
I was looking at this today and I think this is really the starting point for a more overarching compatibility with GraphQL, happy to leave this for now but will keep researching around it and having a play in a branch with the way it will work when I have time.

nathanschepers:
To echo what @lonelycode said above. I think there could be great value in offering 'speed limits' or 'traffic limits' in addition to 'rate limiting'.

Especially for endpoints that may return large amounts of data, clients may want to 'speed limit' that endpoint by MB/s rather than #requests/sec.

[...]

PR #2144, opened 2019-02-27
TCP proxying

buger:
The current code adds initial code for TCP proxy which multiplex to different services. 

Additionally refactored all code related to how we start a web server, in particular now you can run multiple APIs on different ports, and Tyk will dynamically open or close port listeners.

Added 2 new fields to API definition: `listen_port` and `protocol`. 
Valid protocol values: "", "http", "https", "tcp", "tls". 
By default, protocol is selected based on `http_server_options.use_ssl`, and can be either http or https.

Additionally in order to tell that your upstream should be `tcp` or `tls` one, in target URL you can specify protocol like this: "tls://upstream:<port>". So you can have GW listening on TLS, but pointing to TCP upstreaming, or the other way. 

Example service description:
```
"listen_port": 30001,
"protocol": "tls",
"certificate": ["<cert-id>"],
"proxy": {
    "target_url": "tls://upstream:9191"
}
```

All the TLS related features, like mutual TLS or certificate pinning work as expected. 

Adding "listen_port", means that you can now start HTTPS server on one port, HTTP on another port, and some TCP services on another port as well. The only requirement that each port should serve the same protocol (GW has checks preventing it).

Additionally, TCP proxying, support multiplexing based on SNI information, e.g. you can serve multiple TCP services on different domains, pointing to different upstream. 

TODO:
- [ ] Analytics support
- [ ] Way to specific fixed ports and protocols they are support
- [ ] Enable custom ports via config
- [ ] Support load balancing and service discovery
- [ ] Health checks
- [ ] Graceful restarts
- [ ] Proxy protocol https://github.com/TykTechnologies/tyk/issues/2300
- [ ] Rate limiting based on bytes out https://github.com/TykTechnologies/tyk/issues/760

Fix https://github.com/TykTechnologies/product/issues/17


asoorm:
If port is already in use, and there is an error binding, we get the following error, but gateway doesn't exit.

`ERROR main: Can't start listener error=listen tcp :8080: bind: address already in use`

So when we kill the gateway process, it panics.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x18eb97d]

goroutine 1 [running]:
main.(*proxyMux).swap(0x269a0e0, 0xc0007acfe8)
	/Users/ahmet/go/src/github.com/TykTechnologies/tyk/proxy_muxer.go:163 +0x20d
main.main()
	/Users/ahmet/go/src/github.com/TykTechnologies/tyk/main.go:897 +0x4cb
```

asoorm:
If we decide not to bind on a custom port - we get the following info msg:

`INFO main: API bind on custom port:0` I think we can suppress this message if not using custom port.

Issue #1510, opened 2018-03-05
Modify Response Header/Body should support Context Variables


zeldal:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Modify Response Header/Body doesn't support Context Variables

**What is the expected behavior?**
Modify Response Header/Body should support Context Variables

buger:
This PR #1511 adds new `replaceTykVariables`, which should be straightforward to use inside `HeaderInjector` and `HeaderTransform` middlewares which used for responses. 

Regarding body transformation, it is done using `ResponseTransformMiddleware` and we should took context variable support code from `TransformMiddleware`, looking at the code there can be a lot of re-use.

ilijabojanovic:
Verified

PR #2145, opened 2019-02-27
Add option to meta data on response

letzya:
Fix https://github.com/TykTechnologies/tyk/issues/1510


buger:
Do not think we need to have it configurable. Just make it work without "UseSession", as our request header transformation works. 

Also, we have `replaceTykVariables` helper, which you can see at mw_modify_headers.go , and in addition to tyk_meta, it will also add support for tyk_context.

IMHO it can be as simple as:
```
	for h, v := range h.config.AddHeaders {
		r.Header.Set(k, replaceTykVariables(r, v, false))
	}
```

Do not forget to apply this modification for both endpoint plugin and global plugin (within same file).

Issue #2045, opened 2018-12-31
feature: support custom sha256 and md5 signature validation


asoorm:
**Do you want to request a *feature* or report a *bug*?**

feature

**What is the current behavior?**

In order to support custom signature validation, we need to implement custom middleware. This can introduce extra latency.

**What is the expected behavior?**

Natively support custom signature validation methods such as MD5Sum and Sha256Sum with configurable clock skew via api definition.

```
"validate_signature": {
  "mode": "MasherySha256", // which algo to run
  "signature_header_key": "x-Signature", // location to look for the signature in request
  "session_meta_key": "secret", // location to look for shared secret inside token metadata
  "allowed_clock_skew": 300 / allowable clock skew in seconds
}
```

The following appended to the auth object in api definition should configure this feature. On-Request, after validating a token, extract signature `X-Signature` header, then validate that signature attempt against the shared secret stored inside token meta `secret` field with `+/-` allowable clock skew 300s using the mashery Sha256 signature validation algorithm.

Pseudocode algo

```
apiKey \:= '2fvmer3qbk7f3jnqneg58bu2';  
sharedSecret \:= "qvxkmw57pec7"; 
signature \:= sha256sum(apiKey + sharedSecret + time.Now().Unix()) 
```

PR #2147, opened 2019-02-28
Release 27/custom signature validation

asoorm:
#2045

Allows optional addition of `validate_signature` field to
`apidefintion.auth` object.

Currently supports mashery signature validation modes `MasherySha256`
and `MasheryMd5`.

In order to test, add something like the following to auth field of api definition. Declarative syntax, in that signature validation logic is only enabled based upon the presence of the validate_signature object.

```json
"validate_signature": {
  "mode": "MasherySha256",
  "signature_header_key": "x-Signature",
  "session_meta_key": "secret",
  "allowed_clock_skew": 30,
  "error_code": 403,
  "error_message": "your signature is invalid"
}
```

Ensure to add a shared secret to the token meta.

Benchmarks:

```
BenchmarkMasherySha256Sum_Hash-4          500000              2094 ns/op
208 B/op          4 allocs/op
```

### DIFF
--- a/mw_auth_key.go
+++ b/mw_auth_key.go
@@ -2,12 +2,19 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"net/http"
 	"strings"
 
 	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/certs"
 	"github.com/TykTechnologies/tyk/request"
+	"github.com/TykTechnologies/tyk/signature_validator"
+)
+
+const (
+	defaultErrorCode    = http.StatusForbidden
+	defaultErrorMessage = "Not Authorized"
 )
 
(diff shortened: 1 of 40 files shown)


mw_auth_key.go
```diff
@@ -2,12 +2,19 @@ package main
...
 	"strings"
 
 	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/certs"
 	"github.com/TykTechnologies/tyk/request"
+	"github.com/TykTechnologies/tyk/signature_validator"
+)
+
+const (
+	defaultErrorCode    = http.StatusForbidden
```
asoorm: Default should be `http.StatusUnauthorized`.

Because if signature fails, we do not believe the user is who they say that they are - even if they have correct auth token.

PR #2163, opened 2019-03-13
POC Can we use a different caching library inside gateway

joshblakeley:
WIP


buger:
One of the biggest users of go-cache is our regexp package, so test there too.

buger:
I guess it should go though load testing by @asoorm or someone else

PR #2172, opened 2019-03-15
Refactor project packaging

furkansenharputlu:
This PR renames package names to `tyk` in root folder. The `main.go` will be in `bin` folder and the gateway will be started from there. 

This code will provide the followings:
- We can fix the package structure because everything will not be in `main` package.
- Tests can be easily written everywhere.
- Debugging tool will work in main binary. Let's say you're using Goland and want to make some debugging. You can  add breakpoints and debug the code easily.


adelowo:
Possible to have this merged soon? 

> Travis uses utils/ci-tests.sh but this PR renames the package name utils to bin so it should be changed in Travis.

@furkansenharputlu can you can update this in this PR since it causes a failure

furkansenharputlu:
@joshblakeley It would but I updated the PR.

PR #2180, opened 2019-03-22
Add liveness check

### LINKED ISSUE #857: Improved healthcheck API
feature
We have `/<api>/hello` endpoint, which returns 200 if API loaded. 

This type of health checking is not enough because it does not check redis/dashboard/RPC connectivity. We should check this stuff for each "hello" request, and return the meanungfull error if something wrong happened.

adelowo:
- [x] Redis
- [x] Dashboard
- [x] Rpc

Fixes #857


adelowo:
Will take a look and update as needed

adelowo:
Will update this according to the RFC today

adelowo:
The currently available url is a “readiness check” while this one is a “liveness check. 

https://github.com/TykTechnologies/tyk/issues/857#issuecomment-467162318

gateway/health_check.go
```diff
@@ -0,0 +1,182 @@
...
+
+				case <-ticker.C:
+					gatherHealthChecks()
+				}
+			}
+		}(ctx)
+	}
+}
+
+func gatherHealthChecks() {
```
buger: I would be careful here since in case of network issue you can get a timeout, and health check endpoint will take too long to respond. Instead, I would suggest running all three check-in 3 separate goroutines, like every second, and health check API will just depend on this data, and respond immediately.

cli/linter/schema.json
```diff
@@ -1023,6 +1023,18 @@
     },
     "ssl_force_common_name_check":{
 	    "type":"boolean"
+    },
```
buger: We already have an existing health check endpoint, see `HealthCheckEndpointName`. I do not like introducing new option here, and rather will think how to make it backward compatible. Any thoughts on it?

Regarding separate port, `HealthCheckEndpointName` we have now already running on ControlAPI router, so if you move it to the separate port/domain, health check gets moved too.

gateway/health_check.go
adelowo: That's how it is implemented, API responds on the status, `gatherHealthChecks` runs periodically in another goroutine

gateway/health_check.go
```diff
@@ -0,0 +1,182 @@
...
+	Time          string            `json:"time"`
+}
+
+func initHealthCheck(ctx context.Context) {
+	if config.Global().LivenessCheck.Enabled {
+
+		setCurrentHealthCheckInfo(make(map[string]HealthCheckItem, 3))
+
+		go func(ctx context.Context) {
+			ticker := time.NewTicker(time.Second * 10)
```
buger: Needs be configured by variable

config/config.go
```diff
@@ -126,6 +126,10 @@ type HealthCheckConfig struct {
 	HealthCheckValueTimeout int64 `json:"health_check_value_timeouts"`
 }
 
+type LivenessCheckConfig struct {
```
buger: Lets make it turned on all the time, and just remove config option.

Issue #2169, opened 2019-03-14
2.7.6 http: proxy error: http: ContentLength=XX with Body length 0


bitsofinfo:
**Do you want to request a *feature* or report a *bug*?**

bug

**What is the current behavior?**

Note this references the same scenario in #2143. 

The API def referenced in #2143 was updated to fix the endpoint designer url-rewrite/whitelist paths to be relative from the slug. 

```
          "extended_paths": {
            "white_list": [
              {
                "path": ".*",
                "method_actions": {
                  "GET": {
                    "action": "no_action",
                    "code": 200,
                    "headers": {}
                  },
                  "POST": {
                    "action": "no_action",
                    "code": 200,
                    "headers": {}
                  }
                }
              }
            ],
            "url_rewrites": [
              {
                "path": ".*",
                "method": "GET",
                "match_pattern": ".*",
                "rewrite_to": "$tyk_context.headers_X_My_Proxytourl",
                "triggers": [],
                "MatchRegexp": null
              },
              {
                "path": ".*",
                "method": "POST",
                "match_pattern": ".*",
                "rewrite_to": "$tyk_context.headers_X_My_Proxytourl",
                "triggers": [],
                "MatchRegexp": null
              }
            ]
          }
```

The following request works fine against 2.6.4
```
  curl -X POST \
    https://my-tyk-gw/my/1.0/api/token \
    -H 'accept: application/json' \
    -H 'authorization: Basic xxxxxxxx' \
    -H 'cache-control: no-cache' \
    -H 'content-type: application/x-www-form-urlencoded' \
    -d 'grant_type=client_credentials&scope=execute'
```

The same request now fails against 2.7.6:

With client error:
```
{
    "error": "There was a problem proxying the request"
}
```

With server error in logs:
```
http: proxy error: http: ContentLength=43 with Body length 0 
```

```
my-tyk-gw    | time="Mar 14 16:06:40" level=debug msg="Received DRL data: {XXXX solo-XXXX 1 0 }"
my-tyk-gw    | time="Mar 14 16:06:40" level=debug msg="[Active Nodes]: 17 [Token Bucket Value]: 1100 [Current Load p/s]: 1 [Current Load]: 0.000000"
my-tyk-gw    | time="Mar 14 16:06:40" level=debug msg="Running: common"
[...]

PR #2182, opened 2019-03-27
retain body for requests 

adelowo:
Fixes https://github.com/TykTechnologies/tyk/issues/2169

This PR makes sure to retain request body for requests with Content-Type of `application/x-www-form-urlencoded` as it is drained by ParseForm which is called in the `URLRewrite` middleware. 

https://golang.org/pkg/net/http/#Request.ParseForm says

```
For POST, PUT, and PATCH requests, it also parses the request body as a
form and puts the results into both r.PostForm and r.Form. 
Request body parameters take precedence over URL query string values in r.Form.

For other HTTP methods, or when the Content-Type is not application/x-www-form-urlencoded, 
the request Body is not read, and r.PostForm is initialized to a non-nil, empty value.

``` 

To verify, you can swap the request a bit by using `application/json` instead of 
`application/x-www-form-urlencoded` as the content-type, you will notice the request does not fail

```sh
curl -X POST \
    http://localhost:9000/XXXXXXXX/my/1.0/oauth/token/ \
    -H 'accept: application/json' \
    -H 'authorization: Basic xxxxxxxx' \
    -H 'cache-control: no-cache' \
    -d '{"grant_type":"client_credentials","scope":"execute"}' \
-H 'content-type: application/json'

```

cc @bitsofinfo

### DIFF
--- a/coprocess_id_extractor.go
+++ b/coprocess_id_extractor.go
@@ -50,7 +50,7 @@ func (e *BaseExtractor) ExtractHeader(r *http.Request) (headerValue string, err 
 
 // ExtractForm is used when a FormSource is specified.
 func (e *BaseExtractor) ExtractForm(r *http.Request, paramName string) (formValue string, err error) {
-	r.ParseForm()
+	parseForm(r)
 
 	if paramName == "" {
 		return "", errors.New("no form param name set")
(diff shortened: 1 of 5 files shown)


buger:
Can it cause same issues in rest of the places like https://github.com/TykTechnologies/tyk/blob/master/mw_virtual_endpoint.go#L153 ?

adelowo:
I strongly believe so, also https://github.com/TykTechnologies/tyk/blob/dede56574956c3886635f440dd9e523a75072aad/coprocess_id_extractor.go#L53

I will fix usage in those places too

PR #2194, opened 2019-04-04
mw_basic_auth: avoid missing header errors when using body basic auth…

### LINKED ISSUE #2193: bug: body basic auth credentials

bug

when extracting credentials from body is enabled, we still look in headers first - as such we always get info level log `Attempted access with malformed header, no auth header found.`.

We should not see this log message.

master

matiasinsaurralde:
Fix #2193

### DIFF
--- a/mw_basic_auth.go
+++ b/mw_basic_auth.go
@@ -79,8 +79,6 @@ func (k *BasicAuthKeyIsValid) basicAuthHeaderCredentials(w http.ResponseWriter, 
 	logger := k.Logger().WithField("key", obfuscateKey(token))
 	if token == "" {
 		// No header value, fail
-		logger.Info("Attempted access with malformed header, no auth header found.")
-
 		err, code = k.requestForBasicAuth(w, "Authorization field missing")
 		return
 	}
@@ -156,6 +154,8 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
 		if k.Spec.BasicAuth.ExtractFromBody {
 			w.Header().Del("WWW-Authenticate")
 			username, password, err, code = k.basicAuthBodyCredentials(w, r)
+		} else {
+			k.Logger().Warn("Attempted access with malformed header, no auth header found.")
 		}
 
 		if err != nil {


mw_basic_auth.go
```diff
@@ -77,6 +77,11 @@ func (k *BasicAuthKeyIsValid) requestForBasicAuth(w http.ResponseWriter, msg str
 func (k *BasicAuthKeyIsValid) basicAuthHeaderCredentials(w http.ResponseWriter, r *http.Request) (username, password string, err error, code int) {
 	token := r.Header.Get("Authorization")
 	logger := k.Logger().WithField("key", obfuscateKey(token))
+	if token == "" && k.Spec.BasicAuth.ExtractFromBody {
```
dencoded: this edge case is a good candidate to cover by test

matiasinsaurralde: Seems that we already have tests with empty headers + body extraction enabled in `TestBasicAuthFromBody`.

dencoded: why that test wasn't failing without this fix then?

matiasinsaurralde:
@dencoded I've replaced this with a simpler solution, now we should discuss if we want to log auth errors on `basicAuthBodyCredentials` in the same way we do for `basicAuthHeaderCredentials`, think this should be a separate issue/PR.

mw_basic_auth.go
```diff
@@ -79,7 +79,9 @@ func (k *BasicAuthKeyIsValid) basicAuthHeaderCredentials(w http.ResponseWriter,
 	logger := k.Logger().WithField("key", obfuscateKey(token))
 	if token == "" {
 		// No header value, fail
-		logger.Info("Attempted access with malformed header, no auth header found.")
+		if !k.Spec.BasicAuth.ExtractFromBody {
+			logger.Info("Attempted access with malformed header, no auth header found.")
```
dencoded: I think this message should be ERROR or WARN level (not info).
also, it seems to be extra have it here with reference to out auth extraction from body.
how about moving this log message to `ProcessRequest` and put this warning in log in appropriate situation ( `if k.Spec.BasicAuth.ExtractFromBody` is already there).

dencoded:
@matiasinsaurralde I left comments. also, CI is failing again.

mw_basic_auth.go
matiasinsaurralde: @dencoded Updated, can you review?

matiasinsaurralde: Think we don't need to check for `token` with the new change, this scenario will occur when `basicAuthHeaderCredentials` fails.

matiasinsaurralde:
The CI error isn't related to this change, looks like a linking issue:
```
/home/travis/.gimme/versions/go1.11.8.linux.amd64/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: /tmp/go-link-995235108/000016.o: unrecognized relocation (0x2a) in section `.text'
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
```

dencoded:
@matiasinsaurralde yes, coprocess build is failing

dencoded:
@matiasinsaurralde yes, coprocess build is failing

dencoded:
@matiasinsaurralde yes, coprocess build is failing

dencoded:
was closed by exident

buger:
From code point of view looks good. Coprocess builds seems to be unrelated, and looks like CI issue.

PR #2196, opened 2019-04-05
Make linter schema a separate json file

furkansenharputlu:
This PR moves linter schema to a json file. This will provide readability in json. Also, it changes package name from `lint` to `linter` to make it compatible with other cli package names.


cli/cli.go
```diff
@@ -73,7 +74,11 @@ func Init(version string, confPaths []string) {
 	// Linter:
 	lintCmd := app.Command("lint", "Runs a linter on Tyk configuration file")
 	lintCmd.Action(func(c *kingpin.ParseContext) error {
-		path, lines, err := lint.Run(lint.ConfSchema, confPaths)
+		confSchema, err := ioutil.ReadFile("cli/linter/schema.json")
```
matiasinsaurralde: Should we use a more specific path here? Just thinking about potential issues

furkansenharputlu: The working directory is root. I think we should say where the json is.

Issue #2203, opened 2019-04-10
gRPC fails when sending large messages


matiasinsaurralde:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When using a gRPC plugin and sending a message that's larger than 4 MB, the request fails.
4 MB is the default value for maximum length as mentioned [here](https://github.com/googleapis/google-cloud-node/issues/1991).

**What is the expected behavior?**
It should be possible to configure this parameter.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
All
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**

buger:
There was additional PR which fix small issue https://github.com/TykTechnologies/tyk/pull/2855

matiasinsaurralde:
@ilijabojanovic I have created this test program that you could use, instructions in the README:
https://github.com/matiasinsaurralde/grpc-large-msg

PR #2204, opened 2019-04-10
Make maximum gRPC message size configurable

matiasinsaurralde:
Potential fix for #2203 

This introduces two new parameters in `tyk.conf`:
```json
{
  "grpc_recv_max_size": 100000,
  "grpc_send_max_size": 100000
}
```
When no parameter is set (and both parameters equal `0`), we call `WithDefaultCallOptions` so default values are used.
The value is a number of bytes.

### DIFF
--- a/cli/linter/schema.json
+++ b/cli/linter/schema.json
@@ -257,6 +257,12 @@
         },
         "python_version": {
           "type": "string"
+        },
+        "grpc_recv_max_size": {
+          "type": "integer"
+        },
+        "grpc_send_max_size": {
+          "type": "integer"
         }
       }
     },
--- a/config/config.go
+++ b/config/config.go
@@ -214,6 +214,8 @@ type ServiceDiscoveryConf struct {
 type CoProcessConfig struct {
 	EnableCoProcess     bool   `json:"enable_coprocess"`
 	CoProcessGRPCServer string `json:"coprocess_grpc_server"`
+	GRPCRecvMaxSize     int    `json:"grpc_recv_max_size"`
+	GRPCSendMaxSize     int    `json:"grpc_send_max_size"`
 	PythonPathPrefix    string `json:"python_path_prefix"`
 	PythonVersion       string `json:"python_version"`
 }
--- a/coprocess/grpc/coprocess_grpc_test.go
+++ b/coprocess/grpc/coprocess_grpc_test.go
@@ -353,12 +353,23 @@ func TestGRPCDispatch(t *testing.T) {
 		})
 	})
 
-	t.Run("Post Hook with long message", func(t *testing.T) {
+	t.Run("Post Hook with allowed message length", func(t *testing.T) {
 		s := randStringBytes(20000000)
 		ts.Run(t, test.TestCase{
 			Path:    "/grpc-test-api-3/",
 			Method:  http.MethodGet,
 			Code:    http.StatusOK,
+			Headers: headers,
+			Data:    s,
+		})
+	})
+
+	t.Run("Post Hook with with unallowed message length", func(t *testing.T) {
+		s := randStringBytes(150000000)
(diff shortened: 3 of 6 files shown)


matiasinsaurralde:
It's important to note that the gRPC server needs to have the same value (or a higher value) than what's set in `tyk.conf` but this depends on the user's implementation. Just adding this for documentation.

matiasinsaurralde:
Updated schema

matiasinsaurralde:
CI is ok now 👍

coprocess_grpc_test.go
```diff
@@ -303,6 +309,16 @@ func TestGRPCDispatch(t *testing.T) {
 		})
 	})
 
+	t.Run("Post Hook with long message", func(t *testing.T) {
```
dencoded: I think we need to add one more test with message size `>grpcTestMaxSize` to check that reply is error (not `http.StatusOK`)

coprocess_grpc.go
```diff
@@ -85,13 +85,29 @@ func (d *GRPCDispatcher) Reload() {}
 // HandleMiddlewareCache isn't used by gRPC.
 func (d *GRPCDispatcher) HandleMiddlewareCache(b *apidef.BundleManifest, basePath string) {}
 
+func grpcCallOpts() grpc.DialOption {
+	recvSize := config.Global().CoProcessOptions.GRPCRecvMaxSize
+	sendSize := config.Global().CoProcessOptions.GRPCSendMaxSize
+	if recvSize > 0 && sendSize > 0 {
+		return grpc.WithDefaultCallOptions(
+			grpc.MaxCallRecvMsgSize(recvSize),
```
dencoded: looking at `grpc` package we see that both `MaxRecvMsgSizeCallOption` and `MaxSendMsgSizeCallOption` have comment `// This is an EXPERIMENTAL API.` which is kind of worrying. If we decide to go with this approach then might wanna keep an eye on it.

matiasinsaurralde: I think that this will really limit the use cases for our users (like the current implementation does). For users that don't want to override these settings, we use the defaults and compatibility remains the same.

coprocess_grpc.go
```diff
@@ -85,13 +85,29 @@ func (d *GRPCDispatcher) Reload() {}
 // HandleMiddlewareCache isn't used by gRPC.
 func (d *GRPCDispatcher) HandleMiddlewareCache(b *apidef.BundleManifest, basePath string) {}
 
+func grpcCallOpts() grpc.DialOption {
+	recvSize := config.Global().CoProcessOptions.GRPCRecvMaxSize
+	sendSize := config.Global().CoProcessOptions.GRPCSendMaxSize
+	if recvSize > 0 && sendSize > 0 {
```
asoorm: use case, override default SendSize without having to also set Receive Size?

buger:
@matiasinsaurralde pls resolve conflicts and respond to comments if needed.

matiasinsaurralde:
I've rebased and fixed a conflict here, will add what @asoorm mentioned now, think it's a valid use case

coprocess_grpc_test.go
matiasinsaurralde: Updated tests

coprocess_grpc.go
matiasinsaurralde: Added this

jeffrngu:
@matiasinsaurralde Thank you for working on this enhancement. Can we please target this change to go into release 8.3? Thnx!

christtyk:
@matiasinsaurralde I have a customer that would like an update on this issue, please let me know any details - will add to 2.9.3 backlog to review & prioritise.

matiasinsaurralde:
Rebased this!

matiasinsaurralde:
Fixed tests ✅

Issue #2208, opened 2019-04-13
gRPC: can only send message no larger than 2MB eventhough grpc limit is 4MB


jeffrngu:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When using a gRPC plugin and sending a message with size 2.5 MB, the request fails with grpc error saying the message exceeded 4MB limit

It looks like the effective payload is only 50% of the defined gRPC Max message size limit (default 4MB.)

Suspecting the new attribute MiniRequestObject.RawBody is what's taking up the other 50% of the max message size.

In the scenario tested below, the payload sent was only 2.5MB but the total size calculated by grpc was 5MB (perhaps 2.5 for Body and another 2.5 for RawBody)

time=“Apr 10 05:04:45” level=error msg=“rpc error: code = 8 desc = grpc: received message larger than max (5097416 vs. 4194304)”

time=“Apr 10 05:04:45” level=error msg=“Dispatch error” error=“rpc error: code = 8 desc = grpc: received message larger than max (5097416 vs. 4194304)”

**What is the expected behavior?**
We should be able to fully utilize the max message size (be able to send a message with 4MB in size

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Same scenario as https://github.com/TykTechnologies/tyk/issues/2203

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
observed in 2.7.4

matiasinsaurralde:
Thanks, the issue is related to these two body fields, we will find a solution.

matiasinsaurralde:
A potential solution could be a flag that avoids populating the `body` field and only makes use of `raw_body`.
The reason why we have two fields is described in #1724. `raw_body` is always populated and `body` is only used if the data contains is a valid UTF-8 string. When this happens we end up with duplicate data and this only leaves you with half of the actual gRPC message limit.

jeffrngu:
Having an extra field that takes up 50% of the payload just for accommodating format difference seems expensive. Is it possible to inspect the raw_body to see if it's UTF-8?

In the meantime, what is the work-around to fully utilize the 4MB limit if our data is UTF-8? Should we leave the body empty and only fill the raw_body?

matiasinsaurralde:
> Having an extra field that takes up 50% of the payload just for accommodating format difference seems expensive. Is it possible to inspect the raw_body to see if it's UTF-8?

[...]

PR #2213, opened 2019-04-15
Avoid duplicate request body in gRPC request object

matiasinsaurralde:
Potential fix for #2208, described [here](https://github.com/TykTechnologies/tyk/issues/2208#issuecomment-483262129).
Need to add tests.

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -271,6 +271,7 @@ type MiddlewareDefinition struct {
 	Name           string `bson:"name" json:"name"`
 	Path           string `bson:"path" json:"path"`
 	RequireSession bool   `bson:"require_session" json:"require_session"`
+	RawBodyOnly    bool   `bson:"raw_body_only" json:"raw_body_only"`
 }
 
 type MiddlewareIdExtractor struct {
--- a/gateway/api_loader.go
+++ b/gateway/api_loader.go
@@ -289,7 +289,7 @@ func processSpec(spec *APISpec, apisByListen map[string]int,
 			)
 		} else if mwDriver != apidef.OttoDriver {
 			coprocessLog.Debug("Registering coprocess middleware, hook name: ", obj.Name, "hook type: Pre", ", driver: ", mwDriver)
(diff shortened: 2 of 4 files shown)


coprocess.go
```diff
@@ -102,7 +102,7 @@ func (c *CoProcessor) ObjectFromRequest(r *http.Request) *coprocess.Object {
 	if r.Body != nil {
 		defer r.Body.Close()
 		miniRequestObject.RawBody, _ = ioutil.ReadAll(r.Body)
```
dencoded: worth to catch and log error here

matiasinsaurralde: Have extended `ObjectFromRequest` to also return errors 👍

matiasinsaurralde:
Fixed all the issues, the failing test seems to be`TestRateLimitForAPIAndRateLimitAndQuotaCheck` (not related)

matiasinsaurralde:
A note about this, when reconstructing the request object after the hook is called, we always take `raw_body`:
```go
// ObjectPostProcess does CoProcessObject post-processing (adding/removing headers or params, etc.).
func (c *CoProcessor) ObjectPostProcess(object *coprocess.Object, r *http.Request) {
        r.ContentLength = int64(len(object.Request.RawBody))
        r.Body = ioutil.NopCloser(bytes.NewReader(object.Request.RawBody))
```
I don't think that this behavior should be modified when`raw_body_only` flag is in use.
We should clarify in the docs that when a hook needs to modify the body, they should modify `raw_body`, perhaps we can mention that the `body` field is "read-only" (we don't have a way of enforcing this, as the bindings are generated by PB, but any hook that modifies `body` won't make any effect, `raw_body` will still be used).

matiasinsaurralde:
Would appreciate any comments on this ☝️ 
It's mostly about the docs.

matiasinsaurralde:
Fixed to match the new structures
Tests ok @dencoded 👍

dencoded:
👍 to go then

buger:
@matiasinsaurralde not quite sure how it works. How I can set `RawBodyOnly` option?  Can you provide an example?

matiasinsaurralde:
@buger this seems broken after the rebase, I will add additional tweaks and clarify the functionality

PR #2218, opened 2019-04-18
Read log browser response body if compressed

furkansenharputlu:
Fixes https://github.com/TykTechnologies/tyk-analytics/issues/1215

The solution is to read compressed body.

### DIFF
--- a/handler_success.go
+++ b/handler_success.go
@@ -193,6 +193,8 @@ func (s *SuccessHandler) RecordHit(r *http.Request, timing int64, code int, resp
 			// mw_redis_cache instead? is there a reason not
 			// to include that in the analytics?
 			if responseCopy != nil {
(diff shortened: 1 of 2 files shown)


handler_success.go
```diff
@@ -193,6 +193,8 @@ func (s *SuccessHandler) RecordHit(r *http.Request, timing int64, code int, resp
 			// mw_redis_cache instead? is there a reason not
 			// to include that in the analytics?
 			if responseCopy != nil {
+				responseCopy.Body = respBodyReader(r, responseCopy)
```
komalsukhani: You also need to set ContentLength header to 0 here as well

furkansenharputlu: It will be set in `respBodyReader` function so it is not necessary to set again.

komalsukhani: ok

PR #2230, opened 2019-04-25
Added support of WSDL import

komalsukhani:
User can now import WSDL file to create API definition. 
[#Reference](https://github.com/TykTechnologies/tyk-analytics/issues/1160)

**Supported features:**
1. WSDL file with multiple services
2. SOAP/SOAP12 binding
3. HTTP binding

**Generated API Definition will be as follows:**
1. API `name` is same as that of the service name. If WSDL contains multiple services, then the name of the first service in the file will be used.

2. `listen-path` of the API will be 

> "/service name/"

   If WSDL contains multiple services, then the name of the first service in the file will be used.

3. For each supported operation, an endpoint will be created. Endpoint will have the path

> /service_name/end_path

In case of **SOAP/SOAP12** binding, _end_path_ will be operation name.
In case of **HTTP** binding, _end_path_ will be value of `location` attribute of the operation. 

4. Each endpoint will use url-rewrite middleware to direct the request to the upstream. This was necessary to support multiple services

If service has multiple Ports type, name of the Port needs to be specified. If Port name is not provided then first Port of the service is used.

https://github.com/TykTechnologies/tyk/issues/2255

### DIFF
--- a/apidef/importer/wsdl.go
+++ b/apidef/importer/wsdl.go
@@ -3,7 +3,6 @@ package importer
 import (
 	"encoding/xml"
 	"errors"
-	"fmt"
 	"github.com/TykTechnologies/tyk/apidef"
 	uuid "github.com/satori/go.uuid"
 	"strings"
@@ -37,11 +36,11 @@ type WSDLAddress struct {
 }
 
 type WSDLBinding struct {
-	Name       string           `xml:"name,attr"`
-	Operations []*WSDLOperation `xml:"http://schemas.xmlsoap.org/wsdl/ operation"`
-	Protocol   string
-	Method     string
-	isProcess  bool
+	Name                string           `xml:"name,attr"`
+	Operations          []*WSDLOperation `xml:"http://schemas.xmlsoap.org/wsdl/ operation"`
+	Protocol            string
+	Method              string
+	isSupportedProtocol bool
 }
 
 type WSDLOperation struct {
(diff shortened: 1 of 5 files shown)


apidef/importer/wsdl.go
```diff
@@ -0,0 +1,430 @@
...
+									if attr.Name.Local == "verb" {
+										b.Method = attr.Value
+										break
+									}
+								}
+
+							}
+						default:
+							{
+								log.Debug("Unsupported binding protocol is used:", t.Name.Space, ":", t.Name.Local)
```
matiasinsaurralde: Perhaps use similar syntax to the other log lines, should be safe as these two `Name` fields are `string`:
```go
log.Debugf("Unsupported binding protocol is used: %s : %s", t.Name.Space, t.Name.Local)
```

apidef/importer/wsdl.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+
+	//Fetch protocol specific data
+	//If soap/soap12 is used, set Method to POST
+	//If http is used, get value of verb attribute
+	//If any other protocol is used, then skip
+	for {
+		tok, err := d.Token()
+		if err != nil {
+			log.Error("Error will parsing wsdl file: ", err)
```
matiasinsaurralde:
```go
log.Error("Error while parsing WSDL file:", err)
```

buger:
It definitely misses some tests, when it reads example WSDL file, and compare converted version with expected.

buger:
@komalsukhani pls create UI related ticket

komalsukhani:
@buger Already created. Here is the link https://github.com/TykTechnologies/tyk-analytics-ui/issues/1182

Issue #2236, opened 2019-04-27
Async session writes cause goroutine leak


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
On every API reload a number of goroutines significantly grows, linearly with API count. It happens because for each API we start own session update workers pool. 
**What is the expected behavior?**
Reload with the same APIs should not affect goroutine count. 
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Enable `optimisations_use_async_session_write`, and watch how the number of goroutines grows with each reload.
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.8

PR #1757, opened 2018-06-06
Improve async session updates 

joshblakeley:
Fixes #1758 

Current solution for session updates when async is enabled is just to spawn a new goroutine per redis operation. 

This change allows setting a configurable pool of workers to handle the session updates - with the slower parts of the session update such as json marshalling moved into the pool. Default value if unset is currently 15 and worker pool will gracefully exit on gateway shutdown.

Also changed the dummySessionManager in policy_test.go pass by ref to stop go vet failing because of passed mutex.

PR #2237, opened 2019-04-27
Fix goroutine leak with async session update

buger:
Session update worker pool was starting for every "DefaultSessionStore" object. And it was happening for every API initialization. So if you have 1000 APIs, and your worker pool is 100, each API reload will trigger at least 100k goroutines.

This change maintains a single worker pool which initialized only once.

The issue itself was introduced by https://github.com/TykTechnologies/tyk/pull/1757

Fix https://github.com/TykTechnologies/tyk/issues/2236

### DIFF
--- a/auth_manager.go
+++ b/auth_manager.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"strings"
 	"sync"
-	"sync/atomic"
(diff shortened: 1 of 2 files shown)


buger:
@joshblakeley ^

Issue #2238, opened 2019-04-27
Leak in basic auth middleware


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
On every API load, for each API, it starts the new gocache instance for BasicAuth middleware. gocache by itself starts own pool of goroutines, which never clean up.
**What is the expected behavior?**
Basic auth cache should be created only once.
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Issue visible even if you do not use basic auth caching. Just reload few times.
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
2.8

PR #2239, opened 2019-04-27
Fix basic auth caching

buger:
Basic auth cache was initialized for every API on every API reload.
go-cache by itself consume quite lot of resources, when started in such amounts.

Fix https://github.com/TykTechnologies/tyk/issues/2238


buger:
@asoorm ^

Issue #2268, opened 2019-05-08
UrlRewrite/Method transform doesn't work if caching is enabled


komalsukhani:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
If you enable URL Rewrite/Method transform and Cache plugin for an endpoint, URL Rewrite/Method transform doesn't work.

**What is the expected behavior?**
URL Rewrite/Method transform plugin should work

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**

API Defintion

```json
{
  "api_model": {},
  "api_definition": {
    "api_id": "10a5aa802e0346a653eeb71f8e9d9234",
    "jwt_issued_at_validation_skew": 0,
    "upstream_certificates": {},
    "use_keyless": true,
    "enable_coprocess_auth": false,
    "custom_middleware": {
      "pre": [],
      "post": [],
      "post_key_auth": [],
      "auth_check": {
        "name": "",
        "path": "",
        "require_session": false
      },
      "response": [],
      "driver": "",
      "id_extractor": {
        "extract_from": "",
        "extract_with": "",
        "extractor_config": {}
      }
    },
    "disable_quota": false,
    "custom_middleware_bundle": "",
    "cache_options": {
      "cache_timeout": 60,
      "enable_cache": true,
      "cache_all_safe_requests": false,
      "cache_response_codes": [],
      "enable_upstream_cache_control": false,
      "cache_control_ttl_header": ""
    },
    "enable_ip_blacklisting": false,
    "tag_headers": [],
    "pinned_public_keys": {},
    "expire_analytics_after": 0,
    "domain": "",
    "openid_options": {
      "providers": [],
      "segregate_by_client": false
    },
    "active": true,
    "jwt_expires_at_validation_skew": 0,
    "config_data": {},
    "notifications": {
      "shared_secret": "",
      "oauth_on_keychange_url": ""
    },
    "auth": {
      "auth_header_name": "",
      "signature": {
        "algorithm": "",
        "header": "",
        "secret": "",
        "allowed_clock_skew": 0,
        "error_code": 0,
        "error_message": ""
      },
      "validate_signature": false,
      "use_certificate": false
    },
    "check_host_against_uptime_tests": false,
    "auth_provider": {
      "name": "",
      "storage_engine": "",
      "meta": {}
    },
    "blacklisted_ips": [],
    "hmac_allowed_clock_skew": -1,
[...]

PR #2270, opened 2019-05-09
consistent mw version order

### LINKED ISSUE #2317: Inconsistent middlware execution in case of open and secured API
Bug

If caching and URL rewrite plugin are enabled, for an endpoint, open keyless API and secured API behave differently.

_In case of open keyless API_ : URL Rewrite is never executed

_In case of protected API_ : Caching is never enabled

Execution of middlewares should be consistent

master

asoorm:
- Consistent MW execution order - previously, the keyless APIs executed in a different order to protected APIs.
- Remove duplicate code - When creating new plugins - you no longer need to create it in 2 places
- Resolves #2317

### DIFF
--- a/gateway/api_loader.go
+++ b/gateway/api_loader.go
@@ -267,85 +267,37 @@ func processSpec(spec *APISpec, apisByListen map[string]int,
 	cacheStore.Connect()
 
 	var chain http.Handler
+	var chainArray []alice.Constructor
+	var authArray []alice.Constructor
 
 	if spec.UseKeylessAccess {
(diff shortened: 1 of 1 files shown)


komalsukhani:
#2268 will NOT get resolved.

URL Rewrites changes r.URL. 
Every middleware, before executing, checks if it is enabled by checking the value of r.URL
In URL rewrite case, cache plugin will be enabled for the path before URL rewrite, while cache middleware will be checking it against new traget

komalsukhani:
Should fix #2317

Issue #1505, opened 2018-03-02
tyk.conf - when using the dashboard, shouldnt require to set the its uri twice in`policies` and `db_app_conf_options`


letzya:
**Do you want to request a *feature* or report a *bug*?**
Feature, could cause behavioral bugs.

**What is the current behavior?**
Conf file requires to define the dashboard in 2 places:

>      "policies": {
>          "policy_source": "service",
>           "**policy_connection_string**": "http://my-tyk-instance.com:3000",
>           "policy_record_name": "",
>           "allow_explicit_policy_id": false
>      },
>       "use_db_app_configs": true,
>       "db_app_conf_options": {
>           "**connection_string**": "http://my-tyk-instance.com:3000",
>           "node_is_segmented": false,
>           "tags": null
>       },

**What is the expected behavior?**
db_app_conf_options should be enough and _policy_connection_string_ can get it's value from there if  _"policy_source" is "service"._
**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
the regular set up for GW that is not running headless.

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
All (2.5.2 at the moment)

letzya:
Put as Priority: Low since it's working as it is (though it can make our users' lives easier and less error prone)

l3ender:
Commenting as to not let this drop. These types of fixes make the product better and more usable in the long term.

andyo-tyk:
This fix was merged to master in 2019 so I'm closing the ticket.

@l3ender if you think there's still something here please don't hesitate to reopen/comment and we'll look into it again.

PR #2276, opened 2019-05-10
if dash is enabled, auto-configure policy record

asoorm:
Addresses #1505


main.go
```diff
@@ -833,6 +833,13 @@ func initialiseSystem() error {
 		if globalConf.PIDFileLocation == "" {
 			globalConf.PIDFileLocation = "/var/run/tyk/tyk-gateway.pid"
 		}
+		if globalConf.UseDBAppConfigs && globalConf.Policies.PolicySource != "service" {
+			globalConf.Policies.PolicySource = "service"
+			globalConf.Policies.PolicySource = globalConf.DBAppConfOptions.ConnectionString
```
dencoded: looks like it is supposed to be `globalConf.Policies.PolicyConnectionString `

main.go
```diff
@@ -833,6 +833,13 @@ func initialiseSystem() error {
 		if globalConf.PIDFileLocation == "" {
 			globalConf.PIDFileLocation = "/var/run/tyk/tyk-gateway.pid"
 		}
+		if globalConf.UseDBAppConfigs && globalConf.Policies.PolicySource != "service" {
+			globalConf.Policies.PolicySource = "service"
```
dencoded: might worth to introduce constants for for string values like `"service"` and `"tyk_policies"`, maybe in config package (not sure about that)

main.go
asoorm: good spot

buger:
Merged manually

PR #2277, opened 2019-05-10
[TT-3154] Strip listen path when it is regex too

### LINKED ISSUE #2249: Stripping listen paths does not work if listen path contains regexp
bug
If I define listen path like `/{_:.*}/listen`, it works as expected, except stripping listen paths. 
At the moment it can strip only "static" values, and not handle regexps. 
It should detect if regexp is defined, and strip the path using regexp.
See above
All till 2.8

furkansenharputlu:
Fixes #2249

### DIFF
--- a/gateway/handler_success.go
+++ b/gateway/handler_success.go
@@ -10,10 +10,13 @@ import (
 	"strings"
 	"time"
 
+	"github.com/TykTechnologies/tyk/regexp"
+	"github.com/gorilla/mux"
+	cache "github.com/pmylund/go-cache"
+
 	"github.com/TykTechnologies/tyk/config"
 	"github.com/TykTechnologies/tyk/request"
 	"github.com/TykTechnologies/tyk/user"
-	cache "github.com/pmylund/go-cache"
 )
 
 // Enums for keys to be stored in a session context - this is how gorilla expects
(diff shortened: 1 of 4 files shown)


handler_error.go
```diff
@@ -94,7 +94,11 @@ func (e *ErrorHandler) HandleError(w http.ResponseWriter, r *http.Request, errMs
 		}
 
 		if e.Spec.Proxy.StripListenPath {
-			r.URL.Path = strings.Replace(r.URL.Path, e.Spec.Proxy.ListenPath, "", 1)
```
buger: Are you sure that RawPath should be changed here?
If so does it mean that you solved some bug?

handler_success.go
```diff
@@ -332,6 +338,21 @@ func (s *SuccessHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) *http
 	return nil
 }
 
+func stripListenPath(r *http.Request, listenPath string) error {
+	tmp := new(mux.Route).PathPrefix(listenPath)
```
buger: Not quite sure if understand how it works, can you add a comment?
Also, I feel like Regexp can be compiled only once, and cached somewhere in API def.

gateway/handler_success.go
```diff
@@ -332,6 +338,21 @@ func (s *SuccessHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) *http
...
 }
 
+func stripListenPath(r *http.Request, listenPath string) error {
+	tmp := new(mux.Route).PathPrefix(listenPath)
+	s, err := tmp.GetPathRegexp()
+	if err != nil {
+		return err
+	}
+
+	reg := regexp.MustCompile(s)
```
matiasinsaurralde: Not sure about this particular scenario but our `regexp` package might be useful here? https://github.com/TykTechnologies/tyk/blob/master/regexp/regexp.go

cc @dencoded

buger: It is already used, see above

buger:
@furkansenharputlu this can be very slow. We at least need to store cached regexp variable in API def package.

furkansenharputlu:
Not stale

lghiur:
This issue has been addressed under this PR https://github.com/TykTechnologies/tyk/pull/4104

Issue #2271, opened 2019-05-10
Panic during key update


buger:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
Updating the key can cause panic like this:

```
fatal error: concurrent map read and map writetime="May  9 14:23:14" level=info msg="Reset quota for key." inbound-key="****2877" key=quota-705ee35e 


goroutine 3560 [running]:
runtime.throw(0xf1ef5b, 0x21)
	/usr/local/go/src/runtime/panic.go:616 +0x81 fp=0xc4241d4988 sp=0xc4241d4968 pc=0x42cb11
runtime.mapaccess2(0xdb7ba0, 0xc422b02240, 0xc4203cccd0, 0x3, 0xd)
	/usr/local/go/src/runtime/hashmap.go:409 +0x225 fp=0xc4241d49d0 sp=0xc4241d4988 pc=0x409585
reflect.mapaccess(0xdb7ba0, 0xc422b02240, 0xc4203cccd0, 0xc422b02240)
	/usr/local/go/src/runtime/hashmap.go:1195 +0x3f fp=0xc4241d4a08 sp=0xc4241d49d0 pc=0x40b83f
reflect.Value.MapIndex(0xdb7ba0, 0xc422c3f8b8, 0x195, 0xd67ac0, 0xc4203cccd0, 0x98, 0x20, 0x40, 0x10000000000000a)
	/usr/local/go/src/reflect/value.go:1081 +0x126 fp=0xc4241d4a90 sp=0xc4241d4a08 pc=0x4bbb26
encoding/json.(*mapEncoder).encode(0xc42000e2c8, 0xc4200f3130, 0xdb7ba0, 0xc422c3f8b8, 0x195, 0x100)
	/usr/local/go/src/encoding/json/encode.go:684 +0x40d fp=0xc4241d4bf0 sp=0xc4241d4a90 pc=0x58bd4d
encoding/json.(*mapEncoder).(encoding/json.encode)-fm(0xc4200f3130, 0xdb7ba0, 0xc422c3f8b8, 0x195, 0x100)
	/usr/local/go/src/encoding/json/encode.go:700 +0x64 fp=0xc4241d4c30 sp=0xc4241d4bf0 pc=0x5962f4
encoding/json.(*structEncoder).encode(0xc422259980, 0xc4200f3130, 0xee9900, 0xc422c3f860, 0x199, 0x100)
[...]

PR #2278, opened 2019-05-11
send in async update channel a clone of SessionState

dencoded:
added changes for https://github.com/TykTechnologies/tyk/issues/2271

I had hard times to replicate this issue but I clearly see reasons of that panic:  by the time async-update worker does session JSON marshalling the session might being updated.
So I added changes to send deep-copy of session as part of async update message.


buger:
Frankly, I'm really skeptical about this async option in general and have a feeling that we should just remove it. We just need confirmation that it does not give any performance boost. 

@asoorm during your performance testing have you tried this variable, and if not, and you have env, can you try it?

asoorm:
@buger 👍 will check it

PR #2283, opened 2019-05-15
coprocess: extend ReturnOverrides headers usage (fixes #2282)

### LINKED ISSUE #2282: ReturnOverrides does not work for response code > 400
Looks like a bug to me, don't know if it's designed this way

If I use `ReturnOverrides` with `status` > 400 (let's say 401), then the `headers` and `error` overrides are ignored.

`headers` and `error` in `ReturnOverrides` should be used no matter what the response code is.

This is the code causing the issue:
https://github.com/matiasinsaurralde/tyk/blob/master/coprocess.go#L293

matiasinsaurralde:
Fixes #2282.

This was a case we didn't cover while building this feature.

It was available in different `ReturnOverrides` scenarios.


matiasinsaurralde:
Updated this to fit structure in #2172

PR #2284, opened 2019-05-16
Create testutil.go

furkansenharputlu:
This PR creates `testutil.go`. The purpose is to be able to import test functions in other projects.

For example, in dashboard, this PR provides us to write the following test:

```go
func TestAPIContext_RevokeDeveloperKey(t *testing.T) {
	g := gateway.StartTest()
	defer g.Close()

	Tyk.(*TykApi).Port = strconv.Itoa(g.GlobalConfig.ListenPort)
	sessionID, _ := UserSessions.CreateAPISession(&testUser)

	ts := newTykTestServer(tykTestServerConfig{accessKey: sessionID})
	defer ts.Close()

	api := gateway.BuildAndLoadAPI(func(spec *gateway.APISpec) {
		spec.Proxy.ListenPath = "/"
		spec.UseKeylessAccess = false
	})[0]

	key := gateway.CreateSession(func(s *user.SessionState) {
		s.AccessRights = map[string]user.AccessDefinition{api.APIID: {
			APIID: api.APIID,
		}}
	})

	developer := createSampleDeveloper(func(developer *PortalDeveloper) {
		developer.OrgId = testOrg.Id.Hex()
		developer.ApiKeys = map[string]string{
			api.APIID: key,
		}
	})

	authHeaders := map[string]string{
		"authorization": key,
	}

	g.Run(t, test.TestCase{Path: "/", Headers: authHeaders, Code: http.StatusOK})

	headers := map[string]string{"authorization": testUser.AccessKey}
	revokePath := fmt.Sprintf("/api/portal/developers/key/%s/%s/%s", api.APIID, key, developer.Id.Hex())

	_, _ = ts.Run(t, []test.TestCase{
		{Method: http.MethodDelete, Headers: headers, Path: revokePath, Code: http.StatusOK, BodyMatch: "User key revoked"},
	}...)

	g.Run(t, test.TestCase{Path: "/", Headers: authHeaders, Code: http.StatusForbidden})
}
```

### DIFF
--- a/gateway/testutil.go
+++ b/gateway/testutil.go
@@ -7,8 +7,10 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
-	"github.com/TykTechnologies/tyk/cli"
+
+
 	"golang.org/x/net/context"
+
 	"io"
 	"io/ioutil"
 	"math/rand"
@@ -23,11 +25,13 @@ import (
 	"testing"
 	"time"
 
-	"github.com/dgrijalva/jwt-go"
+	"github.com/TykTechnologies/tyk/cli"
+
+	jwt "github.com/dgrijalva/jwt-go"
 	"github.com/garyburd/redigo/redis"
 	"github.com/gorilla/mux"
 	"github.com/gorilla/websocket"
-	"github.com/satori/go.uuid"
+	uuid "github.com/satori/go.uuid"
 
 	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/config"
@@ -431,7 +435,7 @@ func firstVals(vals map[string][]string) map[string]string {
 	return m
 }
 
-type tykTestServerConfig struct {
+type MockConfig struct {
 	sepatateControlAPI bool
 	delay              time.Duration
 	hotReload          bool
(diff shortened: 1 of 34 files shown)


gateway/testutil.go
```diff
@@ -278,25 +429,25 @@ func firstVals(vals map[string][]string) map[string]string {
...
+type MockConfig struct {
 	sepatateControlAPI bool
 	delay              time.Duration
 	hotReload          bool
 	overrideDefaults   bool
 	coprocessConfig    config.CoProcessConfig
 }
 
-type tykTestServer struct {
+type Mock struct {
```
dencoded: I think `Mock` is not the best name because here it doesn't mocking anything, in fact it runs real network listener with tyk

dencoded:
We need better explanation why do we need to write in dashboard tests like this. I think provided piece of code doesn't show the idea.

furkansenharputlu:
@dencoded I added a more meaningful test example.

furkansenharputlu:
@dencoded In general, you're right. We need to update dashboard test util in the way they use gateway's API for future. However, for this test, even if everything isn't integrated, there is stil integration, the revoke key operation is fired by dashboard API and gateway listens it.

furkansenharputlu:
In the next steps of refactoring, I will update dashboard's test util step by step. Every operation will be done by APIs.

PR #2292, opened 2019-05-19
PoC/proposal: support JSON Patch requests to update apidef

asoorm:
some linting:
- moving api handlers to own struct.
- allowing users to delete cache of specific api. more logical to target `DELETE /apis/{apiId}/cache`
- moving org handlers to own struct.
- reduced cyclomatic complexity - removed a lot of if/else/switch statements.

implement JSON Patch for api handlers. This allows us to perform partial updates to an API definition object.
```
curl -X PATCH http://localhost:8080/tyk/apis/httpbin -H 'x-tyk-authorization: aaa' \
  -d '[
  {"op": "replace", "path": "/use_keyless", "value": false},
  {"op": "replace", "path": "/auth/auth_header_name", "value": "Apikey"}
]'
```

It negates the need for a client to "parse" the "GET /apis/httpbin" response, modify the payload, then PUT/POST entire object back.

If the above is seen as a good idea, proposal to implement json patch for all Tyk APIs.


buger:
In general I like the idea. I would be very careful with adding too much OOP and abstractions: it is a powerful concepts which should be used only if you really need too, otherwise code readability may drop. For example `HanderAPI` struct do not have any state, and can be just replaced by functions like: `registerAPIRoutes`, `updateAPIHandler`, `patchAPIHandler`. 

Also `repository/cache/cache.go` seems like overkill. Better move this code back.

Also worth moving all API specific and Org Keys specific functions from api.go to the new related files.

PR #2294, opened 2019-05-20
Notify KeySpaceUpdateNotification when a key deleted or updated

### LINKED ISSUE #2293: Revoking developer key doesn't delete cache
Bug

When a user key is revoked, it continues to work until cache expires

Revoking operation should flush cache.

1. Set `"disable_cached_session_state": false` in `tyk.conf`
2. Create a developer key
3. Make a request with that key
4. Revoke the key
5. Make another request in 10 secs.(Should fail but success)

Master

furkansenharputlu:
Fixes #2293


buger:
I know existing GW instance eventually will get the message too, but lets do a local cache cleanup directly as well. Additionally can you resolve the same issue with updating the key?

PR #2307, opened 2019-05-26
ux: gateway logs

asoorm:
Fixing some of the startup log order
Organizing some of the startup service checks by moving into the checkup package

### DIFF
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -825,6 +825,8 @@ func initialiseSystem() error {
 		mainLog.Debug("No configuration file defined, will try to use default (tyk.conf)")
 	}
 
+	mainLog.Infof("Tyk API Gateway %s", VERSION)
+
 	if !runningTests {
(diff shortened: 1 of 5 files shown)


gateway/server.go
```diff
@@ -993,11 +992,14 @@ func Start() {
 		mainLog.Warn("The control_api_port should be changed for production")
 	}
 
-	start()
+	checkup.FileDescriptors()
```
letzya: I would put everything in one “checkup.Run” public function. This way every time we add something wouldn’t require an update in here and all the checkups would be incapsulated inside “Run”.

buger: In general, agree. It will require modifying code only in one place if we want to add similar warnings in future.

buger: In general, agree. It will require modifying code only in one place if we want to add similar warnings in future.

asoorm:
@buger PTAL

PR #2310, opened 2019-05-27
Fix JWT expiration

### LINKED ISSUE #2304: JWT: expiration time on policy conflicts with jwt token exp time
BUG
JWT tokens does not respect expiration time on policy
Tokens assigned to JWT tokens should respect policy expiration time
If policy expiration time is 0 (never expires) and jwt token exp time is 3600, internal token will use jwt exp time.

Seems regression introduced with this fix:
https://github.com/TykTechnologies/tyk/issues/1805
Release 2.8 / Master

buger:
With code added in #1849 we made JWT keys set expiration on keys (pick bigger value either policy or token value)

However, this change introduce a bug in situation when policy is set to
"Never Expire" (or just lower then JWT value): if JWT key expire its expiration value gets hold in session value and never gets overridden, even if JWT token gets re-issued.

The bug itself consists of two parts.
1) Code which should update the session for already existing JWT token
gets run ONLY if policy ID has changed
2) Code above updates only local session cache, but not propagate the change
to the Redis

This PR fixes both 1) and 2), and adds test covering this issue.

Fix #2304

### DIFF
--- a/gateway/mw_jwt.go
+++ b/gateway/mw_jwt.go
@@ -286,6 +286,7 @@ func (k *JWTMiddleware) processCentralisedJWT(r *http.Request, token *jwt.Token)
 	data := []byte(baseFieldData)
 	keyID := fmt.Sprintf("%x", md5.Sum(data))
 	sessionID := generateToken(k.Spec.OrgID, keyID)
+	updateSession := false
 
 	k.Logger().Debug("JWT Temporary session ID is: ", sessionID)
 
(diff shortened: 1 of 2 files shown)


buger:
Steps to replicate bug:
1) Create a policy with "no expire"
2) Call API with short-lived JWT and wait for it to expire
3) After the first token expired, generate fresh and call API again, it will say "Key expired"

PR #2318, opened 2019-05-29
Add JWTDefaultPolicies field and default policy feature

### LINKED ISSUE #2009: Allow JWT auth method to verify the signature but not enforce the policy
Feature

In order to use this method, you must have permission to change the JWT and **inject** the policy id to it.
The problem is that JWTs are created and **signed** with a private key by the IdP, so injecting more info could be a problem to the API owner.

To be able to add a public key or shared key to the api def without choosing a specific policy.

Please note that this feature is not relevant to individual keys since in this case, you would need to update the JWT anyway.
 
2.7

furkansenharputlu:
This PR adds `JWTDefaultPolicies` to api definition. This means in `JWT` authentication, a user can apply some default policies not having conflict to the session created by `JWT` authentication.

Note: Error messages start with lowercase. I refactored the messages in the related part. There should be a big refactor to make error messages compatible with Go conventions.

Fixes #2009

### DIFF
--- a/gateway/mw_jwt.go
+++ b/gateway/mw_jwt.go
@@ -206,6 +206,10 @@ func (k *JWTMiddleware) getBasePolicyID(r *http.Request, claims jwt.MapClaims) (
 		return pols[0], true
 	}
 
+	if len(k.Spec.JWTDefaultPolicies) > 0 {
+		return k.Spec.JWTDefaultPolicies[0], true
+	}
+
 	return "", false
 }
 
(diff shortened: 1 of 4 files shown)


gateway/mw_jwt.go
```diff
@@ -290,23 +294,41 @@ func (k *JWTMiddleware) processCentralisedJWT(r *http.Request, token *jwt.Token)
...
-			return errors.New("Key not authorized: no matching policy found"), http.StatusForbidden
+			return errors.New("key not authorized: no matching policy found"), http.StatusForbidden
 		}
 
-		newSession, err := generateSessionFromPolicy(basePolicyID,
+		session, err = generateSessionFromPolicy(basePolicyID,
 			k.Spec.OrgID,
 			true)
 
+		for _, pol := range k.Spec.JWTDefaultPolicies {
```
buger: Default policy should be assigned ONLY if it not found in JWT token

PR #2321, opened 2019-05-29
Support response middleware when using rich plugins

matiasinsaurralde:
Added new `response` hook type

```
{
  "file_list": [
  ],
  "custom_middleware": {
    "pre": [
      {
        "name": "MyPreHook"
      }
    ],
    "response": [{
      "name": "MyResponseHook"
    }]
  }
}
```

```
from tyk.decorators import *
@Hook
def MyResponseHook(request, response, session, metadata, spec):
  response.raw_body = b'newbody'
  return response
```


l3ender:
Thanks for the initial work on this, @matiasinsaurralde.

We're very interested in these changes. Is there a way we'd be able to pull them in and test in our environment?

Also, is there any way in which we could support these efforts? I'd be happy to help if you will accept PRs and will share some details on the current state of PR and necessary efforts left.

Thanks!

matiasinsaurralde:
Hi @l3ender, feel free to try this branch, have you previously used pre/post hooks?
This basically works in the same way, the main difference is that there's a new field called `response` in the manifest file, like this:
```json
{
  "file_list": [
  ],
  "custom_middleware": {
    "pre": [
      {
        "name": "MyPreHook"
      }
    ],
    "response": [{
      "name": "MyResponseHook"
    }]
}
```
In this scenario `MyResponseHook` will be called after Tyk proxies your request to the upstream API, and the object that your hook receives will include the response data too.
I'm planning to incorporate other changes like #2213 before this gets merged, also extend tests and polish some code.

Best.

l3ender:
Thanks for the detail @matiasinsaurralde.

We have other middleware that we're using, so I can configure the `response` one too. But how do I pull the changes you've made to the gateway itself so that the middleware is supported? Do I need to copy your modifications to an installation directory, or perform some sort of compilation that will then get used to upgrade our install to this beta version?

Ideally, being able to pull a pre-release [Docker image](https://hub.docker.com/r/tykio/tyk-gateway/builds) would greatly facilitate being able to trial/test new features.

Thanks!

matiasinsaurralde:
Rebased and updated this to incorporate #2213, I'm doing some additional changes related to errors during response middleware and it will be ready.

l3ender:
How is this feature coming along? Any ETA on when it will be merged and released?

matiasinsaurralde:
Rebased and fixed conflicts

l3ender:
Hi @matiasinsaurralde, I was wondering if you could elaborate on what type of plugin that the response middleware will be supported under. Will the feature include Javascript and Go plugins? I'm hoping the answer is yes, but looking through the PR I have doubt. If the feature doesn't currently support JS/Go, what can be done to have them added as well to ensure consistency across plugins?

Thanks again for these changes--looking forward to when they are released.

gateway/server.go
```diff
@@ -573,10 +574,16 @@ func loadCustomMiddleware(spec *APISpec) ([]string, apidef.MiddlewareDefinition,
 		mwPostKeyAuthFuncs = append(mwPostKeyAuthFuncs, mwObj)
 	}
 
-	return mwPaths, mwAuthCheckFunc, mwPreFuncs, mwPostFuncs, mwPostKeyAuthFuncs, mwDriver
+	// Load response hooks
```
letzya: We need to check also the setting of mwPaths for response hook

letzya: would be great to have tests

PR #2326, opened 2019-06-03
basic auth cache - thundering heard

asoorm:
Addition to https://github.com/TykTechnologies/tyk/pull/2323


buger:
Show we at least extend it with user name details?

PR #2328, opened 2019-06-04
Request signing middleware

### LINKED ISSUE #2234: New MW - hmac signature to the message between tyk and the upstreem
feature
Doesn't exists. 
At the moment if your upstream cannot do mTLS, they can't properlly verify the client.  This solution will enable upstream to verify requests coming from Tyk if they will implement the code to verify this signature.

Tyk is doing a similar thing when verifying the signature in the request, as described [here](https://tyk.io/docs/security/your-apis/hmac-signatures/) (using [draft 5](https://tools.ietf.org/html/draft-cavage-http-signatures-05) ).
We can possiblly add support to sign the request that is coming out of tyk after all the other MW have been executed.
Special cases: 
- Post plugin - possiblly run this signature after running the plugin.
- Virtual endpoint - possiblly not working at all along with this MW.

Note:
- This is the latest [draft](https://tools.ietf.org/html/draft-cavage-http-signatures-10#section-4.1.1), but it has expired last year.
- Later stage - consider supporting RSA based signatures to both request and response.

2.8

komalsukhani:
The feature is implemented using [Draft 10](https://tools.ietf.org/html/draft-cavage-http-signatures-10)

`(request-target)` and all the headers of the request will be used for generating signature string. 
If request doesn't contain `Date` header, middleware will add one as it is required according to above draft.

A new config option `request_signing` is added in API Definition to enable/disable request signing. It has following format

```json
"request_signing": {
  "is_enabled": true,
  "key": "xxxx",
  "key_id": "1",
  "algorithm": "hmac-sha256"
}
```
Following algorithms are supported:
1. `hmac-sha1`
2. `hmac-sha256`
3. `hmac-sha384`
4. `hmac-sha512`

Fixes #2234

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -499,7 +499,7 @@ type BundleManifest struct {
 
 type RequestSigningMeta struct {
 	IsEnabled bool   `bson:"is_enabled" json:"is_enabled"`
-	Key       string `bson:"key" json:"key"`
+	Secret    string `bson:"secret" json:"secret"`
 	KeyId     string `bson:"key_id", json:"key_id"`
 	Algorithm string `bson:"algorithm", json:"algorithm"`
 }
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -500,8 +500,8 @@ type BundleManifest struct {
 type RequestSigningMeta struct {
 	IsEnabled bool   `bson:"is_enabled" json:"is_enabled"`
 	Secret    string `bson:"secret" json:"secret"`
-	KeyId     string `bson:"key_id", json:"key_id"`
-	Algorithm string `bson:"algorithm", json:"algorithm"`
+	KeyId     string `bson:"key_id" json:"key_id"`
+	Algorithm string `bson:"algorithm" json:"algorithm"`
 }
 
 // Clean will URL encode map[string]struct variables for saving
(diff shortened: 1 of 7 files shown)


gateway/mw_request_signing_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/TykTechnologies/tyk/test"
+	"github.com/TykTechnologies/tyk/user"
+)
+
+var algoList = []string{"hmac-sha1", "hmac-sha256", "hmac-sha384", "hmac-sha512"}
```
letzya: @komalsukhani can we have a fixed array

buger:
@komalsukhani do no see that much reason specify KeyID as link here. 
I would rather keep "keyID" as string field, and will add one more config option to specify the secret. 

However maybe I missing smth?

apidef/api_definitions.go
```diff
@@ -488,6 +489,13 @@ type BundleManifest struct {
 	Signature        string            `bson:"signature" json:"signature"`
 }
 
+type RequestSigningMeta struct {
+	IsEnabled bool   `bson:"is_enabled" json:"is_enabled"`
+	Key       string `bson:"key" json:"key"`
+	KeyId     string `bson:"key_id", json:"key_id"`
```
komalsukhani: @buger I am using string type for keyId

PR #2334, opened 2019-06-10
opentracing support 

gernest:
Address https://github.com/TykTechnologies/product/issues/187

The whole point behind tracing is tracking a request as it is being propagated
across process boundaries, so we have two kinds of traces that are important.

- [x] The gateway api . This is endpoints under the `/tyk` endpoints.
- [x] The proxy api. This is all loaded specs by the gateway.

The level of details and information covered varies between those two because
of different requirements.

These are servers where the collected traces are sent.

- [x]  [appdash](https://github.com/sourcegraph/appdash)
- [x] [jaeger](https://www.jaegertracing.io/)

This is only PoC . So it is not fine tuned to handle all cases. We are only
tracing the amount of time taken to complete each request/response roundtrip.
This includes all middlewares applied in the process (for the request and
response).

TODO: write more about this as I grok the gateway internals

To configure tracing, a new setting `tracing` must be added to the tyk
configuration. The configuration options are like this,

```json
{
    ... the rest of tyk config

    "tracing":{
        "name":"appdash",
        "enabled":true,
        "options":{
            "conn":"appdash:7701"
        }
    }
}
```

 - `name` is the name of the tracer, for now only `appdash` is implemented for
 testing. `jaeger` or `zipkin` will be nice for production.
- `enabled` : when false then tracing will be disabled.
 - `options`: key/value pairs which are used to configure the tracer in our
 example we set `conn` which is needed for the appdash remote collector.

## Questions to address

- [ ] Should we be able to apply tracing on a per-path basis within API definition
- [x]  Should the gateway itself send all metrics to open tracing

Opentracing is just a standard, I think the gateway should record the traces and leave options on where the traces are collected. After my research I think `jaeger` looks pretty solid for trace collection and visualisation. 
In this implementation two collectors are implemented( [appdash]()  for development and [jaeger]() for production ) with room to add support for more. The idea is to allow the gateway to be configured for enabling and disabling traces based on configuration.

- [x]  How to configure trace sampling

Since I picker jaeger for production , here is what jaeger offers for sampling. Note that we sample traces in the tyk `tracing` configuration property.

- *constant* : this can sample all or none
- *probabilistic*: uses probability to decide what to sample
- *rate limiting*: decides how many samples per second ( this is very handy to reduce loads on high volume paths)
- *remote*: this allows external service to offer sampling configuration( probably the best option when we are dealing with service meshes I think)

This is an example tyk tracing config using constant sampler

```json
{
    ... the rest of tyk config
    "tracing": {
        "name": "jaeger",
        "enabled": true,
        "options": {
            "serviceName": "tyk-gateway",
            "disabled": false,
            "rpc_metrics": false,
            "tags": null,
            "sampler": {
                "type": "const",
                "param": 1,
                "samplingServerURL": "",
                "maxOperations": 0,
                "samplingRefreshInterval": 0
            },
            "reporter": {
                "queueSize": 0,
                "BufferFlushInterval": 0,
                "logSpans": true,
                "localAgentHostPort": "jaeger:6831",
                "collectorEndpoint": "",
                "user": "",
                "password": ""
            },
            "headers": null,
            "baggage_restrictions": null,
            "throttler": null
        }
    }
}
```

`sampler.param=1` says to sample all traces

- [x]  Define what is a span, what tags should be applied, what data to send to the tracer

A span is a unit of work performed by the gateway in the request lifecycle. Tags provided are implementation details, I think it depends on the work performed. For instance the to most work of serving a request attaches tags with metadata about the request at hand, things like `size` of the request content, the `ip` , the `method` etc.

## How it looks like on the UI
For show casing , I log the warning for un authorised  access.

```
curl http://localhost:8181/tyk/not-authorized
```

Produces this trace on jaeger ui

Screen Shot 2019-06-14 at 14 47 13

- [ ] Add tests


joshblakeley:
Really cool!

Are you handling case where there will be multiple Tyk gateway proxies in a mesh pattern where we need to track propogation through the mesh - it's normally done by header injection and tracking with the opentracing agent.

gernest:
>Are you handling case where there will be multiple Tyk gateway proxies in a mesh pattern where we need to track propogation through the mesh - it's normally done by header injection and tracking with the opentracing agent

I am extracting headers at the main entry point and send them down the drain through `context.Context` , however I haven't thought much about service mesh setting and I am just figuring out more about opentracing.

I see we generate some kind of id on a running gateway, I will probably find a good way to attach this id on the spans generated by the running gateway and span contexts send outside the gateway.

buger:
The way how its implemented now can already do all that work, since open tracing basically just injects some headers, containing request path. 

In this case test will be as simple as API1 which target url points to API2, which target URL points to API3. Inside OpenTracing visualisation software you should see single request paths coming 3 points.

buger:
@gernest I actually believe it should be possible to write a unit test as I described above. I think opentracking library you are using can be used not only to write request data, but also to read it from headers.

gernest:
>I actually believe it should be possible to write a unit test as I described above. I think opentracking library you are using can be used not only to write request data, but also to read it from headers. Worth reading opentracing-go test files.

@buger 
Interesting,  that makes sense. I will try to read and see. Because it will be nice to have tests that don't depend on an external dependency (on this case opentracing server).

gernest:
Favor #2384

PR #2337, opened 2019-06-11
Escape invalid chars for xmlMarshal

furkansenharputlu:
Extending #2336

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -3,6 +3,7 @@ package apidef
 import (
 	"encoding/base64"
 	"encoding/json"
+	"encoding/xml"
 	"text/template"
 
 	"github.com/clbanning/mxj"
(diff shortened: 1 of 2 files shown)


apidef/api_definitions.go
```diff
@@ -763,9 +764,16 @@ var Template = template.New("").Funcs(map[string]interface{}{
 		var xmlValue []byte
 		mv, ok := v.(mxj.Map)
 		if ok {
+			mxj.XMLEscapeChars(true)
 			xmlValue, err = mv.Xml()
```
matiasinsaurralde: Perhaps `return` after this line and remove the block in `else` like?
```go
if ok {
  mxj.XMLEscapeChars(true)
  xmlValue, err = mv.Xml()
  return string(xmlValue), err
}
res, ok := v.(map[string]interface{})
...
```

PR #2349, opened 2019-06-17
Get key endpoint should return `key_hash`, if hashing is enabled

komalsukhani:
Fixes [tyk-analytics-ui/issues/1118](https://github.com/TykTechnologies/tyk-analytics-ui/issues/1118)

The response of get key endpoint is modified to 

```
{                                                                       
   Session user.SessionState                                                    
   KeyHash string                                                          
 } 
```


komalsukhani:
Found another way to fix the issue

PR #2369, opened 2019-06-30
(analytics) stop spamming redis with PurgeLoop

asoorm:
Every second, we check ttl of the `analytics-tyk-system-analytics` key.
If <= 0, we assume that the key has expired, and set the TTL to
`StorageExpirationTime`.

This is incorrect, because for an empty list, `TTL key == -2`

As such, we are setting `EXPIRE` to `StorageExpirationTime` even when we
have an empty list which does nothing.

---

Now, we only call `SetExp` when the list does not expire
If the list has an expiry, we sleep for exp

---

`StorageExpirationTime` default is now configured inside `checkup/checkup.go`.


asoorm:
Untested - Do not merge yet.

PR #2387, opened 2019-07-10
Fix detailed logging when overriding response with rich plugins

matiasinsaurralde:
Based on the virtual endpoint approach. Some tweaks are needed in order to initialize the success handler correctly in `api_loader.go`.

After this is ready, we can add it to #2321


matiasinsaurralde:
CI is green 👍

PR #2389, opened 2019-07-10
Fix race condition that occurs during TestVirtualEndpoint

### LINKED ISSUE #2374: TestVirtualEndpoint fails with race
This race causes test to fail. However, it is a bug. While `SuccessHandler.RecordHit` runs in a different goroutine and modifies response. It causes the race.

furkansenharputlu:
Fixes #2374


buger:
Not a fan of such a solution, because it is quite slow...
I wonder if we can just remove using goroutine here, and run it directly? 
Analytics pipeline is async anyway and has own goroutine pool.

PR #2396, opened 2019-07-12
fix panic when failed target lookup

### LINKED ISSUE #2382: Timeout on service discovery endpoint causes panic
bug
Panick when url is not accessable/available etc.

```Failed target lookup: Get https://blabla.com/app: dial tcp 52.209.145.203:443: i/o timeout"
2019/07/05 22:23:58 http: panic serving 172.20.0.1:50350: runtime error: invalid memory address or nil pointer dereference
```

If the SD is not available, Tyk should handle the timeout without panicking

Create an api with an SD to a non-existing endpoint and call this api

1.8.1

gernest:
fixes #2382


ilijabojanovic:
@buger ^^^

PR #2398, opened 2019-07-12
accept oauth credentials as json on /token endpoint

### LINKED ISSUE #2375: oAuth token generation endpoint should support JSON format
feature
At the moment we do support only Form values, e.g. `grant_type=client_credentials&client_id=...&client_secret=...`
Support JSON format if Content-Type set to application/json:
`{"client_id":"...","client_secret":"...", "grant_type":"client_credentials"}`
2.8

gernest:
When Content-Type header is set to application/json . This will decode
and adds them to r.Form to be used by osin

closes #2375


buger:
👌

PR #2400, opened 2019-07-15
migrate to go modules

gernest:
We can `go build`  and `go install`.  The `vendor` directory is gone for now, we can bring it back but It should be with `go mod vendor` so it can work seamless with the latest go tooling

# Takeaways

- we are using ancient libraries, some of which have been deprecated and no longer supported. Example is  `rsc.io/letsencrypt v0.0.2` which was deleted entirely on `v0.0.3`  they recommend 
using `golang.org/x/crypto/acme/autocert`


gernest:
see #2405

Issue #2402, opened 2019-07-16
Python custom loader warns about non-local imports


matiasinsaurralde:
**Do you want to request a *feature* or report a *bug*?**
Enhancement

**What is the current behavior?**
When a custom Python plugin is used, Tyk provides a custom loader that works as an import hook (more details [here](https://docs.python.org/3/reference/import.html#import-hooks)).
This allows us to isolate imports between different middleware that might share module names and keeps the global module dictionary (`sys.modules`) in order.
A minor issue occurs when a system module (or a third party module) that's used in a middleware, tries to load a nonexistent one, for example, the `ldap` package does [the following check](https://github.com/python-ldap/python-ldap/blob/python-ldap-3.2.0/Lib/ldap/__init__.py#L57):
```python
try:
  # Check if Python installation was build with thread support
  import thread
except ImportError:
  LDAPLockBaseClass = DummyLock
else:
  import threading
  LDAPLockBaseClass = threading.Lock
```
The idea behind this check is to test whether the Python version you're running is 2.x or 3.x, it first loads `thread` which only exists in Python 2.x and uses the exception handler to fallback to 3.x (`threading` module).

What happens in Tyk is that even the `import thread` call is passed to our hook. Because there might be no `thread.py` in our middleware/plugin bundle, the following message gets logged:
```
[Jul 16 03:44:17] ERROR python: Your bundle doesn't contain 'thread'
```
However the `ldap` module works perfectly, somehow this becomes an unnecessary warning.

In order to fix this it might be useful to patch the custom loader to ignore `import` (and let the default behavior happen) when the calls come from system or third party packages, not explicitly written in the middleware code.

**What is the expected behavior?**
The custom loader should ignore non-locals `import` calls.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
Create a Python plugin with the following code (no implementation is required because the error will show up on startup):
```python
from tyk.decorators import *
from gateway import TykGateway as tyk
import ldap

@Hook
def TestHook(request, session, metadata, spec):
  pass
```

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
All

matiasinsaurralde:
@buger should we close this? PR was merged 👍

PR #2403, opened 2019-07-16
Make the Python custom loader handle only local calls

matiasinsaurralde:
For #2402, need to do some more tests before merging.


buger:
@matiasinsaurralde what is the state?

matiasinsaurralde:
@buger This is ready, waiting for reviews

buger:
I just was concerned about `need to do some more tests before merging.`

PR #2405, opened 2019-07-16
 use go modules to manage deps

### LINKED ISSUE #2549: Migrate to go mod for dependency management
govendor  is the tool which we use for dependency management, this has been superceeded with official `go mod` tool

Use `go mod` for dependency management

n/a

**Additional context**
n/a

gernest:
Closes #2549


buger:
I see that packages like jwt-go actually changed. Why it does not use the same version as we had?

gernest:
That is me again, I can enforce the same version If I wanted to. running `go get packake@commit` ensures the package/commit is preserved.

There is a lot of manual shell wizardly I had to do to arrive at this .

buger:
If you want this PR merged, it needs to contain exactly the same versions as we have now :)

gernest:
Okay, this was just an experiment for now.I will have to verify all dependencies matches by hand later also I have to check the swagger tool we are using .

So far I'm happy tests are passing.

gernest:
swagger tool we are using doesnt work with go modules. I just can't make it work. So I won't be pursuing this.

gernest:
@buger anything else you need here? I removed unused deps with `go mod tidy` and I rebased this against master.

buger:
Done, I assume we should do the same for the dashboard?

Issue #2395, opened 2019-07-11
BUG - Tyk, in Service Discovery, doesn't use "cache timeout" at the API level, uses the Default Cache TImeout in the conf only


sedkis:
**Do you want to request a *feature* or report a *bug*?**
BUG

**What is the current behavior?**
Tyk, in Service Discovery, doesn't use "cache timeout" at the API level, uses the Default Cache TImeout in the conf only

**What is the expected behavior?**
I would expect Tyk to use my cache timeout if I've set it at the API level. Otherwise, fall back to the default timeout in the CONF

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
here in the code checks the conf default value only:
https://github.com/TykTechnologies/tyk/blob/master/gateway/reverse_proxy.go#L192
doesn't look at the value we set in API def:
![image](https://user-images.githubusercontent.com/8964621/61067602-f547dd80-a3d5-11e9-84d3-8d366df89b3c.png)

PR #2406, opened 2019-07-16
Service Discovery - use cache timeout at the API level before looking at the global configs

sedkis:
Service Discovery - Uses the cache timeout at the API level before looking at the global one in the context of ServiceDiscovery


sedkis:
fixes #2395

PR #2424, opened 2019-07-24
Fix oauth endpoint headers

### LINKED ISSUE #2423: Missing Secure headers on Oauth Token endpoint
Bug

Secure and Cache headers are not set for `/oauth/token` endpoint

Headers should be added

komalsukhani:
Fixes #2423 

Add secure headers even if request to `/oauth/token` fails


ilijabojanovic:
@buger ^^^

Issue #2300, opened 2019-05-22
Support proxy protocol


buger:
**Do you want to request a *feature* or report a *bug*?**
feature
**What is the current behavior?**
Proxy protocol is a way to forward client information through the set of proxies. For HTTP(s) we usually use `X-Forwarded-For` header, but some Proxy protocol provides protocol agnostic solution and works even with any TCP based protocols.
**What is the expected behavior?**
There should be an option in API definition to enable it, once TCP proxying is merged.

buger:
Related info:
https://www.haproxy.com/blog/haproxy/proxy-protocol/
https://inkel.github.io/posts/proxy-protocol/
https://github.com/inkel/viaproxy

excieve:
Please consider an option to enable it globally too. This is especially useful when e.g. load balancer is in TLS pass through mode to the gateway, which terminates it for HTTPS. In this configuration there's no way to get the original IP without PROXY protocol being honoured at both.

ldmonster:
> Please consider an option to enable it globally too. This is especially useful when e.g. load balancer is in TLS pass through mode to the gateway, which terminates it for HTTPS. In this configuration there's no way to get the original IP without PROXY protocol being honoured at both.

Is there anyone else need in this function?

We already add this to our fork of tyk, so we can share it.

Why does we need it?
Because we want to terminate ha-proxy-protocol on tyk, not on nginx before or after.

buger:
@ldmonster would love to see your changes! Lets at least try to review them, on how deeply they are in codebase, and how dangerous is merging smth like this. Thanks!

ldmonster:
> @ldmonster would love to see your changes! Lets at least try to review them, on how deeply they are in codebase, and how dangerous is merging smth like this. Thanks!

Well, hope it gives more context, @buger 

Issue:
https://github.com/TykTechnologies/tyk/issues/5463

PR:
https://github.com/TykTechnologies/tyk/pull/5464

PR #2426, opened 2019-07-24
 tcp proxy

gernest:
successor to #2144 

TODO:
- [x] Analytics support
- [x] ~Way to specific fixed ports and protocols they are support~Way to specific fixed ports and protocols they don't  support
- [x] Support load balancing and service discovery
- [x] Health checks
- [x] Graceful restarts
- [x] Proxy protocol https://github.com/TykTechnologies/tyk/issues/2300


buger:
Note that the analytics part will require integration with proxy protocol, since right now you just skipping this info.

Issue #2429, opened 2019-07-24
JSVM redirects not working


matiasinsaurralde:
**Branch/Envrionment/Version**
- Branch/Version: Master
- Environment: All

**Describe the bug**
It's not possible to trigger HTTP redirects when using JSVM `ReturnOverrides`. The rich plugins code (gRPC/Python) works fine.

**Reproduction steps**

1. Enable JSVM in `tyk.conf` and use the following API definition:
```js
{
    "name": "Tyk Test API",
    "api_id": "1",
    "org_id": "default",
    "definition": {
        "location": "header",
        "key": "version"
    },
    "use_keyless": true,
    "auth": {
        "auth_header_name": "Authorization"
    },
    "version_data": {
        "not_versioned": true,
        "versions": {
            "Default": {
                "name": "Default",
                "expires": "3000-01-02 15:04",
                "use_extended_paths": true,
                "extended_paths": {
                    "ignored": [],
                    "white_list": [],
                    "black_list": []
                }
            }
        }
    },
    "proxy": {
        "listen_path": "/tyk-api-test/",
        "target_url": "http://httpbin.org",
        "strip_listen_path": true
    },
    "custom_middleware_": {
        "pre": [
            {
                "name": "sampleMiddleware",
                "path": "middleware/sample.js",
                "require_session": false
            }
        ],
        "post": []
    },
    "enable_batch_request_support": true
}
```

2. Create the file `middleware/sample.js` with the following contents:

```js
var sampleMiddleware = new TykJS.TykMiddleware.NewMiddleware({});
sampleMiddleware.NewProcessRequest(function(request, session, spec) {
  console.log("sampleMiddleware is called")
  request.ReturnOverrides.ResponseError = "Response body"
  request.ReturnOverrides.ResponseCode = 301
  return sampleMiddleware.ReturnData(request, session.meta_data);
});
```

3. Access the API. The redirect won't happen.

**Expected behavior**
It should be possible to trigger HTTP redirects. When using Python plugins, [this code](https://github.com/matiasinsaurralde/tyk-python-redirect-hook/blob/master/middleware.py) works.

ilijabojanovic:
@buger please merge in release-2.8

PR #2430, opened 2019-07-25
mw_js_plugin: extend ReturnOverrides HTTP code check to allow redirects

matiasinsaurralde:
Fix for #2429, with this patch the following code is able to trigger a redirect:
```js
var sampleMiddleware = new TykJS.TykMiddleware.NewMiddleware({});

sampleMiddleware.NewProcessRequest(function(request, session, spec) {
  console.log("sampleMiddleware is called")
  request.ReturnOverrides.ResponseCode = 301
  request.ReturnOverrides.ResponseHeaders = {
    "Location": "http://anotherurl.com/"
  }
  return sampleMiddleware.ReturnData(request, session.meta_data);
});
```

### DIFF
--- a/gateway/mw_js_plugin.go
+++ b/gateway/mw_js_plugin.go
@@ -255,8 +255,7 @@ func (d *DynamicMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reques
 		return errors.New(newRequestData.Request.ReturnOverrides.ResponseError), newRequestData.Request.ReturnOverrides.ResponseCode
 	}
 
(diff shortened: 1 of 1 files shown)


matiasinsaurralde:
*Note:* this is the way we do it on rich plugins (gRPC/Python): https://github.com/TykTechnologies/tyk/blob/master/gateway/coprocess.go#L341

We check for ~400 HTTP codes just when using custom authentication and report this as failed auth events but in general we allow the user to use `ReturnOverrides` with any HTTP code: https://github.com/TykTechnologies/tyk/blob/master/gateway/coprocess.go#L341

PR #2434, opened 2019-07-27
don't use Request.GetBody

### LINKED ISSUE #2375: oAuth token generation endpoint should support JSON format

gernest:
Decode json directly from r.Body because r.GetBody is nil

The tests missed this because they used `http.NewRequest` which sets `r.GetBody`.
 
Fixes #2375
Fixes #2398


ilijabojanovic:
@buger ^^^

PR #2445, opened 2019-08-04
[PoC] implementing a runtime configurable memory ballast

asoorm:
The memory ballast should reduce pressure on the GC, & reduce request latency.

Some very basic tests indicate c. 30% decrease in request latency at the expense of increasing heap.

TODO - Further tests to gain better understanding of how best to tune / whether to use in conjunction with GOGC.

### DIFF
--- a/cli/linter/schema.json
+++ b/cli/linter/schema.json
@@ -501,6 +501,10 @@
       "type": "string",
       "format": "host-no-port"
     },
+    "memory_ballast_gb": {
+      "type": "integer",
+      "min": 0
+    },
     "listen_port": {
       "type": "integer"
     },
--- a/config/config.go
+++ b/config/config.go
@@ -283,6 +283,7 @@ type Config struct {
 	VersionHeader             string                  `json:"version_header"`
 	UseAsyncSessionWrite      bool                    `json:"optimisations_use_async_session_write"`
 	SuppressRedisSignalReload bool                    `json:"suppress_redis_signal_reload"`
+	MemoryBallastGB           int                     `json:"memory_ballast_gb"`
 
 	// Gateway Security Policies
 	HashKeys                bool           `json:"hash_keys"`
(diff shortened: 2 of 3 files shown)


buger:
So basically it allocates big chunk on heap from the start, GC free it up on next cycle (since this variable not used anywhere), but it helps with heap growth issues in future? 

Am I correct that it mostly helps with cold start?

jensneuse:
> So basically it allocates big chunk on heap from the start, GC free it up on next cycle (since this variable not used anywhere), but it helps with heap growth issues in future?
> 
> Am I correct that it mostly helps with cold start?

It doesn't really allocate memory. It only makes virtual memory larger.
This means nothing else but more heap allocations and less frequent GC cycles because it takes longer until heap size triggers a GC cycle due to larger "virtual" memory.

asoorm:
It doesn't get cleaned up by GC, because it's in main function.

Without this memory ballast:

- Heap size at start is relatively small. 
- Under load, GC mark & sweep runs quite often
- GO GC uses a pacer to determine when to run next GC cycle, (GOGC=100 default), it runs every time the Heap doubles in size.


With the memory ballast

- We start application requesting lets say 2GB heap
- This means that GC will be run when heap gets to 4GB (2GB heap + 2GB of other allocations)
- After GC, ballast memory will not be swept, because reference still held in main.
- So most of the 2GB of other allocations will be swept, and as such, GC will only run again once heap gets to ~4GB

So the ballast increases base-size of heap so that GC is delayed.
And because we don't actually read from or write to the ballast, it only uses Virtual memory, rather than physical memory.

buger:
Nice. I think we need to rename this variable to smth else and understand how to explain this approach in docs, without getting too much in details.

asoorm:
Makes sense - And that's why I want to do some proper benchmarks so that we can give guidance as to what size ballast to set for different scenarios / instance types.

buger:
What you described ^

dencoded:
I am a little bit late to the party. This looks like an interesting idea. Let me do a couple of experiments with this branch and then I will post my review.

asoorm:
@spetrunin @jensneuse FYI

PR #2454, opened 2019-08-12
(log) update logrus to fix casing issue

asoorm:
- As per title - Updated logrus package
- updating evalphobia
- updating gemnasium graylog hook
- updating logrus-logstash-hook
- removing unused dependencies
- no need to escape json from msg
- JSVM log message now contains prefix - and logger no longer leaves erroneous space at the end
- Fixes issue where `TextFormatter` shows log as seconds since start, rather than timestamp.


gernest:
I think  the ci error isn't related to this PR. I get the same error on my other PR. coveralls is choking under heavy load I think

```html
<h2>This website is under heavy load (queue full)</h2><p>We're sorry, too many people are accessing this website at the same time. We're working on this problem. Please try again later.</p>
```

PR #2457, opened 2019-08-16
Fix APIDefinition JWTIdentityBaseField bson/json key inconsisitency i…

singleye:
When I tried to create app definition by dumping MongoDB, I found the JWT identity base field doesn't match between MongoDB (jwt_identit_base_field) and JSON (jwt_identity_base_field) file. After reviewing other field definition, I think this should be a typo issue.


buger:
Hi! In order to achieve backward compatibility, we need to support both fields like:
```
OldJWTIdentityBaseField       string               `bson:"jwt_identit_base_field" json:"-"`
JWTIdentityBaseField       string               `bson:"jwt_identity_base_field" json:"jwt_identity_base_field"`
```

And `DecodeFromDB` function, do smth like:
```
if JWTIdentityBaseField =="" && OldJWTIdentityBaseField != "" && {
  JWTIdentityBaseField = OldJWTIdentityBaseField
}
```

PR #2463, opened 2019-08-21
account  for data expiration on orgid failed cache lookup

### LINKED ISSUE #2370: [TT-2393] - Data expiration inside OrgId ignored on failed cache lookups
bug
If [Time based capping](https://tyk.io/docs/analyse/capping-analytics-data-storage/#a-name-time-based-cap-a-time-based-cap) enabled, in some cases, you can see analytics records with default `604800` expiration, rather then what specified in organization object.
It should always use value from org object.
See above
2.8

gernest:
Fixes #2370

### DIFF
--- a/gateway/middleware.go
+++ b/gateway/middleware.go
@@ -15,6 +15,7 @@ import (
 	"github.com/paulbellamy/ratecounter"
 	cache "github.com/pmylund/go-cache"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/singleflight"
 
 	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/config"
@@ -27,6 +28,8 @@ import (
 const mwStatusRespond = 666
 
 var GlobalRate = ratecounter.NewRateCounter(1 * time.Second)
+
+var orgSessionExpiryCache singleflight.Group
 
 type TykMiddleware interface {
 	Init()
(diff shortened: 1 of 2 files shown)


gateway/middleware.go
```diff
@@ -28,6 +29,8 @@ const mwStatusRespond = 666
 
 var GlobalRate = ratecounter.NewRateCounter(1 * time.Second)
 
+var orgSessionExpiryCache singleflight.Group
```
asoorm:
```
var (
        GlobalRate = ratecounter.NewRateCounter(1 * time.Second)
        orgSessionExpiryCache singleflight.Group
)
```

gernest:
@asoorm PTAL

gateway/middleware.go
```diff
@@ -232,10 +236,17 @@ func (t BaseMiddleware) OrgSessionExpiry(orgid string) int64 {
 	cachedVal, found := ExpiryCache.Get(orgid)
 	if !found {
 		// Cache failed attempt
-		t.SetOrgExpiry(orgid, 604800)
-		go t.OrgSession(orgid)
-		t.Logger().Debug("no cached entry found, returning 7 days")
-		return 604800
+		id, _, _ := orgSessionExpiryCache.Do(orgid, func() (interface{}, error) {
```
asoorm: Maybe I don't understand business logic - but it looks like singleflight should run on `ExpiryCache.Get(orgid)` logic?

asoorm:
LGTM

@gernest As per discussions, some tests to ensure we didn't break anything.
@buger Please review for business logic

gateway/api_definition.go
```diff
@@ -1161,6 +1164,17 @@ func (a *APISpec) CheckSpecMatchesStatus(r *http.Request, rxPaths []URLSpec, mod
 	return false, nil
 }
 
+func (a *APISpec) stripListenPath(path string) string {
```
asoorm: @gernest does this belong in this PR?

gernest: nope, sorry was some other work.

PR #2465, opened 2019-08-22
Kv store for secrets in configuration

adelowo:
See https://github.com/TykTechnologies/product/issues/137

This is the first step to allow for the removal of plain passwords in the config file e.g https://tyk.io/docs/configure/tyk-gateway-configuration-options/#a-name-secret-a-secret

This PR will add implementations for

- [x] Consul
  - [x] Body/URL rewrite
- [x] Vault
  - [x] Body/URL rewrite
- [x] Environment
   - [x] Body/URL rewrite
- [x] `tyk.conf` secrets - dictionary of values in the `tyk.conf`
  - [x] Body/URL rewrite
- [x] Support kv storage in an API spec's listen path and target URL.

For body/URL rewrites, the syntax is :

- `$secret_vault.` for Vault
- `$secret_consul.` for consul
- `$secret_env.` for environment values.
- `$secret_conf.` for dictionary in `tyk.conf`

> You can still make use of `$tyk_context.` as before though

Example URL ==> `https://jsonplaceholder.typicode.com/users/1/posts?key1=$secret_vault.secret/data/lanre/adelowo.god&yikes=$secret_cohttps://github.com/TykTechnologies/tyk/pull/2465/filesnsul.god/name&key2=$tyk_context.request_id`

### Configuration

For environment values, use `env://yikes` which will be expanded to `TYK_SECRET_YIKES`
For values set in the dashboard, use `secrets://value`.

For consul, you need to specify like `consul://path/to/secret`
For Vault, you need to specify like `vault://engine/path/to/secret.actual_secret_name`. Why the dot notation, Vault is a little different as per how it keeps secrets, multiple secrets can be under one key. So we use the dot notation for retrieving the exact one we need. Take for example

```sh
vault kv get --format json secret/hello | jq
```
gives us

```json
{
  "request_id": "0c7e44e1-b71d-2102-5349-b5c60c13fb02",
  "lease_id": "",
  "lease_duration": 0,
  "renewable": false,
  "data": {
    "data": {
      "excited": "yes",
      "foo": "world",
      "man": "yes"
    },
    "metadata": {
      "created_time": "2019-08-28T14:18:44.477126Z",
      "deletion_time": "",
      "destroyed": false,
      "version": 1
    }
  },
  "warnings": null
}

```

Take a look at the `data` attribute. In this case, if we want to retrieve the `excited` key, we use `secret/data/hello.excited`.


buger:
Love the idea! 

We need to add support for KV values inside url or body rewrites.
Should be easy to do just by modifying `replaceTykVariables` function to understand values like: `$consul.some_var`.

adelowo:
Got it!!

buger:
@adelowo would love if you consider adding support for values from environment variables and extend this "kv" usage to a lot of places in API spec, so I would be able to specify via kv API even things like upstream URL, or for example, setting values inside API config_data object.

adelowo:
Will fix CI tomorrow

buger:
In general looks good to me! 

Would love if you add support for 2 more simple stores:
1) Environment variable store: If you have env variable TYK_SECRET_FOO, it should be available like `env://foo` and `$secret_env.foo`
2) Individual gateway config storage. Add a new section to tyk.conf call secrets, which is just map[string]string, like `"secrets": {"foo": "bar"}`, and can be accessed like: `conf://foo`and `$secret_conf.foo` (no need for nested values). Can be very handy when you have multiple environments like staging and production, and things like domain or listen path will be set from secrets storage.

adelowo:
@buger it is ready for review now

adelowo:
Need to update documentation

tomshy:
Trying to use this amazing feature but am still not successful. 

Here's what I have done:
- updated Tyk to v3.0.0
- included my kv object to tyk.conf as:
```...
"kv":{
    "secrets": {
        "gateway": "secret"
    }
  }
```
- and trying to use the gateway secret in body transform as:

`{{$secret_conf.gateway}}`

However, tyk is raising errors, where did I go wrong?

PR #2467, opened 2019-08-23
Add support of RSA in request signing and request check middleware

### LINKED ISSUE #2452: Add support of RSA in request signing middleware
Right now, request signing middleware only supports HMAC algorithm. 
Extend this feature to add support of RSA.

komalsukhani:
Fixes #2452 

 ### **Added new fields in below structure**

**Session**
1. _EnableHTTPSignatureValidation bool_
            This field should be used to enable signature validation for both `HMAC` and `RSA`.
**Note**: `HMACEnabled` field will still be supported for backward compatibility, but will be deprecated in future.

2. _RSACertificateId string_
            Public key certificate ID/path to be used for RSA signature validation.

**Policy**
_1. EnableHTTPSignatureValidation bool_

**Request Signing Middleware**
1. _CertificateId string_
      Public key certificate ID/path to be used for RSA signature generation.
2. _HeaderList []string_
      Optinal field. List of headers to be used for generating signature

### **Added RSA support in Request Checker middleware**
You can now validate HTTP message signature signed using RSA private key.
You need to set `EnableHTTPSignatureValidation` to true and `RSACertificateId` to uploaded RSA Public key certificate or to the path of public key pem file.

### **Added RSA support in Request Signing middleware**
Now user can sign requests using RSA key pair. They need to set `CertificateId` to uploaded RSA certificate ID or to the path of certificate file itself.

### **Specify Header List to be used while signing the request**
Earlier request signing middleware considered below headers while generating signature string
          ` All the headers of the request + (request-target) + date `

Now you can specify headers to be used by setting `HeaderList` field of request signing middleware. 

**Cases**

- If an header is specified in HeaderList but is not set in the request, then that header is ignored.

- If non of the header from the HeaderList is set, then below headers are used for signing request

                     `(request-target) + date`

- If HeaderList is empty, then below headers are used for signing request

              ` All the headers of the request + (request-target) + date `


apidef/api_definitions.go
```diff
@@ -498,10 +498,12 @@ type BundleManifest struct {
...
-	IsEnabled bool   `bson:"is_enabled" json:"is_enabled"`
-	Secret    string `bson:"secret" json:"secret"`
-	KeyId     string `bson:"key_id" json:"key_id"`
-	Algorithm string `bson:"algorithm" json:"algorithm"`
+	IsEnabled     bool     `bson:"is_enabled" json:"is_enabled"`
+	Secret        string   `bson:"secret" json:"secret"`
+	KeyId         string   `bson:"key_id" json:"key_id"`
+	Algorithm     string   `bson:"algorithm" json:"algorithm"`
+	HeaderList    []string `bson:"header_list" json:"header_list"`
+	CertificateId string   `bson:"certificate_id" json:"certificate"`
```
buger: bson and json fields should be the same to avoid confusion

buger:
I do not think that we need `RSAEnabled` boolean fields inside policy or session since in general, it is just part of HMAC in our case. 

`rsa_certificate_id` in session object can be renamed to `hmac_public_key_id` (and if it set to non empty it just repaces `RSAEnabled` field.

komalsukhani:
RSA is not part of HMAC. If HMAC is enabled, we are signing data using HMAC algorithm while if RSA is enabled, we are signing it using RSA algorithm. 

Using `HMACEnabled` field for RSA might cause confusion.
I thought of renaming the `HMACEnabled` field to something generic but then it will break backward compatibilty.

komalsukhani:
@buger Updated with latest master

PR #2473, opened 2019-08-27
support of imported keys with new format added

dencoded:
added changes for https://github.com/TykTechnologies/product/issues/165

I required to modify tests a lot as now operation with key requires carefully set and matched orgID

### DIFF
--- a/gateway/auth_manager.go
+++ b/gateway/auth_manager.go
@@ -286,20 +286,25 @@ func (b *DefaultSessionManager) SessionDetail(keyName string, hashed bool) (user
 	if hashed {
 		jsonKeyVal, err = b.store.GetRawKey(b.store.GetKeyPrefix() + keyName)
 	} else {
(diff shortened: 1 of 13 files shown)


gateway/auth_manager.go
```diff
@@ -281,7 +286,21 @@ func (b *DefaultSessionManager) SessionDetail(keyName string, hashed bool) (user
 	if hashed {
 		jsonKeyVal, err = b.store.GetRawKey(b.store.GetKeyPrefix() + keyName)
 	} else {
-		jsonKeyVal, err = b.store.GetKey(keyName)
+		// try to get legacy and new format key at once
```
buger: We can actually detect if key is "imported". Smt like:
`If storage.TokenOrg(key) == ""`

buger: Mget needed only for custom keys

PR #2478, opened 2019-08-30
[perf] goroutine optimisations

asoorm:
- CPU time is spent on scheduling GoRoutines.
- Memory allocations for every request due to copyBuffer.
- Every request allocates memory on the hot path due to ranging over `rxPaths`

---

1. Setting the default worker size from 50 to `runtime.NumCPU`. This makes tunny more efficient. There are significantly less go routines in-use by Tunny now.

2. Updating the rate counter library because the old version spins up too many unnecessary go routines - under load - this was 484. This change reduces e2e latency by circa 10% when rate-limiting is enabled.

Screenshot 2019-08-30 at 14 39 40

Screenshot 2019-08-30 at 14 41 20

3. optimise `copyBuffer` so that memory isn't allocated for every request and now is obtaining from a buffer pool & returning once no longer needed.

4. Access `rxPaths` by index

```
OLD: BenchmarkDefaultVersion-8   	    1233	   1029221 ns/op	  226285 B/op	    1527 allocs/op
NEW: BenchmarkDefaultVersion-8   	    1336	    815036 ns/op	  155331 B/op	    1507 allocs/op
```

### DIFF
--- a/gateway/api_definition.go
+++ b/gateway/api_definition.go
@@ -1011,14 +1011,14 @@ func (a *APISpec) getURLStatus(stat URLStatus) RequestStatus {
 // URLAllowedAndIgnored checks if a url is allowed and ignored.
 func (a *APISpec) URLAllowedAndIgnored(r *http.Request, rxPaths []URLSpec, whiteListStatus bool) (RequestStatus, interface{}) {
 	// Check if ignored
-	for _, v := range rxPaths {
-		if !v.Spec.MatchString(r.URL.Path) {
+	for i := range rxPaths {
+		if !rxPaths[i].Spec.MatchString(r.URL.Path) {
 			continue
 		}
 
-		if v.MethodActions != nil {
+		if rxPaths[i].MethodActions != nil {
 			// We are using an extended path set, check for the method
-			methodMeta, matchMethodOk := v.MethodActions[r.Method]
+			methodMeta, matchMethodOk := rxPaths[i].MethodActions[r.Method]
 			if !matchMethodOk {
 				continue
 			}
@@ -1028,7 +1028,7 @@ func (a *APISpec) URLAllowedAndIgnored(r *http.Request, rxPaths []URLSpec, white
 			switch methodMeta.Action {
 			case apidef.NoAction:
 				// NoAction status means we're not treating this request in any special or exceptional way
-				return a.getURLStatus(v.Status), nil
+				return a.getURLStatus(rxPaths[i].Status), nil
 			case apidef.Reply:
 				return StatusRedirectFlowByReply, &methodMeta
 			default:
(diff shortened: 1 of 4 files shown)


gateway/api_definition.go
```diff
@@ -988,14 +988,14 @@ func (a *APISpec) getURLStatus(stat URLStatus) RequestStatus {
 // URLAllowedAndIgnored checks if a url is allowed and ignored.
 func (a *APISpec) URLAllowedAndIgnored(r *http.Request, rxPaths []URLSpec, whiteListStatus bool) (RequestStatus, interface{}) {
 	// Check if ignored
-	for _, v := range rxPaths {
-		if !v.Spec.MatchString(r.URL.Path) {
+	for i := range rxPaths {
+		if !rxPaths[i].Spec.MatchString(r.URL.Path) {
```
dencoded: this is quite interesting optimization - so access by index is faster than just sequential iteration over array?

dencoded: would be interesting to see `BenchmarkVersioning` cmp-results vs master

gateway/api_definition.go
```diff
@@ -1084,84 +1084,84 @@ func (a *APISpec) CheckSpecMatchesStatus(r *http.Request, rxPaths []URLSpec, mod
 	}
 
 	// Check if ignored
-	for _, v := range rxPaths {
-		if mode != v.Status {
+	for i := range rxPaths {
+		if mode != rxPaths[i].Status {
```
dencoded: the same idea - would be good to see benchmark difference vs master where this code involved, i.e. `BenchmarkResponseHeaderInjection`

gateway/reverse_proxy.go
```diff
@@ -910,16 +918,17 @@ func (p *ReverseProxy) CopyResponse(dst io.Writer, src io.Reader) {
...
 
-	p.copyBuffer(dst, src, nil)
+	p.copyBuffer(dst, src)
 }
 
-func (p *ReverseProxy) copyBuffer(dst io.Writer, src io.Reader, buf []byte) (int64, error) {
-	if len(buf) == 0 {
-		buf = make([]byte, 32*1024)
-	}
+func (p *ReverseProxy) copyBuffer(dst io.Writer, src io.Reader) (int64, error) {
```
dencoded: I think it is worth to support this change with new benchmark vs master (would have to add it to master manually to have numbers `"before"`)

jensneuse:
You could use benchcmp to compare benchmarks:
https://godoc.org/golang.org/x/tools/cmd/benchcmp

gateway/api_definition.go
asoorm: It's not just sequential access over array - before you are making the allocation to `v`.

Master: `list UrlAllowedAndIgnored`
```
(pprof) list URLAllowedAndIgnored
Total: 3.43MB
ROUTINE ======================== github.com/TykTechnologies/tyk/gateway.(*APISpec).URLAllowedAndIgnored in /Users/ahmet/go/src/github.com/TykTechnologies/tyk/gateway/api_definition.go
       7kB        7kB (flat, cum)   0.2% of Total
         .          .    986:}
         .          .    987:
         .          .    988:// URLAllowedAndIgnored checks if a url is allowed and ignored.
         .          .    989:func (a *APISpec) URLAllowedAndIgnored(r *http.Request, rxPaths []URLSpec, whiteListStatus bool) (RequestStatus, interface{}) {
         .          .    990:	// Check if ignored
       7kB        7kB    991:	for _, v := range rxPaths {
         .          .    992:		if !v.Spec.MatchString(r.URL.Path) {```
```

New: it doesn't even show, because we are not allocating to `v`.

dencoded:
@jensneuse we've been using `benchcmp` to compare for all our performance related PRs, I don't see it here so put those comments.
There was also idea to add it to CI tool chain to run benchmarks and fail if benchmarks are degrading (I believe we have ticket for that)

jensneuse:
I don't know if you want this enabled by default. Adding new features could decrease performance. Improving code quality might decrease performance too. I'd decide from case to case if it's worth it.

gateway/api_definition.go
letzya: But rxPath[i] is like v
That feels like it’s the optimiser job

asoorm: This was done using a large `[]int`. But rxPaths is considerably larger.

```
BenchmarkRangeCopy-8   	     490	   2497576 ns/op	     282 B/op	       3 allocs/op
BenchmarkRangeIndex-8   	     470	   2476958 ns/op	     262 B/op	       2 allocs/op
```

asoorm: @dencoded 

```
OLD: BenchmarkDefaultVersion-8   	    1233	   1029221 ns/op	  226285 B/op	    1527 allocs/op
NEW: BenchmarkDefaultVersion-8   	    1336	    815036 ns/op	  155331 B/op	    1507 allocs/op
```

Not created specific benchmark - but here is an example bench which touches `URLAllowedAndIgnored`

PR #2488, opened 2019-09-05
[perf] avoid memory allocations from converting []byte to string

asoorm:
```
if encoded, err := msgpack.Marshal(record); err != nil {
```

`msgpack.Marshal` returns a byte slice. We currently convert to string which causes memory allocations.

`ClusterTransaction` args are of type interface and underlying interface `writeArg` function can write bytes directly to Redis.

```
func (c *conn) writeArg(arg interface{}, argumentTypeOK bool) (err error) {
	switch arg := arg.(type) {
	case string:
		return c.writeString(arg)
	case []byte:
		return c.writeBytes(arg)
	case int:
		return c.writeInt64(int64(arg))
	case int64:
```

---

20% decrease in GC Cycles when Analytics is enabled

Current Master: 348 GC Cycles after 20,000 requests:

```
gc 345 @43.610s 0%: 0.97+2.0+0.040 ms clock, 7.8+5.3/3.9/0+0.32 ms cpu, 10->10->5 MB, 11 MB goal, 8 P
gc 346 @43.663s 0%: 0.15+1.7+0.003 ms clock, 1.2+5.9/3.1/0+0.028 ms cpu, 11->11->6 MB, 12 MB goal, 8 P
gc 347 @43.720s 0%: 0.012+1.0+0.002 ms clock, 0.10+0.46/1.8/5.0+0.021 ms cpu, 11->11->5 MB, 12 MB goal, 8 P
gc 348 @44.174s 0%: 0.073+2.5+0.005 ms clock, 0.58+2.1/4.6/11+0.044 ms cpu, 9->9->4 MB, 10 MB goal, 8 P
```

After this change: 281 GC Cycles after 20,000 requests:

```
gc 278 @14.364s 2%: 1.6+3.8+0.013 ms clock, 13+5.2/5.9/0.85+0.11 ms cpu, 12->13->6 MB, 13 MB goal, 8 P
gc 279 @14.432s 2%: 0.070+1.5+0.003 ms clock, 0.56+2.9/3.0/4.6+0.025 ms cpu, 12->12->6 MB, 13 MB goal, 8 P
gc 280 @14.506s 2%: 0.014+1.3+0.003 ms clock, 0.11+1.4/2.6/5.3+0.027 ms cpu, 12->12->6 MB, 13 MB goal, 8 P
gc 281 @14.573s 2%: 0.013+0.99+0.002 ms clock, 0.11+0.19/1.9/5.2+0.022 ms cpu, 12->12->5 MB, 13 MB goal, 8 P
```

---

Pprof output: Before

RedisAnalyticsHandler recordWorker-old

Pprof output: After

RedisAnalyticsHandler recordWorker-new


buger:
Why you added new interface instead of changing storage.Handler ?

asoorm:
> Why you added new interface instead of changing storage.Handler ?

The bigger the interface, the weaker the abstraction

checkup/checkup.go
```diff
@@ -79,3 +84,21 @@ func defaultSecrets(c config.Config) {
...
 	}
 }
+
+func defaultAnalytics(c config.Config) {
+	if !c.EnableAnalytics {
+		return
+	}
+
+	if c.AnalyticsConfig.PoolSize == 0 {
+		log.Warningf("AnalyticsConfig.PoolSize unset. Defaulting to number of available CPUs (%d)", runtime.NumCPU())
```
dencoded: more logrus-ish would be:
```
log.WithField("numCPU", runtime.NumCPU()).Warning("AnalyticsConfig.PoolSize unset. Defaulting to number of available CPU")
```

checkup/checkup.go
```diff
@@ -79,3 +84,21 @@ func defaultSecrets(c config.Config) {
...
+		c.AnalyticsConfig.PoolSize = runtime.NumCPU()
+	}
+
+	if c.AnalyticsConfig.RecordsBufferSize < minRecordsBufferSize {
+		log.Warningf("AnalyticsConfig.RecordsBufferSize < minimum (%d) - Overriding", minRecordsBufferSize)
```
dencoded: same `Withfiled` log-entry here

asoorm:
> some minor notes about logging

Not sure what you think about this, appreciate it's cleaner, and more logrusy, but I think we should remove all `logrus.WithField` or `logrus.WithFields` due to memory allocations.

```
// Add a map of fields to the Entry.
func (entry *Entry) WithFields(fields Fields) *Entry {
	data := make(Fields, len(entry.Data)+len(fields))
	for k, v := range entry.Data {
		data[k] = v
	}
	fieldErr := entry.err
	for k, v := range fields {
		isErrField := false
		if t := reflect.TypeOf(v); t != nil {
			switch t.Kind() {
			case reflect.Func:
				isErrField = true
			case reflect.Ptr:
				isErrField = t.Elem().Kind() == reflect.Func
			}
		}
		if isErrField {
			tmp := fmt.Sprintf("can not add field %q", k)
			if fieldErr != "" {
				fieldErr = entry.err + ", " + tmp
			} else {
				fieldErr = tmp
			}
		} else {
			data[k] = v
		}
	}
	return &Entry{Logger: entry.Logger, Data: data, Time: entry.Time, err: fieldErr, Context: entry.Context}
}
```

dencoded:
I think it should log less info in general, only useful info like warnings in this PR. So thats minify logrus overhead.

According to memory optimizations - formatted output into string might be heavier on CPU or RAM.

Issue #2379, opened 2019-07-05
[TT-2392] - Extend the REST API to provide policy management


matiasinsaurralde:
**Do you want to request a *feature* or report a *bug*?**
Feature/Enhancement

**What is the current behavior?**
Only file-based policies are supported at the moment, as described [here](https://tyk.io/docs/with-tyk-community-edition/tutorials/create-security-policy/).

**What is the expected behavior?**
It would be useful to implement an endpoint similar to the one that's used [for keys](https://tyk.io/docs/with-tyk-community-edition/tutorials/create-api-token/).

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
Tyk Gateway (CE)

cajund:
Use Case #1 (GET):

After policies are loaded via the file, it is necessary to obtain the policy id so that they can be given to the identity provider. The IDP then uses these policy ids to embed them into the JWT that is sent to users of varying privileges.

Use Case #2 (PUT)

Endpoint based permissions schemas can be updated on the fly from an administrative UI is the whitelisted endpoints can be updated on the fly. The alternative would be to regenerate the policy file and reload the gateway (storing the policy -> endpoint definitions in some other manner (which would still need to be done, but using this type of endpoint would prevent reloading of the gateway service.

swchoi727:
Would it be possible to support POST and DELETE APIs as well to create new and delete existing policies?

matiasinsaurralde:
> Would it be possible to support POST and DELETE APIs as well to create new and delete existing policies?

It should be possible, a CRUD behavior could be implemented here

williamokano:
At least the documentation could have been updated to warn that policies management is not possible yet through the REST API, or at least on the CE version

williamokano:
Hi @matiasinsaurralde , could you please code review my pull request? Thanks

christtyk:
not stale

haoyuanli:
@christtyk Thanks for continuing the discussion here

JavaBloxx:
Would be very useful

PR #2493, opened 2019-09-07
[Feature] Add policies rest api

williamokano:
This pull request intention is to add the missing REST API endpoints for listing and managing the Tyk Gateway policies.

I've added endpoints for managing and listing policies. Modifying any policy requires reloading the policies.json file. Won't work if using Tyk Dashboard.

| Method | URL                   | Payload     | Response            | Description                                                              |
|--------|-----------------------|-------------|---------------------|--------------------------------------------------------------------------|
| GET    | /tyk/policies         | nil         | []user.Policy       | Get a list of all policies                                               |
| GET    | /tyk/policies/{polID} | nil         | user.Policy         | Get a single policy by ID                                                |
| POST   | /tyk/policies         | user.Policy | apiModifyKeySuccess | Create a policy (append to policies.json). Only works if using headless. |
| PUT    | /tyk/policies/{polID} | user.Policy | apiModifyKeySuccess | Update an existing policy.                                               |
| DELETE | /tyk/policies/{polID} | nil         | apiModifyKeySuccess | Delete an existing policy                                                |

This pull request fixed the this isse https://github.com/TykTechnologies/tyk/issues/2379


gateway/api.go
```diff
@@ -847,6 +847,236 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func policiesHandler(w http.ResponseWriter, r *http.Request) {
+	policyID := mux.Vars(r)["policyID"]
+
+	var obj interface{}
+	var code int
+
+	switch r.Method {
+	case "GET":
```
matiasinsaurralde: Better to use `http.MethodGet` and similar constants.

gateway/api.go
```diff
@@ -847,6 +847,236 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
...
+		if policyID != "" {
+			log.Debug("Requesting Policy definition for", policyID)
+			obj, code = handleGetPolicy(policyID)
+		} else {
+			log.Debug("Requesting Policies list")
+			obj, code = handleGetPoliciesList()
+		}
+	case "POST":
```
matiasinsaurralde: `http.MethodPost` here

gateway/api.go
```diff
@@ -847,6 +847,236 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
...
+		log.Debug("Creating new policy")
+		obj, code = handleAddOrUpdatePolicy(policyID, r)
+	case "PUT":
```
matiasinsaurralde: `http.MethodPut` here

gateway/api.go
williamokano: Sure

gateway/api.go
williamokano: Sure

gateway/api.go
williamokano: Sure

sedkis:
Hi @williamokano ,
I will review and get back to you.
Thanks

sedkis:
Thanks for your hard work @williamokano.  I've made a note to add these to our docs should (when) this PR get approved by an engineer

williamokano:
@sedkis Hi, Thanks.

sedkis:
It's obvious the hard work you've put into this.  Thank-you for your contribution!  But can't accept at this time because of the following reasons:
- This approach only works on a single node
- These APIs already exist in the Dashboard

Thank-you.

williamokano:
Hi, thanks for the feedback.

About working on a single node, it's true, although I just kind of copied
the way the "api" rest API already works.

And about it working with the dashboard, it's also true, but, according to
the Tyk's licence, I need a licence in order to use Tyk, which is not
required when using it headless.

But thanks for the feedback :)

Em ter, 17 de set de 2019 15:45, Sedky Abou-Shamalah <
notifications@github.com> escreveu:

haoyuanli:
> It's obvious the hard work you've put into this. Thank-you for your contribution! But can't accept at this time because of the following reasons:
> 
> * This approach only works on a single node
> * These APIs already exist in the Dashboard
> 
> Thank-you.

@sedkis

- Do existing Tyk dashboard APIs work in multi nodes? If they do, but this one doesn't, what should be done to make it support multiple nodes?

- While the APIs exist in the dashboard already, this API would be very helpful for people who are using tyk-gateway standalone.

Thank you so much!

hinorashi:
ok let's use the dashboard 🐧

sedkis:
@haoyuanli Let me explain by how by going through the 10k foot view.  Essentially what the Dashboard does is introduce a higher level concept of an "organization".  This allows you to manage all your API definitions, policies, users, developers, etc by an org level.  The Gateway is quiet low level in comparison.  When a Gateway boots up, it connects to the Control Plane (Dashboard) and loads all the run-time configurations.  This becomes your single source of truth.   The OSS Gateway is the work-horse of the whole operation, but it's a dummy client that does what it's told.  What you are missing is the orchestration platform that syncs them all up. 

To answer your question directly, to keep all your Gateways in sync you need some kind of orchestration platform or CICD workflow to manage all your Gateways, via APIs or Tyk Sync etc.

haoyuanli:
> It's obvious the hard work you've put into this. Thank-you for your contribution! But can't accept at this time because of the following reasons:
> 
> * This approach only works on a single node
> * These APIs already exist in the Dashboard
> 
> Thank-you.

Btw, the existing `/tyk/apis` also only works for single-node installation

Per documentation at https://tyk.io/docs/tyk-gateway-api/

```These methods only work on a single API node. If updating a cluster, it is important to ensure that all nodes are updated before initiating a reload.```

**Comment by bmonteiro** [comment]
There is an initiative from Tyk -> https://github.com/TykTechnologies/tyk-operator that will need to also have the policies managed via API (Tyk GW API to be clear) like it is today for the APIs. So why is this PR not ok ?
CC: @asoorm

**Comment by buger** [comment]
You raised a good point. 
We actually internally re-thinking this PR now. 
I will re-open it to keep everyone in the loop.

**Comment by bmonteiro** [comment]
> You raised a good point.
> We actually internally re-thinking this PR now.
> I will re-open it to keep everyone in the loop.

Please keep the chat updated with the outcome :) I believe the implementation maybe different but it is a needed feature. thx!

File: gateway/api.go [hunk 4]
```diff
@@ -847,6 +847,236 @@ func apiHandler(w http.ResponseWriter, r *http.Request) {
...
+
+	log.WithFields(logrus.Fields{
+		"prefix":   "policies",
+		"policyID": policyID,
+	}).Error("Policy doesn't exist.")
+	return apiError("Policy not found"), http.StatusNotFound
+}
+
+func handleGetPoliciesList() (interface{}, int) {
+	policiesMu.RLock()
```
vaske, line 900: where this one has been defined?

vaske:
We are addressing this one at this PR https://github.com/TykTechnologies/tyk/pull/3302 thanks for rising this in the first place

PR #2494, opened 2019-09-10
Updating supported versions

alephnull:
precise was EOL a couple of years ago
bionic is the newest LTS
jessie is reaching EOL but not yet


buger:
@alephnull ^

PR #2505, opened 2019-09-13
Fix GetKeys filter in RedisCluster

furkansenharputlu:
Fixes https://github.com/TykTechnologies/tyk/issues/2508

### DIFF
--- a/storage/redis_cluster.go
+++ b/storage/redis_cluster.go
@@ -446,6 +446,21 @@ func (r *RedisCluster) DeleteKey(keyName string) bool {
 	return n.(int64) > 0
 }
 
+// DeleteAllKeys will remove all keys from the database.
+func (r *RedisCluster) DeleteAllKeys() bool {
+	r.ensureConnection()
+	n, err := r.singleton().Do("FLUSHALL")
+	if err != nil {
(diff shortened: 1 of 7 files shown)


storage/redis_cluster.go
```diff
@@ -361,6 +363,7 @@ func (r *RedisCluster) GetKeys(filter string) []string {
 	filterHash := ""
 	if filter != "" {
 		filterHash = r.hashKey(filter)
```
buger: not quite sure why hashKey needed here

storage/redis_cluster.go
```diff
@@ -361,6 +363,7 @@ func (r *RedisCluster) GetKeys(filter string) []string {
...
+		filterHash = base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf(`{"org":"%s"`, filterHash)))
```
buger: We should make it backward compatible way, with keys which was generated with old format (e.g. murmur32 hash), so we should have both old and new filter. 
I propose to move this code directly to Keys List API method, which will look like:
```
orgIDB64 := fmt.Sprintf(`{"org":"%s"`, filterHash)
keys := storage.GetKeys(orgID)
keys = append(keys, orgIDB64...)

PR #2522, opened 2019-09-19
allow polices to be loaded from dash irrespective of where APIs are loaded from

adelowo:
Right now, if APIs are loaded from the filesystem, policies cannot be loaded from the dash.

How to replicate:

- set `policies.policy_source: "service"` and `use_db_app_config: false`
- APIs will be loaded from files while policies won't be loaded from the dashboard.

Fixes https://github.com/TykTechnologies/tyk/issues/2545

### DIFF
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -1188,7 +1188,7 @@ func dashboardServiceInit() {
 }
 
 func handleDashboardRegistration() {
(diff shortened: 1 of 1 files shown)


buger:
Can you pls create a ticket for it, so we can track in project board?

PR #2528, opened 2019-09-19
Supporting dlpython

alephnull:
- dropped arm64 builds
- dropped Lua builds
- python init script identical to vanilla gateway, provided for
  backward compatibility

Multilib setup is in the environment section of the buddy pipeline.


matiasinsaurralde:
Note that `coprocess` tag isn't required anymore. See [here](https://github.com/TykTechnologies/tyk/pull/1875#issue-209697750).

For Lua: `go build -tags 'lua'`.

PR #2533, opened 2019-09-20
Add support for whitelisting ports

### LINKED ISSUE #2529: TCP proxy should use port whitelisting instead of blacklisting
master/release-2.9

**Describe the bug**
At the moment it exposes security issue because by default you can bind any service to any port. At tyk.conf level, we should "whitelist" available ports. It means that in default installation TCP proxy won't work unless you whitelist ports.

gernest:
Add support for whitelisting ports

This change introduces `ports_whitelist` configuration option to `tyk.conf`, which
defines which ports should the gateway allow services to be bind to.

example configuration is like this

```json
{
  "ports_whitelist": {
    "http": {
      "ranges": [
        {
          "from": 8000,
          "to": 9000
        }
      ]
    },
    "tcp": {
      "ranges": [
        {
          "from": 7001,
          "to": 7900
        }
      ]
    },
    "tls": {
      "ports": [
        6000,
        6015
      ]
    }
  }
}
```

You define the protocol `http`, `https`, `tcp` or `tls` which maps to an object 
with schema

```json
{
    "type": [
        "object"
    ],
    "additionalProperties": false,
    "properties": {
        "ranges": {
            "type": "array",
            "items": {
                "type": "object",
                "additionalProperties": false,
                "properties": {
                    "from": {
                        "type": "integer"
                    },
                    "to": {
                        "type": "integer"
                    }
                }
            }
        }
    },
    "ports": {
        "type": "array",
        "items": {
            "type": "integer"
        }
    }
}
```

# Defining specific ports

You can  configure a specific port to be available like this.

```json
{
    "ports_whitelist": {
        "http": {
            "ports": [
                8089,
                8090
            ]
        }
    }
}
```

This configuration tells the gateway to allow `http` services only on port `8089`
and `8090`

# Defining port ranges

port ranges is the boundary of allowed ports inclusively. This helps to avoid 
tedious specifying each possible ports.

Example

```json
{
    "ports_whitelist": {
        "http": {
            "ranges": [
                {
                    "from": 8000,
                    "to": 9000
                }
            ]
        }
    }
}
```

This configures the gateway to allow any port withing `8000` `...` `9000`
boundary inclusively meaning `8000` ,`8050`, `9000` will all be allowed.

# Precedence

Ports defined in `ports` property takes precedence over `ranges`. This means if
a port is allowed inside `ports` property no further checks will be done to see if
its within range.

Fixes #2529

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -266,15 +266,15 @@ type PortWhiteList struct {
 	Ports  []int       `json:"ports,omitempty"`
 }
 
-// Ok returns true if port is acceptable from the PortWhiteList.
-func (p PortWhiteList) Ok(port int) bool {
+// Match returns true if port is acceptable from the PortWhiteList.
+func (p PortWhiteList) Match(port int) bool {
 	for _, v := range p.Ports {
 		if port == v {
 			return true
 		}
 	}
 	for _, r := range p.Ranges {
-		if r.Ok(port) {
+		if r.Match(port) {
 			return true
 		}
 	}
@@ -287,8 +287,8 @@ type PortRange struct {
 	To   int `json:"to"`
 }
 
-// Ok returns true if port is within the range
-func (r PortRange) Ok(port int) bool {
+// Match returns true if port is within the range
+func (r PortRange) Match(port int) bool {
 	return r.From <= port && r.To >= port
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -325,7 +325,7 @@ type Config struct {
 	EnableAPISegregation    bool           `json:"enable_api_segregation"`
 	TemplatePath            string         `json:"template_path"`
 	Policies                PoliciesConfig `json:"policies"`
-	EnablePortWhiteList     bool           `json:"enable_ports_whitelist"`
+	DisablePortWhiteList    bool           `json:"disable_ports_whitelist"`
 	// Defines the ports that will be available for the api services to bind to.
 	// This is a map of protocol to PortWhiteList. This allows per protocol
 	// configurations.
--- a/config/config.go
+++ b/config/config.go
@@ -325,6 +325,7 @@ type Config struct {
 	EnableAPISegregation    bool           `json:"enable_api_segregation"`
 	TemplatePath            string         `json:"template_path"`
 	Policies                PoliciesConfig `json:"policies"`
+	EnablePortWhiteList     bool           `json:"enable_ports_whitelist"`
 	// Defines the ports that will be available for the api services to bind to.
 	// This is a map of protocol to PortWhiteList. This allows per protocol
 	// configurations.
--- a/config/config.go
+++ b/config/config.go
@@ -260,6 +260,38 @@ type ServicePort struct {
 	Port     int    `json:"port"`
 }
 
+// PortWhiteList defines ports that will be allowed by the gateway.
+type PortWhiteList struct {
+	Ranges []PortRange `json:"ranges,omitempty"`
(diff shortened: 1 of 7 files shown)


gateway/gateway_test.go
```diff
@@ -844,6 +844,9 @@ func TestProxyProtocol(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	EnablePort(p, "tcp")
```
buger: Should not it be "http" instead?

gateway/proxy_muxer.go
```diff
@@ -404,15 +404,28 @@ func target(listenAddress string, listenPort int) string {
...
+				return nil
+			}
 		}
 	}
+	return fmt.Errorf("%s:%d trying to open disabled port", protocol, listenPort)
+}
 
+func (m *proxyMux) generateListener(listenPort int, protocol string) (l net.Listener, err error) {
+	listenAddress := config.Global().ListenAddress
+	if err := CheckPortWhiteList(config.Global(), listenPort, protocol); err != nil {
```
buger: Do not see reasons bassing config object as argument.

gateway/proxy_muxer.go
```diff
@@ -404,15 +404,28 @@ func target(listenAddress string, listenPort int) string {
...
 }
 
-func (m *proxyMux) generateListener(listenPort int, protocol string) (l net.Listener, err error) {
-	listenAddress := config.Global().ListenAddress
-	disabled := config.Global().DisabledPorts
-	for _, d := range disabled {
-		if d.Protocol == protocol && d.Port == listenPort {
-			return nil, fmt.Errorf("%s:%d trying to open disabled port", protocol, listenPort)
+func CheckPortWhiteList(cfg config.Config, listenPort int, protocol string) error {
+	if protocol == "" || protocol == "http" || protocol == "https" {
```
buger: Instead of this logic, lets just set PortWhiteList vars on the gateway start. Like (pseudocode):
```
protocol = "http"
if cfg.SSLEnabled {
protocol = "https"
}

   ports := cfg.PortWhiteList[protocol].Ports
   ports := append(ports, cfg.ListenPort)
   cfg.PortWhiteList[protocol].Ports = ports
```

config/config.go
```diff
@@ -260,6 +260,38 @@ type ServicePort struct {
...
+}
+
+// PortRange defines a range of ports inclusively.
+type PortRange struct {
+	From int `json:"from"`
+	To   int `json:"to"`
+}
+
+// Ok returns true if port is within the range
+func (r PortRange) Ok(port int) bool {
```
buger: Lets change it to more idiomatic name, like "Between"

config/config.go
```diff
@@ -260,6 +260,38 @@ type ServicePort struct {
...
 }
 
+// PortWhiteList defines ports that will be allowed by the gateway.
+type PortWhiteList struct {
+	Ranges []PortRange `json:"ranges,omitempty"`
+	Ports  []int       `json:"ports,omitempty"`
+}
+
+// Ok returns true if port is acceptable from the PortWhiteList.
+func (p PortWhiteList) Ok(port int) bool {
```
buger: Same as below, lets change name to "Match" or similar.

buger:
Additionally pls add a flag to disable whitelisting.

gateway/gateway_test.go
gernest: nope, the test case is for `tcp` service.

PR #2536, opened 2019-09-23
Abandoning tyk-build-env, images will live in their respective repos

alephnull:
Moving code from tyk-build-env (henceforth abandoned).
Makes the pipelines simpler. 
See pipeline at https://app.buddy.works/tyk-projects/tyk/pipelines/pipeline/212040/execution/5d8895a4aab2cb43429d7461

### DIFF
--- a/images/README.md
+++ b/images/README.md
@@ -1,0 +1,50 @@
+# build-env
+
+Docker environment used to build official images and plugins.
+
+This is the base image that will slowly be used in all our builds. It
+is not capable of handling i386 or arm64 builds. Those builds are
+handled by installing additional components in the environment section
+of the pipeline.
+
+This image will need to be updated only when upgrading the go version
+or if some system dependencies for building change. This image is
+mainly used internally at Tyk for CD pipelines.
+
+# plugin-compiler
+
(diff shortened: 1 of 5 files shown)


buger:
Can we remove .tar.gz files from this PR?

alephnull:
Those are the older libraries required for Python3.4. They will not be updated, and it would good to keep them around.

buger:
Why do we need them if our master and 2.9 does not depend on this python installations?

alephnull:
The tarballs for these ancient versions were not easily available and its not going to get easier. The reasoning behind checking them in was if we have to reproduce an old environment.

I'm happy to remove them but it is a small price to pay for having them around.

buger:
We do not need to maintain compatibility for old environment here. Just clone current Buddy pipelines (if we decide to do a release for old release), and update new ones to use only new build env, without this dependencies.

excieve:
Agree with @buger on this. If it's becoming difficult to get the original source tarballs for these dependencies, we could put them on some internal S3 or just CI virual FS.

alephnull:
@buger Is this ok to merge?

PR #2544, opened 2019-09-24
Bugfix/fix custom auth

### LINKED ISSUE #2543: I shouldn't need to set "token" metadata when using ID extractor in Custom Auth
All - when I am using custom authentication plugin and returning a session to Tyk, Tyk gateway will throw an error if the session doesn't include metadata with a "token" and ANY value.  

**Describe the bug**
If I'm using ID extractor, I don't need to set "token".  "token" only needs to be set if I am NOT using ID Extractor.

```javascript
object.Session = &coprocess.SessionState{

		LastUpdated:         time.Now().String(),
		Rate:                5,
		Per:                 10,
		QuotaMax:            int64(0),
		QuotaRenews:         time.Now().Unix(),
		IdExtractorDeadline: extractorDeadline,
		Metadata: map[string]string{
			"jwt":   jot,
			"Hello": "world",
		},
		ApplyPolicies: policiesToApply,
	}
```

This throws an error by Tyk because we didn't set `token`

sedkis:
fix #2543

### DIFF
--- a/gateway/coprocess.go
+++ b/gateway/coprocess.go
@@ -374,7 +374,7 @@ func (m *CoProcessMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Requ
 	// Is this a CP authentication middleware?
 	if m.Spec.EnableCoProcessAuth && m.HookType == coprocess.HookType_CustomKeyCheck {
 		// The CP middleware didn't setup a session:
-		if returnObject.Session == nil || token == "" {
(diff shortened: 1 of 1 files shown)


sedkis:
Which branch does this need to get merged into?

buger:
Everything should be merged to master first

sedkis:
aha. ok i will rebase

sedkis:
done, @buger

buger:
Do you want to try to modify existing tests to confirm this fix?

PR #2562, opened 2019-09-26
When CGO is disabled only build an API compat stub, enables arm64 builds

excieve:
Cross-compiling CGO requires a complete GCC tool chain along with libc. This is a bit more difficult with arm/aarch64 when the build host is x86. We don't currently set up such an environment.

This PR makes sure we can still cross-compile aarch64 (arm64) with CGO disabled. In order to make this work the Python coprocess dispatcher is stubbed to maintain API compatibility.

### DIFF
--- a/bin/dist_build.sh
+++ b/bin/dist_build.sh
@@ -58,7 +58,9 @@ do
 done
 
 echo "Building Tyk binaries"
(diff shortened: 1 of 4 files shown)


excieve:
@alephnull FYI

buger:
Looks good!

PR #2576, opened 2019-09-27
Go redis driver

### LINKED ISSUE #2497: /tyk/keys is not returning all the keys, if redis cluster mode is enabled
master

**Describe the bug**
Key listing api `tyk/keys` is not returning all the keys stored in redis cluster. It is fetching keys from single node, instead of fetching from all the nodes of the cluster

**Reproduction steps**

1. Create redis cluster with 3 master and 3 slave nodes.
2. Connect gateway to redis cluster
3. Create key with id `l7xxcfcb9079118c4358b45b4db203699601`
4. Call `/tyk/keys` endpoint. 

The key will not be listed in the response

**Actual behavior**
l7xxcfcb9079118c4358b45b4db203699601 is not listed in response of  `/tyk/keys` endpoint

**Expected behavior**
l7xxcfcb9079118c4358b45b4db203699601 should be returned in `/tyk/keys` response

komalsukhani:
Fixes #2497 
1. Replace RedigoCluster driver with go-redis driver
2. Replace `Keys pattern` command with `Scan pattern`, as it is recommended in production environment.

Note: Go-redis driver doesn't have an option for MaxIdle connections, instead it provides option for MinIdle connections. So, `optimisation_max_idle` field of tyk.conf could not be used.


storage/redis_cluster.go
```diff
@@ -123,46 +160,69 @@ func NewRedisClusterPool(isCache bool) *rediscluster.RedisCluster {
...
+			Addr:         seedHosts[0],
+			Password:     cfg.Password,
+			DB:           cfg.Database,
+			DialTimeout:  timeout,
+			ReadTimeout:  timeout,
+			WriteTimeout: timeout,
+			IdleTimeout:  240 * time.Second,
+			PoolSize:     poolSize,
+			TLSConfig:    tlsConfig,
+		})
```
asoorm: Is it worth using `NewUniversalClient` which is capable of connecting to single redis server, redis cluster and also sentinel?

```
// NewUniversalClient returns a new multi client. The type of client returned depends
// on the following three conditions:
//
// 1. if a MasterName is passed a sentinel-backed FailoverClient will be returned
// 2. if the number of Addrs is two or more, a ClusterClient will be returned
// 3. otherwise, a single-node redis Client will be returned.
func NewUniversalClient(opts *UniversalOptions) UniversalClient {
```

komalsukhani: @asoorm Thank you. Great catch. I was not happy with interface idea.

PR #2588, opened 2019-10-04
Fix ci failure caused by transitioning to go modules

gernest:


.travis.yml
```diff
@@ -48,7 +45,6 @@ script:
   - go build -tags 'coprocess python'
   - go build -tags 'coprocess lua'
   - go build -tags 'coprocess grpc'
-  - ./bin/ci-swagger.sh
```
adelowo: We still need to validate the swagger file is valid at all times. Not sure these lines should go off,

adelowo: We still need to validate the swagger file is valid at all times. Not sure these lines should go off,

buger: We need to find a way work with modules. Can be separate task.

PR #2596, opened 2019-10-07
Add analytics upstream latency info 

### LINKED ISSUE #2590: Extend analtyics data to store latency information
Store below latency information in the analytics
1. total latency
2. upstream latency

Data will have precision of ms

gernest:
Closes #2590 
Related https://github.com/TykTechnologies/tyk-pump/issues/151
Related  https://github.com/TykTechnologies/tyk-analytics/issues/1526

### DIFF
--- a/gateway/analytics.go
+++ b/gateway/analytics.go
@@ -39,7 +39,7 @@ func (n *NetworkStats) Flush() NetworkStats {
 
 type Latency struct {
 	Total    int64
-	Upstream time.Duration
+	Upstream int64
 }
 
 // AnalyticsRecord encodes the details of a request
--- a/gateway/analytics.go
+++ b/gateway/analytics.go
@@ -330,3 +330,7 @@ func (r *RedisAnalyticsHandler) recordWorker() {
 		}
 	}
 }
+
+func DurationToMillisecond(d time.Duration) float64 {
+	return float64(d) / 1e6
+}
--- a/gateway/analytics.go
+++ b/gateway/analytics.go
@@ -38,7 +38,7 @@ func (n *NetworkStats) Flush() NetworkStats {
 }
 
 type Latency struct {
-	Total    time.Duration
+	Total    int64
 	Upstream time.Duration
 }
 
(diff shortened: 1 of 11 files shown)


buger:
I assume require pump or too? Also does it handle tcp analytics part too?

gernest:
@buger This is still a work in progress will ping you when ready. Only works for http services. Reasoning about tcp right now.

> I assume require pump or too?
Correct, a PR for that is underway.

One use case is already covered `RequestTime` which is the total request time. Only needed to handle upstream latency.

gernest:
After much thought, I believe this makes sense only to cover tcp services. Because of the following reasons

- For tcp analytics stats are not per connection, they are fired during the lifetime of the connection. This makes `upstream latency` blurry, because the connection can last forever, something that comes that close is maybe measuring the time it takes to read /write to upstream which IMHO is not relevant here as its something else entirely.

- Total request time ( see my point above). `n` number of stats will be fired during a single tcp connection. For http only one stat, meaning it is possible to record meaningful total duration in a single analytic record.

So @buger This is ready to be merged. It is backward compatible so everything should just work no breaking changes (no need to wait for tyk-pump)

komalsukhani:
@buger @gernest If `RequestTime` and Total Latency are same then can we skip Total Latency field?

asoorm:
> After much thought, I believe this makes sense only to cover http services. Because of the following reasons

If I configure Tyk as a TCP proxy to a http service, then latency will be useful for me @gernest 

e.g. a L4 load balancer scenario? Considerably more performant than L7.

so dependent on the service type - is it a long/short lived connection - latency might be handy. because http can still go over tcp.

buger:
@komalsukhani I think it’s done for backward compatibility

buger:
@asoorm in TCP mode it will be just packets not HTTP messages. One HTTP message can be 10 TCP packets, and latency, in this case, will show a completely different picture. Not telling if its a bit more complex case like HTTP2 with multiple streams.

buger:
Looks good to me at this point

PR #2606, opened 2019-10-09
key session update fixed

dencoded:
added changed for https://github.com/TykTechnologies/tyk-analytics/issues/1535

we needed to do update for legacy and new format key as well. might worth to implement MSET if this looks good.


buger:
Will merge, but we really need to add tests for it.

PR #2608, opened 2019-10-09
Remove references to tyk-gateway-lua

alephnull:
Completely remove references to lua version.


install/post_install.sh
```diff
@@ -39,7 +36,6 @@ if [ -d "$SYSV1" ]; then
 	[ -f /etc/default/tyk-gateway-python ] || cp $DIR/inits/sysv/default/tyk-gateway-python /etc/default/
 	[ -f /etc/default/tyk-gateway-lua ] || cp $DIR/inits/sysv/default/tyk-gateway-lua /etc/default/
```
excieve: Looks like there are more references still.

PR #2609, opened 2019-10-09
Remove references to tyk-gateway-lua

alephnull:
Completes removal of lua references


install/post_remove.sh
```diff
@@ -12,11 +12,9 @@ if [ -f "/lib/systemd/system/tyk-gateway.service" ]; then
 	echo "Stopping the service"
 	systemctl stop tyk-gateway.service
 	systemctl stop tyk-gateway-python.service
-	systemctl stop tyk-gateway-lua.service
```
excieve: Stopping and deleting init files is still needed. Think of upgrading from an older version, where things are first removed and then installed. However, make sure to ignore any errors if those files are not found.

PR #2616, opened 2019-10-09
release API spec resources when reload APIs

dencoded:
this PR just brings back this change: https://github.com/TykTechnologies/tyk/pull/2256/files#diff-ad28d87199f22b3b034c7fc355476a6dR672

also, we need re-enable this test if we want to make sure no leaks happening (connected with CB logic): https://github.com/TykTechnologies/tyk/blob/master/gateway/gateway_test.go#L831

### DIFF
--- a/gateway/api_loader.go
+++ b/gateway/api_loader.go
@@ -692,6 +692,12 @@ func loadApps(specs []*APISpec) {
 
 	// Swap in the new register
 	apisMu.Lock()
+
+	// release current specs resources before overwriting map
+	for _, curSpec := range apisByID {
+		curSpec.Release()
+	}
+
 	apisByID = tmpSpecRegister
 	apisMu.Unlock()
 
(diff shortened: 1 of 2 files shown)


gateway/api_loader.go
```diff
@@ -692,6 +692,12 @@ func loadApps(specs []*APISpec) {
...
 	// Swap in the new register
 	apisMu.Lock()
+
+	// release current specs resources before overwriting map
+	for _, curSpec := range apisByID {
+		curSpec.Release()
+	}
+
 	apisByID = tmpSpecRegister
 	apisMu.Unlock()
```
gernest: I think the locking logic here can be improved. 

```go
	apisMu.RLock()
	// release current specs resources before overwriting map
	for _, curSpec := range apisByID {
		curSpec.Release()
	}
	apisMu.RUnlock()

	apisMu.Lock()
	apisByID = tmpSpecRegister
	apisMu.Unlock()
```

We are first reading from `apisByID` so we use `RLock`  and then writing to it using `Lock`

buger: In this implementation, it should be fine.

Issue #2619, opened 2019-10-11
Proxy on internal api does not wrok


ilijabojanovic:
**Branch/Environment/Version**
- Branch/Version: Release-2.9
- Environment: All

**Describe the bug**
When we set target_path to internal api (`tyk://interal`) proxying does not work

**Reproduction steps**

1. Add internal api
2. Add regular api and in target specify url of internal api
3. Send request to regular api

**Actual behavior**
Request ins't forwarded to internal api

**Expected behavior**
Request should pass

**Screenshots/Video**

**Logs (debug mode or log file):**
```
[Oct 11 14:22:51] DEBUG Outbound Request: tyk://internal/headers
[Oct 11 14:22:51] ERROR proxy: http: proxy error: unsupported protocol scheme "tyk" api_id=59927da0564c42097d5a7ff2cc1a30ec org_id=5d9f58489633d7135818b5b5 server_name=internal user_id=-- user_ip=127.0.0.1 user_name=
```

**Configuration (tyk config file):**
Testing definition
```
{
    "id": "5d9f58679633d7135818b5b9",
    "name": "api",
    "slug": "api",
    "listen_port": 0,
    "protocol": "",
    "enable_proxy_protocol": false,
    "api_id": "59927da0564c42097d5a7ff2cc1a30ec",
    "org_id": "5d9f58489633d7135818b5b5",
    "use_keyless": true,
    "use_oauth2": false,
    "use_openid": false,
    "openid_options": {
        "providers": [],
        "segregate_by_client": false
    },
    "oauth_meta": {
        "allowed_access_types": [],
        "allowed_authorize_types": [],
        "auth_login_redirect": ""
    },
    "auth": {
        "use_param": false,
        "param_name": "",
        "use_cookie": false,
        "cookie_name": "",
        "auth_header_name": "Authorization",
        "use_certificate": false,
        "validate_signature": false,
        "signature": {
            "algorithm": "",
            "header": "",
            "secret": "",
            "allowed_clock_skew": 0,
            "error_code": 0,
            "error_message": ""
        }
    },
    "use_basic_auth": false,
    "basic_auth": {
        "disable_caching": false,
        "cache_ttl": 0,
        "extract_from_body": false,
        "body_user_regexp": "",
        "body_password_regexp": ""
    },
    "use_mutual_tls_auth": false,
    "client_certificates": [],
    "upstream_certificates": {},
    "pinned_public_keys": {},
    "enable_jwt": false,
[...]

PR #2628, opened 2019-10-23
gateway: allow proxy on internal APIs (#2619)

matiasinsaurralde:
After inspecting different versions I didn't really find this functionality, seems the code from #1859 didn't include this. I've reused some of its logic in order to achieve what's expected in #2619.
Suggestions are welcome.


matiasinsaurralde:
CI is fixed 👍

komalsukhani:
@matiasinsaurralde Can you add test cases?

buger:
Are you sure this change does the job? Lets add a test as mentioned above.
The thing that worries me that on line 492 we already check if it is looping or not https://github.com/TykTechnologies/tyk/pull/2628/files#diff-5c368c424b98887710e47547295e3706L492.

PR #2630, opened 2019-10-24
Don't skip the irrelevant middlewares with key when ignore plugin is …

### LINKED ISSUE #2579: Ignore plugin can't be used with URL rewriting
- Branch/Version: [e.g. Master/Release/Stable/Feature branch]
- Environment: [e.g. On-prem/Hybrid/MDCB]

**Describe the bug**
The main purpose of "ignore" plugin is to bypass authentification, however, it also bypass lot of other middlewares like URL rewrite. 

**Reproduction steps**
- Add endpoint with URL rewrite and Ignore plugins
- URL rewrite won't work

furkansenharputlu:
Save `RequestStatus` to context and add check for `Ignore` to skip in middlewares requiring auth header.

Fixes #2579

### DIFF
--- a/ctx/ctx.go
+++ b/ctx/ctx.go
@@ -32,6 +32,7 @@ const (
 	Trace
 	CheckLoopLimits
 	Definition
+	RequestStatus
 )
 
 func setContext(r *http.Request, ctx context.Context) {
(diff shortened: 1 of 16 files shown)


buger:
What about the rest of the auth middlewares, like JWT or custom auth?

PR #2631, opened 2019-10-24
Update ci-test.sh

furkansenharputlu:
This PR removes redundant `LATEST_GO` checks in `ci-test.sh` script. Also, it renames the script name as  `ci-tests.sh` to be compatible with `tyk-analytics`.


buger:
Why it was updated like this? Can you update description?

PR #2638, opened 2019-10-27
coprocess: enhance error handling in ObjectPostProcess

matiasinsaurralde:
This small tweak could be useful in scenarios where the request URL is modified by a plugin (and the URL is invalid or empty), for example:

```python
from tyk.decorators import *
print("middleware.py is loaded")
@Hook
def myhook(request, session, spec):
  print("myhook called")
  request.object.url = ''
  return request, session
```
In the current implementation, when `ParseRequestURI` fails, the following line panics because of a nil `r.URL`:
```golang
r.URL.RawQuery = values.Encode()
```

### DIFF
--- a/gateway/coprocess.go
+++ b/gateway/coprocess.go
@@ -172,7 +172,7 @@ func (c *CoProcessor) BuildObject(req *http.Request, res *http.Response) (*copro
 }
 
 // ObjectPostProcess does CoProcessObject post-processing (adding/removing headers or params, etc.).
-func (c *CoProcessor) ObjectPostProcess(object *coprocess.Object, r *http.Request) {
+func (c *CoProcessor) ObjectPostProcess(object *coprocess.Object, r *http.Request) (err error) {
 	r.ContentLength = int64(len(object.Request.RawBody))
 	r.Body = ioutil.NopCloser(bytes.NewReader(object.Request.RawBody))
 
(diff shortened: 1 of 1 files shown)


gateway/coprocess.go
```diff
@@ -193,8 +193,12 @@ func (c *CoProcessor) ObjectPostProcess(object *coprocess.Object, r *http.Reques
 		values.Set(p, v)
 	}
 
-	r.URL, _ = url.ParseRequestURI(object.Request.Url)
+	r.URL, err = url.ParseRequestURI(object.Request.Url)
+	if err != nil {
+		return err
```
dencoded: having named return parameter and use it in `return` statement looks a little odd

matiasinsaurralde: Updated

gateway/coprocess.go
```diff
@@ -193,8 +193,12 @@ func (c *CoProcessor) ObjectPostProcess(object *coprocess.Object, r *http.Reques
 		values.Set(p, v)
 	}
 
-	r.URL, _ = url.ParseRequestURI(object.Request.Url)
+	r.URL, err = url.ParseRequestURI(object.Request.Url)
+	if err != nil {
+		return
```
letzya: @matiasinsaurralde don't we want to print this error or return it?

letzya: @matiasinsaurralde consider using Parse which calls internal parse with `viaRequest` set to false

letzya: Also maybe we should add a test (since we don't check if `r.url == nil` and we trust that function that if err is nil then the pointer isn't nil, but someone might change it at some point, maybe to allow empty string or nil as an input etc.)

letzya: Last thing (I hope :) ) - could it be that the right fix would be to change the function to use Parse?  https://golang.org/src/net/url/url.go line 472 since the request to the plugin can come from anywhere so maybe it's too much to assume, and we should allow this use case?

matiasinsaurralde: @letzya About printing/returning the error, with this change, it will achieve that.
The first call is `ProcessRequest`, which dispatches the object and calls `ObjectPostProcess` with the modified request object. In case `ObjectPostProcess` fails, we will see the error!

PR #2641, opened 2019-10-28
gateway: provide a fallback for Python inspect.stack() (#2640) (WIP)

### LINKED ISSUE #2640: Python inspect module changes
- Branch/Version: All
- Environment: On-prem

**Describe the bug**
The [`inspect` module ](https://docs.python.org/3/library/inspect.html#inspect.stack) has changed its syntax after Python 3.4.x/3.5x. We use `inspect.stack` for handling local imports in plugins (see #2402).

When using these versions, the following issue may occur:
```
AttributeError: 'tuple' object has no attribute 'function'
```

We should add fallback logic to support this.

**Reproduction steps**
- Setup Python 3.4.9 or 3.5.6.
- Create a Python plugin bundle.
- Setup an API and start or reload the gateway.

**Actual behavior**
Python plugin initialization fails with `inspect.stack` related errors.

**Expected behavior**
The custom loader should have a fallback for this scenario.

matiasinsaurralde:
Fixes #2640

### DIFF
--- a/coprocess/python/tyk/loader.py
+++ b/coprocess/python/tyk/loader.py
@@ -10,9 +10,9 @@ class MiddlewareLoader():
       # Inspect the stack and verify if the "import" call is local (direct call from middleware code) or not:
       is_local = False
       for fr in stack:
-        if fr.function != "<module>":
+        if fr[3] != "<module>":
           continue
-        if self.base_path not in fr.filename:
+        if self.base_path not in fr[1]:
           break
         is_local = True
       return is_local


matiasinsaurralde:
Sorry, I replied on the wrong place:
@buger I have updated this PR after several tests with Python 3.5.x/3.4.x/3.5.x. A fallback wasn't the best solution, instead we can rely on tuple indexes, they're available in versions that support named attributes too 🙌

PR #2644, opened 2019-10-30
SSL Force Common Name Check

komalsukhani:
Added new option in gateway configuration `ssl_force_common_name_check`.

If an SSL certificate has non-empty SAN, it is used for validating hostname, ignoring Common Name field of Certificate.(This is recommended, as most client have dropped support of CN)

But if a user has legacy certificate and wants to enforce check against Common Name, he can enable `ssl_force_common_name_check`

### DIFF
--- a/apidef/schema.go
+++ b/apidef/schema.go
@@ -51,9 +51,9 @@ const Schema = `{
         "use_standard_auth": {
             "type": "boolean"
         },
-		"use_go_plugin_auth": {
-			"type": "boolean"
-		},
+        "use_go_plugin_auth": {
+            "type": "boolean"
+        },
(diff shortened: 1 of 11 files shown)


buger:
@komalsukhani modified PR to allow configuration on API level

apidef/schema.go
```diff
@@ -319,6 +319,9 @@ const Schema = `{
                         },
                         "proxy_url": {
                             "type": "string"
+                        },
+                        "ssl_force_common_name_check": {
+                            "type": "bool"
```
adelowo: This should be `boolean`

buger: done

Issue #2381, opened 2019-07-05
Make oAuth error code configurable


buger:
**Do you want to request a *feature* or report a *bug*?**
feature
**What is the current behavior?**
At the moment we respond with 403 error code, and its hardcoded here https://github.com/TykTechnologies/tyk/blob/632d40500d3a7519a7077f236d3b477e8bb0ae24/gateway/server.go#L488

RFC at the same time says that default option should be 400 https://tools.ietf.org/html/rfc6749#page-43, but leaves the window for custom codes as well.
**What is the expected behavior?**
It should be possible to tune oAuth error code via configuration.
**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
All

PR #2648, opened 2019-10-31
Make oauth error status code configurable

furkansenharputlu:
This PR 
- makes oauth error status code configurable by adding a new config variable `oauth_error_status_code`
- changes default value from `403` to `400` as mentioned in the issue.

Fixes https://github.com/TykTechnologies/tyk/issues/2381

### DIFF
--- a/cli/linter/schema.json
+++ b/cli/linter/schema.json
@@ -668,6 +668,9 @@
     "oauth_token_expired_retain_period": {
       "type": "integer"
     },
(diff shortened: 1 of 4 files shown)


buger:
I think I confused you a bit with the ticket definition. 

This change breaks compatibility, so make default oauth code exactly the same as before.

We just need a way to override it.

furkansenharputlu:
@buger updated

PR #2650, opened 2019-11-01
Bug coprocess auth check policy

asoorm:
CoProcess Auth should check the policy. Should also be fixed in master.

### DIFF
--- a/gateway/coprocess.go
+++ b/gateway/coprocess.go
@@ -362,12 +362,13 @@ func (m *CoProcessMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Requ
 		if returnObject.Session == nil || token == "" {
 			authHeaderValue := r.Header.Get(m.Spec.Auth.AuthHeaderName)
 			AuthFailed(m, r, authHeaderValue)
-			return errors.New(http.StatusText(http.StatusForbidden)), http.StatusForbidden
+			return errors.New(http.StatusText(http.StatusUnauthorized)), http.StatusUnauthorized
 		}
 
 		returnedSession := TykSessionState(returnObject.Session)
 
 		if err := m.ApplyPolicies(returnedSession); err != nil {
+			AuthFailed(m, r, r.Header.Get(m.Spec.Auth.AuthHeaderName))
 			return errors.New(http.StatusText(http.StatusForbidden)), http.StatusForbidden
 		}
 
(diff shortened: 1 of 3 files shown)


gateway/coprocess.go
```diff
@@ -362,11 +362,15 @@ func (m *CoProcessMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Requ
 		if returnObject.Session == nil || token == "" {
 			authHeaderValue := r.Header.Get(m.Spec.Auth.AuthHeaderName)
 			AuthFailed(m, r, authHeaderValue)
-			return errors.New("Key not authorised"), 403
+			return errors.New(http.StatusText(http.StatusForbidden)), http.StatusForbidden
 		}
 
 		returnedSession := TykSessionState(returnObject.Session)
 
+		if err := m.ApplyPolicies(returnedSession); err != nil {
```
letzya: shouldn't this be before TykSessionState?

matiasinsaurralde: @letzya Actually before line 368, we don't know the contents of the session. So it's ok to call `ApplyPolicies` after it.

gateway/coprocess.go
```diff
@@ -362,11 +362,15 @@ func (m *CoProcessMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Requ
 		if returnObject.Session == nil || token == "" {
 			authHeaderValue := r.Header.Get(m.Spec.Auth.AuthHeaderName)
 			AuthFailed(m, r, authHeaderValue)
-			return errors.New("Key not authorised"), 403
+			return errors.New(http.StatusText(http.StatusForbidden)), http.StatusForbidden
```
adelowo: Not sure `http.StatusText` is the best approach here. It will return "Forbidden". I'd say ___Key not authorised___ is way more descriptive and easier to debug.

asoorm: I think that this is a security leak and introduces an attack vector. If I hit the api with a bad key i will know that the server responding is Tyk.

Also, i've tweaked it a bit to use NotAuthorized & Forbidden. PTAL

christtyk:
@buger are you happy adding to 2.9.2 as ahmet says fix done?

asoorm:
Bump @buger @christtyk

buger:
@asoorm can you rebase this PR so it will go to master first?

asoorm:
@buger Closing in favour of #2798 because rebasing is a bit of a mess

PR #2651, opened 2019-11-04
Add AuthConfigs field to APISpec

### LINKED ISSUE #2580: JWT auth can't be combined with Auth token
master/2.9

**Describe the bug**
The goal is to perform JWT validation and log user using standard auth token.

JWT authentification use auth.Header value to determine HTTP header name. 
Auth token middleware use exactly the same field to the same purpose.
So you can't have multiple auth mode using both, because you can't set them to read multiple headers. 

**Expected behavior**
It should be possible to set distinct headers for both MWs. We should add a separate header field for JWT middleware.
Also for Oidc, oauth, a certificate authorization.

furkansenharputlu:
Fixes #2580 

This PR adds a new field 

`AuthConfigs  map[string]AuthConfig` 

to `APISpec`. The idea is to have special auth configuration for authentication types needing Auth token. 

Right now the following middlewares need Auth token, please let me know if I miss something:

- `JWTMiddleware` -> `jwt`
- `AuthKey` -> `authToken`
- `HTTPSignatureValidationMiddleware` -> `hmac`
- `Oauth2KeyExists` -> `oauth`
- `BasicAuthKeyIsValid` -> `basic`
- `CoProcessMiddleware` -> `coprocess`

Example configuration:

```
"auth_configs": {
	"basic": {"auth_header_name": "My-Basic-Auth-Key"},
	"authToken": {"auth_header_name": "My-Auth-Key"}
},
```

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -366,8 +366,9 @@ type APIDefinition struct {
 		AllowedAuthorizeTypes  []osin.AuthorizeRequestType `bson:"allowed_authorize_types" json:"allowed_authorize_types"`
 		AuthorizeLoginRedirect string                      `bson:"auth_login_redirect" json:"auth_login_redirect"`
 	} `bson:"oauth_meta" json:"oauth_meta"`
-	Auth         Auth `bson:"auth" json:"auth"`
-	UseBasicAuth bool `bson:"use_basic_auth" json:"use_basic_auth"`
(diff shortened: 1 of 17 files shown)


matiasinsaurralde:
Would this affect [coprocess.go](https://github.com/TykTechnologies/tyk/blob/master/gateway/coprocess.go#L376), when using CP auth?

furkansenharputlu:
@matiasinsaurralde thanks for the warning. I changed the auth header value getting for coprocess as well.

PR #2656, opened 2019-11-05
fix opentracing headers propagation

### LINKED ISSUE #2655: opentracing headers are not propagated  to upstream services
When open tracing is enabled. Calling api's doesn't properly propagate tracing headers.<br/><br/>This has been verified for both zipkin and jaegerChris Taylor

gernest:
Fixes #2655

For some reason the changes which was calling `trace.AddTracer` were lost and as a result, traces for api's were never initialised.

This change ensures that we call `trace.AddTracer` for all http api's.


gernest:
This is ready for review. The test failure has nothing to do with this, it seems there is a data race detected when running plugin tests.

PR #2660, opened 2019-11-06
Circuit breaker improvement

tbuchaillot:
Tyk's Circuit Breaker only works for 500 errors. This is pretty limiting. e.g. the breaker will not trip for failed upstreams.

This PR fixes that in order to break the circuit for any 5XX errors or if RoundTrip returns an error.

### DIFF
--- a/gateway/reverse_proxy_test.go
+++ b/gateway/reverse_proxy_test.go
@@ -2,6 +2,7 @@ package gateway
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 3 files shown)


gateway/reverse_proxy_test.go
```diff
@@ -347,6 +348,38 @@ func TestWrappedServeHTTP(t *testing.T) {
...
 }
 
+func TestCircuitBreaker5xxs(t *testing.T) {
+	ts := StartTest()
+	defer ts.Close()
+
+	t.Run("Extended Paths", func(t *testing.T) {
+		BuildAndLoadAPI(func(spec *APISpec) {
+			UpdateAPIVersion(spec, "v1", func(v *apidef.VersionInfo) {
+				json.Unmarshal([]byte(`[
```
buger: Feels like formatting error.

buger:
Apart from small formatting bug, it works

gateway/reverse_proxy_test.go
tbuchaillot: fixed

PR #2674, opened 2019-11-18
Hybrid DRL

gernest:
Fixes https://github.com/TykTechnologies/product/issues/229
Fixes https://github.com/TykTechnologies/product/issues/230

This adds `drl_threshold` configuration value that is used to fine tune selection of rate limiting strategy 
between distributed rate limiter and redis rate limiter.

There are some cases based on the number of servers that distributed rate limiter 
works best and there are cases it doesn't. 

For a single server `drl` is used by default. If more than one gateway are running
then if `numer_of_servers*drl_threshold > rate` where rate is the global rate limit
we will use distributed rate limiter and redis otherwise.

By default `drl_threshold =5`


gateway/session_manager.go
```diff
@@ -129,54 +209,26 @@ func (l *SessionLimiter) ForwardMessage(r *http.Request, currentSession *user.Se
...
-			userBucket, err := l.bucketStore.Create(bucketKey, rate, time.Duration(per)*time.Second)
-			if err != nil {
-				log.Error("Failed to create bucket!")
-				return sessionFailRateLimit
-			}
-
-			if dryRun {
-				// if userBucket is empty and not expired.
-				if userBucket.Remaining() == 0 && time.Now().Before(userBucket.Reset()) {
+			if n <= 1 || n > apiLimit.Rate {
```
buger: It should be more complex than this. 
We need to introduce constant, which is configurable via the config. 
Additionally, it is not only about `Rate`, but it also depends on `Per`. We need to have Request per second in this formula. So formula will be `N*C > (apiLimit.Rate/apiLimit.Per)`. And C by default will be smth like 5.

gateway/session_manager.go
```diff
@@ -129,54 +209,28 @@ func (l *SessionLimiter) ForwardMessage(r *http.Request, currentSession *user.Se
...
+			var n float64
+			if DRLManager.Servers != nil {
+				n = float64(DRLManager.Servers.Count())
 			}
+			rate := apiLimit.Rate / apiLimit.Per
 
-			if dryRun {
-				// if userBucket is empty and not expired.
-				if userBucket.Remaining() == 0 && time.Now().Before(userBucket.Reset()) {
+			if n <= 1 || n > rate {
```
buger: Still, it require adding a configurable constant.

apidef/api_definitions.go
```diff
@@ -499,8 +499,9 @@ type SignatureConfig struct {
 }
 
 type GlobalRateLimit struct {
-	Rate float64 `bson:"rate" json:"rate"`
-	Per  float64 `bson:"per" json:"per"`
+	Rate         float64 `bson:"rate" json:"rate"`
+	Per          float64 `bson:"per" json:"per"`
+	DRLThreshold float64 `bson:"drl_threshold" json:"drl_threshold"`
```
buger: Pls move it to the config root, since it is not related to global rate limit.

Issue #2669, opened 2019-11-12
Enhance OAuth error logging


matiasinsaurralde:
**Is your feature request related to a problem? Please describe.**
When using Tyk OAuth handlers, the `osin` package describes several error scenarios, however these aren't integrated with our logging feature.

Example, a `token` request with an invalid `grant_type` value, *note:* the right value is `client_credentials`:
```
curl "http://localhost:8080/oauth-test/oauth/token?client_id=5c93a5ac503b4db8b0e7bdb35bbe623c&redirect_url=http://yourserver.com/redirect&grant_type=client_credential" -X POST -H 'Authorization: Basic NWM5M2E1YWM1MDNiNGRiOGIwZTdiZGIzNWJiZTYyM2M6TVRRM1l6ZG1NalV0WkRBMFpDMDBObUk1TFRrM09ETXRNamN5TW1FMU1XTTJPRGxq'
```
This is the HTTP error:
```js
{"error":"unsupported_grant_type","error_description":"The authorization grant type is not supported by the authorization server."}
```

It could be useful to log this in `stderr`/`stdout`.

christtyk:
Commentary from Client on #4328: Can we please have the Incoming_source_IP + client_id of the failed transaction + the URL accessed + the error code returned printed, so that we can identify the consumer ? Also it would be great if the logs gets to log browser.

We do not want the secret printed, as its against security standard to print secrets in INFO mode

christtyk:
@matiasinsaurralde was this working in previous versions?

christtyk:
Would like to keep open, sla request

PR #2677, opened 2019-11-20
[WIP] gateway: log OAuth error on HandleAccessRequest

matiasinsaurralde:
Simple patch for #2669, don't merge yet, we might need to add more details.


buger:
@matiasinsaurralde does it require more details?

matiasinsaurralde:
@buger We should decide what would make sense to add. From the customer:

> Commentary from Client on #4328: Can we please have the Incoming_source_IP + client_id of the failed transaction + the URL accessed + the error code returned printed, so that we can identify the consumer ? Also it would be great if the logs gets to log browser.

Issue #2687, opened 2019-11-22
Infinite loop in rate limit middleware


jlucktay:
**Branch/Environment/Version**
- Branch/Version: tag `v2.8.5`
- Environment: on-premise, though likely to affect others also

**Describe the bug**
In certain circumstances, requests passing through the rate limit/quota check middleware may fall into an infinite loop, due to a recursive function call that is not properly guarded and as such does not have an adequate exit condition.

**Reproduction steps**
TBD

**Actual behavior**
Some requests entering the gateway get stuck on an infinite/recursive loop, never get past the rate limit/quota check middleware, and cause a gradual increase in CPU usage and allocated memory.

**Expected behavior**
All requests entering the gateway are limit-/quota-checked and throttled/retried/completed appropriately.

**Screenshots/Video**
`pprof` output from the `inuse_space` sample of two `heap` profiles of an affected gateway are below.

**Profile 1** is still receiving new traffic and **Profile 2** is after the point where new traffic is no longer entering the gateway.

Corresponding increases in CPU time spent were also identified on the CPU profile in `pprof` (not shown).

![image](https://user-images.githubusercontent.com/460028/69441252-454e8d80-0d42-11ea-902b-78ff96ee38ac.png)
![image](https://user-images.githubusercontent.com/460028/69441260-4aabd800-0d42-11ea-9328-0d5c5ab24db0.png)

CPU usage of the instance running this affected gateway, in contrast to some other healthy instances/gateways:

![image](https://user-images.githubusercontent.com/460028/69441267-4f708c00-0d42-11ea-8a22-8e6bb078e9c7.png)

**Configuration (tyk config file):**
TBD

PR #2696, opened 2019-11-26
Fix for CPU slow burn issue

jlucktay:
Fix https://github.com/TykTechnologies/tyk/issues/2687

Move guard statement forward to prevent infinite recursive loop.
Test coverage is forthcoming.

### DIFF
--- a/gateway/mw_api_rate_limit_test.go
+++ b/gateway/mw_api_rate_limit_test.go
@@ -7,11 +7,12 @@ import (
 	"testing"
 	"time"
 
-	"github.com/justinas/alice"
-	uuid "github.com/satori/go.uuid"
-
 	"github.com/TykTechnologies/tyk/config"
 	"github.com/TykTechnologies/tyk/test"
+	uuid "github.com/satori/go.uuid"
+
+	"github.com/justinas/alice"
+
 	"github.com/TykTechnologies/tyk/user"
 )
 
@@ -205,6 +206,7 @@ func TestRequestThrottling(t *testing.T) {
 		t.Run("SentinelRateLimiter", requestThrottlingTest("SentinelRateLimiter", "APILevel"))
 		t.Run("RedisRollingRateLimiter", requestThrottlingTest("RedisRollingRateLimiter", "APILevel"))
 	})
+
 }
 
 func TestRLClosed(t *testing.T) {
--- a/gateway/mw_api_rate_limit_test.go
+++ b/gateway/mw_api_rate_limit_test.go
@@ -7,12 +7,11 @@ import (
 	"testing"
 	"time"
 
-	"github.com/TykTechnologies/tyk/config"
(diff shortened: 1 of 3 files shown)


buger:
While in general, it fixes the issue, what about the deeper issue happening inside ForwardMessage when you run it with `dryrun`?

jlucktay:
> While in general, it fixes the issue, what about the deeper issue happening inside ForwardMessage when you run it with `dryrun`?

I don't have a solid answer for that just yet, still getting my head around all of the moving parts.

jlucktay:
Talking to @gernest the other day, and from some previous conversations as well, sorting out the DRL seems like a much larger piece of work.

jlucktay:
Some existing test coverage already runs through this code, so I'm currently focusing around `TestRequestThrottling` like so:

``` shell
TYK_LOGLEVEL=debug go test \
    -count=1 \
    -run "^TestRequestThrottling$" \
    -timeout 30s \
    -v \
    github.com/TykTechnologies/tyk/gateway
```

buger:
@gernest does it looks good now?

gernest:
@buger I had forgotten about this, please wait I'm looking at it now.

gernest:
@jlucktay I have take a look at this, I apologise that my review comment was less informative I thought you could refer to our last chat regarding the tests.

So, I ran the same test with the old code and your change and they all passed. So This looks like you solved a problem that doesn't exist.

I believe we agreed you should write a test case to showcase the problem with the old code, then submit this patch. Meaning, when you run the test with old code the test should fail, highlighting the infinity loop nature of the code segment. And when you run the test with the patch it should pass highlighting the fix.

The patch itself is harmless(on `RateLimitAndQuotaCheck`), but I'm against incorporating the  tests, they don't serve any purpose with respect to this issue. And saying this patch fixes the issue without steps to reproduce it isn't a good idea.

jlucktay:
I am getting a `Server closed` warning after the test comes back with `PASS` but before the whole `go test` execution wraps up.

It doesn't do this when checked out on the `v2.8.5` tag.

```
--- PASS: TestRequestThrottling (2.74s)
    --- PASS: TestRequestThrottling/APILevel (2.74s)
        --- PASS: TestRequestThrottling/APILevel/InMemoryRateLimiter (2.74s)
PASS
time="Dec 03 13:55:42" level=warning msg="testServer.ListenAndServe() err: http: Server closed"
ok      github.com/TykTechnologies/tyk/gateway  2.771s
```

However, it does happen regardless of whether the `if ... break` is first or second out of the two `if`s inside `ProcessRequest` so I'm wondering if the `test` package is wrapping up the tests prematurely while throttled requests are still pending inside the test gateway.

`go test` still gives an exit code of 0 so everything from the outside would look fine to a CI process.

I'm not sure why I didn't spot this before.

Issue #2695, opened 2019-11-25
Multi auth (basic auth + auth token) not working


maciejwojciechowski:
**Branch/Environment/Version**
- Branch/Version: release-2.9
- Environment: on prem

**Describe the bug**
This is regression from 2.8
Multi auth mechanism does not work in 2.9. If I try to chain Basic Auth with Auth Token I get this error in response
`"error": "Attempted access with malformed header, header not in basic auth format"`

**Reproduction steps**

1. add api that uses 'Multiple Auth Mechanism' - Auth Token + Basic Auth
For Basic Auth modify 'Auth Key Header Name' to some custom field.
api def [my_api_def.txt](https://github.com/TykTechnologies/tyk/files/3889123/my_api_def.txt)
2. Add Basic Auth key
3. Add Auth token key
4. try sending some traffic via api 

**Actual behavior**
You get error message

**Expected behavior**
Traffic should go to upstream no problem.

**Screenshots/Video**
Compare 2.8
![Zrzut ekranu 2019-11-25 o 23 59 47](https://user-images.githubusercontent.com/39672152/69585428-f9207900-0fdf-11ea-855c-ba5dd7077d8c.png)

vs 2.9
![Zrzut ekranu 2019-11-26 o 00 00 28](https://user-images.githubusercontent.com/39672152/69585442-02114a80-0fe0-11ea-9382-60326ad1688d.png)

furkansenharputlu:
@maciejwojciechowski @ilijabojanovic `auth` field is deprecated, so please use `auth_configs` field. There is no UI for this. So, you need to use `raw api definition` to edit api definition. See the PR description: https://github.com/TykTechnologies/tyk/pull/2651

buger:
@furkansenharputlu so, while its deprecated, it should still work with old auth field too, or I'm missing smth?

maciejwojciechowski:
@furkansenharputlu sorry but this way we brake backwards compatibility, don't we?

furkansenharputlu:
@buger `Auth` header wasn't used only by `AuthKey`. It was used by `JWT` as well. However, the other ones weren't using `Auth` config at all. That was an inconsistency and bug. To fix it, I created a method `getAuthToken`. For all authentication types which needs auth header, to get a auth header name the following is how it works:

1. Look at `auth_configs`
2. Look at `auth`

Why does this work in 2.8?
`AuthKey` looks at `auth` config and gets `x-mw-auth`
[...]

PR #2697, opened 2019-11-26
Maintain backward compatibility during auth header name set

furkansenharputlu:
Fixes https://github.com/TykTechnologies/tyk/issues/2695


maciejwojciechowski:
verified https://github.com/TykTechnologies/tyk/issues/2695 on this PR

PR #2712, opened 2019-12-04
check for expiration time of certificate. If expired, do not accept it

adelowo:
Fixes https://github.com/TykTechnologies/tyk-analytics/issues/1638 and https://github.com/TykTechnologies/tyk-analytics/issues/1637

> The certificate used in the test was generated by @maciejwojciechowski in https://github.com/TykTechnologies/tyk-analytics/issues/1638


adelowo:
Should add a test with an expired certificate

certs/manager_test.go
```diff
@@ -121,17 +122,30 @@ func TestAddCertificate(t *testing.T) {
 	cert2Raw, _ := pem.Decode(cert2Pem)
 	cert2ID := HexSHA256(cert2Raw.Bytes)
 
+	buf := new(bytes.Buffer)
+
+	f, err := os.Open("testdata/expired-cert.pem")
```
buger: Needs be replaced by dynamically generated certificate. See `genCertificate` usage in cert package.

adelowo: Done in https://github.com/TykTechnologies/tyk/pull/2712/commits/d84230485dae20592cd11acf67473d7bbe89cbee

PR #2725, opened 2019-12-10
Inherit metadata from policy to key

### LINKED ISSUE #2717: Please include a nested JSON object on the policy
Regarding https://support.tyk.io/a/tickets/3793,

Could we please request an enhancement to include a nested JSON object, similar to this one, on the policy(https://tyk.io/docs/security/security-policies/policies-guide/) ? Can this also be prioritised to be bundled with the 2.8.3 please ? 

We just want a nested JSON object, similar to this one, included on the Oauth client & on the policy definition.

  "meta_data" :{      "key1": "value1",      "key2": "value2",      "key3": "value3"}

furkansenharputlu:
Fixes #2717

### DIFF
--- a/gateway/middleware.go
+++ b/gateway/middleware.go
@@ -474,6 +474,10 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 			tags[tag] = true
 		}
 
+		for k, v := range policy.MetaData {
+			session.MetaData[k] = v
+		}
+
 		if policy.LastUpdated > session.LastUpdated {
 			session.LastUpdated = policy.LastUpdated
 		}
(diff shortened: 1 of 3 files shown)


gateway/middleware.go
```diff
@@ -474,6 +474,10 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 			tags[tag] = true
 		}
 
+		for k, v := range policy.MetaData {
+			session.MetaData[k] = v
```
buger: Let's ensure that key can override metadata if needed. So first check if it has non-empty value.

furkansenharputlu: On each update, `ApplyPolicies` is used. So, we need to keep a separate field to hold overrides. Another thing, `tags` also works like this. If user wants to change something he should go to policy to make change. Therefore, it is better for us to keep it consistent.

buger: I see, makes sense!

Issue #2679, opened 2019-11-20
Oauth /oauth/authorize-client returns invalid redirect_uri


komalsukhani:
**Branch/Environment/Version**
master

**Describe the bug**
`/oauth/authorize-client` endpoint replaces `&` character with it's unicode character while returning redirect_uri

**Reproduction steps**

1. Import [APIDef](https://github.com/TykTechnologies/tyk/files/3867978/APIDef.txt)
2. Create oauth client
3. Send below curl command 
``` curl -X POST 'http://localhost:8181/test/tyk/oauth/authorize-client/?response_type=token&client_id={CLIENT_ID}&redirect_uri=http://httpbin.org' -H 'x-tyk-authorization: {TYK_SECRET}' ```

**Note**: Replace `CLIENT_ID` with clientID generated in step 2, and `TYK_SECRET` with your secret.

**Actual behavior**
redirect_uri will contain unicode of & instead of actual character.

**Expected behavior**
redirect_uri should not contain unicode

maciejwojciechowski:
verified, current response looks like this

`{"access_token":"5dfa54d8fe238f624dc5c35590ba38fd9a764f90a40d8a2ef1c71af8","expires_in":3600,"redirect_to":"https://url.org#access_token=5dfa54d8fe238f624dc5c35590ba38fd9a764f90a40d8a2ef1c71af8&expires_in=3600&token_type=bearer","token_type":"bearer"}`

PR #2726, opened 2019-12-10
Escape unicode while generating redirect_to url for oauth

furkansenharputlu:
- Fixes https://github.com/TykTechnologies/tyk/issues/2679


buger:
Looks good! The only suggestion that probably instead of adding new `BodyMatchRegex` I would make `BodyMatch` to support regexp by default.

furkansenharputlu:
@buger At the beginning, I implemented as you said. However, some special chars should have `\` in regex match like `?` so I changed it to this way. However, if you say, I can return back.

buger:
I see, In general, if we do not use these characters in BodyMatch right now, or it is possible to fix such cases, better keep one helper, and just update test docs to mention such matcher behaviour.

PR #2743, opened 2019-12-17
Targeted detailed recording

adelowo:
Fixes https://github.com/TykTechnologies/product/issues/168

- [x] Add flag on API Definition
- [x] Flag on key level

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -469,15 +469,16 @@ type APIDefinition struct {
 		OptionsPassthrough bool     `bson:"options_passthrough" json:"options_passthrough"`
 		Debug              bool     `bson:"debug" json:"debug"`
 	} `bson:"CORS" json:"CORS"`
-	Domain            string                 `bson:"domain" json:"domain"`
-	Certificates      []string               `bson:"certificates" json:"certificates"`
-	DoNotTrack        bool                   `bson:"do_not_track" json:"do_not_track"`
-	Tags              []string               `bson:"tags" json:"tags"`
-	EnableContextVars bool                   `bson:"enable_context_vars" json:"enable_context_vars"`
-	ConfigData        map[string]interface{} `bson:"config_data" json:"config_data"`
-	TagHeaders        []string               `bson:"tag_headers" json:"tag_headers"`
-	GlobalRateLimit   GlobalRateLimit        `bson:"global_rate_limit" json:"global_rate_limit"`
-	StripAuthData     bool                   `bson:"strip_auth_data" json:"strip_auth_data"`
+	Domain                  string                 `bson:"domain" json:"domain"`
+	Certificates            []string               `bson:"certificates" json:"certificates"`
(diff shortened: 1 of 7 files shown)


apidef/api_definitions.go
```diff
@@ -469,15 +469,16 @@ type APIDefinition struct {
...
+	Domain                  string                 `bson:"domain" json:"domain"`
+	Certificates            []string               `bson:"certificates" json:"certificates"`
+	DoNotTrack              bool                   `bson:"do_not_track" json:"do_not_track"`
+	Tags                    []string               `bson:"tags" json:"tags"`
+	EnableContextVars       bool                   `bson:"enable_context_vars" json:"enable_context_vars"`
+	ConfigData              map[string]interface{} `bson:"config_data" json:"config_data"`
+	TagHeaders              []string               `bson:"tag_headers" json:"tag_headers"`
+	GlobalRateLimit         GlobalRateLimit        `bson:"global_rate_limit" json:"global_rate_limit"`
+	StripAuthData           bool                   `bson:"strip_auth_data" json:"strip_auth_data"`
+	EnableDetailedAnalytics bool                   `bson:"enable_detailed_recording" json:"enable_detailed_recording"`
```
gernest: `EnableDetailedAnalytics` doesn't reflect `enable_detailed_recording` I think we need to be consistent here and name it `EnableDetailedRecording`

adelowo: Fixed.

buger:
So, we already had `EnableDetailedRecording` on session level (just not used), correct?

adelowo:
Yes, we only use it when it is set and organisation tracking/logging is enabled

buger:
So to clarify one more time, we have 3 ways to enable detailed logging now:
1) API level
2) Session level
3) Org level (org keys)

Correct?

adelowo:
4. There's also the global config option

apidef/api_definitions.go
letzya: so the field is `enable_detailed_recording` and the variable is EnableDetailed`Analytics`?

adelowo: You seem to have reviewed an old commit or something. It was updated before this PR was merged. https://github.com/TykTechnologies/tyk/commit/5e74e19c6759f25d6edc2544b1f88f18634d9a7a

letzya:
@adelowo would be good to add some info in the description of the pr explaining a bit the logic and how these config being applied.
This later can be use in the docs or for early adopters of users of the gateway

apidef/api_definitions.go
letzya: Thanks! Got this link from the release channel

letzya:
@adelowo This is the logic we need to follow: 
image

lghiur:
@adelowo please provide more info about this in the docs, as per @adelowo comment above

adelowo:
Also, docs are at https://github.com/TykTechnologies/tyk-docs/pull/1235

adelowo:
targeted_detailed_recording-2

PR #2746, opened 2019-12-18
Fix and enhance Python call protection

matiasinsaurralde:
I've found several issues when using `dlpython` GIL helper functions, mostly because there might be internal calls that weren't correctly mapped when preparing the bindings. I did many experiments with it and the simplest approach is to control the lock in the Go side, with this we can avoid additional `cgo` overhead (in 2.8 we were controlling the GIL using the CPython calls).

Performance gets slightly better, less context switches and CPU usage as well, I'm comparing 2.8.6 (with previous approach) with this 2.9.x patch. It should protect all calls and ensure no crashes related to `PyCallObject`:

test1
test2

cpuinterrupts
cpuusage

Fix for: https://github.com/TykTechnologies/tyk/issues/2760

### DIFF
--- a/gateway/coprocess_python.go
+++ b/gateway/coprocess_python.go
@@ -28,7 +28,7 @@ import (
 var (
 	dispatcherClass    unsafe.Pointer
 	dispatcherInstance unsafe.Pointer
-	mwCacheLock        = sync.Mutex{}
+	pythonLock         = sync.Mutex{}
 )
 
 // PythonDispatcher implements a coprocess.Dispatcher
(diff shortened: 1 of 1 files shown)


christtyk:
@furkansenharputlu will this be in 2.9.3?

furkansenharputlu:
@christtyk Yes, already merged to release branch.

christtyk:
@furkansenharputlu @matiasinsaurralde will this fix issue for ticket 4492? Client is asking for more context

christtyk:
4631 is the related ticket

Issue #2708, opened 2019-11-29
Websocket connection requests fail when highly concurrent


bpiper:
**Branch/Environment/Version**
- Branch/Version: 2.9.1
- Environment: on-premise gateway in Docker

**Describe the bug**

When Tyk proxies multiple websocket connection requests simultaneously, occasionally one of them will fail, either with a 500 response, or a response containing an invalid Sec-WebSocket-Accept header.  The following is logged by Tyk:

```
time="Nov 29 00:52:52" level=error msg="Hijack error: http: connection has been hijacked" origin=******** path="/api/some/endpoint"
time="Nov 29 00:52:52" level=error msg="http: proxy error: http: connection has been hijacked" api_id=my-api org_id=default prefix=proxy server_name=api.server.com user_id=-- user_ip="********" user_name=
2019/11/29 00:52:52 http: response.WriteHeader on hijacked connection from github.com/TykTechnologies/tyk/gateway.(*ErrorHandler).HandleError (handler_error.go:82)
2019/11/29 00:52:52 http: response.Write on hijacked connection from text/template.(*state).walk (exec.go:269)
```

I originally encountered this when running integration tests that (due to parallel execution) can open up to around 10 websocket connections simultaneously.  We run two gateway servers behind an AWS load balancer with clustering.

**Reproduction steps**

1. Run Tyk gateway in front of a server with a websocket endpoint.
2. Open a whole bunch of websocket connections at once.  If you have `wscat` installed, something like this will encounter a single connection failure at least 50% of the time:
```
for i in {1..40}; do wscat -c wss://api.server.com/api/some/endpoint < /dev/zero & done
```

**Configuration (tyk config file):**
See attached API configuration.  Tyk config is just the default from tykio/tyk-gateway:v2.9.1 with the following env variables set:
```
TYK_GW_STORAGE_HOST = some.redis.endpoint
TYK_GW_STORAGE_PORT = <redis-port>
TYK_GW_COPROCESSOPTIONS_ENABLECOPROCESS = false
TYK_GW_HEALTHCHECK_ENABLEHEALTHCHECKS = true
TYK_GW_SECRET = ********
TYK_GW_STORAGE_ENABLECLUSTER = true
```

bpiper:
[tyk_api_config.json.txt](https://github.com/TykTechnologies/tyk/files/3903321/tyk_api_config.json.txt)

candux:
[...]

PR #2747, opened 2019-12-18
Fix websockets

buger:
Removed all custom websocket code, and instead backported default Go ReverseProxy code, which now supports websockets by default.
Websockets now use the same dialer, and transport, which makes it scale and configure way easier.

Fix https://github.com/TykTechnologies/tyk/issues/2708


gateway/reverse_proxy.go
```diff
@@ -1025,6 +1008,56 @@ func (p *ReverseProxy) copyBuffer(dst io.Writer, src io.Reader) (int64, error) {
...
+	return nil
+}
+
+// switchProtocolCopier exists so goroutines proxying data back and
+// forth have nice names in stacks.
+type switchProtocolCopier struct {
+	user, backend io.ReadWriter
+}
+
+func (c switchProtocolCopier) copyFromBackend(errc chan<- error) {
```
jensneuse: Why copy by value and not reference?

gateway/reverse_proxy.go
```diff
@@ -1146,3 +1179,26 @@ func nopCloseResponseBody(r *http.Response) {
...
+		return true, ""
+	}
+
+	connection := strings.ToLower(strings.TrimSpace(req.Header.Get(headers.Connection)))
+	if connection != "upgrade" {
+		return false, ""
+	}
+
+	upgrade := strings.ToLower(strings.TrimSpace(req.Header.Get("Upgrade")))
+	if upgrade != "" {
```
jensneuse: Upgrade should be checked for correct upgrade I think:
https://en.wikipedia.org/wiki/HTTP/1.1_Upgrade_header
The original function did this correctly.

gateway/reverse_proxy.go
```diff
@@ -1025,6 +1008,56 @@ func (p *ReverseProxy) copyBuffer(dst io.Writer, src io.Reader) (int64, error) {
...
+	errc := make(chan error, 1)
+	spc := switchProtocolCopier{user: conn, backend: backConn}
+	go spc.copyToBackend(errc)
+	go spc.copyFromBackend(errc)
+	<-errc
+
+	return nil
+}
+
+// switchProtocolCopier exists so goroutines proxying data back and
```
jensneuse: I understand the reason. Maybe name it different thought? E.g. websocketConnectionCopier? Makes it more clear what it's related to.

gateway/reverse_proxy.go
buger: This is purely copy-pate from default Go reverse proxy package

gateway/reverse_proxy.go
buger: It is not only for websockets, but all connections which could be either upgraded or steamed. For example Server-side events, or gRPC streaming.

gateway/reverse_proxy.go
buger: On of the reasons I do not check it now, because it is more general-purpose, and not limited to websockets. Original reverse proxy Go implementation does the same logic, and just checks if field non empty.

gateway/reverse_proxy.go
jensneuse: I see, makes sense.

PR #2761, opened 2019-12-23
make sure payload is released and log message without locking

letzya:


gateway/dashboard_register.go
```diff
@@ -149,6 +149,7 @@ func (h *HTTPDashboardHandler) Register() error {
 	req := h.newRequest(h.RegistrationEndpoint)
 	c := initialiseClient(5 * time.Second)
 	resp, err := c.Do(req)
+	defer resp.Body.Close()
```
gernest: This will panic if there was an error doing the client call because `resp` will be nil.

PR #2781, opened 2020-01-06
Copy arguments to RedisCluster.GetMultiKey

gernest:
The method was modifying original slice, this results into bugs if caller
fails to pay attention.

The test was Skipped, after enabling it it was failing because of the
above assumptions.


buger:
Where exactly you found this bug?

gernest:
@buger I was writing tests , then noticed `TestRedisClusterGetMultiKey` had `t.Skip()` there was no need really to skip the test (was probably there by mistake), I was suprised to see the tests were failing. I investigated and noticed this bug.

Issue #2780, opened 2020-01-05
Tyk spams Quota Checks despite having this behaviour turned off


sedkis:
**Branch/Environment/Version**
- Branch/Version: Master
- Environment: On-Prem

**Describe the bug**
When I create a key that has a policy with many APIs, there is (atleast) one quota check for each API in my policy, even though I've disabled Quotas in my Policy by giving a "-1" value, and turned off the Quota Per API feature.  This is resulting in many TPS which cause my GW to become unstable and crash.

**Reproduction steps**
1.  Create 10 APIs
2. create a Policy with access rights to all 10 APIs
3. Create a Key with policy in step 2 

Here are the logs for step 3.
see logs:
```
[Jan 05 12:58:04]  INFO auth-mgr: Reset quota for key. inbound-key=****0859 key=quota-cdd92825
[Jan 05 12:58:04]  INFO auth-mgr: Reset quota for key. inbound-key=****0859 key=quota-cdd92825
[Jan 05 12:58:04]  INFO auth-mgr: Reset quota for key. inbound-key=****0859 key=quota-cdd92825
[Jan 05 12:58:04]  INFO auth-mgr: Reset quota for key. inbound-key=****0859 key=quota-cdd92825
[Jan 05 12:58:04]  INFO auth-mgr: Reset quota for key. inbound-key=****0859 key=quota-cdd92825
[Jan 05 12:58:04]  INFO auth-mgr: Reset quota for key. inbound-key=****0859 key=quota-cdd92825
[Jan 05 12:58:04]  INFO auth-mgr: Reset quota for key. inbound-key=****0859 key=quota-cdd92825
[Jan 05 12:58:04]  INFO Both event_trigers_defunct and event_triggers_defunct are configured in the config, event_triggers_defunct will be used.
[Jan 05 12:58:04]  INFO api: Generated new key: (****0859) api_id=-- key=****0859 org_id=5d431f8d06ac9e183bf49f9c path=-- server_name=system status=ok user_id=system user_ip=127.0.0.1
[Jan 05 12:58:04]  INFO api: Can't retrieve key quota key=****2825 message=key not found status=ok
[Jan 05 12:58:04]  INFO api: Can't retrieve api limit quota apiID=1bf19808226348be7283c1b6cd241947 error=key not found key=****0859
[Jan 05 12:58:04]  INFO api: Can't retrieve api limit quota apiID=545e7ebdbac84e295ea574648304287f error=key not found key=****0859
[...]

PR #2789, opened 2020-01-08
don't do quota checks on API level or key level if they have -1 value

### LINKED ISSUE #2777: Remove redundant quota check
related to #2780 
- Branch/Version: All
- Environment: All

**Describe the bug**
When retrieving a Key, Tyk will attempt to check for a Quota pool, despite it being disabled.  I've disabled the Quota by giving it a value of "-1"

**Reproduction steps**
1.  Create a Key, with Quota Max value "-1" (disabled)
2.  Get key details using the API 
`/api/apis/927605639bc54ee346956bba1742a1ce/keys/5d431f8d06ac9e183bf49f9c1d5a1e3e458d4349bf048b3d603cc240`

**Actual behavior**
Check logs, you will see this:
```
[Jan  3 16:23:49]  INFO api: Can't retrieve key quota error=key not found key=****c240 status=ok
[Jan  3 16:23:49]  INFO api: Can't retrieve api limit quota apiID=927605639bc54ee346956bba1742a1ce error=key not found key=****c240
[Jan  3 16:23:49]  INFO api: Retrieved key detail. key=****c240 status=ok
[Jan  3 16:23:55]  INFO api: Can't retrieve key quota error=key not found key=****c240 status=ok
[Jan  3 16:23:55]  INFO api: Can't retrieve api limit quota apiID=927605639bc54ee346956bba1742a1ce error=key not found key=****c240
[Jan  3 16:23:55]  INFO api: Retrieved key detail. key=****c240 status=ok
```

**Expected behavior**
I don't want Tyk to check for Quota if I have it disabled

sedkis:
fixes #2777 , #2780  

I think this line here is the culprit.

We create a new instance of access rights for each API in the policy and thus the GO runtime sets the default values to 0.  This essentially "turns on" limits for each API.
https://github.com/TykTechnologies/tyk/blob/master/gateway/middleware.go#L371

I believe this function (which runs after key creation), is supposed to cancel out the API level Limits, however it doesn't work correctly and never gets hit because the line above in `middleware.go` instantiates the Limit object.
https://github.com/TykTechnologies/tyk/blob/master/gateway/api.go#L198

This fix leaves the current logic as is, and adds an IF checks to see if Quotas are enabled before performing a quota check.

### DIFF
--- a/gateway/api.go
+++ b/gateway/api.go
@@ -443,32 +443,34 @@ func handleGetDetail(sessionKey, apiID string, byHash bool) (interface{}, int) {
 	mw := BaseMiddleware{Spec: spec}
 	mw.ApplyPolicies(&session)
 
-	quotaKey := QuotaKeyPrefix + storage.HashKey(sessionKey)
-	if byHash {
-		quotaKey = QuotaKeyPrefix + sessionKey
-	}
-
-	if usedQuota, err := GlobalSessionManager.Store().GetRawKey(quotaKey); err == nil {
-		qInt, _ := strconv.Atoi(usedQuota)
-		remaining := session.QuotaMax - int64(qInt)
-
-		if remaining < 0 {
-			session.QuotaRemaining = 0
-		} else {
-			session.QuotaRemaining = remaining
+	if session.QuotaMax != -1 {
+		quotaKey := QuotaKeyPrefix + storage.HashKey(sessionKey)
+		if byHash {
+			quotaKey = QuotaKeyPrefix + sessionKey
 		}
-	} else {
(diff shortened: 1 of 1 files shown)


christtyk:
@buger if we can't get this into 2.9.3 the client will need unstable packages as their go live is end of January

sedkis:
bump

buger:
Incremental API updates will not be available in 2.9.3, so I would rather move this to 2.9.4 and generate unstable packages.

sedkis:
I've marked it as 2.9.4 👍

Issue #2713, opened 2019-12-04
Can't disable a token generated through JWT


sedkis:
**Branch/Environment/Version**
2.9.1 / 1.9.1

**Describe the bug**
I can't update the 'is_inactive' status of a Session Token that is created via the use of JWT.
I need to do this so that I can de-activate through Tyk a JWT.

**Reproduction steps**
1. Add JWT API
2. Create JWT Policy
3. Create JWT with access to API and POL defined from step 2
4. Add global header with `$tyk_meta.TykJWTSessionID`
5. Query API through Tyk to `http://httpbin.org/headers` and retrieve the session ID from step 4
6. Try to deactivate the JWT by updating the Session ID, by setting "is_inactive" from `false` to `true` using the:
```
PUT http://<DASHBOARD_HOST>/api/apis/<JWT_API_ID>/keys/<SESSION_ID>
```

The value "true" doesn't persist.

**Expected behavior**
I would expect the value to become "true" so that I may deactivate a Key!

**Additional context**

Here is what I'm trying to do:
https://community.tyk.io/t/blacklisting-jwt-token/986

**JWT**
```
eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJsb2NhbGhvc3QiLCJzdWIiOiJqYXNkODEyajRrYWQ4MWo5ZGE4MSIsImF1ZCI6InJlZnJlc2giLCJleHAiOjE2MDcwNzYxMTQsImlhdCI6MTU3NTQ1MzcxNCwianRpIjoiZWU4MjBjMmM0ZTA3NTk3NDU0MDZjZjA4ZDYxODBkYWJkYjhhMGI0ZTNhMjE2NjdhOGE3NDliYzE1YmFmODJiOTNmYzhiNmY1MTc0NTMzMGVlMzczODczNTJlNDM3NjU4NmQ1MmM2NmJiZGRmOTI0NWJiYzdkZWFhNWUzYWZjYTEiLCJwb2wiOiI1ZGU3Yzg3N2Y1NmUxYTY1NGFhZDFhYWUifQ.fMyP3dJeTJq9BgwMFrrMhI45O-DDlf16jFmRZB0RjNU
```
**API Definition**
```javascript
{
    "id": "5de7c281f56e1a647de671d9",
    "name": "jwt-test",
    "slug": "jwt-test",
    "listen_port": 0,
    "protocol": "",
    "enable_proxy_protocol": false,
[...]

Issue #2784, opened 2020-01-06
[TT-1963] - Disabling a policy disables the key - issue with multiple policies for a key


asoorm:
**Branch/Environment/Version**

all

**Describe the bug**

A key has access to 2 policies.
Deactivate one policy, and access to the other policy is also disabled.

https://github.com/TykTechnologies/tyk/blob/master/gateway/middleware.go#L468-L472

**Expected behavior**

Disabling one policy, should not completely disable the session token, only access to the services protected by the policy. especially if more than one policy is applied to the key.

asoorm:
bump

christtyk:
do you know if it's still valid @asoorm

andyo-tyk:
This will be fixed in Tyk 5.0.4 and 5.1.1.

PR #2791, opened 2020-01-08
Fix session's inactivate value setting during policy apply

furkansenharputlu:
Fixes https://github.com/TykTechnologies/tyk/issues/2713


asoorm:
Problem is that in investigating this issue, we uncovered another related bug: https://github.com/TykTechnologies/tyk/issues/2784

PR #2798, opened 2020-01-11
CoProcess Auth should check the policy

asoorm:
Replaces: https://github.com/TykTechnologies/tyk/pull/2650
Fixes: https://github.com/TykTechnologies/tyk/issues/2884


letzya:
@asoorm wasn't it fixed before?

buger:
Do you have a ticket for this one?

buger:
@asoorm we need a task with a full context on this, if we want it get intcluded to the patch.

Issue #2468, opened 2019-08-25
404 error logs when no matched API


asoorm:
**Branch/Envrionment/Version**

ALL

**Describe the bug**

When sending a request to the gateway, where the gateway is unable to match against an api, the response is a 404 not found.

This is fine, but there is currently no way of identifying the offending client. 
This potentially could mean increased load on the gateway server, but with no ability to detect these erroneous calls, the gateway relies on the clients resolving the issue themselves.

**Reproduction steps**

```
curl http://gateway.ahmet:8080/non-existent-api/get -v
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to gateway.ahmet (127.0.0.1) port 8080 (#0)
> GET /non-existent-api/get HTTP/1.1
> Host: gateway.ahmet:8080
> User-Agent: curl/7.54.0
> Accept: */*
>
< HTTP/1.1 404 Not Found
< Content-Type: text/plain; charset=utf-8
< X-Content-Type-Options: nosniff
< Date: Sun, 25 Aug 2019 12:20:30 GMT
< Content-Length: 19
<
404 page not found
```

**Expected behavior**

Expose a config option to enable raw error request logs.
This log entry should contain a client IP address, along with method, route & possibly optional request headers.

An example from NGINX request logs

```
185.134.69.59 - - [25/Aug/2019:11:30:04 +0000] "POST /cgi-bin/ViewLog.asp HTTP/1.1" 404 9 "-" "Ankit" "-"
```

ilijabojanovic:
Verified.

We have new config option
`"track_404_logs": true`
In gateway system log we see this:

```
[Apr 13 18:07:40] ERROR Not Found origin=127.0.0.1:59330 request=GET /zikica/ip HTTP/1.1
[Apr 13 18:07:41] ERROR Not Found origin=127.0.0.1:59338 request=GET /zikica/ip HTTP/1.1
[Apr 13 18:09:17] ERROR Not Found origin=127.0.0.1:59868 request=PUT /zikica/ip HTTP/1.1
[Apr 13 18:09:20] ERROR Not Found origin=127.0.0.1:59881 request=POST /zikica/ip HTTP/1.1
[Apr 13 18:09:24] ERROR Not Found origin=127.0.0.1:59902 request=PATCH /zikica/ip HTTP/1.1
[Apr 13 18:09:27] ERROR Not Found origin=127.0.0.1:59919 request=OPTIONS /zikica/ip HTTP/1.1
```

PR #2800, opened 2020-01-13
Log when page not found

furkansenharputlu:
Example log:

``` 
time="Jan 13 16:07:06" level=error msg="404 page not found" request="GET /non-existing-api/get HTTP/1.1" 
```

Fixes https://github.com/TykTechnologies/tyk/issues/2468

### DIFF
--- a/gateway/api_loader.go
+++ b/gateway/api_loader.go
@@ -683,6 +683,7 @@ func loadApps(specs []*APISpec) {
 
 	muxer := &proxyMux{}
 	router := mux.NewRouter()
+	router.NotFoundHandler = http.HandlerFunc(muxer.handle404)
 	loadAPIEndpoints(router)
 	muxer.setRouter(port, "", router)
 
--- a/gateway/proxy_muxer.go
+++ b/gateway/proxy_muxer.go
@@ -148,6 +148,14 @@ func (m *proxyMux) setRouter(port int, protocol string, router *mux.Router) {
 	}
 }
 
+func (m *proxyMux) handle404(w http.ResponseWriter, r *http.Request) {
+	requestMeta := fmt.Sprintf("%s %s %s", r.Method, r.URL.Path, r.Proto)
+	log.WithField("request", requestMeta).Error("404 page not found")
+
+	w.WriteHeader(http.StatusNotFound)
+	fmt.Fprintf(w, "404 page not found")
+}
+
 func (m *proxyMux) addTCPService(spec *APISpec, modifier *tcp.Modifier) {
 	hostname := spec.GlobalConfig.HostName
 	if spec.GlobalConfig.EnableCustomDomains {
--- a/gateway/proxy_muxer.go
+++ b/gateway/proxy_muxer.go
@@ -15,8 +15,8 @@ import (
 	"github.com/TykTechnologies/again"
 	"github.com/TykTechnologies/tyk/config"
 	"github.com/TykTechnologies/tyk/tcp"
-	proxyproto "github.com/pires/go-proxyproto"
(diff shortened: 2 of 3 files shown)


gateway/api_loader.go
```diff
@@ -676,10 +676,12 @@ func loadApps(specs []*APISpec) {
...
 	muxer := &proxyMux{}
 
-	globalConf := config.Global()
 	r := mux.NewRouter()
-	muxer.setRouter(globalConf.ListenPort, "", r)
+	r.NotFoundHandler = http.HandlerFunc(muxer.handle404)
+
+	globalConf := config.Global()
 	if globalConf.ControlAPIPort == 0 {
+		muxer.setRouter(globalConf.ListenPort, "", r)
```
gernest: Why did you move this?  I think you are breaking the logic here.

So, `r` is bound on `globalConf.ListenPort` by default, if we don't have a separate `globalConf.ControlAPIPort` we will load control endpoints on `r` . When that control port is provided then we bind control endpoints there.

Please leave `muxer.setRouter(globalConf.ListenPort, "", r)` above the `if` condition as it was before.

furkansenharputlu: If it enter if, there is nothing change.
if it enters else statement, It is overriden in else statement. 

So I think it doesn't break.

furkansenharputlu: Let me make it more simple.

gernest:
@furkansenharputlu please only leave the changes related to `404` page. Leave everything else as it was. The `404` changes already look good. You can make any unrelated changes in a separate PR( With more details on why you made them) .

furkansenharputlu:
Yes, you're actually right.  However, the current code forces me to make redundant changes while implementing 404. Therefore, I will open a separate PR and after merging it. I will make the 404 changes on top of it.

furkansenharputlu:
@gernest The other PRs are merged. I think this is ready to go. Can you please review?

gateway/proxy_muxer.go
```diff
@@ -148,6 +148,14 @@ func (m *proxyMux) setRouter(port int, protocol string, router *mux.Router) {
 	}
 }
 
+func (m *proxyMux) handle404(w http.ResponseWriter, r *http.Request) {
+	requestMeta := fmt.Sprintf("%s %s %s", r.Method, r.URL.Path, r.Proto)
+	log.WithField("request", requestMeta).Error("404 page not found")
+
+	w.WriteHeader(http.StatusNotFound)
+	fmt.Fprintf(w, "404 page not found")
```
gernest: lets use standard library here `fmt.Fprint(w, http.StatusText(http.StatusNotFound))` . Use `fmt.Fprint` because we are not formatting the string.

gateway/proxy_muxer_test.go
```diff
@@ -289,3 +291,18 @@ func TestHTTP_custom_ports(t *testing.T) {
...
+	defer g.Close()
+
+	BuildAndLoadAPI(func(spec *APISpec) {
+		spec.Proxy.ListenPath = "/existing"
+		spec.UseKeylessAccess = true
+	})
+
+	_, _ = g.Run(t, []test.TestCase{
+		{Path: "/existing", Code: http.StatusOK},
+		{Path: "/nonexisting", Code: http.StatusNotFound, BodyMatch: "404 page not found"},
```
gernest: use `http.StatusText(http.StatusNotFound)` on `BodyMatch`

asoorm:
@furkansenharputlu This should be configurable - we should be able to turn it on and off. Can get extremely noisy if we just blindly log everything...

PR #2802, opened 2020-01-14
Expand TestContolListener to cover user api cases

furkansenharputlu:


buger:
How it differ from `TestControlListener` test?

furkansenharputlu:
It doesn't load a new API.

buger:
Can we extend it then, instead of adding new test?
Note that it does very important thing, which your test currently missing: `doReload` - which do lot of things with a rounder.

gateway/gateway_test.go
```diff
@@ -832,6 +832,17 @@ func TestControlListener(t *testing.T) {
 	ts.RunExt(t, tests...)
 	DoReload()
 	ts.RunExt(t, tests...)
+
+	// Moved here because DoReload overrides it
+	BuildAndLoadAPI(func(spec *APISpec) {
+		spec.Proxy.ListenPath = "/user-api"
+		spec.UseKeylessAccess = true
+	})
```
gernest: Please move this block up below 
```
	ts := StartTest(TestConfig{
		sepatateControlAPI: true,
	})
	defer ts.Close()
```

gateway/gateway_test.go
```diff
@@ -832,6 +832,17 @@ func TestControlListener(t *testing.T) {
...
+
+	_, _ = ts.Run(t, []test.TestCase{
+		{Path: "/user-api", Code: http.StatusOK},
+		{Path: "/user-api", ControlRequest: true, Code: http.StatusNotFound},
```
gernest: please add this inside the `tests` variable and remove this block entirely. We are only extending the tests so this block isn't needed.

furkansenharputlu: Don't you see my comment there?

furkansenharputlu: `// Moved here because DoReload overrides it`

buger: I think @gernest meant smth like this:
You more tests cases inside tests object and your test will look as simple as :

``` 
	BuildAndLoadAPI(func(spec *APISpec) {
		spec.Proxy.ListenPath = "/user-api"
		spec.UseKeylessAccess = true
	})

        ts.RunExt(t, tests...)
        doReload()

	// DoReload overrides APIs, needs loading again
	BuildAndLoadAPI(func(spec *APISpec) {
		spec.Proxy.ListenPath = "/user-api"
		spec.UseKeylessAccess = true
	})
        ts.RunExt(t, tests...)
```

buger: So basically we verify that API actually loaded before the reload, and still working after reload.

gernest: @furkansenharputlu ☝️ see @buger  comment  above that is what I meant.

buger: Also, a better way will be extending `LoadAPI` to actually allow keep API between reloads.

furkansenharputlu: For sure I firstly added inside the tests object. However, after seeing that `DoReload` overrides it, I moved it to there. I am not sure this way is simpler and covers any new thing but I will update as you said.

furkansenharputlu: Nope, your way will also not work because `RunExt` fails with user api.

```
func TestDummy(t *testing.T) {
	ts := StartTest(TestConfig{
		sepatateControlAPI: true,
	})
	defer ts.Close()

	BuildAndLoadAPI(func(spec *APISpec) {
		spec.Proxy.ListenPath = "/user-api"
		spec.UseKeylessAccess = true
	})

	tests := []test.TestCase{
		{Path: "/user-api", Code: http.StatusOK},
	}

	ts.RunExt(t, tests...)
}```

This doesn't work.

buger: Hmm, it should. Needs investigation.

furkansenharputlu: So, I am in favor of merging the current version.

PR #2808, opened 2020-01-17
Fix up broken docs link in checkup package

jlucktay:


buger:
@jlucktay pls make a task for it, and attach to 2.9.4 project board

jlucktay:
Fixes #2875.

PR #2811, opened 2020-01-17
Replace client in set instead of adding new one if redis cluster enabled

### LINKED ISSUE #2804: Redis-cluster: Edit/Update oauth client duplicates oauth client in redis
- Branch/Version: Master/Release-2.9
- Environment: All

**Describe the bug**
When gateway is connected to redis cluster, edit or update of oauth client duplicates entry redis database

**Reproduction steps**

1. Start gateway in redis cluster mode
2. Over dashboard create api with oauth auth
3. Create oauth client
4. Open auth client page and add new metadata filed or update policy
5. Click update

**Actual behavior**
We have two oauth clients on page and also in database

**Expected behavior**
Edit or update existing oauth client should not create new entry in database

**Screenshots/Video**
Screenshot 2020-01-15 at 14 07 09

Log from console or from log file.

Attach tyk configuration file

**Additional context**

furkansenharputlu:
Fixes #2804

### DIFF
--- a/gateway/oauth_manager.go
+++ b/gateway/oauth_manager.go
@@ -9,10 +9,11 @@ import (
 	"math"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/lonelycode/osin"
-	uuid "github.com/satori/go.uuid"
+	"github.com/satori/go.uuid"
 	"golang.org/x/crypto/bcrypt"
 
 	"strconv"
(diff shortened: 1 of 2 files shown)


gateway/oauth_manager.go
```diff
@@ -690,6 +691,15 @@ func (r *RedisOsinStorageInterface) SetClient(id string, client osin.Client, ign
 	log.Debug("Storing copy in set")
 
 	keyForSet := prefixClientset + prefixClient // Org ID
+
+	// In set, there is no option for update so the existing client should be removed before adding new one.
+	set, _ := r.store.GetSet(keyForSet)
+	for _, v := range set {
+		if strings.Contains(v, client.GetId()) {
+			r.store.RemoveFromSet(keyForSet, v)
```
buger: Do we really need this check? 
Feels like calling `r.store.RemoveFromSet(keyForSet, v)` every time, without getting fitst will do the same job, and will be even faster (because you do not need perform get)

furkansenharputlu: We need. There may be more than one client and user can update randomly.

PR #2817, opened 2020-01-21
Use redis client  api to set expiration of keys

gernest:
This avoid making 2 calls to set key and expiration since redis client
api for set is

```
Set(key string, value interface{}, expiration time.Duration) *StatusCmd
```

This allows us to directly set the expiration. If timeout is 0
then expiration will be 0


komalsukhani:
@gernest Actually we have to set expiry only if timeout is greater than 0

gernest:
Okay

PR #2818, opened 2020-01-21
Fix panic during AuthConfigs assignment

furkansenharputlu:
Fixes https://github.com/TykTechnologies/tyk-analytics/issues/1682


buger:
Any chance we can write test for it?

PR #2820, opened 2020-01-21
Refactor loadAPIEndpoints to loadControlAPIEndpoints

furkansenharputlu:
This PR renames `loadAPIEndpoints` func to `loadControlAPIEndpoints` to make the code more understandable. When you say just `API` it results in a confusion with user APIs.


buger:
Makes sense!

PR #2825, opened 2020-01-23
Fix storage expiration time

### LINKED ISSUE #2805: storage_expiration_time not working correctly
- Branch/Version: tested on release-2.9
- Environment: on prem

**Describe the bug**
It seems the record in redis is stored longer than time defined by storage_expiration_time
This is not 100%

**Reproduction steps**

Set this is tyk.conf
```
enable_analytics: true,
analytics_config.storage_expiration_time: 15
```
1. Add api and key
2. Switch off pump so the record from redis is not processed
2. Send 1 request to that api
3. Observe redis record and its TTL

**Actual behavior**
Sometimes the TTL is counted 2 times so the record resides in redis longer than expected

**Expected behavior**
When TTL goes down to 0 redis record should be removed.

**Screenshots/Video**
redis_ttl_3

[tyk_conf.txt](https://github.com/TykTechnologies/tyk/files/4070769/tyk_conf.txt)

asoorm:
Fixes #2805

Only sets TTL to a list if enabled in tyk.conf & not currently set.
The TTL is now set when shipping the analytics rather than separate goroutine.

---

Moved config check to checkup package.


buger:
But I'm not sure how it actually fix the issue. Goroutine basically does the same.

asoorm:
Go routine logic was bugged. As part of fixing that bug, I moved the logic to only fire if When analytics are sent to Redis. Now you don’t need the go routine. I can move it back if you prefer?

buger:
@maciejwojciechowski I want it to be tested before merging.

maciejwojciechowski:
@buger we tested on this PR, looks ok.

PR #2830, opened 2020-01-24
make sure json tags is consistent across boards

adelowo:
While working on https://github.com/TykTechnologies/tyk-docs/pull/1002 , I 
noticed that the tls config json tags were expected to start with an uppercase
letter as against lowercase all through.


buger:
👌

buger:
@adelowo

Issue #2762, opened 2019-12-24
Secure payload validation fails on key events


ilijabojanovic:
**Branch/Environment/Version**
- Branch/Version: Stable/release-2.9
- Environment: All

**Describe the bug**
When we enable secure payloads on gateway (https://tyk.io/docs/configure/securing-system-payloads/), pub-sub messages for key events fails on validation.

**Reproduction steps**
1. Enable payload validation in tyk.conf
2. Start gateway
3. Over dashboard generate new token

**Actual behavior**
Gateway errors on crate/update/delete keys actions pushed over redis pub-sub channel

**Expected behavior**
Gateway should throw errors

**Screenshots/Video**
If applicable, add screenshots or video to help explain your problem.

**Logs (debug mode or log file):**
```
[Dec 24 14:05:34] ERROR api: Failed to remove the key key=1ef6b5eadf70e532 status=fail
[Dec 24 14:05:34] ERROR pub-sub: Could not verify notification: crypto/rsa: verification error: {KeySpaceUpdateNotification 1ef6b5eadf70e532 }
[Dec 24 14:05:34] ERROR pub-sub: Payload signature is invalid!
[Dec 24 14:05:34] ERROR api: Failed to remove the key key=1ef6b5eadf70e532 status=fail
[Dec 24 14:05:34] ERROR pub-sub: Could not verify notification: crypto/rsa: verification error: {KeySpaceUpdateNotification 1ef6b5eadf70e532 }
[Dec 24 14:05:34] ERROR pub-sub: Payload signature is invalid!
[Dec 24 14:05:34] ERROR api: Failed to remove the key key=eyJvcmciOiI1ZGIyYWJhOGNjZTlmMzBiZmRiMjllN2MiLCJpZCI6IjFlZjZiNWVhZGY3MGU1MzIiLCJoIjoibXVybXVyNjQifQ== status=fail
[Dec 24 14:05:34] ERROR pub-sub: Could not verify notification: crypto/rsa: verification error: {KeySpaceUpdateNotification eyJvcmciOiI1ZGIyYWJhOGNjZTlmMzBiZmRiMjllN2MiLCJpZCI6IjFlZjZiNWVhZGY3MGU1MzIiLCJoIjoibXVybXVyNjQifQ== }
[Dec 24 14:05:34] ERROR pub-sub: Payload signature is invalid!
[Dec 24 14:05:34] ERROR api: Failed to remove the key key=1ef6b5eadf70e532 status=fail
[...]

PR #2819, opened 2020-01-21
Sign gateway notifications with private key before sending

furkansenharputlu:
Fixes #2762

PR #2834, opened 2020-01-26
Fix cross-gateway notification validation

buger:
Alternative approach to https://github.com/TykTechnologies/tyk/pull/2819 which does not require new config value.
Gateways now sign their messages using `node_secret` value.
Added new "SignatureAlgo" notification object field, and if it set to `sha256`, it generates SHA256(type+message+secret) hash, and validates it the same way.

Fix https://github.com/TykTechnologies/tyk/issues/2762

### DIFF
--- a/gateway/gateway_test.go
+++ b/gateway/gateway_test.go
@@ -880,12 +880,18 @@ func TestManagementNodeRedisEvents(t *testing.T) {
 		msg := redis.Message{
 			Payload: `{"Command": "NoticeGatewayDRLNotification"}`,
 		}
+
+		callbackRun := false
 		shouldHandle := func(got NotificationCommand) {
+			callbackRun = true
 			if want := NoticeGatewayDRLNotification; got != want {
 				t.Fatalf("want %q, got %q", want, got)
 			}
 		}
 		handleRedisEvent(&msg, shouldHandle, nil)
+		if !callbackRun {
+			t.Fatalf("Should run callback")
+		}
 		globalConf.ManagementNode = true
 		config.SetGlobal(globalConf)
 		notHandle := func(got NotificationCommand) {
@@ -908,13 +914,18 @@ func TestManagementNodeRedisEvents(t *testing.T) {
 		payload, _ := json.Marshal(n)
 		msg.Payload = string(payload)
 
+		callbackRun := false
 		shouldHandle := func(got NotificationCommand) {
+			callbackRun = true
 			if want := NoticeGroupReload; got != want {
 				t.Fatalf("want %q, got %q", want, got)
 			}
 		}
 
 		handleRedisEvent(&msg, shouldHandle, nil)
+		if !callbackRun {
+			t.Fatalf("Should run callback")
+		}
 
 		n.Signature = "wrong"
 		payload, _ = json.Marshal(n)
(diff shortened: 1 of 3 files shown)


buger:
@furkansenharputlu let me know what do you think about this approach.

gateway/redis_signals.go
```diff
@@ -36,9 +38,16 @@ const (
 
 // Notification is a type that encodes a message published to a pub sub channel (shared between implementations)
 type Notification struct {
-	Command   NotificationCommand `json:"command"`
-	Payload   string              `json:"payload"`
-	Signature string              `json:"signature"`
+	Command       NotificationCommand `json:"command"`
+	Payload       string              `json:"payload"`
+	Signature     string              `json:"signature"`
+	SignatureAlgo string              `json:"algorithm"`
```
furkansenharputlu: Let's make this field `cyrpto.Hash`

gateway/redis_signals.go
```diff
@@ -36,9 +38,16 @@ const (
...
+}
+
+func (n *Notification) Sign() {
+	n.SignatureAlgo = "secret-sha256"
```
furkansenharputlu: If we make the field `crypto.Hash`, we can assign `cyrpto.SHA256` here.

gateway/redis_signals.go
```diff
@@ -132,45 +141,51 @@ var redisInsecureWarn sync.Once
...
-		// Gateway to gateway
+	if config.Global().AllowInsecureConfigs {
 		return true
 	}
 
-	if notification.Signature == "" && config.Global().AllowInsecureConfigs {
-		return true
-	}
+	switch notification.SignatureAlgo {
+	case "secret-sha256":
```
furkansenharputlu: `crypto.SHA256`

gateway/redis_signals.go
```diff
@@ -132,45 +141,51 @@ var redisInsecureWarn sync.Once
...
+		hash := sha256.Sum256([]byte(string(notification.Command) + notification.Payload + config.Global().NodeSecret))
+		expectedSignature := hex.EncodeToString(hash[:])
 
-	if config.Global().PublicKeyPath != "" && notificationVerifier == nil {
-		var err error
+		if expectedSignature == notification.Signature {
+			return true
+		} else {
+			pubSubLog.Error("Notification signer: Failed verifying pub sub signature using node_secret: ")
```
furkansenharputlu: Error logs should start with lower case. Same for all logs in this PR.

gateway/gateway_test.go
```diff
@@ -875,21 +875,60 @@ func TestManagementNodeRedisEvents(t *testing.T) {
...
 		}
-	}
-	handleRedisEvent(&msg, shouldHandle, nil)
-	globalConf.ManagementNode = true
-	config.SetGlobal(globalConf)
-	notHandle := func(got NotificationCommand) {
-		t.Fatalf("should have not handled redis event")
-	}
-	handleRedisEvent(msg, notHandle, nil)
+		shouldHandle := func(got NotificationCommand) {
```
furkansenharputlu: (*) We don't if it is entered here. I think a bool should be defined, set in the func and check after the func.

gateway/gateway_test.go
```diff
@@ -875,21 +875,60 @@ func TestManagementNodeRedisEvents(t *testing.T) {
...
+		n := Notification{
+			Command: NoticeGroupReload,
+			Payload: string("test"),
+		}
+		n.Sign()
+		msg := redis.Message{}
+		payload, _ := json.Marshal(n)
+		msg.Payload = string(payload)
+
+		shouldHandle := func(got NotificationCommand) {
```
furkansenharputlu: Same with (*)

buger:
@furkansenharputlu fixed the comments

PR #2836, opened 2020-01-27
fix use of singleflight in  RedisCacheMiddleware

gernest:
`m.CacheStore.GetKey(key)` is called under singleFlight. This remove
duplicate call of `m.CacheStore.GetKey(key)`  which nullifies the purpose of singleflight.


buger:
Makes sense

buger:
@gernest pls create a ticket, and put to 2.9.4 project

gernest:
@buger done #2838

PR #2851, opened 2020-01-29
Fix request signing middleware

### LINKED ISSUE #2779: Request signing middleware does not strip listen path
- Branch/Version: master/release-2.9

**Describe the bug**
When we sign a request to the upstream, we must use the exact final path/method. 
However, right now, if Tyk API has listen path, it does not get stripped out, so request signature is generated with "ListenPath + Path", while it should be "Path" (only if "Strip listen path" is turned on).

Currently, there is a workaround if you are using URL rewrite, it uses the right path. And current tests are passing only because they use URL rewrite.

**Reproduction steps**
Create API with non-empty listen path, enable request signing (no matter if HMAC or RSA). Sent request to the upstream, and see gateway debug logs for  "Generated sig string: ..." message. Signature in the gateway will contain listen path, even if listen path stripping is enabled.

komalsukhani:
Fixes #2779 

1. Handle case of url-rewrite and strip listen path while generating signature string
2. Adds additional checks for input
3. Added more test cases

### DIFF
--- a/gateway/mw_request_signing.go
+++ b/gateway/mw_request_signing.go
@@ -128,8 +128,13 @@ func (s *RequestSigning) ProcessRequest(w http.ResponseWriter, r *http.Request, 
 	var encodedSignature string
 
 	if strings.HasPrefix(s.Spec.RequestSigning.Algorithm, "rsa") {
+		if s.Spec.RequestSigning.CertificateId == "" {
+			log.Error("CertificateID is empty")
+			return errors.New("CertificateID is empty"), http.StatusInternalServerError
+		}
+
 		certList := CertificateManager.List([]string{s.Spec.RequestSigning.CertificateId}, certs.CertificatePrivate)
-		if len(certList) == 0 {
+		if len(certList) == 0 || certList[0] == nil {
 			log.Error("Certificate not found")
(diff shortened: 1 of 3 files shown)


gateway/mw_request_signing.go
```diff
@@ -101,7 +101,12 @@ func (s *RequestSigning) ProcessRequest(w http.ResponseWriter, r *http.Request,
 	}
 
 	headers := generateHeaderList(r, s.Spec.RequestSigning.HeaderList)
-	signatureString, err := generateHMACSignatureStringFromRequest(r, headers)
+
+	path := r.URL.Path
+	if s.Spec.Proxy.StripListenPath {
+		path = s.Spec.StripListenPath(r, r.URL.Path)
+	}
+	signatureString, err := generateHMACSignatureStringFromRequest(r, headers, path)
```
buger: Do not think it is only about stripping listen path. What if user used URL rewrite, and updated the path, will your change pick this up? Pls test.

komalsukhani: Will check

komalsukhani:
@buger I have made changes to handle case of url-rewrite and added few more checks on input.

PR #2863, opened 2020-02-03
Python session handlers (experimental, don't merge)

matiasinsaurralde:
Sample code:
```python
from tyk.decorators import *
from session import TykSession

tyk.log("middleware.py is loaded", "info")

@Hook
def authhook(request, session, metadata, spec):
  tyk.log("authhook is called", "info")
  token = request.get_header('Authorization')

  # Check if there's a session associated with this token:
  session = tyk.get_session(spec, token)
  if session:
    tyk.log("found a session for token {0}".format(token), "info")
    print("session=",session)
  else:
    tyk.log("no session found for token {0}".format(token), "error")

  # Perform the authentication:
  if token == '47a0c79c427728b3df4af62b9228c8ae':
    tyk.log("auth ok", "info")

    # If there was no session associated with this token, create one and return:
    if session == None:
      tyk.log("initializing a new session", "info")
      session = TykSession()
      session.rate = 1000.0
      session.per = 1.0
      session.metadata["user"] = "matias"
      metadata["token"] = token
      tyk.set_session(token, session, True)

    # The session already exists
    tyk.log("re-using previously created session", "info")
    # This is still required (for now):
    metadata["token"] = token

  # We reach this point if the auth fails:
  tyk.log("auth failed", "error")

```

### DIFF
--- a/coprocess/python/tyk/gateway.py
+++ b/coprocess/python/tyk/gateway.py
@@ -1,10 +1,22 @@
+from session import TykSession
+
 import ctypes
+
 parent = ctypes.cdll.LoadLibrary(None)
 
 parent.TykGetData.argtypes = [ctypes.c_char_p]
 parent.TykGetData.restype = ctypes.c_char_p
 
 parent.TykStoreData.argtypes = [ctypes.c_char_p, ctypes.c_char_p, ctypes.c_int]
+
+class tyk_get_session_ret(ctypes.Structure):
+    _fields_ = [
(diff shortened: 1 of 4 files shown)


buger:
@matiasinsaurralde what is the context of it?

matiasinsaurralde:
@buger it's mostly a draft/experiment to allow interactions with Tyk sessions from plugins, might be useful in the future, for ex. if virtual endpoints support rich plugins or to support more advanced use cases when using auth plugins.

PR #2866, opened 2020-02-06
Fix JWT multi-auth usage

buger:
it was trying to work with uninitialized session object, ignoring the fact that it can be done only if based identity provided by JWT

Fix https://github.com/TykTechnologies/tyk/issues/2580

### DIFF
--- a/gateway/multiauth_test.go
+++ b/gateway/multiauth_test.go
@@ -10,9 +10,12 @@ import (
 	"testing"
 	"time"
 
+	jwt "github.com/dgrijalva/jwt-go"
 	"github.com/justinas/alice"
 	"github.com/lonelycode/go-uuid/uuid"
 
+	"github.com/TykTechnologies/tyk/apidef"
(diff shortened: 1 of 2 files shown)


gateway/multiauth_test.go
```diff
@@ -255,3 +258,69 @@ func TestMultiSession_BA_Standard_FAILAuth(t *testing.T) {
...
+	})[0]
+
+	LoadAPI(spec)
+
+	jwtToken := CreateJWKToken(func(t *jwt.Token) {
+		t.Claims.(jwt.MapClaims)["user_id"] = "user"
+		t.Claims.(jwt.MapClaims)["exp"] = time.Now().Add(time.Hour * 72).Unix()
+	})
+
+	key := CreateSession()
```
letzya: @buger `// Deprecated: Use Test.CreateSession instead.`

buger: Missed this change ;)

But test works, so that should be fine.

Issue #2874, opened 2020-02-13
Plugin bundle download base URL doesn't respect query strings


excieve:
**Branch/Environment/Version**
- Branch/Version: Master/Stable
- Environment: Any

**Describe the bug**
When plugin bundle download URL (`TYK_GW_BUNDLEBASEURL` env var) has a query string parameters in it, the bundle ID is not appended correctly. This is required e.g. to pass some access key as a part of the download URL.

**Reproduction steps**

1. Set `TYK_GW_BUNDLEBASEURL="http://localhost:8081/?api_key=supersecret"`  (or corresponding config file option)
2. Add an API and set the auth mode to Python (or Go) plugin
3. In advanced options set bundle ID to e.g. `mybundle` and save
4. Check gateway logs

**Actual behavior**
Gateway is not able to fetch the bundle because the download URL will look like `http://localhost:8081/?api_key=supersecretmybundle`.

**Expected behavior**
Gateway successfully fetches the bundle and constructs download URL by respecting all its components, e.g. `http://localhost:8081/mybundle?api_key=supersecret`

**Logs (debug mode or log file):**
In this case the error log is:
```
time="Feb 13 01:35:36" level=info msg="----> Fetching Bundle: 8561db0c-b697-4844-8caa-8a3c123bb041" prefix=main
time="Feb 13 01:35:36" level=error msg="Couldn't fetch bundle: HTTP Error, got status code 403" api_id=ff21b4e4e11c4e6b5f836b49651b33f1 org_id=5e4481cad71e960001a807db path=- prefix=main server_name="http://httpbin.org/" user_id=- user_ip=-
```
But the exact message might look differently based on the setup.

**Additional context**
This needs to be changed from a simple concatenation to URL parsing with `net/url` and processing of the bundle name as a concatenation to the path component, while keeping the rest intact:
https://github.com/TykTechnologies/tyk/blob/ba2c9387e3bc283fd693d6faf7ab329313addf57/gateway/coprocess_bundle.go#L201

ilijabojanovic:
Verified

PR #2876, opened 2020-02-13
Parse bundle download URL and allow optional insecure TLS for it

excieve:
Bundle fetcher now parses the bundle base URL and joins the bundle ID to its path component instead of simple string concatenation. Resulting URL is then rendered from the parsed structure. This ensures correct URL will be constructed if the base contains e.g. query string parameters.

This also adds a "bundle_insecure_skip_verify" config option to enable downloading from e.g. test sources that might not have a trusted certificate. Disabled by default.

Addresses #2874

### DIFF
--- a/cli/linter/schema.json
+++ b/cli/linter/schema.json
@@ -278,6 +278,9 @@
     "enable_bundle_downloader": {
       "type": "boolean"
     },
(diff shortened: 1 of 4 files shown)


buger:
@excieve I need to understand priority of it, and when it should be released.

PR #2879, opened 2020-02-14
Fix memory leak

buger:
We stored HTTP handler inside API definition, which probably introduced some circular dependency, confusing the GC.
Handler access itself was needed for the looping, so even if API not loaded to the router (internal), it still could be accessed.

Now handler access moved to the map, where key is API id.
And map is updated in the same maneer as apisByID

### DIFF
--- a/gateway/api_loader.go
+++ b/gateway/api_loader.go
@@ -517,15 +517,15 @@ func (d *DummyProxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 		var handler http.Handler
 		if r.URL.Hostname() == "self" {
-			if d.SH.Spec.middlewareChain != nil {
-				handler = d.SH.Spec.middlewareChain.ThisHandler
+			if h, found := apisHandlesByID[d.SH.Spec.APIID]; found {
+				handler = h
 			}
 		} else {
 			ctxSetVersionInfo(r, nil)
 
 			if targetAPI := fuzzyFindAPI(r.URL.Hostname()); targetAPI != nil {
-				if targetAPI.middlewareChain != nil {
-					handler = targetAPI.middlewareChain.ThisHandler
+				if h, found := apisHandlesByID[targetAPI.APIID]; found {
+					handler = h
 				}
 			} else {
 				handler := ErrorHandler{*d.SH.Base()}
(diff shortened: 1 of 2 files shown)


gateway/api_loader.go
```diff
@@ -517,15 +517,15 @@ func (d *DummyProxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 		var handler http.Handler
 		if r.URL.Hostname() == "self" {
-			if d.SH.Spec.middlewareChain != nil {
-				handler = d.SH.Spec.middlewareChain.ThisHandler
+			if h, found := apisHandlesByID[d.SH.Spec.APIID]; found {
```
gernest: `apisHandlesByID` needs to be protected against concurrent use

gateway/api_loader.go
```diff
@@ -551,12 +551,12 @@ func (d *DummyProxyHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	if d.SH.Spec.target.Scheme == "tyk" {
 		if targetAPI := fuzzyFindAPI(d.SH.Spec.target.Host); targetAPI != nil {
-			if targetAPI.middlewareChain != nil {
+			if h, found := apisHandlesByID[d.SH.Spec.APIID]; found {
```
gernest: protect `apisHandlesByID` against concurrent use

ilijabojanovic:
@buger on this PR

We have panics which crashes gw:
```
time="Feb 17 11:44:16" level=debug msg="Using RPC singleton for connection"
panic: sync: WaitGroup is reused before previous Wait has returned

goroutine 14 [running]:
sync.(*WaitGroup).Wait(0xc0053ae920)
	/usr/local/go/src/sync/waitgroup.go:132 +0xae
github.com/TykTechnologies/tyk/gateway.gatherHealthChecks()
	/develop/go/src/github.com/TykTechnologies/tyk/gateway/health_check.go:177 +0x2c7
github.com/TykTechnologies/tyk/gateway.initHealthCheck.func1(0x1673780, 0xc000032840)
	/develop/go/src/github.com/TykTechnologies/tyk/gateway/health_check.go:89 +0x140
created by github.com/TykTechnologies/tyk/gateway.initHealthCheck
	/develop/go/src/github.com/TykTechnologies/tyk/gateway/health_check.go:69 +0x53
```

This panic is visible only on hybrid gateway.

buger:
@gernest now it use sync.Map instead.

PR #2883, opened 2020-02-18
Rotating signing key

alephnull:
Corresponding changes have been made in rpm-sign.sh in the BuddyFS. Please merge to all release branches to handle historical builds in the future.

### DIFF
--- a/README.signatures
+++ b/README.signatures
@@ -1,0 +1,202 @@
+tl;dr
+=====
+
+RHEL does not support signing RPMs with subkeys. This is supposedly
+fixed in RHEL 8. Thus, we use the same key to sign rpms and debs.
+
+These keys DO NOT expire.
+
+Fingerprints:
+Main key                : 12B5D62C28F57592D1575BD51ED14C59E37DAC20
+Signing subkey (deb)    : 9ADE11DA6DD70355E7C1C270543ABE02AC6AC40A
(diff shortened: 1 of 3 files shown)


excieve:
Should this change the default key ID that the script is using?

alephnull:
I had originally just replaced the file key so this commit was not required. 

This commit is required to keep with the naming convention mentioned by @lonelycode

alephnull:
> Should this change the default key ID that the script is using?

:sweat_smile: Thanks for catching that. Pushed fix.

buger:
Does the `deb` packages have the same signing mechanism?

alephnull:
> Does the `deb` packages have the same signing mechanism?

That is handled by package cloud with their signing key. Debian signing is tied to the repo, not the file.

buger:
Lets put it as part of 2.10 in this case.

buger:
@alephnull pls create a ticket with proper description (and way to test it for the QA), and put it to 2.10 project board.

excieve:
@buger This **needs** to go into the next release (whichever it is).

alephnull:
Failing due to 

```
time="Feb 18 11:48:44" level=info msg="[HOST CHECKER] Stopping poller"
--- FAIL: TestChecker_triggerSampleLimit (28.00s)
    host_checker_test.go:529: expected host down to be fired once got 0 instead
```
Seems unrelated to this change.

PR #2888, opened 2020-02-20
Istio initial work

furkansenharputlu:
- This PR is the initial work for Istio.
- It uses Istio metric template to collect metrics and writes them to `out.txt` for now. 
- For now, I added some metrics but we can customize it and add more.
- Adapter is totally embedded into the gateway. It runs in `4242` port as gRPC server. The port can be made configurable.
- To enable Istio mixer adapter in `tyk.conf`, set `"enable_istio_mixer_adapter" : true`.
- Authentication etc. can be added easily on top of it. This PR is elementary.
- Used `go mod vendor` to get all dependencies in vendor directory


joshblakeley:
@buger @furkansenharputlu what is the plan for this?

PR #2889, opened 2020-02-21
GraphQL middleware template

buger:


apidef/api_definitions.go
```diff
@@ -231,6 +231,70 @@ type ExtendedPathsSet struct {
...
+}
+
+// GraphQLExecution defines the GraphQL Execution Mode as well as its configuration
+type GraphQLExecution struct {
+	// Mode is the mode in which the Middleware should operate
+	Mode       GraphQLExecutionMode       `bson:"mode" json:"mode"`
+	// Validation defines the behaviour regarding GraphQL request validation
+	Validation GraphQLExecutionValidation `bson:"validation" json:"validation"`
+	// Config is the GraphQLExecutionMode specific configuration object
+	Config     json.RawMessage            `bson:"config" json:"config"`
```
buger: It probably makes sense to have it as interface{} instead (like with do with metadata for all objects), otherwise, you will have to parse it each time, since it is just byte slice.

buger:
Not needed anymore

PR #2896, opened 2020-03-01
Feature revoke oauth tokens

sredxny:
Added endpoints to allow users to revoke access and refresh tokens, the endpoints are exposed in 2 levels: dashboard access (its already connected using another endpoint) and exposed directly via api as `OAuthHandler`  (this one is compliant of https://tools.ietf.org/html/rfc7009)

To be specific, the endpoints added are for:

- revoke single token, describing the token and token_type_hint
- revoke all the tokens related to one oauth client, for this the user must provide: apiId, clientSecret and clientID

**Revoke single token**
Method: POST
Path: /{api-id}/oauth/revoke
Content type: application/x-www-form-urlencoded
Body:

- token: string | required
- token_type_hint: string | optional | accepted values = {access_token, refresh_token}

**Revoke all tokens**
Method: POST
Path: /{api-id}/oauth/revoke_all
Content type: application/x-www-form-urlencoded
Body:

- client_id : ID of Oauth client
- client_secret: Secret of Oauth client, used to check the veracity of the request

Related to: https://github.com/TykTechnologies/product/issues/264

### DIFF
--- a/gateway/api.go
+++ b/gateway/api.go
@@ -1883,10 +1883,8 @@ func handleDeleteOAuthClient(keyName, apiID string) (interface{}, int) {
 const oAuthNotPropagatedErr = "OAuth client list isn't available or hasn't been propagated yet."
 const oAuthClientNotFound = "OAuth client not found"
 
-// List Clients
-func getOauthClients(apiID string) (interface{}, int) {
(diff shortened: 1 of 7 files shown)


furkansenharputlu:
Can you please add unit tests with high code coverage?

sredxny:
@furkansenharputlu yes, totally we need that and I have it in consideration :)

Issue #2894, opened 2020-03-01
Python issues in 2.9


matiasinsaurralde:
**Branch/Environment/Version**
- Branch/Version: 2.9
- Environment: All

**Describe the bug**
When using Python 2.9 with the dynamic Python loader, memory usage increases quite fast. A potential issue is that the refcount isn't properly handled in the `Dispatch` method.

**Reproduction steps**

1. Setup an API with the following Python middleware:

```python
from tyk.decorators import *
import gc
@Hook
def myhook(request, session, spec):
  gc.collect()
  print(len(gc.get_objects()))
  return request, session
```

2. The above will force the GC collection and print the number of objects which is expected to increase substantially over time without proper refcount management. This happened in 2.8 and was fixed in #1886. 

**Expected behavior**
The refcount handlers should be added, just like #1886. The main difference is that we use our own bindings in 2.9.

**Logs (debug mode or log file):**
This is what the test plugin output looks like:

```
myhook called
29631
myhook called
29631
myhook called
29632
myhook called
29633
myhook called
29634
myhook called
29635
myhook called
29636
myhook called
29637
myhook called
29638
myhook called
29639
myhook called
29640
myhook called
29641
```

ilijabojanovic:
Original issue is fixed and delivered. New improvements are moved in separate ticket and we will handle them in release 2.10 #2917

PR #2898, opened 2020-03-02
Avoid the "dispatch_hook" reference on every request

matiasinsaurralde:
This gets the reference on initialization and keeps it global, just like other dispatcher references.
Complementary to #2895 (#2894).

Fix for: #2917


matiasinsaurralde:
This is part of #2903 now.

furkansenharputlu:
@matiasinsaurralde If this PR is still valid, can you please add a test to it and try to merge?

PR #2904, opened 2020-03-04
Fix request signing middleware

komalsukhani:
Fixes [2779#issuecomment-593985740](https://github.com/TykTechnologies/tyk/issues/2779#issuecomment-593985740)

1. Handle the case of listen plan containing trailing slash
2. Add test cases
3. Fix test cases of URLRewrite case


buger:
Looks good!

Issue #2803, opened 2020-01-14
Gateway should synchronise only changed objects from the dashboard or MDCB


buger:
**Is your feature request related to a problem? Please describe.**
At the moment, if you have lot of policies or APIs (thousands), each gateway reload takes a lot of time (up to a minute). And any policy or api change/update can trigger such reload. 

In some of the flows, it is required dynamically update policies, and such delays could be unacceptable. 

**Describe the solution you'd like**
It will be great if we will be able to just sync only the APIs/Policies, which actually got changed. 
Dashboard/MDCB can define new APIs, which accept gateway current state like this:
```
{
  "policies": {"<policy-id>": "<last-updated-date>"},
  "apis": {"<api-id>": "<last-updated-date>"} 
}
```
And in response it should get response like:
{
   "updated": {"<policy-id>": <policy-object>},
   "deleted": ["<policy-id>"],
   "created": {"<policy-id>": <policy-object>},
}
The same response should be for the APIs. 

We can make it backward compatible, and first check new "POST" endpoint, and after old "GET" endpoint.

matiasinsaurralde:
Update: I've started to work on this, first on the gateway side. For now I've isolated all the functionality into a function inside `DoReload` so that's called instead of `syncPolicies` and `syncAPISpecs` (from the above description I've assumed that we'll unify both calls).
An interesting finding is that we don't currently have a `last_updated` field in the API definition. Will need to handle this on the dashboard side. I will focus on the dashboard side today.

PR #2906, opened 2020-03-04
New reload mechanism (delta hot reload)

matiasinsaurralde:
For #2803


matiasinsaurralde:
I've refactored the implementation a bit more and merged the standard reload mechanism as well as the new one under a `syncAll` method so that it becomes easier to test. Currently checking the CI with latest changes.

PR #2907, opened 2020-03-05
Certificate and oAuth client listing optimization

tbuchaillot:
Related [#1740 tyk-analytics](https://github.com/TykTechnologies/tyk-analytics/issues/1740)

The idea behind this change is to create an index list with the keys values of certs / OAuth clients to avoid SCAN or KEYS operations.

### DIFF
--- a/certs/manager.go
+++ b/certs/manager.go
@@ -380,13 +380,13 @@ func (c *CertificateManager) ListAllIds(prefix string) (out []string) {
 			c.storage.AppendToSet(indexKey, key)
 			out = append(out, strings.TrimPrefix(key, "raw-"))
 		}
+
 	} else {
 		keys, _ := c.storage.GetListRange(indexKey, 0, -1)
 		for _, key := range keys {
 			out = append(out, strings.TrimPrefix(key, "raw-"))
 		}
 	}
-
 	return out
 }
 
--- a/certs/manager.go
+++ b/certs/manager.go
@@ -13,6 +13,7 @@ import (
 	"encoding/hex"
 	"encoding/pem"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"strings"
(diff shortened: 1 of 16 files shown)


certs/manager.go
```diff
@@ -488,10 +501,18 @@ func (c *CertificateManager) Add(certData []byte, orgID string) (string, error)
...
 	}
 
+	c.storage.AppendToSet(orgID+"-index", "raw-"+certID)
+
 	return certID, nil
 }
 
 func (c *CertificateManager) Delete(certID string) {
+
+	if len(certID) >= sha256.Size*2 {
```
buger: Can we remove this somehow? Maybe move this logic from CertificateManager to Certificate API handler level, where we can have access to the orgID?

gateway/oauth_manager.go
```diff
@@ -688,10 +712,28 @@ func (r *RedisOsinStorageInterface) SetClient(id string, client osin.Client, ign
...
+		for _, v := range keys {
+			if strings.Contains(v, key) {
+				r.store.RemoveFromList(indexKey, v)
+			}
+		}
+	}
+	// append to oauth client index list
+	r.store.AppendToSet(indexKey, key)
+
+	log.Debug("Storing copy in set")
```
buger: Remove debug logs

gateway/oauth_manager.go
```diff
@@ -688,10 +712,28 @@ func (r *RedisOsinStorageInterface) SetClient(id string, client osin.Client, ign
...
+	log.Debug("Storing key in oauth clients index list")
+	indexKey := prefixClientIndexList + orgID
+	//check if the indexKey exists
+	exists, err := r.store.Exists(indexKey)
+	if err != nil {
+		return err
+	}
+	// if it exists, delete the key to avoid duplicity in the client index list
+	if exists {
+		keys, _ := r.store.GetListRange(indexKey, 0, -1)
```
buger: Can we instead of retrieving the list all the time, we can just do the "lrem" of this value?

tbuchaillot: solved

PR #2908, opened 2020-03-06
Add global option to ignore endpoint case

### LINKED ISSUE #2897: Allow configure whitelist/blacklist ignore option on global and API level.
- Branch/Version: 2.9

**Describe the bug**
When we introduced "ignore" option for whitelist/blacklist plugin, we actually broke compatibility with previous versions. These plugins now started respecting the case by default (while previously it was not the case). https://github.com/TykTechnologies/tyk/pull/2339/files

We need a way to make it behave like previously, by adding `ignore` param on the API and Global levels (tyk.conf)

furkansenharputlu:
This PR adds new config parameter `ignore_endpoint_case`. If it is `true`, all endpoint matches will work by ignoring case no matter what is set for the `ignore_endpoint_case` value in API level and the endpoint `ignore_case` value. On UI, maybe we should disable it when the global one is true.  

Also, it adds new API level parameter `ignore_enpoint_case`. If it is `true`, all endpoint matches for the API will work by ignoring case no matter what is set for the endpoint `ignore_case` value.

Fixes #2897


buger:
Should be added on API level too.

PR #2927, opened 2020-03-13
Override oauth errors with global config

### LINKED ISSUE #2897: Allow configure whitelist/blacklist ignore option on global and API level.

furkansenharputlu:
This PR defines a new config variable: `override_messages` to override error messages and codes. Example usage:
```
"override_messages": {
        "oauth.auth_field_missing" : {
            "code": 401,
            "message": "Token is not authorised"
        }
    }
```
The following are the configurable errors for now:
```
"auth.auth_field_missing"
"auth.key_not_found"

"oauth.auth_field_missing"
"oauth.auth_field_malformed"
"oauth.key_not_found"
"oauth.client_deleted"
```

Also, it is possible to override only error code or only error message. Like this:
 ```
"override_messages": {
        "oauth.auth_field_missing" : {
            "code": 401,
        }
    }
```
Fixes https://github.com/TykTechnologies/tyk/issues/2900

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -465,11 +465,9 @@ type Config struct {
 	// Secrets are key-value pairs that can be accessed in the dashboard via "secrets://"
 	Secrets map[string]string `json:"secrets"`
 
-	// TykErrors is used to override returned API error codes and messages.
-	TykErrors map[TykErrorType]TykError `bson:"tyk_errors" json:"tyk_errors"`
+	// OverrideErrors is used to override returned API error codes and messages.
+	OverrideErrors map[string]TykError `bson:"override_errors" json:"override_errors"`
 }
-
-type TykErrorType string
 
 type TykError struct {
 	Message string `json:"message"`
--- a/config/config.go
+++ b/config/config.go
@@ -465,8 +465,8 @@ type Config struct {
 	// Secrets are key-value pairs that can be accessed in the dashboard via "secrets://"
 	Secrets map[string]string `json:"secrets"`
 
-	// OverrideErrors is used to override returned API error codes and messages.
-	OverrideErrors map[string]TykError `bson:"override_errors" json:"override_errors"`
+	// OverrideMessages is used to override returned API error codes and messages.
+	OverrideMessages map[string]TykError `bson:"override_messages" json:"override_messages"`
 }
 
 type TykError struct {
--- a/config/config.go
+++ b/config/config.go
@@ -464,6 +464,16 @@ type Config struct {
 
 	// Secrets are key-value pairs that can be accessed in the dashboard via "secrets://"
 	Secrets map[string]string `json:"secrets"`
+
+	// TykErrors is used to override returned API error codes and messages.
+	TykErrors map[TykErrorType]TykError `bson:"tyk_errors" json:"tyk_errors"`
+}
+
+type TykErrorType string
+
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: config/config.go [hunk 1]
```diff
@@ -464,6 +464,16 @@ type Config struct {
 
 	// Secrets are key-value pairs that can be accessed in the dashboard via "secrets://"
 	Secrets map[string]string `json:"secrets"`
+
+	// TykErrors is used to override returned API error codes and messages.
+	TykErrors map[TykErrorType]TykError `bson:"tyk_errors" json:"tyk_errors"`
+}
+
+type TykErrorType string
```
buger: Maybe rename it to`TykErrorID`?

config/config.go
```diff
@@ -464,6 +464,16 @@ type Config struct {
 
 	// Secrets are key-value pairs that can be accessed in the dashboard via "secrets://"
 	Secrets map[string]string `json:"secrets"`
+
+	// TykErrors is used to override returned API error codes and messages.
+	TykErrors map[TykErrorType]TykError `bson:"tyk_errors" json:"tyk_errors"`
```
buger: Maybe align it with what we already have in python plugins, and name it `override_errors`? 
Note that in future we can add a notation to override upstream server errors too.

gateway/handler_error.go
```diff
@@ -20,6 +23,54 @@ const (
 	defaultContentType    = headers.ApplicationJSON
 )
 
+const (
+	OAuthAuthorizationFieldMissing config.TykErrorType = "oauth.auth_field_missing"
+	OAuthBearerTokenMalformed      config.TykErrorType = "oauth.bearer_token_malformed"
+	OAuthKeyNotAuthorised          config.TykErrorType = "oauth.key_not_authorised"
+	OAuthClientDeleted             config.TykErrorType = "oauth.client_deleted"
```
buger: Suggest to add Err postfix or prefix, to have consistent notation: `OAuthAuthorizationFieldMissingErr` or `ErrOAuthAuthorizationFieldMissing`

gateway/handler_error.go
```diff
@@ -20,6 +23,54 @@ const (
...
+)
+
+var TykErrors = map[config.TykErrorType]config.TykError{
```
buger: We probably should initialize map somewhere else. And each file can define its errors inside `init` function or similar.

config/config.go
buger: On the other hand, maybe we do not need a separate type for it at all

buger:
Let's normalize messages like this:

"auth.auth_field_missing" -> stays as it is
"auth.nonexistent_key" -> "auth.key_not_found"

"oauth.auth_field_missing" -> stays as it is
"oauth.bearer_token_malformed" -> "oauth.auth_field_malformed"
"oauth.key_not_authorised" -> "oauth.key_not_found"
"oauth.client_deleted" -> stays as it is

PR #2929, opened 2020-03-16
Certs ListAllIds fix

tbuchaillot:
Fixing the way we get cert listing when no orgID is given.


furkansenharputlu:
@tbuchaillot Is this PR still valid? If so, can you please rebase and ask re-review?

tbuchaillot:
This is no longer valid

PR #2932, opened 2020-03-17
add json schema for graphql

pvormste:
This pr adds the json schema for graphql in its initial form.

It includes:
- configuration of graphql endpoint
- configuration of validation
- configuration of playground
- configuration of execution (proxyOnly as for now)

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -527,35 +527,35 @@ type RequestSigningMeta struct {
 	SignatureHeader string   `bson:"signature_header" json:"signature_header"`
 }
 
-// GraphQLConfig is the root config object for GraphQL middlewares
+// GraphQLConfig is the root config object for GraphQL middlewares.
 type GraphQLConfig struct {
-	// Enabled indicates if GraphQL should be enabled
+	// Enabled indicates if GraphQL should be enabled.
 	Enabled bool `bson:"enabled" json:"enabled"`
-	// GraphQLAPI is the API specific configuration
+	// GraphQLAPI is the API specific configuration.
 	GraphQLAPI GraphQLAPI `bson:"api" json:"api"`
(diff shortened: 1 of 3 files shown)


apidef/api_definitions.go
```diff
@@ -591,6 +527,75 @@ type RequestSigningMeta struct {
 	SignatureHeader string   `bson:"signature_header" json:"signature_header"`
 }
 
+// GraphQLConfig is the root config object for GraphQL middlewares
+type GraphQLConfig struct {
+	// Enabled indicates if GraphQL should be enabled
+	Enabled bool `bson:"enabled" json:"enabled"`
+	// GraphQLApi is the API specific configuration
+	GraphQLApi GraphQLApi `bson:"api" json:"api"`
```
furkansenharputlu: Better to change to `GraphQLAPI` with all references

apidef/api_definitions.go
```diff
@@ -591,6 +527,75 @@ type RequestSigningMeta struct {
 	SignatureHeader string   `bson:"signature_header" json:"signature_header"`
 }
 
+// GraphQLConfig is the root config object for GraphQL middlewares
```
furkansenharputlu: GoDoc convention is to finish with `.`

Better to finish all GoDoc with `.` in this PR.

apidef/api_definitions.go
jensneuse: @furkansenharputlu what do you mean by "with all references"?

pvormste: In my understanding, it means to rename the type and field to `GraphQLAPI`, so it is consistent with the other definitions regarding naming.
I have changed it now.

apidef/api_definitions.go
pvormste: Done.

Issue #2453, opened 2019-08-11
(mw) global response header transform middleware


asoorm:
**Is your feature request related to a problem? Please describe.**

We can modify global request headers, but not response headers

**Describe the solution you'd like**

Ability to delete or add a response header globally.

For example, a common security issue would be that upstream servers typically expose information which could provide an attack vector by exposing known vulnerabilities:

`X-Powered-By: PHP/some-old-version`
`Server: Nginx`

This plugin will enable the gateway to protect sensitive information leaked from upstream servers.

**Describe alternatives you've considered**

It is possible to use the modify headers plugin, but this means modifying the header for every path. It's unmaintainable and needs to be exposed globally.

sedkis:
@asoorm , in this context, when you say "Global" you mean at the Global API level, not the Global Tyk level, correct?

asoorm:
Correct - Global for the API.

asoorm:
<img width="468" alt="Screenshot 2020-01-08 at 10 00 43" src="https://user-images.githubusercontent.com/1465130/71969058-de0b7080-31fd-11ea-957a-61e7cb67645d.png">

buger:
BE already supports this functionality, just set it to smth like:
```
    "response_processors": [
      {
        "name": "header_injector",
        "options": {
            "add_headers": {"X-Add": "test"},
            "remove_headers": ["Content-Type"]
        }
      }
    ]
```

Note that when you add response headers to the endpoint via plugin, it also adds 'response_processors' section. Also, note that there should be only one `header_injector` processor. In other words, just ensure that both plugin + global work together.

letzya:
@buger when I use the UI I get:
`          "transform_response_headers": [
              {
                "delete_headers": [
                  "X-Ratelimit-Reset"
                ],
                "add_headers": {},
                "path": "get",
                "method": "GET",
                "act_on": false
              }`

What is the difference?

ilijabojanovic:
On latest 2.10/1.10 branch this feature does not work as expected:

Here is testing def:
```
{
    "id": "5eb9586b9633d70aa027e1fd",
    "name": "test-api-4",
    "slug": "test-api-4",
    "listen_port": 0,
    "protocol": "",
    "enable_proxy_protocol": false,
    "api_id": "0853785e1d804ccb4ce2182d32c313a2",
    "org_id": "5dcaa92a9633d70d2397e557",
    "use_keyless": false,
    "use_oauth2": false,
[...]

PR #2935, opened 2020-03-18
Add global response headers adding and removing

furkansenharputlu:
This PR adds new way of adding and removing response global headers in versions level.
```
"versions": {
        "Default": {                
          "global_response_headers": {},
          "global_response_headers_remove": [],       
        }
      }
```

`response_processor` will continue to be the other way. This feature doesn't break it.

Gateway change for: https://github.com/TykTechnologies/tyk/issues/2453


l3ender:
Will the doc be updated when 2.10 is released? https://tyk.io/docs/advanced-configuration/transform-traffic/response-headers/

Issue #2884, opened 2020-02-18
plugin: coprocess auth doesn't apply policy check


asoorm:
**Branch/Environment/Version**

all

**Describe the bug**

It is impossible to use policies with coprocess auth.

**Expected behavior**

An auth plugin should be able to check security policies. This will enable access rights (Authorization), rate-limits & quotas to be managed at the policy level, not the responsibility of the Auth plugin.

ilijabojanovic:
@buger @asoorm 
Original issue is fixed, but we have problem with quotas.

On release-2.9 branch
Reproduction steps:
1. Create policy with quota rates e.g 50 per hour
2. Add that policy in middleware 
3. Generate traffic to that api

**Actual:**
When we reach quota limit, quota counter resets and user can consume api without restrictions 

**Expected:**
When user reach quota limit we should stop traffic to that api due quota limits

**Here is testing bundle:**
[policy.zip](https://github.com/TykTechnologies/tyk/files/4309288/policy.zip)

**Here is video:**
![bug](https://user-images.githubusercontent.com/8557292/76263906-72c33680-6260-11ea-9ac1-6a9a1dab0e40.gif)

ilijabojanovic:
@buger 
On latest release-2.9 branch Issue is still visible
![rate](https://user-images.githubusercontent.com/8557292/76574790-946d2980-64bd-11ea-9b89-21eea0f16dd0.gif)


Worth noticing that we apply in response headers `X-Ratelimit-Reset` but again issue is still visible

buger:
I can't replicate so far. 
Do you use an ID extractor or similar?
Can you try using a "new" unique key?
How are your api and policy look like?

ilijabojanovic:
@buger 
I we tried with new unique key, same issue `X-Ratelimit-Reset` was not set

Here is middleware:
[ilijafix-10.zip](https://github.com/TykTechnologies/tyk/files/4338291/ilijafix-10.zip)

Here is policy obj:
```
{
    "auth_type": "custom",
    "_id": "5e66c4d29633d71135a3b00a",
    "id": "",
    "org_id": "5dcaa92a9633d70d2397e557",
    "rate": 20,
    "per": 60,
    "throttle_interval": -1,
    "throttle_retry_limit": -1,
    "quota_max": 50,
    "quota_renewal_rate": 3600,
    "access_rights": {
        "273df2d6308d457e59cb5b8d3e391cd7": {
            "api_name": "auth",
[...]

PR #2939, opened 2020-03-19
coprocess: update token management logic when using custom authentica…

matiasinsaurralde:
Alternative fix for #2924

We should probably extend the first block to cover other type of token updates, or we could implement them if a requirement is raised in the future.

I've tried different scenarios like the ones described in #2884.


ilijabojanovic:
@buger it's safe for merge. All working as expected

furkansenharputlu:
@matiasinsaurralde Needs a rebase and maybe a test.

matiasinsaurralde:
@furkansenharputlu For solving the issue, we ended up using another approach (https://github.com/TykTechnologies/tyk/pull/2924)

PR #2948, opened 2020-03-20
inherit throttle values from policy

sredxny:
In order to give a fix to https://github.com/TykTechnologies/tyk-analytics/issues/1774 was added logic to inherit the throttle values from policy


buger:
pls fix test

PR #2949, opened 2020-03-23
inherit rate values on rate partition policy

sredxny:
when a policy for rates partition is applied, set the correct values in key values, this aso fix the inheritance of values when more than one partition policy is applied 

Related to: https://github.com/TykTechnologies/tyk-analytics/issues/1748


ilijabojanovic:
@gernest @buger why this fix is merged on release-2.9 branch?

PR #2950, opened 2020-03-23
Adding packer to tyk-build-env

alephnull:
Also noted that build pipelines were using wrong version of tyk-build-env. New version `1.3` pushed and tagged as `latest`. All pipelines should now be using latest tag.

### DIFF
--- a/images/build-env/Dockerfile
+++ b/images/build-env/Dockerfile
@@ -1,9 +1,10 @@
 FROM golang:1.12
 LABEL io.tyk.vendor="Tyk" \
-      version="1.1" \
+      version="1.3" \
       description="Base image for builds"
 
 ENV GOPATH=/
+ENV PACKER_VERSION="1.5.4"
 
 RUN apt-get update && apt-get dist-upgrade -y && \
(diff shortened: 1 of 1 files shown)


excieve:
@alephnull  Could you please elaborate on why is packer needed in this one?

IIRC this image is also the base image for plugin builder. Packer is really only needed to our internal pipelines that build AMIs, no?

alephnull:
> Packer is really only needed for our internal pipelines that build AMIs, no?
Yes, correct. 

packer used to be installed via a custom env section in buddy but when I moved everything to use the build-env base image, I removed that section as it was 
- easy to forget that it was there
- updating the packer version was cumbersome

Perhaps I should make another image that can be used for AMI builds.

excieve:
@alephnull OK, let's keep it in this image for now — it's silly huge anyway :) Split later if it becomes an issue.

alephnull:
> @alephnull OK, let's keep it in this image for now — it's silly huge anyway :)

Agreed. With a little work I might be able to hive off all of ruby from this. python-dev will still keep it large though.

PR #2962, opened 2020-03-26
Create schema in GraphQL middleware and check

furkansenharputlu:


gateway/mw_graphql.go
```diff
@@ -17,17 +21,19 @@ func (m *GraphQLMiddleware) EnabledForSpec() bool {
...
-func (m *GraphQLMiddleware) Destructor() {
+	m.Schema = schema
 }
 
 func (m *GraphQLMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Request, _ interface{}) (error, int) {
-	logger := m.Logger()
-	logger.Info("I'm loaded")
+
+	if m.Schema == nil {
+		return errors.New("schema is not created"), http.StatusInternalServerError
```
jensneuse: Will this error message be returned to the client? I would avoid disclosing this information to the API user.

furkansenharputlu: If there was a problem with schema, when user makes a request, it is returned to the user

PR #2965, opened 2020-03-26
Add murmur64 to the example as it's preferred setting over murmur3

letzya:
Setting murmur64 to be default as it's preferred setting over murmur3. 
With murmur3 you encounter old bugs that you can use a key even if you add to it a few characters. So you think you are on the latest release but actually the old setting causing that.


tyk.conf.example
```diff
@@ -32,6 +32,7 @@
     "policy_source": "file"
   },
   "hash_keys": true,
+  "hash_key_function": “murmur64",
```
excieve, line 35:
```suggestion
  "hash_key_function": "murmur64",
```

asoorm:
Recommend to remove the boiler-plate / template commit message body. Otherwise approved.

furkansenharputlu (approved):

PR #2971, opened 2020-03-30
Debug rate limit algorithm being used

gernest:


buger:
In such sensitive places, we can write it like this:
```
if log.Level == logrus.Debug {
    log.Debug(....)
}
```

gernest:
@asoorm do you have an alternative  suggestion? This was requested by QA team while trying to test rate limiting algorithms feature.

There are about `265` use of `logrus.Fields` in the gateway right now.

gernest:
Ooops, sorry @asoorm I typed my comment faster before @buger comment showed up.

buger:
@gernest, for now, we can apply the trick I mentioned only to rate limiter code (since it is very sensitive)

gernest:
@buger done PTAL

PR #2975, opened 2020-03-30
Implement GraphQL request validation

furkansenharputlu:
Related to https://github.com/TykTechnologies/product/issues/281

### DIFF
--- a/gateway/handler_error.go
+++ b/gateway/handler_error.go
@@ -23,7 +23,7 @@ const (
 	defaultContentType    = headers.ApplicationJSON
 )
 
-var customizableErr = errors.New("customizable error")
+var errCustomBodyResponse = errors.New("errCustomBodyResponse")
 
 var TykErrors = make(map[string]config.TykError)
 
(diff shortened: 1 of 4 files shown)


gateway/handler_error.go
```diff
@@ -23,6 +23,8 @@ const (
 	defaultContentType    = headers.ApplicationJSON
 )
 
+var customizableErr = errors.New("customizable error")
```
pvormste, line 26: I think this error should have a better name (and error text) as it will be hard to remember in 6 months what the purpose of this error was. I could even think about it to add an explaining comment to it.

something like `errCustomBodyResponse` could work.

gateway/mw_graphql.go
```diff
@@ -43,5 +45,19 @@ func (m *GraphQLMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reques
...
+	result, err := gqlRequest.ValidateForSchema(m.Schema)
+	if err != nil {
+		m.Logger().Errorf("Error while validating GraphQL request: '%s'", err)
+		return errors.New("there was a problem proxying the request"), http.StatusInternalServerError
+	}
+
+	if result.Errors != nil && result.Errors.Count() > 0 {
+		w.Header().Set(headers.ContentType, headers.ApplicationJSON)
+		w.WriteHeader(http.StatusBadRequest)
+		_, _ = result.Errors.WriteResponse(w)
```
pvormste, line 57: is it good not to handle a write error here?

furkansenharputlu: Errors are built-in in library and they are tested,  not handling them in gateway makes code clear I think

PR #2977, opened 2020-03-30
filter by orgId on search apis for oauthClients

sredxny:
When looking for the apis that uses certain oauthClient then first we filter by OrgId to make the search faster.

Related to: https://github.com/TykTechnologies/tyk-analytics/issues/1817


tbuchaillot:
IMO org_id should be a **required** field because, in another way, it's going to timeout like now

PR #2978, opened 2020-03-31
Tests fix

tbuchaillot:
Fix some test in order to stabilize master tests.

### DIFF
--- a/certs/manager.go
+++ b/certs/manager.go
@@ -374,16 +374,17 @@ func (c *CertificateManager) ListRawPublicKey(keyID string) (out interface{}) {
 func (c *CertificateManager) ListAllIds(prefix string) (out []string) {
 	indexKey := prefix + "-index"
 	exists, _ := c.storage.Exists(indexKey)
-	if !exists {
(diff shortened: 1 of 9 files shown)


sredxny:
This PR also constains a fix for https://github.com/TykTechnologies/tyk-analytics/issues/1815

certs/manager.go
```diff
@@ -374,16 +374,17 @@ func (c *CertificateManager) ListRawPublicKey(keyID string) (out interface{}) {
...
-	if !exists {
-		keys := c.storage.GetKeys("raw-" + prefix + "*")
+	if exists && prefix != "" {
+		keys, _ := c.storage.GetListRange(indexKey, 0, -1)
 		for _, key := range keys {
-			c.storage.AppendToSet(indexKey, key)
 			out = append(out, strings.TrimPrefix(key, "raw-"))
 		}
-
 	} else {
```
matiasinsaurralde, line 382: Would it make sense to remove this block and make "if exists" return earlier:
```go
if exists && prefix != "" {
  // ...
  return out
}
keys := c.storage.GetKeys("raw-" + prefix + "*")
// ...
return out

PR #2988, opened 2020-04-02
Fire event on authentication failure

larxxor:
When Custom Go plugin returns 403 StatusForbidden, fire an event indicating Auth failure.

### DIFF
--- a/gateway/mw_go_plugin.go
+++ b/gateway/mw_go_plugin.go
@@ -3,13 +3,14 @@ package gateway
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
+	"net/http"
+	"time"
+
 	"github.com/TykTechnologies/tyk/ctx"
 	"github.com/TykTechnologies/tyk/goplugin"
 	"github.com/TykTechnologies/tyk/request"
 	"github.com/sirupsen/logrus"
-	"io/ioutil"
-	"net/http"
-	"time"
 )
 
 // customResponseWriter is a wrapper around standard http.ResponseWriter
--- a/gateway/mw_go_plugin.go
+++ b/gateway/mw_go_plugin.go
@@ -151,10 +151,12 @@ func (m *GoPluginMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reque
 		if rw.statusCodeSent >= http.StatusForbidden {
 
 			m.logger.WithError(err).Error("Authentication error in Go-plugin middleware func")
-
-			m.Base().FireEvent(EventAuthFailure, EventMetaDefault {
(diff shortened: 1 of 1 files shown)


gateway/mw_go_plugin.go
```diff
@@ -148,7 +148,19 @@ func (m *GoPluginMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reque
...
+
+			m.logger.WithError(err).Error("Authentication error in Go-plugin middleware func")
+			m.Base().FireEvent(EventAuthFailure, EventKeyFailureMeta{
+				EventMetaDefault: EventMetaDefault{Message: "Auth Failure", OriginatingRequest: EncodeRequestToEvent(r)},
+				Path:             r.URL.Path,
+				Origin:           request.RealIP(r),
+				Key:              "n/a",
+			})
+			respCode = rw.statusCodeSent
+			err = fmt.Errorf("plugin function sent auth failed response code: %d", rw.statusCodeSent)
```
spetrunin, line 161: we are not logging this error anymore

gateway/mw_go_plugin.go
```diff
@@ -148,7 +148,19 @@ func (m *GoPluginMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Reque
 	// check if response was sent
 	if rw.responseSent {
 		// check if response code was an error one
-		if rw.statusCodeSent >= http.StatusBadRequest {
+		if rw.statusCodeSent >= http.StatusForbidden {
```
spetrunin, line 151: I propose to replace this whole `if` statement with a `switch`
And add `fallthrough` to keep error processing same as before and reduce duplication
```go
    // check if response code was an error one
		switch {
		case rw.statusCodeSent == http.StatusForbidden:
			m.logger.WithError(err).Error("Authentication error in Go-plugin middleware func")
			m.Base().FireEvent(EventAuthFailure, EventKeyFailureMeta{
				EventMetaDefault: EventMetaDefault{Message: "Auth Failure", OriginatingRequest: EncodeRequestToEvent(r)},
				Path:             r.URL.Path,
				Origin:           request.RealIP(r),
				Key:              "n/a",
			})
			fallthrough
		case rw.statusCodeSent >= http.StatusBadRequest:
			// base middleware will report this error to analytics if needed
			respCode = rw.statusCodeSent
			err = fmt.Errorf("plugin function sent error response code: %d", rw.statusCodeSent)
			m.logger.WithError(err).Error("Failed to process request with Go-plugin middleware func")
		default:
			// record 2XX to analytics
			m.successHandler.RecordHit(r, Latency{Total: int64(ms)}, rw.statusCodeSent, rw.getHttpResponse(r))

			// no need to continue passing this request down to reverse proxy
			respCode = mwStatusRespond
		}
```

PR #2991, opened 2020-04-03
[TT-1954] Add the errCode (http status code) to error view model

Bouwdie:
I would like to render the http status code in the error response body. And i mean; why not? :-)

### DIFF
--- a/gateway/handler_error.go
+++ b/gateway/handler_error.go
@@ -82,6 +82,7 @@ func overrideTykErrors() {
 
 // APIError is generic error object returned if there is something wrong with the request
 type APIError struct {
+	Status  int
 	Message template.HTML
 }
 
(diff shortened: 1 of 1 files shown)


furkansenharputlu:
@Bouwdie It will  already be returned as a separate field in the response object, why would you need to put it in body as well?

Bouwdie:
@furkansenharputlu Glad you asked; we would like this for the following reason:

-  some of our customers are not able to access response headers easily from their client implementations
- standards like problem+json advocate for adding the status to the response body

furkansenharputlu:
Hi @Bouwdie I created a ticket and added it to our `2.10.1` release board. I will discuss it with team. You can see the status of the issue: 

https://github.com/TykTechnologies/tyk/issues/3011

andyo-tyk:
We have implemented a comprehensive Error Overrides feature in Tyk 5.13.0. Please check that out.

PR #3015, opened 2020-04-13
on bind to slugs=true then add a path separator on oauth paths

sredxny:
Behaviour of `/{yourapi}/oauth/token` change depending of the value of `slave_options.bind_to_slugs`

## Description
On bind to slugs we are not setting a slash at the end of the route, then, in `addOAuthHandlers` we concatenate a route to that path without adding a separator between slug and `oauthPath`, so, was added that slash depending of the value of `slave_options.bind_to_slugs` otherwise we would need to hit the route `/{yourapi}oauth/token` to have the same result (current behaviour). Other Oauth routes were affected by this same behaviour, so was added the separator as well. Now, we need to be able to revoke tokens issued locally, currently we revoke tokens issued from dashboard/master GW, this change will be done in a different PR

## Related Issue
https://github.com/TykTechnologies/tyk-analytics/issues/1842

## Motivation and Context
Solves https://github.com/TykTechnologies/tyk-analytics/issues/1842 so in slave GW we are allowed to issue oauth tokens no matter what is the value of `bind_to_slugs`

## How This Has Been Tested
- Created an API (from dashboard)
- Created an Oauth Client (from dashboard)
- Run Slave gateway
- Request a token with `slave_options.bind_to_slugs=true`
- Request a token with `slave_options.bind_to_slugs=false`

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist

  - [x] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -493,7 +493,7 @@ func generateOAuthPrefix(apiID string) string {
 func addOAuthHandlers(spec *APISpec, muxer *mux.Router) *OAuthManager {
 	var pathSeparator string
 	if !strings.HasSuffix(spec.Proxy.ListenPath, "/") {
-		pathSeparator = ""
+		pathSeparator = "/"
 	}
 
 	apiAuthorizePath := spec.Proxy.ListenPath + pathSeparator + "tyk/oauth/authorize-client{_:/?}"
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -493,7 +493,7 @@ func generateOAuthPrefix(apiID string) string {
 func addOAuthHandlers(spec *APISpec, muxer *mux.Router) *OAuthManager {
 	var pathSeparator string
 	if !strings.HasSuffix(spec.Proxy.ListenPath, "/") {
-		pathSeparator = "/"
+		pathSeparator = ""
 	}
 
 	apiAuthorizePath := spec.Proxy.ListenPath + pathSeparator + "tyk/oauth/authorize-client{_:/?}"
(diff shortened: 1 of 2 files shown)


gateway/server.go
```diff
@@ -491,11 +491,16 @@ func generateOAuthPrefix(apiID string) string {
...
 // Create API-specific OAuth handlers and respective auth servers
 func addOAuthHandlers(spec *APISpec, muxer *mux.Router) *OAuthManager {
-	apiAuthorizePath := spec.Proxy.ListenPath + "tyk/oauth/authorize-client{_:/?}"
-	clientAuthPath := spec.Proxy.ListenPath + "oauth/authorize{_:/?}"
-	clientAccessPath := spec.Proxy.ListenPath + "oauth/token{_:/?}"
-	revokeToken := spec.Proxy.ListenPath + "oauth/revoke"
-	revokeAllTokens := spec.Proxy.ListenPath + "oauth/revoke_all"
+	var pathSeparator string
+	if config.Global().SlaveOptions.BindToSlugsInsteadOfListenPaths {
+		pathSeparator = "/"
```
buger, line 496: I believe you should use Slug value instead of "/"

tbuchaillot:
LGTM

gateway/server.go
sredxny: Slug is already set in ListenPath:

```
if config.Global().SlaveOptions.BindToSlugsInsteadOfListenPaths {
   newListenPath := "/" + def.Slug //+ "/"
```

Issue #3008, opened 2020-04-09
Health check: RPC status isn't updated when mdcb is down


ilijabojanovic:
**Branch/Environment/Version**
- Branch/Version: Master/release-2.10
- Environment: MDCB

**Describe the bug**
When MDCB goes down, gateway status for RPC isn't updated. 

**Reproduction steps**

1. Start MDCB
2. Connect slave gateway to MDCB
3. Stop MDCB

**Actual behavior**
RPC status and general status if wrong
```
{"status":"pass","version":"v2.9.3","description":"Tyk GW","details":{"redis":{"status":"pass","componentType":"datastore","time":"2020-04-09T11:51:21+02:00"},"rpc":{"status":"pass","componentType":"system","time":"2020-04-09T11:51:21+02:00"}}}
```
**Expected behavior**
Since MDCB is down we should have `Warn` state and display proper `RPC` status message

**Screenshots/Video**
If applicable, add screenshots or video to help explain your problem.

**Logs (debug mode or log file):**
```
[Apr 09 11:51:50]  INFO RPC Conn Mgr: Can't connect to RPC layer error=gorpc.Client: [localhost:9090]. Cannot obtain response during timeout=3s
[Apr 09 11:51:51]  WARN Keyspace warning: gorpc.Client: [localhost:9090]. Cannot obtain response during timeout=3s
[Apr 09 11:51:53]  INFO RPC Conn Mgr: Can't connect to RPC layer error=gorpc.Client: [localhost:9090]. Cannot obtain response during timeout=3s
[Apr 09 11:51:54] ERROR Can't purge cache, failed to ping RPC error=gorpc.Client: [localhost:9090]. Cannot obtain response during timeout=3s
[Apr 09 11:51:56]  INFO RPC Conn Mgr: Can't connect to RPC layer error=gorpc.Client: [localhost:9090]. Cannot obtain response during timeout=3s
[Apr 09 11:51:59]  INFO RPC Conn Mgr: Can't connect to RPC layer error=gorpc.Client: [localhost:9090]. Cannot obtain response during timeout=3s
[Apr 09 11:52:02]  INFO RPC Conn Mgr: Can't connect to RPC layer error=gorpc.Client: [localhost:9090]. Cannot obtain response during timeout=3s
[Apr 09 11:52:04] ERROR Can't purge cache, failed to ping RPC error=gorpc.Client: [localhost:9090]. Cannot obtain response during timeout=3s
[...]

PR #3019, opened 2020-04-15
change way how we detect that we are connected to rpc for healtCheck

sredxny:
## Description
Changed condition on how we check that we are connected to RPC, before we used the `Connect` method of RPC but it contains certain conditions that doesn't match with the expected result, then, as we constantly check if we can login to the group, then its a good criteria to use this value to check the status of rpc.

## Related Issue
https://github.com/TykTechnologies/tyk/issues/3008

## Motivation and Context
Give solution to https://github.com/TykTechnologies/tyk/issues/3008

## How This Has Been Tested
- Run MDCB setup
- Stop Sink
- Check health status for the slave GW pointing to http://tyk-gateway:8182/hello
- Check logs and ensure that we are printing a Warning log giving information that we are not connected to RPC

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/health_check.go
+++ b/gateway/health_check.go
@@ -175,7 +175,7 @@ func gatherHealthChecks() {
 				Time:          time.Now().Format(time.RFC3339),
 			}
 
-			if !rpc.GroupLogin() {
+			if !rpc.Login() {
 				checkItem.Output = "Could not connect to RPC"
 				checkItem.Status = Fail
 			}
--- a/gateway/health_check.go
+++ b/gateway/health_check.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"github.com/TykTechnologies/tyk/rpc"
 	"net/http"
 	"sync"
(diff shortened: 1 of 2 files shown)


gateway/health_check.go
```diff
@@ -173,9 +175,7 @@ func gatherHealthChecks() {
 				Time:          time.Now().Format(time.RFC3339),
 			}
 
-			rpcStore := RPCStorageHandler{KeyPrefix: "livenesscheck-"}
-
-			if !rpcStore.Connect() {
+			if !rpc.GroupLogin() {
```
buger, line 178: We also have 'Login' function there. And we run either group login or standard login depending on the setup. 
Also, this logic is built in inside rpc Login function:
```
	// If we have a group ID, lets login as a group
	if config.GroupID != "" {
		return GroupLogin()
	}
```
So feels like replacing `GroupLogin` with `Login` should do the job.

sredxny: Change done and tested

Issue #3017, opened 2020-04-14
[Task] Implement field restriction on Gateway


furkansenharputlu:
Combine all the graphql library, gateway and dashboard, made them work properly and make api based field restrictions.

PR #3041, opened 2020-04-22
Feature/field restrictions

furkansenharputlu:
Fixes https://github.com/TykTechnologies/tyk/issues/3017


gateway/api_definition.go
```diff
@@ -17,6 +17,8 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/jensneuse/graphql-go-tools/pkg/graphql"
```
spetrunin, line 20: could we group imports?
like
```
import (
  std_imports
 
  third party

  tyk
)
```

gateway/middleware.go
```diff
@@ -403,6 +403,20 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 							}
 						}
 
+						for _, t := range v.RestrictedTypes {
```
spetrunin, line 406: do we support partitioning of restricted types?

spetrunin: I know it is unrelated to this PR but could we make iteration variables names more meaningful?

like `v -> accessRight` and `k -> apiID`

gateway/mw_granular_access.go
```diff
@@ -30,34 +32,56 @@ func (m *GranularAccessMiddleware) ProcessRequest(w http.ResponseWriter, r *http
...
+			w.Header().Set(headers.ContentType, headers.ApplicationJSON)
+			w.WriteHeader(http.StatusBadRequest)
+			_, _ = result.Errors.WriteResponse(w)
+			m.Logger().Errorf("Error during GraphQL request restricted fields validation: '%s'", result.Errors)
+			return errCustomBodyResponse, http.StatusBadRequest
+		}
+
+		return nil, http.StatusOK
+
+	} else {
```
spetrunin, line 58: `else` branch is not required here as you already have returns in `if`

gateway/middleware.go
furkansenharputlu: You are right, it is very unclear but better to make it in another PR.

PR #3049, opened 2020-04-27
Generate list of env var names for Config

gernest:
Many people are not so sure about env var names to use
when overiding config values. This  patch generates config_env_vars.md.

The file lists all possible env var names for tyk config.
A test is present to ensure the file is updated when config changes.

## Description
Based on our [docs](https://tyk.io/docs/tyk-configuration-reference/environment-variables/) some variables will be wrongly constructed for instance `TYK_GW_TRACING_ENABLED` might be assumed to be correct while `TYK_GW_TRACER_ENABLED` is the correct env var.
 
## Related Issue
https://github.com/TykTechnologies/tyk-docs/issues/1187

## Motivation and Context
This has been asked so many times on slack.

## How This Has Been Tested
The variables are automatically generated by the library which we use. Tests are included in the library.

## Types of changes

## Checklist

- [x] When updating library version must provide reason/explanation for this update.
  - [ ] `go fmt -s`
  - [ ] `go vet`


sedkis:
You.
Are.
A.
Legend

marksou:
I'm going to take the markdown file and cross reference with existing GW Config page - https://tyk.io/docs/tyk-configuration-reference/tyk-gateway-configuration-options/. I'll add the equivalent env var to the json value. We can then see if there are any gaps in Config page. Thanks @gernest

buger:
Can we get full JSON key, e.g. with parent keys?

gernest:
@buger  Nope, we can't. That will require separate work outside of the `envconfig` API and I don't see any point of it as it is beyond the scope of this PR.

PR #3057, opened 2020-04-29
rpc: ensure recovery from emergency mode

### LINKED ISSUE #3050: APIs and policies not being loaded from Dashboard after restart
- Branch/Version: 2.9.3.1
- Environment: Hybrid

**Describe the bug**
Due to network connectivity problems after a Gateway restart, it couldn't connect to the Dashboard via RPC.  The Gateway moved to an 'EmergencyMode' and it performed a cold start, retrieving the APIs, policies and keys from Redis:

```
time="Apr 24 10:14:26" level=info msg="Setting new RPC connection!"
time="Apr 24 10:14:46" level=info msg="Starting Poller" prefix=host-check-mgr
time="Apr 24 10:14:56" level=error msg="RPC Login failed" error="gorpc.Client: [hybrid.cloud.tyk.io:9091]. Cannot obtain response during timeout=30s"
time="Apr 24 10:14:56" level=info msg="[RPC Conn Mgr] Starting keepalive watcher..." prefix="RPC Conn Mgr"
time="Apr 24 10:14:56" level=info msg="[RPC] Starting keyspace poller"
time="Apr 24 10:14:56" level=warning msg="[RPC Store] --> Detected cold start, attempting to load from cache"
time="Apr 24 10:14:56" level=warning msg="[RPC Store] ----> Found APIs... beginning emergency load"
```

After this, the Gateway continued to try to reconnect to RPC but due to the network problems, during this period it was not possible to load new APIs, Policies or API Keys:

```
time="Apr 24 10:15:26" level=info msg="Can't connect to RPC layer" error="gorpc.Client: [hybrid.cloud.tyk.io:9091]. Cannot obtain response during timeout=30s" prefix="RPC Conn Mgr"
time="Apr 24 10:15:26" level=warning msg="Keyspace warning: gorpc.Client: [hybrid.cloud.tyk.io:9091]. Cannot obtain response during timeout=30s"
```

After a while they solved the connectivity issues and the RPC client could reconnect again. New API Keys were loaded correctly, however, it appears APIs and Policies were still loaded from Redis instead of the Dashboard:

```
time="Apr 24 11:08:38" level=warning msg="[RPC STORE] Received Reload instruction!"
time="Apr 24 11:08:38" level=info msg="Reloading endpoints" prefix=pub-sub
time="Apr 24 11:08:38" level=info msg="Reload queued" prefix=main
time="Apr 24 11:08:38" level=info msg="reload: initiating" prefix=main
time="Apr 24 11:08:38" level=info msg="Loading policies" prefix=main
time="Apr 24 11:08:38" level=info msg="[RPC] Loading Policies from backup"
...
time="Apr 24 11:08:38" level=info msg="[RPC] --> Loading API definitions from backup"
...
```

It looks like the Gateway was still in 'EmergencyMode' and consequently would not resync the APIs and Policies with the dashboard. It kept loading the APIs from Redis.
Restarting the Gateway eventually solved this issue, but do we think this is a bug that the Gateway would not reload Policies and APIs from the dashboard after a cold start, even if the RPC connection is restored after a while??

**Reproduction steps**

1. Simulate network connectivity issue to Hybrid Dashboard 
2. Restart Gateway

**Actual behavior**
New keys are loaded, but APIs and Policies continue to be loaded from Redis

**Expected behavior**
APIs and Policies should be re-loaded from the Dashboard

[api-gateway-cold-start-logs.gz](https://github.com/TykTechnologies/tyk/files/4544171/api-gateway-cold-start-logs.gz)

gernest:
## Description

Our rpc implementation was using home brewed retry mechanism that was hard to test and prone to flooding rpc server with login requests. Also it was also hard to know/prove if this was able to properly recover from emergency mode after it has succeeded to login after failing to do so.

So in a. gist

- We had racy unprotected global variables
- Home brew retry mechanism
- No test to ensure we properly recover from emergency mode

## Related Issue
fixes #3050

## Motivation and Context
The changes here are important ensure we are properly recovering from emergency mode.

- Global variables use atomics which protects against race condition. This makes it safer for concurrent use.
- Use exponential backoff algorithm for retries using a well tested/solid implementation
- Add tests for recovery logic 

## How This Has Been Tested
Unit test a have been included
```
go test ./rpc
```

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`


rpc/rpc_client.go
```diff
@@ -231,42 +283,50 @@ func Connect(connConfig Config, suppressRegister bool, dispatcherFuncs map[strin
...
+	// Unless someone audits to ensure all of where this appears the parent calls
+	// are not concurrent, this is a much safer solution.
+	v, _, _ := loginFlight.Do("Login", func() (interface{}, error) {
+		return loginBase(), nil
+	})
+	return v.(bool)
+}
 
-	if strings.Contains(err.Error(), "Cannot obtain response during timeout") {
-		reConnect()
```
gernest, line 255: This is a no-op. Which doesn't help  much here as we will never recover from the emergency mode when we get timeout from connecting to remote rpc server.

rpc/rpc_client_test.go
```diff
@@ -0,0 +1,44 @@
+package rpc
+
+import "testing"
+
+func TestRecoveryFromEmregencyMode(t *testing.T) {
+	if IsEmergencyMode() {
+		t.Fatal("expected not to be in emergency mode before initiating login attempt")
+	}
+	hasAPIKey := func() bool { return true }
+	// gript login has the same recovery api so we don't need to test for it.
```
sredxny, line 10: did you mean "group"?

rpc/rpc_client_test.go
```diff
@@ -0,0 +1,44 @@
...
+	}, func() error {
+		return errLogFailed
+	}, hasAPIKey, isGroup)
+	if ok {
+		t.Fatal("expected to fail login")
+	}
+	if !IsEmergencyMode() {
+		t.Fatal("expected to be in emergency mode")
+	}
+	if true {
```
sredxny, line 24: this condition will prevent the code below to be executed

rpc/rpc_client.go
```diff
@@ -231,42 +283,50 @@ func Connect(connConfig Config, suppressRegister bool, dispatcherFuncs map[strin
...
+// if it fails.
+func Login() bool {
+	// I know this is extreme but rpc.Login() appears about 17 times and teh
+	// methods appears to be sometimes called in goroutines.
+	//
+	// Unless someone audits to ensure all of where this appears the parent calls
+	// are not concurrent, this is a much safer solution.
+	v, _, _ := loginFlight.Do("Login", func() (interface{}, error) {
+		return loginBase(), nil
+	})
```
sredxny, line 296: do not omit the error, even if it will return false we must log that here we get an error

gernest: This will never return an error because the error you see returned comes from the  ` func() (interface{}, error) ` where we always return `loginBase(), nil` which means logging error will be meaningless as we no for a fact we always have `nil` returned(It means we will always be doing `if nil==nil` check. Other errors however are properly logged on their respective locations.

gernest:
@sredxny thanks for taking a look. I addressed your comments Please take another look.

Issue #3058, opened 2020-04-29
allow users to set seconds to be checking for key space changes


sredxny:
**Is your feature request related to a problem? Please describe.**
Currently the method in charge of check for key space changes is listening for changes every 10 seconds, the idea is to make this value configurable from config file in the slave options section.

**Describe the solution you'd like**
Add a new field in the config file, inside the section `slave options`. Also if this value is not set then set by default 10 seconds

**Describe alternatives you've considered**

**Additional context**
This will not only benefit tyk gateway users to configure this value, but also is a requisite to build the _mdcb testing framework_, as we will need to check quickly that changes are being synchronized successfully

PR #3059, opened 2020-04-29
read the seconds to check for key space changes from config file, als…

sredxny:
## Description
Allow users to set the number of seconds before GW check again for key space changes

## Related Issue
https://github.com/TykTechnologies/tyk/issues/3058

## Motivation and Context
give solution to https://github.com/TykTechnologies/tyk/issues/3058 and accomplish one of the requisites to build the MDCB testing framework

## How This Has Been Tested
- Starts a MDCB setup
- Set 1 seconds  as value of this field
- Restart slaved GW
- Check logs, we must see every second a record saying that its checking for key space changes

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`

### DIFF
--- a/cli/linter/schema.json
+++ b/cli/linter/schema.json
@@ -735,6 +735,9 @@
         },
         "rpc_pool_size": {
           "type": "integer"
+        },
+        "seconds_to_check_key_space_changes": {
+          "type": "integer"
         }
       }
     },
--- a/cli/linter/schema.json
+++ b/cli/linter/schema.json
@@ -737,7 +737,7 @@
           "type": "integer"
         },
         "key_space_sync_interval": {
-          "type": "float"
+          "type": "number"
         }
       }
     },
--- a/cli/linter/schema.json
+++ b/cli/linter/schema.json
@@ -736,7 +736,7 @@
         "rpc_pool_size": {
           "type": "integer"
         },
-        "seconds_to_check_key_space_changes": {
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -167,6 +167,7 @@ type SlaveOptionsConfig struct {
 	CallTimeout                     int    `json:"call_timeout"`
 	PingTimeout                     int    `json:"ping_timeout"`
 	RPCPoolSize                     int    `json:"rpc_pool_size"`
+	SecondsToCheckKeySpaceChanges   int    `json:"seconds_to_check_key_space_changes"`
```
buger, line 170: Pls name it `keyspace_sync_interval`

sredxny: Done! currently there is a test that doesn't pass but is not related to this ticket. (Hostchecker test)

cli/linter/schema.json
```diff
@@ -735,6 +735,9 @@
         },
         "rpc_pool_size": {
           "type": "integer"
+        },
+        "key_space_sync_interval": {
+          "type": "integer"
```
buger, line 740: I wonder if you want make it float, so tests will be faster?

sredxny: @buger that's a very good idea. I just pushed that change, tested and its ok

sredxny:
Documentation https://github.com/TykTechnologies/tyk-docs/pull/1386

Issue #3061, opened 2020-05-04
Keys: Quota counter does not work when quota renew is never


ilijabojanovic:
**Branch/Environment/Version**
- Branch/Version: Master/Release-2.10
- Environment: All

**Describe the bug**
When quota renewal is set to never, quota counter does not work

**Reproduction steps**
1. Log on dashboard 
2. Go to keys
3. Create token with some quota e.g 50
4. Set renwal Never

**Actual behavior**
When we sent request to gateway only first call will be calculated, after that we don't calculate remaining quota

**Expected behavior**
Quota counter should work and when quota is reached we should stop traffic.

**Screenshots/Video**
![bug](https://user-images.githubusercontent.com/8557292/80976511-d6fb1480-8e23-11ea-800e-936ffb84811b.gif)

**Logs (debug mode or log file):**
Log from console or from log file.

**Configuration (tyk config file):**
Attach tyk configuration file

**Additional context**

ilijabojanovic:
On release-2.10 branch on c07a0d24045b0eb05940f3acfa3cfc2987f40728

This issue is still visible:
![bug2](https://user-images.githubusercontent.com/8557292/81284064-e6fe3880-905d-11ea-817b-1e57573b0f92.gif)

ilijabojanovic:
Verified

PR #3065, opened 2020-05-05
for keys with quota renew set to never, dont renew the periods

sredxny:
## Description
When we reach the quota max, if the key have set "never renew the quota" then we must not restart the period and not renew the quota. 

## Related Issue
https://github.com/TykTechnologies/tyk/issues/3061

## Motivation and Context
Give solution to https://github.com/TykTechnologies/tyk/issues/3061

## How This Has Been Tested
- Created policy with quota set to 5 and renewal rate = never
- Created policy applying that policy
- Use the key, and I should only be allow to use it 5 times, unless the key is updated with renew quota

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`


komalsukhani:
@sredxny When user sets **Never** for **Quota resets every** in the UI, **QuotaRenewalRate** is set to -1 not 0.

Issue #3056, opened 2020-04-29
[Task] Implement execution mode on Gateway


spetrunin:
part of [[Story: 3] Define data sources for schema fields](https://github.com/TykTechnologies/tyk-analytics/issues/1890)

Acceptance criteria:
- [x] we could create datasource from datasource configuration fields in APIDefinition

PR #3077, opened 2020-05-08
Implement GraphQL execution engine

furkansenharputlu:
Fixes https://github.com/TykTechnologies/tyk/issues/3056


gateway/api_loader.go
```diff
@@ -676,6 +678,29 @@ type generalStores struct {
 	redisStore, redisOrgStore, healthStore, rpcAuthStore, rpcOrgStore storage.Handler
 }
 
+func loadGraphQLPlayground(router *mux.Router) {
```
furkansenharputlu, line 681: Needs to be decided where to live, I put it now temporarily.

furkansenharputlu: I implemented this as another PR to fix the sprint5 task separately. I will send the PR after this is merged. Let's keep this in the root playground level for this PR.

gateway/mw_graphql.go
```diff
@@ -27,12 +39,69 @@ func (m *GraphQLMiddleware) Init() {
...
 	}
 
-	m.Spec.graphqlSchema = schema
+	m.Spec.GraphQLExecutor.Schema = schema
+
+	if m.Spec.GraphQL.GraphQLAPI.Execution.Mode == apidef.GraphQLExecutionModeExecutionEngine {
+
+		typeFieldConfigurations := m.Spec.GraphQL.GraphQLAPI.TypeFieldConfigurations
+		typeFieldConfigurations = append(typeFieldConfigurations, datasource.TypeFieldConfiguration{
+			TypeName:  "query",
```
spetrunin, line 48: as Patric fixed lowercase issue, should we use `Query` here?

Issue #3080, opened 2020-05-12
Internal api usage doesn't work


furkansenharputlu:
**Branch/Environment/Version**
- Branch/Version: Master
- Environment: On-prem

**Describe the bug**
Internal api usage makes gateway to crash

**Reproduction steps**

1. Add api with name `first`
2. Add another api with name `second`
3. For `second`, try to set target URL as `tyk://first`
4. Make request `http://tyk-gateway:8181/second/get`

**Actual behavior**
Gateway crashes.

**Expected behavior**
It should route to the other api successfully and return result

ilijabojanovic:
This is duplicate of https://github.com/TykTechnologies/tyk/issues/2619

ilijabojanovic:
Verified

PR #3081, opened 2020-05-12
Fix internal api usage as target

furkansenharputlu:
The problem was that it was always calling the same api's handler and enters endless loop. It should call the target apis handler, so needs to search `apisHandlesByID` with target api id. 

Fixes https://github.com/TykTechnologies/tyk/issues/3080


buger:
Can you explain a bit of how it solves the issue? I'm a bit puzzled. 

Also does the "self" still works?

furkansenharputlu:
I don't think `self` was working for target url. It should be for url-rewrite.

buger:
This function is used not only by target URL but also with URL rewrites too.

buger:
@furkansenharputlu ^

furkansenharputlu:
Merged manually

PR #3083, opened 2020-05-12
On quota or rate set as partitioned policy, set the values in key data

sredxny:
…l data of the key

## Description
if we apply a policy that has partitioned quota or rate then those values will be enforced and set in the global fields of a key. Before, as an example (it applies for rate and quota) if `session.QuotaMax` was -1 and `policy.QuotaMax`any other value then in the key we were not setting this value in the global scope of the key and -1 was set instead, and that is not correct at all as if we have a policy with infinite quota and another with partitioned quota, the partitioned quota should be set.

## Related Issue
https://github.com/TykTechnologies/tyk-analytics/issues/1748 some changes were applied to this same function `ApplyPolicies`, specifically the places where the function `greaterThanInt64` is used is were changed in some way the logic. 

## Motivation and Context
give solution to https://github.com/TykTechnologies/tyk-analytics/issues/1748

## How This Has Been Tested
- Create a policy P1 with quota partitioned
- Create a policy P2 with rate partitioned
- Created key applying the policies P1 and P2
- The key returned, in the global fields for Quota and rate should have the values set in P1 and P2

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`

### DIFF
--- a/gateway/middleware.go
+++ b/gateway/middleware.go
@@ -406,10 +406,11 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 
 				if !usePartitions || policy.Partitions.Quota {
 					didQuota[k] = true
-
 					if greaterThanInt64(policy.QuotaMax, ar.Limit.QuotaMax) {
+
 						ar.Limit.QuotaMax = policy.QuotaMax
-						if greaterThanInt64(policy.QuotaMax, session.QuotaMax) {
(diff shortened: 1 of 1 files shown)


sredxny:
@furkansenharputlu Hi, I saw that you made some changes in this logic recently, so would be great if you take a look at it, so we ensure that we're not gonna break the logic in this new PR ;) Thanks!

PR #3094, opened 2020-05-21
fix listen path bug with kv store

adelowo:
## Description
When a listen path is set as `/consul://api_url`. The listen path stays that way as there isn't a kV store associated with that prefix. `consul://api_url` is what it supposed to be. Take a look at the image below. That api will be inaccessible because the value won't be read from consul

Screen Shot 2020-05-21 at 15 46 02

Also made sure users can have either of these variations in the consul/vault definition:

- `/some-listen-path/`
- `some-listen-path/`

> Both will resolve to `/some-listen-path`/

## How This Has Been Tested

- In the listen path, add `consul://url`
- In consul, create a key as "url" with the actual listen path
- make a request to `http://gateway:8000/<listen path in consul>`

> Try with this PR and without it

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`


adelowo:
Will be fix by UI

PR #3126, opened 2020-06-04
include empty values for graphql api execution mode in json schema

spetrunin:
## Description
Allow execution mode enum values to be null or empty

## Related Issue
https://github.com/TykTechnologies/tyk/pull/3123

### DIFF
--- a/apidef/schema.go
+++ b/apidef/schema.go
@@ -437,12 +437,11 @@ const Schema = `{
 					"type": "boolean"
 				},
 				"execution_mode": {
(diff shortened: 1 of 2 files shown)


apidef/schema.go
```diff
@@ -437,10 +437,12 @@ const Schema = `{
 					"type": "boolean"
 				},
 				"execution_mode": {
-					"type": "string",
+					"type": ["string", "null"],
```
furkansenharputlu, line 440: Why null? Can a string be null ?

PR #3142, opened 2020-06-11
[TT-423] Fix gRPC  proxy and add h2c  proxying support

joshblakeley:
## Description

This draft enables optional H2C support, both in the server and in the reverse proxy.

## Related Issue

https://github.com/TykTechnologies/product/issues/334

## Motivation and Context

When Tyk is proxying gRPC it may be the case that those services are not secure. Currently only HTTP/2 over TLS is supported by Tyk.

In service meshes such as Istio the gRPC service will not be secured in any way as the security in the mesh is offloaded to Envoy. When proxying service to service calls in this scenario it will be over HTTP/2 cleartext (h2c)

This draft enables us to support this in scenarios when needed by wrapping the server handler for the API in the h2c helper handler provided by Go https://godoc.org/golang.org/x/net/http2/h2c

For the upstream request when a h2c request is required the underlying Transport is swapped out for a http2 transport that will allow http calls.

## How This Has Been Tested

Unit test and example client and server code to test will be provided.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


matiasinsaurralde (approved):

buger:
You can use command-line tools from this repo to test it out: https://github.com/thrawn01/h2c-golang-example

joshblakeley:
Also https://github.com/fullstorydev/grpcurl

buger:
Additionally, you can use this tool to validate if h2c service proxied though Tyk is still valid https://github.com/summerwind/h2spec

h2spec -p 8888 http2

sredxny:
Also was added the property `proxy.enable_h2c` at API Level, so, we can control h2c enabling now at whole GW level as well as API level. 

Note that after this changes are merged then we need to update the vendor in Dashboard as well as in sink, the reason behind this is:

- In dashboard: because we load the apis from dashboard api, if dashboard doesn't know anything about this new field then it will be always false
- In sink: to replicate the value across all different nodes

sredxny (approved):

matiasinsaurralde:
Merged into `release-3.1` (f2567e9adff8dd426487c7ee58ce06029057815a) and `release-3-lts` (85d33f70cb01512365cf2478a1175f65f8bc1c11).

Issue #718, opened 2017-05-04
Extend Virtual Paths to call gRPC endpoints


lonelycode:
Currently it's possible to have Tyk proxy to a service and for it to run local JS code to handle orchestrations, however this can be quite expensive and lead to a latency hit.

It would be better (and more extensible) if I could set up an API in tyk that has as it;s target a gRPC server and then set up endpoints that map to functions on the gRPC host, essentially creating an RPC API Gateway.

The feature would mean that Tyk supports Websockets, HTTP Rest and RPC natively, it also significantly extends the capability of Tyk as a proxy for anyone using gRPC as a pattern, or more likely, need to provide glue-code and orchestration code that is more performant than Otto.

l3ender:
As we start to use gRPC more in our environment, being able to use Tyk to proxy gRPC would be a great benefit.

l3ender:
Please keep open. Native gRPC support would be a great feature for Tyk.

andyo-tyk:
I think this ticket should have been closed some time ago; Tyk can now be used as a gRPC proxy, as detailed [here](https://tyk.io/docs/key-concepts/grpc-proxy/).

PR #3157, opened 2020-06-17
Add https case when processing tcp requests

joshblakeley:
This will unblock basic grpc proxying for the gateway.

## Description
Widen check to allow https traffic through when TLS api configured
## Motivation and Context

Allows secured gRPC traffic through the gateway which is currently not possible without breaking the gRPC spec

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/tcp/tcp.go
+++ b/tcp/tcp.go
@@ -224,14 +224,14 @@ func (p *Proxy) handleConn(conn net.Conn) error {
 		} else {
 			rconn, err = net.Dial("tcp", u.Host)
 		}
-	case "tls":
+	case "tls", "https":
 		if p.DialTLS != nil {
 			rconn, err = p.DialTLS("tcp", u.Host)
 		} else {
 			rconn, err = tls.Dial("tcp", u.Host, p.TLSConfigTarget)
 		}
 	default:
-		err = errors.New("Unsupported protocol. Should be empty, `tcp` or `tls`")
+		err = errors.New("Unsupported protocol. Should be empty, `tcp`, `tls` or `https`)
 	}
 	if err != nil {
 		conn.Close()


joshblakeley:
@gernest gRPC client libraries will always use the protocol https or grpc neither of which we can handle currently.

An alternative to doing this in the tcp package is to allow https/grpc apis in tyk to listen on only host:port.

I'm trying to open the conversation up a bit here about what the best solution is so really appreciate you taking a look.

What do you think?

buger:
I'm not sure how it solves h2c issue. 
The only effect of this change, that you will be able to use "https" scheme for the target URL when your API is in TCP mode. But technically Tyk will be in TCP mode, and it will not be performing HTTPS requests to the target: it will be TCP passthrough, and we will not have access to things like routing by URL and etc. 
If it so, then semantically it probably not right to allow setting "https", and it will work the same if you just specific "tls://" scheme. 

Am I missing anything?

joshblakeley:
> I'm not sure how it solves h2c issue.
> The only effect of this change, that you will be able to use "https" scheme for the target URL when your API is in TCP mode. But technically Tyk will be in TCP mode, and it will not be performing HTTPS requests to the target: it will be TCP passthrough, and we will not have access to things like routing by URL and etc.
> If it so, then semantically it probably not right to allow setting "https", and it will work the same if you just specific "tls://" scheme.
> 
> Am I missing anything?

This doesn't solve h2c issue - this is a workaround for gRPC services over HTTPS which currently cannot be called through Tyk either.

I don't actually think we should merge this but this was the only way to make gRPC work with Tyk for a POC and that should definitely not be the case. 

I want this to be a conversation starter in the engineering team about how to cross the gap we have at the moment for anyone with gRPC APIs.

gernest:
>  this is a workaround for gRPC services over HTTPS which currently cannot be called through Tyk either.

I don't see why you can't do this. Because with tcp service we are just tunnelling the traffic, and the transport here is tcp. The trick for changing target to `tls://` will entirely remove the need for this change, because they all serve the same purpose.

With your change, we force tyk to open tcp socket to the api even though the api url says it speaks https, if we explicitly change the target of the api to say it speaks tls  tyk will be happy to open a socket to the api without forcing anything.

joshblakeley:
> > this is a workaround for gRPC services over HTTPS which currently cannot be called through Tyk either.
> 
> I don't see why you can't do this. Because with tcp service we are just tunnelling the traffic, and the transport here is tcp. The trick for changing target to `tls://` will entirely remove the need for this change, because they all serve the same purpose.
> 
> With your change, we force tyk to open tcp socket to the api even though the api url says it speaks https, if we explicitly change the target of the api to say it speaks tls tyk will be happy to open a socket to the api without forcing anything.

Ok fine, but this is not the main point. I would like to be able to use auth and middlewares in the case for HTTPS API where ONLY the host:port is exposed (like we do for TLS API). That is the real problem.

gernest:
> I would like to be able to use auth and middlewares in the case for HTTPS

This is a valid point. We can't do that with tcp api's.

gernest:
@joshblakeley a quick search I found this  #718 which when you follow comments someone mentions native `gRPC` support.

I believe you are asking for native gRPC support right? I think it will a nice feature. I wonder though if the shape of gRPC traffic will be similar to our http stack so we can reuse most of out http stack features.

joshblakeley:
We have documents claiming we can do gRPC proxying here: https://site-dev.tykbeta.com/docs/key-concepts/grpc-proxy/ 
but the implementation is not correct.

gRPC clients will not call on a path - the path information in contained in a special pseudoheader. So we need to be able to listen on an empty listen path for cases where the gateway either has custom domain or custom port. Both for HTTP and HTTPS traffic.

gernest:
@joshblakeley why did you close this? I think its interesting.

If the claim for gRPC is wrong then this is very relevant. I just don't have gRPC expertise yet to say anything about this. Its definitely worth investigating.

A quick search led me  to this project https://github.com/mwitkow/grpc-proxy . What is your thought on it?

joshblakeley:
Closed because i'll take a different code approach to get this fixed hopefully - will find a better avenue for discussion than here too :)

Issue #3162, opened 2020-06-19
API calls takes longer in slaved gw when MDCB is down and analytics are enabled


sredxny:
**Branch/Environment/Version**
- Branch/Version:  Master
- Environment: MDCB

**Describe the bug**
In a MDCB environment when we have set `"enable_analytics": true` the response of the request for any API gets affected and increased when MDCB is down.

**Reproduction steps**

1. Execute Tyk in MDCB environment
2. In the config file for the slave gateway set `enable_analytics` to true
3. Register an API
4. Consume API in slave (and check the time taken to get a response)
5. Shut down tyk-sink (MDCB)
6. Consume again the API (and check again the time taken, it was increased)

**Actual behavior**
Api response time gets increased when MDCB is down and `enable_analytics` is true

**Expected behavior**
The response time should not be affected, or at least it should not take that longer.

**Screenshots/Video**
If applicable, add screenshots or video to help explain your problem.

**Logs (debug mode or log file):**
Log from console or from log file.

**Configuration (tyk config file):**
Attach tyk configuration file

**Additional context**
It was tested using an hybrid pump as well as RPC, the result was the same.

christtyk:
@sredxny was this merged/moved to jira/can we close?

christtyk:
released 3.2

PR #3163, opened 2020-06-19
when rpc in emergency mode then dont check org session in rpc

sredxny:
## Description
When a slave gw has enabled analytics, and sink is down, then dont try to get org session from rpc

## Related Issue
https://github.com/TykTechnologies/tyk/issues/3162

## Motivation and Context
Give solution to https://github.com/TykTechnologies/tyk/issues/3162 also, fix how gw behaves when sink is down and do not affect the response time for the apis.

## How This Has Been Tested
- Setup MDCB environment and ensure that slaved gw have `enable_analytics` set to true
- Consume api in master and slaved gw
- Turn off sink
- Consume API in slaved gw, the time of response should be similar as step 2 (however, consider that it can take some seconds to change the state to emergency mode)
- Consume api in master (to ensure that nothing is broken now)
- Turn on again sink
- Consume api in slave

Some load test were performed using apache benchmark sending 10 request in parallel and 100 request in total (executing  `ab -n 100 -c 10 http://tyk-gateway:8182/1/` ), next are the results after the change:

1- **With Sink up&running:**
```
Requests per second:    13.91 [#/sec] (mean)
Time per request:       718.859 [ms] (mean) (this is for a batch of 10 requests)
Time per request:       71.886 [ms] (mean, across all concurrent requests) (per single request)
```
2- With Sink down:
```
Requests per second:    13.79 [#/sec] (mean)
Time per request:       725.245 [ms] (mean)
Time per request:       72.524 [ms] (mean, across all concurrent requests)
```

As we see, the number are almost the same. And just as a reference these are the numbers with sink down before make this change:

```
Requests per second:    0.97 [#/sec] (mean)
Time per request:       10302.129 [ms] (mean)
Time per request:       1030.213 [ms] (mean, across all concurrent requests)
```

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`

### DIFF
--- a/gateway/middleware.go
+++ b/gateway/middleware.go
@@ -230,7 +230,7 @@ func (t BaseMiddleware) OrgSession(orgID string) (user.SessionState, bool) {
 	var session user.SessionState
 	var found bool
 
-	if !rpc.IsEmergencyMode(){
+	if !rpc.IsEmergencyMode() {
 		// Try and get the session from the session store
 		session, found = t.Spec.OrgSessionManager.SessionDetail(orgID, orgID, false)
 		if found && t.Spec.GlobalConfig.EnforceOrgDataAge {
--- a/gateway/middleware.go
+++ b/gateway/middleware.go
@@ -231,18 +231,20 @@ func (t BaseMiddleware) OrgSession(orgID string) (user.SessionState, bool) {
 	var session user.SessionState
 	var found bool
 
-	if !rpc.IsEmergencyMode() {
-		// Try and get the session from the session store
-		session, found = t.Spec.OrgSessionManager.SessionDetail(orgID, orgID, false)
-		if found && t.Spec.GlobalConfig.EnforceOrgDataAge {
-			// If exists, assume it has been authorized and pass on
-			// We cache org expiry data
-			t.Logger().Debug("Setting data expiry: ", session.OrgID)
-			ExpiryCache.Set(session.OrgID, session.DataExpires, cache.DefaultExpiration)
-		}
-
-		session.SetKeyHash(storage.HashKey(orgID))
+	if rpc.IsEmergencyMode() {
+		return session, false
 	}
+
+	// Try and get the session from the session store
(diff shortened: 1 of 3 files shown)


sredxny:
Captura de Pantalla 2020-06-19 a la(s) 6 13 27 p  m
Its failing `TestGRPCDispatch` in the pipeline but not in local, can we confirm that its not an issue in the cloud?

gernest:
@sredxny does that fix I mention related to the failed test?

sredxny:
> @sredxny does that fix I mention related to the failed test?

No, as the test passed locally. I just uploaded a change that is not related to that test and now it passed in the pipeline...weird

gateway/middleware.go
```diff
@@ -226,16 +228,22 @@ func (t BaseMiddleware) Config() (interface{}, error) {
...
-	session, found := t.Spec.OrgSessionManager.SessionDetail(orgID, orgID, false)
-	if found && t.Spec.GlobalConfig.EnforceOrgDataAge {
-		// If exists, assume it has been authorized and pass on
-		// We cache org expiry data
-		t.Logger().Debug("Setting data expiry: ", session.OrgID)
-		ExpiryCache.Set(session.OrgID, session.DataExpires, cache.DefaultExpiration)
+	var session user.SessionState
+	var found bool
+
+	if !rpc.IsEmergencyMode() {
```
gernest, line 234: so in emergency mode session will always be nil. Won't that impact the gateway?

buger: It is not nil, but empty object, pls we return `found` as false so I guess we should be save. 

But I would rewrite it to the guard statement like this:
```
if rpc.IsEmergencyMode() {
   return session, false
}
```

letzya:
@sredxny woudn't it be cool if our integration testing could do this https://github.com/TykTechnologies/tyk/pull/3163#issue-437320874 for every merge :)

spetrunin:
To keep the release branch green (in case we will have to merge this branch into other release branches) we should have the next order of applying fixes:
#3163
#3306
#3308

Issue #3195, opened 2020-07-09
Not able to load multiple native Go plugins


matiasinsaurralde:
**Branch/Environment/Version**
- Branch/Version: All
- Environment: All

**Describe the bug**
When using the `tyk-plugin-compiler` to build plugins, all generated shared libraries use the same package path.
This causes an error when trying to load more than one plugin, for example, when `plugin1` is used in `api1 and `plugin2` is used in `api2`:
```
level=error msg="Could not load Go-plugin" error="plugin.Open("/tmp/test-plugin/plugin1"): plugin already loaded" mwPath="/tmp/test-plugin/plugin1.so" mwSymbolName=AddFooBarHeader
```
It's important to note that this error isn't related to duplicate handler/hook names but it's caused by a common package path that's used in all Docker-based builds, even if the host directory is a different one, it's mounted under the same path as a Docker volume on every build.

As a workaround my suggestion is to use a dynamic package path, this is also mentioned in [golang discussions](https://github.com/golang/go/issues/19004#issuecomment-288923294).

**Reproduction steps**

1. Create a directory for your sample plugin (`/tmp/sample-plugin`) and a file named `plugin.go` with the following contents:
```go
package main

import (
  "net/http"
)

// AddFooBarHeader adds custom "Foo: Bar" header to the request
func AddFooBarHeader(rw http.ResponseWriter, r *http.Request) {
  r.Header.Add("Foo", "Bar")
}

func main() {}
```
2. Build a Go plugin:
```
$ docker run --rm -v `pwd`:/plugin-source tykio/tyk-plugin-compiler:v2.9.4.2 plugin1.so 
```
3. Build a second version:
```
$ docker run --rm -v `pwd`:/plugin-source tykio/tyk-plugin-compiler:v2.9.4.2 plugin2.so 
```
4. Prepare two API definitions, one will load `plugin.so`, the other one will try to load `plugin2.so`:

**API 1:**
```json
    "custom_middleware": {
        "pre": [],
        "post_key_auth": [],
        "auth_check": {},
        "post": [
            {
                "name": "AddFooBarHeader",
                "path": "/tmp/test-plugin/plugin1.so"
            }
        ],
        "driver": "goplugin"
    },
```

**API 2:**
```json
    "custom_middleware": {
        "pre": [],
        "post_key_auth": [],
        "auth_check": {},
        "post": [
            {
                "name": "AddFooBarHeader",
                "path": "/tmp/test-plugin/plugin2.so"
[...]

PR #3203, opened 2020-07-10
Add some documentation and tests for goplugins

alephnull:
Uses docker-compose

## Description
A Makefile and some documentation to test goplugins. This is for gateway versions _after_ 2.9.4.1 as prior versions exposed `plugin-build` and will probably not work.

## Related Issue
#3195 

## Motivation and Context
Testing goplugins is cumbersome as it requires a bit of setup and the commands are long and easy to mess up.

## How This Has Been Tested
By hand.

## Types of changes

## Checklist

- [x] I have updated the documentation accordingly.
  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/images/plugin-compiler/README.md
+++ b/images/plugin-compiler/README.md
@@ -1,0 +1,28 @@
+# Tyk Plugin compiler
+
+## Building a plugin
+Navigate to where your plugin is and build using a docker volume to
+mount your code into the image. Since the vendor directory needs to be
+identical between the gateway build and the plugin build, this means
+that you should pull the version of this image corresponding to the
+gateway version you are using.
+
+This also implies that if your plugin has vendored modules that are
+also used by Tyk gateway then your module will be overridden by the
+version that Tyk uses.
+
+```shellsession
+% docker run -v `pwd`:/plugin-source tykio/tyk-plugin-compiler:v2.9.4.2 myplugin.so
+```
+
+You will find a myplugin.so in the current directory which is the file
+that goes into the API definition
+
+## Plugin aliasing
+Due to the way that `dl_open(3)` works, the filename needs to change
+for `libdl` to recognise that the plugin has been updated. Else, the
+cached value from `ld.so.cache` will be used.
+
+## Hot reloading
+Using `/tyk/reload/group` will _not_ update the plugin if you have
+compiled a fresh version and used the same pathname.
--- a/images/plugin-compiler/README.md
+++ b/images/plugin-compiler/README.md
@@ -29,3 +29,5 @@ more details.
 ## Hot reloading
 Using `/tyk/reload/group` will _not_ update the plugin if you have
 compiled a fresh version and used the same pathname.
+
+To reload a plugin, you will have to restart the gateway process.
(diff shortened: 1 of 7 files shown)


images/plugin-compiler/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+```shellsession
+% docker run -v `pwd`:/plugin-source tykio/tyk-plugin-compiler:v2.9.4.2 myplugin.so
+```
+
+You will find a myplugin.so in the current directory which is the file
+that goes into the API definition
+
+## Plugin aliasing
+Due to the way that `dl_open(3)` works, the filename needs to change
```
letzya, line 22: Are you referring to the fact you need to give the name `myplugin.so` when compiling? It's not so clear

images/plugin-compiler/README.md
```diff
@@ -0,0 +1,28 @@
...
+mount your code into the image. Since the vendor directory needs to be
+identical between the gateway build and the plugin build, this means
+that you should pull the version of this image corresponding to the
+gateway version you are using.
+
+This also implies that if your plugin has vendored modules that are
+also used by Tyk gateway then your module will be overridden by the
+version that Tyk uses.
+
+```shellsession
```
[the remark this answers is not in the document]
alephnull: ?

images/plugin-compiler/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Plugin aliasing
+Plugins are loaded via `dl_open(3)` and the shared library cache,
+`ld.so.cache` will be used. Therefore,even if a plugin's content
+changes but the filename does not, the cached plugin will be used.
+
+See the manpages for `dl_open(3)` and `ld.so(8)` on your platform for
+more details.
+
+## Hot reloading
```
letzya, line 29: Maybe I missed it but you explain how not to reload a plugin. Can we explain what we should do when we want to update the plugin?
(I assume- same name and restart or new name and hot reload?)

images/plugin-compiler/README.md
```diff
@@ -0,0 +1,33 @@
...
+Using `/tyk/reload/group` will _not_ update the plugin if you have
+compiled a fresh version and used the same pathname.
+
+To reload a plugin, you will have to restart the gateway process.
```
letzya, line 33: This could open a sea of complaints if that’s the only way @alephnull

matiasinsaurralde: This is hard to fix unless we implement our own dynamic loading logic. We should wait for the `plugin` package to implement something that allows the safe unloading/reloading of Go shared libraries, etc.
For now the only strategy is to use different filenames every time the code changes or perform a full process reload.

images/plugin-compiler/README.md
```diff
@@ -0,0 +1,33 @@
+# Tyk Plugin compiler
+
+## Building a plugin
+Navigate to where your plugin is and build using a docker volume to
```
matiasinsaurralde, line 4: Docker

PR #3208, opened 2020-07-13
TT-2 Bugfix/issue 3070 

### LINKED ISSUE #3070: TT-2 Gateway should record response payload also case of an error
- Branch/Version: [e.g. Master/Release/Stable/Feature branch]
- Environment: [e.g. On-prem/Hybrid/MDCB]
On prem, 2.9.4

**Describe the bug**
If there was an error (in tyk or upstream), tyk does not record the payload

From the client:
When an upstream backend replies something, a 4xx/5xx error or 2xx response, we can capture the response body clearly in the Logs Brower UI and via Pump.

But when it is Tyk.io that replies a 4xx error directly to the client, then the response body send by Tyk.io to the client does not show in logs: in either in the Logs Browser UI or via Pump feeding our own ELK instance.
**It appears that such a response message also does not appear in the file logs from the Gateways.**

We need to be able to capture in some logs the response message that Tyk.io sent to our external clients, or at least some information of same value, so we can take appropriate support actions toward them. 
Doesn't make sense that the caller will have more information than the dashboard user that views the analytics

**Reproduction steps**

1. Turn on detailed recording in tyk.conf and maybe sure the pump is running
2. Create an API with authentication
3. Call the API without authorization header and get an error
4. For completeness also make a valid call with the API key
5. Check the "log browser" in the dashboard and see that the payload in the response is missing. Screengrab attached

**Actual behavior**
No payload recording in the analytics in case of an error in the upstream or tyk

**Expected behavior**
Do the same as we do for regular calls - record the response. 

**Screenshots/Video**
The error response as seen by the caller:
image

The missing payload recording in case of an error
image

Log from console or from log file.

Attach tyk configuration file

**Additional context**

tbuchaillot:
## Description

This PR adds a http.Response to the gateway `HandleError` func. It also copy all the content of the response, including headers to forward it as `RawResponse` in the analytic records.

## Related Issue

Closes #3070 

## Motivation and Context

Explained in #3070 

## How This Has Been Tested

Create an API with Auth token. 
Send a request with an invalid token to that API.
Check if the error response is showing on the dashboard logs. It should.

Example of dashboard analytic log.
image

## Types of changes

## Checklist

- [X] All new and existing tests passed.
  - [X] `go fmt -s`
  - [X] `go vet`


gateway/handler_error.go
```diff
@@ -128,23 +132,34 @@ func (e *ErrorHandler) HandleError(w http.ResponseWriter, r *http.Request, errMs
...
 			w.WriteHeader(errCode)
+			response.StatusCode = errCode
+
 			apiError := APIError{template.HTML(template.JSEscapeString(errMsg))}
-			tmpl.Execute(w, &apiError)
+			var log bytes.Buffer
+
+			rsp := io.MultiWriter(w, &log)
+			tmpl.Execute(rsp, &apiError)
+			response.Body = ioutil.NopCloser(bytes.NewBufferString(log.String()))
```
gernest, line 162: I think this will do 
```go
response.Body = ioutil.NopCloser(&log)
```
`bytes.Buffer` implements `io.Reader`

tbuchaillot: done! thanks

PR #3210, opened 2020-07-14
add last_schema_update field to graphql api definition

pvormste:
Introduce `last_schema_update` field to api definition.

This is an optional field that will be used for proxyOnly mode.

Related to https://github.com/TykTechnologies/product/issues/361

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -539,6 +539,8 @@ type GraphQLConfig struct {
 	ExecutionMode GraphQLExecutionMode `bson:"execution_mode" json:"execution_mode"`
 	// Schema is the GraphQL Schema exposed by the GraphQL API/Upstream/Engine.
 	Schema string `bson:"schema" json:"schema"`
+	// LastSchemaUpdate contains the date and time of the last triggered schema update to the upstream
+	LastSchemaUpdate *time.Time `bson:"last_schema_update" json:"last_schema_update,omitempty"`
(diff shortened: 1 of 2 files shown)


apidef/api_definitions.go
```diff
@@ -539,6 +539,8 @@ type GraphQLConfig struct {
 	ExecutionMode GraphQLExecutionMode `bson:"execution_mode" json:"execution_mode"`
 	// Schema is the GraphQL Schema exposed by the GraphQL API/Upstream/Engine.
 	Schema string `bson:"schema" json:"schema"`
+	// LastSchemaUpdate contains the date and time of the last triggered schema update to the upstream
+	LastSchemaUpdate *time.Time `bson:"last_schema_update" json:"last_schema_update,omitempty"`
```
spetrunin, line 543: maybe `SchemaUpdatedAt`?

pvormste: Why do you think `SchemaUpdatedAt` is a better choice than `LastSchemaUpdate`?

spetrunin: it is common to have fields named like `created_at` and `updated_at`

But as I see we use `LastUpdated` and `DateUpdated`

pvormste: I would say that none of them is a superior choice: `updated_at`, `last_modified`, `last_update` are all common terms you find in the wild so I would suggest to not "overengineer" this term here :-)

PR #3211, opened 2020-07-14
Adding a Dockerfile that lets you build from Source

sedkis:
Adds a new Dockerfile that lets you build Tyk from Source.
Final size: 145 Mb

**Purposes:**

1) This can replace the main package-based official image because it will replace the dependency on packages.  This will allow us to easily build ci/cd pipelines based off commits to master without needing to generate packages for every commit.

2) Allows developers to locally build Docker images based off their forks & modifications for any test purposes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,41 @@
+# Start from the latest golang base image
+FROM golang:1.13 as builder
+
+# Set the Current Working Directory inside the container
+WORKDIR /app
+
+# Copy go mod and sum files
+COPY go.mod go.sum ./
+
+# Download all dependencies. Dependencies will be cached if the go.mod and go.sum files are not changed
+RUN go mod download
+
+# Copy the source from the current directory to the Working Directory inside the container
+COPY . .
+
+# Build the Go app
+RUN CGO_ENABLED=0 go build -o tyk .
+
+######## Start a new stage from scratch #######
+FROM debian:buster-slim
+
+RUN apt-get update \
+ && apt-get upgrade -y \
+ && apt-get install -y --no-install-recommends ca-certificates \
+ && apt-get autoremove -y \
+ && rm -rf /root/.cache
+
+RUN mkdir -p /opt/tyk-gateway
+WORKDIR /opt/tyk-gateway
+
+# Copy the Pre-built binary file from the previous stage
+COPY --from=builder /app/tyk /opt/tyk-gateway/tyk
+
+## Copy everything else
+COPY templates /opt/tyk-gateway/templates
+COPY coprocess /opt/tyk-gateway/coprocess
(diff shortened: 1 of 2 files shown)


excieve:
@sedkis Could you please add motivation in the description on why is this needed in addition to the package-based official images? What's the intended use and whether/how would it be tied into CD?

sedkis:
> @sedkis Could you please add motivation in the description on why is this needed in addition to the package-based official images? What's the intended use and whether/how would it be tied into CD?

Done @excieve

excieve:
@sedkis This cannot currently replace package-based builds because it breaks the CD process. Binaries need to be build with specific conditions too in order to enable plugins (both Python and Go).

This can be used for dev builds, but further discussion is needed on which flows would it be integrated with before proceeding.

sedkis:
Maybe this belongs elsewhere as a dev resource if we don't want to bloat the repo?

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+COPY go.mod go.sum ./
+
+# Download all dependencies. Dependencies will be cached if the go.mod and go.sum files are not changed
+RUN go mod download
+
+# Copy the source from the current directory to the Working Directory inside the container
+COPY . .
+
+# Build the Go app
+RUN CGO_ENABLED=0 go build -o tyk .
```
excieve, line 17: This should be something like:
```suggestion
RUN CGO_ENABLED=1 go build -tags 'goplugin' -o tyk .
```

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+######## Start a new stage from scratch #######
+FROM debian:buster-slim
+
+RUN apt-get update \
+ && apt-get upgrade -y \
+ && apt-get install -y --no-install-recommends ca-certificates \
```
excieve, line 24: Much more stuff is needed to be able to use Python plugins. As I'm not sure which flows would this image support I'm not sure which things this needs.

Please take a look at this for reference:
https://github.com/TykTechnologies/tyk-gateway-docker/blob/bdd642e0261402565cd2f0f88ba1ddac7e9cbd25/Dockerfile#L9

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ && rm -rf /root/.cache
+
+RUN mkdir -p /opt/tyk-gateway
+WORKDIR /opt/tyk-gateway
+
+# Copy the Pre-built binary file from the previous stage
+COPY --from=builder /app/tyk /opt/tyk-gateway/tyk
+
+## Copy everything else
+COPY templates /opt/tyk-gateway/templates
```
excieve, line 35: There are a couple more directories that the package installs. Not sure if they're required or not for this use case.

gwotg:
@sedkis you found other resources suitable for your purpose? Wanna keep it or lego?

PR #3206, opened 2020-07-11
fix: adding x5c when not available from Idp

bmonteiro:
## Description
This fix will add the public key x5c when Idp's like Okta just expose 'n'

## Related Issue
Will close this issue #2303 

## Motivation and Context
Will close this issue #2303 

## How This Has Been Tested
Using keys from Okta:
 
https://dev-519462.okta.com/oauth2/default/v1/keys

```
{
  "keys": [
    {
      "kty": "RSA",
      "alg": "RS256",
      "kid": "xA8t81QFWX9dFzFvn3Y6EGfXyuQ8dSMIM1wgysNvMX8",
      "use": "sig",
      "e": "AQAB",
      "n": "z-iP7OCP5lqSB51IDO6RzYlkmQaf9E1qlK5pZ370G44mifFVKnf7XfEuXHvge56vQRPyPUaFU2YYxX23dcdppl53PEZPHkPw9UcKInZjN9Uqq9JJ_CeW3qWtWm0hmQm0lOStv76sYpr-PWFRhIxwL55-2Ydy5m0254RBykUq8WGEhPysFHBIjJZscG8fIFh51c_qb0Q7ucAFXNrkQDCH-jlsnY-8CiN_Y-HmwqipTts6F_M6Nr8GHBMcs1d9VCrE3OFGwuRMzPpwUoKY0FFhmjSjnS9cwerd72trxZqzVKfNSE9Rm-_hBctLQuNHk9kgaA9at0MNgBbSzwolodkvKQ"
    },
    {
      "kty": "RSA",
      "alg": "RS256",
      "kid": "PfuTNFLkNeaFEv-huEzscAcWlXtxGplHeXKfOnS4G6I",
      "use": "sig",
      "e": "AQAB",
[...]

PR #3212, opened 2020-07-15
[TT-645] Use standard in handling of JWK

### LINKED ISSUE #2303: Validate JWT using PEM rather than x509 certificate

Feature

Receive the following error when setting up a Tyk JWT authorizer with Okta as an authorization server:

```
curl 'https://$tyk-api/$path-with-jwt-authorizer' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.14; rv:67.0) Gecko/20100101 Firefox/67.0' -H 'Accept: */*' -H 'Accept-Language: en-US,en;q=0.5' --compressed -H 'Referer: http://localhost:3000/protected' -H 'Authorization: Bearer $okta-access-token' -H 'Origin: http://localhost:3000' -H 'DNT: 1' -H 'Connection: keep-alive'
```

```json
{
    "error": "Key not authorized:no certificates in JWK"
}
```

This is due to Tyk expecting an `x5c` parameter: https://github.com/TykTechnologies/tyk/blob/f69055a931dc50abf3ef884fa9cb7277ae667e4c/gateway/mw_jwt.go#L94-L104 and not using `n` and `e` to generate a PEM if they are present.

Tyk verifies the JWT using the PEM modulo and exponent in Okta's JWT keyset.

Okta do not expose their X509 certificate in their JWKS endpoint anymore (https://stackoverflow.com/questions/38727528/is-okta-oidc-documentation-outdated/38947218#38947218)).

Example keyset response parameters (values omitted):

```
{
    "keys": [
        {
            "alg": "RS256",
            "e": ...,
            "kid": ...,
            "kty": "RSA",
            "n": ...,
            "use": "sig"
        }
    ]
}
```

All versions (`<=v2.8.0` at the time of writing).

gernest:
## Description

This adds support for using JWK with missing certificate field. Currently we
were expecting jwk with the following format.

```json
{
  "keys": [
    {
      "alg": "RS256",
      "kty": "RSA",
      "use": "sig",
      "x5c": [
        "Ci0tLS0tQkVHSU4gUFVCTElDIEtFWS0tLS0tCk1JSUJJakFOQmdrcWhraUc5dzBCQVFFRkFBT0NBUThBTUlJQkNnS0NBUUVBeXFaNHJ3S0Y4cUNFeFM3a3BZNGMKbkphLzM3Rk1rSk5rYWxaM091c2xMQjBvUkw4VDRjOTRrZEY0YWVOelNGa1NlMm45OUlCSTZTc2w3OXZiZk1aYgordDA2TDBROTRrKy9QMzd4NysvUkpaaWZmNHkxVkdqcm5ybk1JMml1OWw0aUJCUll6Tm1HNmVibHJvRU1NV2xnCms1dHlzSGd4QjU5Q1NOSWNEOWdxazFoeDRuL0ZnT212S3NmUWdXSE5sUFNEVFJjV0dXR2hCMi9YZ05WWUcycE8KbFF4QVBxTGhCSGVxR1RYQmJQZkdGOWNIeml4cHNQcjZHdGJ6UHdoc1EvOGJQeG9KN2hkZm4rcnp6dGtzM2Q2KwpIV1VSY3lOVExSZTBtalhqamVlOVo2K2daK0grZlM0cG5QOXRxVDdJZ1U2ZVBVV1Rwam9pUHRMZXhnc0FhL2N0CmpRSURBUUFCCi0tLS0tRU5EIFBVQkxJQyBLRVktLS0tLQo="
      ],
      "n": "xofiG8gsnv9-I_g-5OWTLhaZtgAGq1QEsBCPK9lmLqhuonHe8lT-nK1DM49f6J9QgaOjZ3DB50QkhBysnIFNcXFyzaYIPMoccvuHLPgdBawX4WYKm5gficD0WB0XnTt4sqTI5usFpuop9vvW44BwVGhRqMT7c11gA8TSWMBxDI4A5ARc4MuQtfm64oN-JQodSztArwb9wcmH8WrBvSUkR4pyi9MT8W27gqJ2e2Xn8jgGnswNQWOyCTN84PawOYaN-2ORHeIea1g-URln1bofcHN73vZCIrVbE6iA2D7Ybh22AVrCfunekEDEe2GZfLZLejiZiBWG7enJhcrQIzAQGw",
      "e": "AQAB",
      "kid": "12345",
      "x5t": "12345"
    }
  ]
}
```

Where we used first entry on `x5c` which can either be a valid rsa key or  a `x509` certificate to derive the public key needed for verification of jwt tokens.

However services like okta they don't provide `x5c` instead you get the key from the `n` and `e` field.JWK response from okta looks like this.

```json
{
  "keys": [
    {
      "use": "sig",
      "kty": "RSA",
      "kid": "12345",
      "alg": "RS256",
      "n": "yqZ4rwKF8qCExS7kpY4cnJa_37FMkJNkalZ3OuslLB0oRL8T4c94kdF4aeNzSFkSe2n99IBI6Ssl79vbfMZb-t06L0Q94k-_P37x7-_RJZiff4y1VGjrnrnMI2iu9l4iBBRYzNmG6eblroEMMWlgk5tysHgxB59CSNIcD9gqk1hx4n_FgOmvKsfQgWHNlPSDTRcWGWGhB2_XgNVYG2pOlQxAPqLhBHeqGTXBbPfGF9cHzixpsPr6GtbzPwhsQ_8bPxoJ7hdfn-rzztks3d6-HWURcyNTLRe0mjXjjee9Z6-gZ-H-fS4pnP9tqT7IgU6ePUWTpjoiPtLexgsAa_ctjQ",
      "e": "AQAB"
    }
  ]
}
```

In this PR we are using the library `github.com/square/go-jose` for parsing JWK with missing `s5c`. To maintain backward compatibility we parse the JWK twice with our current implementation and with `go-jose` because tyk doesn't respect `x5c` to be a valid certificate , tyk works with rsa keys inside `x5c` which is not standard so we  leave certificate handling routines to preserve backward compatibility. For the case of missing `x5c` we use `go-jose` parsed keys.
## Related Issue
Based on #3206
fixes #2303

## How This Has Been Tested
`go test ./gateway`

- [ ] TODO with okta?

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/testutil.go
+++ b/gateway/testutil.go
@@ -468,7 +468,7 @@ const (
 	testGraphQLDataSource = TestHttpAny + "/graphql-data-source"
 	testRESTDataSource    = TestHttpAny + "/rest-data-source"
 	testHttpJWK           = TestHttpAny + "/jwk.json"
-	testHttpJWKDER        = TestHttpAny + "/jwk-der.json"
+	testHttpJWKLegacy     = TestHttpAny + "/jwk-legacy.json"
 	testHttpBundles       = TestHttpAny + "/bundles/"
 	testReloadGroup       = TestHttpAny + "/groupReload"
 
@@ -549,9 +549,10 @@ func testHttpHandler() *mux.Router {
 	r.HandleFunc("/jwk.json", func(w http.ResponseWriter, r *http.Request) {
 		io.WriteString(w, jwkTestJson)
 	})
-	r.HandleFunc("/jwk-der.json", func(w http.ResponseWriter, r *http.Request) {
-		io.WriteString(w, jwkTestDERJson)
+	r.HandleFunc("/jwk-legacy.json", func(w http.ResponseWriter, r *http.Request) {
+		io.WriteString(w, jwkTestJsonLegacy)
 	})
+
 	r.HandleFunc("/compressed", func(w http.ResponseWriter, r *http.Request) {
 		response := "This is a compressed response"
 		w.Header().Set("Content-Encoding", "gzip")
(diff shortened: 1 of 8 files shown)


gernest:
Gentle ping @letzya

bmonteiro:
tested looks good

bmonteiro:
Just one additional comment, if we use e and n why keeping x5c :) let's remove it because it is optimal from rfc and n is enough. Makes sense ? @gernest @letzya

gernest:
@bmonteiro we have to maintain backward compatibility.

bmonteiro:
There is no need "n" is a must and x5c optional. So it will work

letzya:
@bmonteiro don't think it's not going to work since we need to keep backward compatibility. If for some reason a user only used x5c, his api now will stop working.

gernest:
> There is no need "n" is a must and x5c optional. So it will work. It will be impossible having an IDP not sending "n" ;)

Prior to this change, tyk actually required `x5c` to work so `x5c` was  a must, that's why this bug manifested. Now we need to keep that behaviour and handle  the edge case of when `x5c` is not there.

bmonteiro:
Current implementation when you consider x5c a must is actually not correct based on the RFC ;)

bmonteiro:
> @bmonteiro don't think it's not going to work since we need to keep backward compatibility. If for some reason a user only used x5c, his api now will stop working. 

The user will also have from the IdP the "n", right ?

gernest:
> Current implementation when you consider x5c a must is actually not correct based on the RFC ;)
I think thats what this PR is trying to address without breaking users who were relying on `x5c`.

bmonteiro:
> > Current implementation when you consider x5c a must is actually not correct based on the RFC ;)
> 
> 
> 
> I think thats what this PR is trying to address without breaking users who were relying on `x5c`.

Ok guys, not to drag too much the chat. Long story short the IdP will pass the public key via "N" and "x5c" is optional and some do pass it others no. So if you remove x5c completely even old customers assuming they are using x5c will start using silently the "N"

bmonteiro:
> @bmonteiro I'm not sure. potentially if you scrap a JWKs with just x5c and alg and kid it should work, so clients that don't follow the standard as-is are still **appreciated** and **close to our hearts** and should still be able to upgrade :)

The fundamental question is: There are out-there IdPs issuing or exposing non standard JWKs ?  What I mean is: if all is based on RFCs such scenario will never happen. Do you have any Idp doing that? Not passing 'N' that is the public key?

buger:
@gernest let's make it use x5c only if N is not presented.

gateway/testutil.go
```diff
@@ -463,16 +467,30 @@ func restDataSourceHandler(w http.ResponseWriter, r *http.Request) {
...
+            "kty": "RSA",
+            "kid": "12345",
+            "alg": "RS256",
+            "n": "yqZ4rwKF8qCExS7kpY4cnJa_37FMkJNkalZ3OuslLB0oRL8T4c94kdF4aeNzSFkSe2n99IBI6Ssl79vbfMZb-t06L0Q94k-_P37x7-_RJZiff4y1VGjrnrnMI2iu9l4iBBRYzNmG6eblroEMMWlgk5tysHgxB59CSNIcD9gqk1hx4n_FgOmvKsfQgWHNlPSDTRcWGWGhB2_XgNVYG2pOlQxAPqLhBHeqGTXBbPfGF9cHzixpsPr6GtbzPwhsQ_8bPxoJ7hdfn-rzztks3d6-HWURcyNTLRe0mjXjjee9Z6-gZ-H-fS4pnP9tqT7IgU6ePUWTpjoiPtLexgsAa_ctjQ",
+            "e": "AQAB"
+        }
+    ]
+}`
+
+const jwkTestJsonLegacy = `{
```
bmonteiro, line 482: I don't believe this test is valid because N is always present. I have never seen an IDP not sending N.

buger:
@bmonteiro re-read the standard one more time.

`n` and `e` presented only for RSA keys. If you have ECDSA it will be another set of params. 
So having `x5c` as a fallback still makes sense. 

Additionally, even if both x5c and `n`/`e` are presented, x5c[0] cert public key match key from `n`,`e`.

bmonteiro:
> ECDSA

@buger sure, I can read but it is also available for you here a good written example: https://tools.ietf.org/html/draft-jones-json-web-key-00

Indeed ECDSA uses curve and x and y (4.2.1.  JWK Key Object Members for ECDSA Keys). 

Having a hard time to understand and see where the x5c is but I'm sure it is me not getting it.

RSA also (4.2.2.  JWK Key Object Members for RSA Keys)

They explain it pretty nice with MUST be present fields but still I cannot see x5c there.

Long story short, no need to continue this because obviously you guys got it right.

letzya:
@gernest is this still on our plan to be merged?

gernest:
@letzya I don't think so.  It's not on jira. You can help create a jira ticket and link this PR to it, maybe it can be squeezed into any of the next release. Its pretty much done by now.

I will try to look at this on monday.

matiasinsaurralde (approved):

edwardsmatt:
Hi there,
FWIW I have successfully tried this branch with AWS Cognito User Pool well-known keys.

Is there anything I can provide specifically to support this PR?

If not, are there any indications of when this may be merged/released?

Thanks very much

matiasinsaurralde:
Conflicts solved but we still need additional reviews for merging.

edwardsmatt (approved):

matiasinsaurralde:
Merged into `release-3-lts` (4bcf6dccdc7fbbc25640f7792758e1cf99ffc8da) and `release-3.1` (d24b90eb7ae72cfa626d8f6b0f1f2e147aec5afd).

tbuchaillot (approved):

Issue #3003, opened 2020-04-08
JWT Scope Claim policy not being applied 


chris-fewtrell:
**Branch/Environment/Version**
- Branch/Version: Release
- Environment: OnPrem

**Describe the bug**
When using "Scope claim" feature, to map a scope value to a policy, the expected policy is not applied.

The name of the claim field used is "roles", and the value that should map to a policy is "automation", as shown below. The Policy name is "scope-claim-policy".

This is an extract of the body JSON from our JWT token. 
 
```
{
  "application": "test",
  "roles": "automation",
  "env": "dev"
}
```
 
This has been configured via the Dashboard as shown on the included screenshot, however the requested Policy does not seem to get applied as expected. The "scope-claim-policy" Policy shown is a policy that allows access to the rest endpoint.
 
API definition has been included.

The following error is returned when calling the API with a valid JWT token where "roles" equals "automation".
 
```
{
    "error": "key not authorized: no matching policy found"
}
```

**Reproduction steps**

1. Create a new API with JWT Authorization mode, using scope claim (scope name = "roles", claim = "automation")
2. Send request with JWT token that has matching roles value of 'automation'.
3. Correct policy applies and 200 Success
4. Submit request with JWT token that has non-matching roles value.
5. 403 "error": "key not authorized: no matching policy found" returned.
6. Try submitting original request with JWT token that has matching roles value.
7. 403 "error": "key not authorized: no matching policy found" still returned. Same error for all subsequent calls no matter what roles value is provided.

**Screenshots/Video**
Scope Claim setup:
![image](https://user-images.githubusercontent.com/59964991/78798776-49a5db00-79b1-11ea-904b-7652378708b0.png)

**Logs (debug mode or log file):**
```
time="Apr 07 12:50:51" level=error msg="Could not identify a policy to apply to this token from field" api_id=1dce6c90893942656e2c5d6aa7f16d04 api_name=scope-claim-test mw=JWTMiddleware org_id=5e87036152190c0021426c7b origin=x.x.x.x path="/scope-claim-test/"
time="Apr 07 12:50:51" level=error msg="No policies for the found session. Failing Request." api_id=1dce6c90893942656e2c5d6aa7f16d04 api_name=scope-claim-test mw=JWTMiddleware org_id=5e87036152190c0021426c7b origin=x.x.x.x path="/scope-claim-test/"
```

**Additional context**
[...]

PR #3219, opened 2020-07-20
Return error If scope is enabled but no match in jwt

furkansenharputlu:
The behavior is like that:
If `Use Scope Claim` is enabled, but it has 0 matches, it will fail. However, if JWT doesn't contain a scope claim, it will use default policy.

In terms of fallback to default policy, It behaves exactly same with `Policy Field Name`. I made them consistent.

Fixes https://github.com/TykTechnologies/tyk/issues/3003


buger:
Looks good!

Issue #3109, opened 2020-05-28
Token update by hash generates new hash entry in redis


ilijabojanovic:
**Branch/Environment/Version**
- Branch/Version: Master/Stable
- Environment: All

**Describe the bug**
When user updates token by hash, gateway will create new hash entry in redis database.

**Reproduction steps**
1. Create new token
2. Search that token by hash
3. On keys edit page just update token

**Actual behavior**
Before token update in redis we will have hash like this `api-3f2736714c0076a2`, when we update token in redis we will have two entries:
```
{
    "data": {
        "keys": [
            "3f2736714c0076a2",
"eyJvcmciOiI1ZWNlZGFlMjQ1YjY0MjBhMGE1ZDQxNTgiLCJpZCI6IjhmNDcxNWMwY2Q5YzMwN2MiLCJoIjoibXVybXVyNjQifQ==",
        ]
    },
```
**Expected behavior**
Update token by hash should not create new token entry in redis.

**Screenshots/Video**
If applicable, add screenshots or video to help explain your problem.

**Logs (debug mode or log file):**
Log from console or from log file.

**Configuration (tyk config file):**
Attach tyk configuration file

**Additional context**

ilijabojanovic:
Verified

PR #3229, opened 2020-07-22
Fix token update generate new hash entry

sredxny:
## Description
On update keys, we should not call twice `doAddOrUpdate` as this create duplicates for the same key in redis, instead, we should check if we are dealing with a custom key or not, and therefore perform the proper action.

## Related Issue
https://github.com/TykTechnologies/tyk/issues/3109

## Motivation and Context
Give solution to https://github.com/TykTechnologies/tyk/issues/3109

## How This Has Been Tested
- Run GW and Dashboard with `"hash_keys": true`
- Create a custom key via gw api
- Update custom key...and check that the value was updated and an additional registry was not created in redis
- Create key via dashboard
- Update key via dashboard, check that the value was updated but an additional registry in redis was not created for the same key
- Do the same but now with `"hash_keys": false`

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`

### DIFF
--- a/gateway/api.go
+++ b/gateway/api.go
@@ -411,13 +411,27 @@ func handleAddOrUpdate(keyName string, r *http.Request, isHashed bool) (interfac
 			return apiError("Failed to create key, ensure security settings are correct."), http.StatusInternalServerError
 		}
 	} else {
-		// update legacy key format one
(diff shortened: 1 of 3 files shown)


buger:
Can we add test for it?

sredxny:
@buger added test coverage

PR #3238, opened 2020-07-26
Discarding the master-slave jargon

bopopescu:
For diversity reasons, it would be nice to try to avoid 'master' and 'slave' terminology in this repository which can be associated to slavery. The master-slave terminology could be problematic for people in several countries which has the history of slavery like Romania, USA and many others. Thank you for considering the proposal. Let me know if any changes in the PR are needed, I would be happy to implement them.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -82,7 +82,7 @@ type StorageOptionsConf struct {
 	Port                  int               `json:"port"`
 	Hosts                 map[string]string `json:"hosts"` // Deprecated: Addrs instead.
 	Addrs                 []string          `json:"addrs"`
-	MasterName            string            `json:"master_name"`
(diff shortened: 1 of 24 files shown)


buger:
Hi! We already started this work on the documentation level, and in future releases planning to update product as well https://github.com/TykTechnologies/tyk-docs/pull/1312

As for the code, this PR should come with backward compatibility layer, so in some period of time, both config variables will be possible to use. 

Thank you!

furkansenharputlu:
@bopopescu Thanks for your contribution! Our team is currently working on this issue. However, we need to be careful while doing these changes as it may break existing users' setups. Also, we don't use `vendor` directories anymore. So, for that we will close this PR but we will handle in a proper way. Thanks a lot again!

Issue #3235, opened 2020-07-24
Composed Graphql api's does not work on Slaves/hybrid


ilijabojanovic:
**Branch/Environment/Version**
- Branch/Version: Stable/Master
- Environment: Hybrid/Mdcb

**Describe the bug**

Data is not returned for composed apis only on slave/hybrid gateway.

**Reproduction steps**
1. Start mdcb or hybrid gateway
2. Create composed graphql api
3. Try to use that api over playground, we will see data
4. Try same queries on slaved gateway 

**Actual behavior**
On slaved gateway we always have this:
```
{
    "data": null
}
```

**Expected behavior**
We should get requested data

**Screenshots/Video**
If applicable, add screenshots or video to help explain your problem.

**Logs (debug mode or log file):**
Log from console or from log file.

**Configuration (tyk config file):**
Attach tyk configuration file

**Additional context**

ilijabojanovic:
@furkansenharputlu @lghiur @jensneuse 

On latest release-3-lts branch, problem is still visible. When we send traffic to slaved/hybrid gateway we expected data structure in response, but data is not visible.

Gateway error:
```
[Aug 05 13:59:31]  INFO main: Initialised API Definitions
[Aug 05 13:59:31]  INFO main: API reload complete
[Aug 05 13:59:37] ERROR HttpJsonDataSource.Resolve.Client.Do error=Get https:: http: no Host in request URL
[Aug 05 13:59:41]  WARN pub-sub: Reconnecting storage: Redis is either down or ws not configured
[Aug 05 13:59:54] ERROR HttpJsonDataSource.Resolve.Client.Do error=Get https:: http: no Host in request URL
[Aug 05 14:01:13] ERROR HttpJsonDataSource.Resolve.Client.Do error=Get https:: http: no Host in request URL
```

Response:

```
{
    "data": {
        "data": null,
        "ip": null
    }
}
}
```

ilijabojanovic:
Verified on Ara and local mdcb

PR #3254, opened 2020-08-04
Vendor latest library changes

furkansenharputlu:
Related to #3235


spetrunin:
@leo we have to revendor library, as vendoring by particular commit is not using the whole branch

Furkan did an update for master branch of the library to include all changes

PR #3266, opened 2020-08-11
Update env for auto-build images

alephnull:
Update terraform env name as we have moved to a new subaccount.

This PR is to be merged into master only.


integration/image/Dockerfile
```diff
@@ -1,16 +1,35 @@
...
 
+RUN apt-get update \
+ && apt-get upgrade -y \
+ && apt-get install -y --no-install-recommends \
+        wget jq curl ca-certificates apt-transport-https gnupg unzip
+
+RUN apt-get install -y --no-install-recommends \
+        build-essential \
+        python3-setuptools \
+        libpython3.7 \
```
excieve, line 15: If Python is for the plugins here, it might also need `python3.7-dev` package — otherwise libpython won't be correctly detected by the coprocess loader (see https://github.com/TykTechnologies/tyk-gateway-docker/blob/f35efea072e3e9a613124537bde9b82c77969d82/Dockerfile#L18 and https://github.com/TykTechnologies/tyk-gateway-docker/issues/35 for context).

integration/image/Dockerfile
```diff
@@ -1,16 +1,33 @@
...
+        python3.7-dev \
+        python3-pip \
+        && pip3 install protobuf grpcio==1.24.0 \
+        && apt-get purge -y build-essential \
+        && rm -rf /root/.cache
+RUN apt-get autoremove -y 
+
 ADD tyk.tar.gz /opt/tyk
```
vesko-tyk, line 23: Why is this ADD operation needed? If I try to build the image I'm getting an error: 
`Step 5/9 : ADD tyk.tar.gz /opt/tyk`
`ADD failed: stat /var/lib/docker/tmp/docker-builder541896976/tyk.tar.gz: no such file or directory`

alephnull: The tarfile is built by https://github.com/TykTechnologies/tyk/blob/master/bin/integration_build.sh and the image is built and pushed in the int-image workflow. This image is meant to be built by automation.

PR #3273, opened 2020-08-14
provide default CORS config in dummy API definition

pvormste:
related to https://github.com/TykTechnologies/tyk-analytics/issues/2056


spetrunin:
@pvormste I'm ok with changes. Could you check that tests pass locally?
seems CI fail probably unrelated to your changes

PR #3283, opened 2020-08-20
[Fix/TT-225] Not encode and reset RawQuery after JSVM if no change on query params

furkansenharputlu:
```
url, _ := url.Parse("http://example.com/get?c=1&a=2&b=3&e=4&d=5")
queries := url.Query()
url.RawQuery = queries.Encode()
fmt.Println(url) // http://example.com/get?a=2&b=3&c=1&d=5&e=4
```

As you see in the above example, `Encode()` function sorts by key. The reason why it does this becase `url.Values` is a `map` and setting order is not saved when you iterate. `Encode` function sorts to prevent random results.

The fix is that if there is no change on query params after JSVM, it doesn't make query parsing and encoding and continue with the existing `RawQuery` value.

Fixes https://tyktech.atlassian.net/browse/TT-225

### DIFF
--- a/gateway/coprocess.go
+++ b/gateway/coprocess.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"net/url"
+	"reflect"
 	"strings"
 	"time"
 	"unicode/utf8"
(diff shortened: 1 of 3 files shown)


furkansenharputlu:
Merged manually to:
- `release-3`
- `release-3-lts`
- `release-3.2`

PR #3287, opened 2020-08-24
[TT-307] Accept combined certificates as public certificates

furkansenharputlu:
https://tyktech.atlassian.net/browse/TT-307


buger:
Looks good!

PR #3289, opened 2020-08-24
Implements jsvm http request timeout configuration

ghost:
Implements jsvm http request timeout configuration

## Description
Adds an additional parameter that allows to configura a timeout for HTTP calls from inside the JS Middleware inside JSVM

## Motivation and Context
The JSVM Timeout is great in order to set a max execution time. However, if a HTTP call is used inside the Middleware there is no way to define a timeout. If a JS Middleware is used for authentication it will be skipped if the time limit is exeded.

## How This Has Been Tested
Create a JS Middleware that uses a `TykMakeHttpRequest` and take the service that is going to be called offline. Set the `jsvm_timeout` to 10 and the `jsvm_http_request_timeout` to 5. The HTTP call will timeout after 5 seconds and the error can be handled in the Middleware.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/cli/linter/schema.json
+++ b/cli/linter/schema.json
@@ -362,6 +362,9 @@
     "jsvm_timeout": {
       "type": "integer"
     },
+    "jsvm_http_request_timeout": {
+      "type": "integer"
+    },
     "enable_non_transactional_rate_limiter": {
       "type": "boolean"
     },
--- a/config/config.go
+++ b/config/config.go
@@ -830,8 +830,11 @@ type Config struct {
 	// Set to true if you are using JSVM custom middleware or virtual endpoints.
 	EnableJSVM bool `json:"enable_jsvm"`
 
-	// Set the execution timeout for JSVM plugins and virtal endpoints
+	// Set the execution timeout for JSVM plugins and virtual endpoints
 	JSVMTimeout int `json:"jsvm_timeout"`
+
+	// Set the execution timeout for HTTP requests inside JSVM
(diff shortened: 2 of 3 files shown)


buger:
Hi! Thank you for contribution! 

I wonder if we really need a new option, of we can rely on existing `proxy_default_timeout` and "Timeout" plugins. 
See how they used inside reverse proxy here https://github.com/TykTechnologies/tyk/blob/master/gateway/reverse_proxy.go#L472

anx-hnezbeda:
I'm not so sure `proxy_default_timeout` will work this way.

Tyk Auth JS Middleware

We are targeting an external service for authentication/authorization that is called from the JS Middleware.
Based on the response of the service we decide if the request should continue or not.

The auth service is configurable in the API and processed by the middleware in our case. If the auth service takes too long to respond and it exceeds the JS middleware timeout it will allow the request to pass trough as the middleware will be ignored. Adding an additional timeout for the request inside the middleware (via `TykMakeHttpRequest`) will allow to control such behavior.

letzya:
@matiasinsaurralde @anx-hnezbeda we need to check backwards compatibility if we add a default value.

nezhar:
I updated the PR according to the curent `master`

lghiur:
@nezhar it looked that the owner of the issue is not on GH anymore and the issue was quite old. Will reopen and also create an internal issue. Thank you!

PR #3290, opened 2020-08-24
[TT-221] Move query depth logic to SessionLimiter

furkansenharputlu:
When per API limit is applied, session global limit values should be ignored. However, it is not working like that for query depth limit. This PR moves the depth limit check inside `ForwardMessage` function to be able to benefit this piece: https://github.com/TykTechnologies/tyk/blob/fafaba25989f2569f4aa02482a338b3210747767/gateway/session_manager.go#L189-L197

Fixes https://tyktech.atlassian.net/browse/TT-221


buger:
Love the refactoring. Looks solid to me 👌

buger (approved):

gernest (approved):
LGTM

PR #3302, opened 2020-09-08
Policy REST API

joshblakeley:
## Description

This PR creates a way to manage policies in Tyk CE using a REST API instead of the single file approach currently. It makes policies act in very much the same way as API when using CE. A directory is specified in the Tyk config in the new `policies.policy_path` field. Policies are then stored in that folder with their ID as a filename i.e. `default-policy.json`. CRUD operations can be carried out by calling `{GATEWAYHOST}:{PORT}/tyk/policies` endpoint with the policy ID passed in the last part of Path.

## Related Issue

Fixes https://github.com/TykTechnologies/tyk/issues/2379

## Motivation and Context

Means simpler way of managing policies - also unblocks development of Kubernetes Operator for the OSS Gateway.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


gernest (changes requested):
Looks good, I left a few comments. After they are addressed I will take another look.

joshblakeley:
Thanks for the review @gernest. Have made suggested changes.

gateway/server.go
```diff
@@ -138,6 +138,13 @@ func setTestMode(v bool) {
 	runningTestsMu.Unlock()
 }
 
+func getPolicy(polID string) user.Policy {
+	apisMu.RLock()
+	pol := policiesByID[polID]
+	apisMu.RUnlock()
```
gernest, line 144:
```suggestion
	policiesMu.RLock()
	pol := policiesByID[polID]
	policiesMu.RUnlock()
```

gernest (changes requested):
I left another comment. We also need tests to cover all the cases.

gateway/server.go
joshblakeley: I was worried you would say that :D

buger (approved):
Fix Linter and CI in general and looks solid 👍🏻

config/config.go
```diff
@@ -85,7 +85,8 @@ type PoliciesConfig struct {
 	// If you set this value to `true`, then the id parameter in a stored policy (or imported policy using the Dashboard API), will be used instead of the internal ID.
 	//
 	// This option should only be used when moving an installation to a new database.
-	AllowExplicitPolicyID bool `json:"allow_explicit_policy_id"`
+	AllowExplicitPolicyID bool   `json:"allow_explicit_policy_id"`
+	PolicyPath            string `json:"policy_path"`
```
buger, line 89: Do not forget adding a comment to all fields, since it will be used to auto-generate docs.

buger (approved):
Add comment for `PolicyPath` option, otherwise looks good!

config/config.go
vaske: Oh didn't know that, adding it

gernest (approved):

vaske:
merged manually to release-4-lts

PR #3321, opened 2020-09-21
TT-496 Bugfix/uptime checks nodes

tbuchaillot:
## Description

Added 2 configurations options for uptime tests.
`disable_management_poller` - bool - Disables host checker manager in that gw.
`poller_group` - string - Adds a postfix to the Poller Cache key. In that way, we can have different pollers for each group.

An example of a config with `poller_group`:
```
 "uptime_tests": {
        "disable": false,
        "poller_group":"external",
        "config": {
            "failure_trigger_sample_size": 1,
            "time_wait": 2,
            "checker_pool_size": 50,
            "enable_uptime_analytics": true
        }
    },
```

An example of a config with `disable_management_poll`:
```
 "uptime_tests": {
        "disable_management_polle":true
    },
```
## Related Issue

https://tyktech.atlassian.net/browse/TT-496

## Motivation and Context

On sharded gateway environments, each shard / node should have their own Poller / Host checker to do the uptimetest of their APIs. For more context, please check https://tyktech.atlassian.net/browse/TT-496

## How This Has Been Tested

Tested with:
2 GWs: poller_group:"a" , db_app_conf_options.node_is_segmented:true and db_app_conf_options.tags:["test"]
1 GWs: poller_group:"b" , db_app_conf_options.node_is_segmented:true and db_app_conf_options.tags:["test-2"]
1 GW : "disable_management_polle":true
1 Pump running with uptime tests configured.

1 API with tag "test" and uptime tests on.
1 API with tag "test-2" and uptime tests on.

In Redis we're going to have 2 Poller: 
image

and in Mongo we're going to have the uptime records for boths APIs.
image

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`

### DIFF
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -169,7 +169,7 @@ func setupGlobals(ctx context.Context) {
 		mainLog.Fatal("Analytics requires Redis Storage backend, please enable Redis in the tyk.conf file.")
 	}
 
-	if config.Global().UptimeTests.DisableManagementPoller && config.Global().ManagementNode {
+	if !config.Global().UptimeTests.DisableManagementPoller && config.Global().ManagementNode {
 		mainLog.Warn("Running Uptime checks in a management node.")
 	}
 	// Initialise our Host Checker if it's enabled to be a management poller
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -169,9 +169,11 @@ func setupGlobals(ctx context.Context) {
 		mainLog.Fatal("Analytics requires Redis Storage backend, please enable Redis in the tyk.conf file.")
 	}
 
-	// Initialise our Host Checker
-	healthCheckStore := storage.RedisCluster{KeyPrefix: "host-checker:"}
-	InitHostCheckManager(&healthCheckStore)
+	// Initialise our Host Checker if it's enabled to be a management poller
+	if !config.Global().UptimeTests.DisableManagementPoller {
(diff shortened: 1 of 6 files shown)


matiasinsaurralde (approved):
LGTM

gateway/server.go
```diff
@@ -169,9 +169,14 @@ func setupGlobals(ctx context.Context) {
 		mainLog.Fatal("Analytics requires Redis Storage backend, please enable Redis in the tyk.conf file.")
 	}
 
-	// Initialise our Host Checker
-	healthCheckStore := storage.RedisCluster{KeyPrefix: "host-checker:"}
-	InitHostCheckManager(&healthCheckStore)
+	if config.Global().UptimeTests.DisableManagementPoller && config.Global().ManagementNode {
```
joshblakeley, line 172: This seems a bit iffy.

The log here should state the opposite I think because we are saying that disabling the management poller is true.

Also could this cause panics down the line if the store is not initialised but uptime checks are still enabled somewhere?

joshblakeley (changes requested):

gateway/server.go
tbuchaillot: @joshblakeley updated the warning condition for managementNode = true and DisableManagementPoller=false.

I also tested the config with uptime_tests.disable_management_poller = true and uptime_tests.disable = false and there are no problems with panics.

joshblakeley (approved):
Thanks @tbuchaillot this looks good to me.

PR #3350, opened 2020-10-06
[TT-716] Handle HostDown check when host checker manager isn't initialized

matiasinsaurralde:
Workaround for TT-716, using the same check that's performed by `HostCheckerManager.AmIPolling` but excluding any additional behavior.
From the logic that's implemented in `server.go` we don't expect the host checker to be initialized when `disable_management_poller` is set to `true`.
Disabling `check_host_against_uptime_tests` could be a config-level workaround for the issue as well.

### DIFF
--- a/gateway/host_checker_manager_test.go
+++ b/gateway/host_checker_manager_test.go
@@ -33,8 +33,8 @@ func TestHostCheckerManagerInit(t *testing.T) {
 func TestAmIPolling(t *testing.T) {
 	hc := HostCheckerManager{}
 
-	pooling := hc.AmIPolling()
-	if pooling {
+	polling := hc.AmIPolling()
+	if polling {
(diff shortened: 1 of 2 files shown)


sredxny (approved):
LGTM

buger:
Can we also rename the param to fix the typo?

buger:
Also, it will be required to cover this functionality with tests.

matiasinsaurralde:
> Can we also rename the param to fix the typo?

Is it the `disable_management_poller` parameter?

tbuchaillot (approved):

buger:
will merge  it now, lets do tests as separate PR

buger:
Merged it manually to release-2.9 branch (bot is not working for some reason). But had issues with merging to release-3 branch, pls do it by yourself (and ensure that Jira has a proper tags assigned)

PR #3373, opened 2020-10-29
TT-653 on delete key by username then loop over all apis to update properly the key

sredxny:
## Description
When we attempt to remove a key and we do not receive the id of the api, then in some cases (such keys by username) we need the orgId to perform the deletion of the key. As we do not receive that information from sink, then we must loop over apis and use the orgId for each api to attempt the removal of the key, otherwise the key will not be updated in slaved nodes

## Related Issue
https://tyktech.atlassian.net/browse/TT-653

## Motivation and Context
give solution to https://tyktech.atlassian.net/browse/TT-653

## How This Has Been Tested
- Create basic API and key
- Search by username and update (check changes are being replicated in slaved nodes)
- Search by hash and update (to check that nothing is broken so changes are replicated)

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/api.go
+++ b/gateway/api.go
@@ -618,11 +618,18 @@ func handleDeleteKey(keyName, apiID string, resetQuota bool) (interface{}, int) 
 	if apiID == "-1" {
 		// Go through ALL managed API's and delete the key
 		apisMu.RLock()
-		removed := GlobalSessionManager.RemoveSession(orgID, keyName, false)
-		GlobalSessionManager.ResetQuota(
-			keyName,
-			user.NewSessionState(),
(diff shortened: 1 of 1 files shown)


matiasinsaurralde (approved):

buger:
I do not think it is 3.0 specific issue. Can you confirm that 2.9 also affected?

matiasinsaurralde:
Applied to `release-3.1` and `release-3-lts`.

sredxny:
@buger I confirm it works in `2.9.4`

PR #3376, opened 2020-11-03
Remove terraform version pin and lock

alephnull:
## Description
The terraform state file encodes the version of terraform CLI used. Later versions of the CLI will use statefiles created by older CLIs but the reverse is not true. Removing the version pin will use the latest packaged version. Hopefully this will help.

Secondly, there are there are always stale locks in the remote s3 state bucket. The source of these locks is not clear but there is no need to obtain a lock for the read-only operations in this workflow.

## Related Issue
Reported on Slack.

## Motivation and Context
See above.

## How This Has Been Tested
By hand.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


sedkis:
I wonder if it would be better to manually update the version each time instead of forcing "latest" ?  Pinning is a good practice

alephnull:
My shell automatically installs the latest which was the root cause of this problem. I'm trying to keep up with the newest syntax until 1.0, which should be the one after 0.14.

sedkis (approved):

PR #3372, opened 2020-10-27
TT-423 fix grpc proxy connection

sredxny:
## Description
Fix grpc proxy over h2c

## Related Issue
https://tyktech.atlassian.net/browse/TT-423

## Motivation and Context
enable grpc proxy

## How This Has Been Tested
- Run gRPC server (I used https://github.com/grpc/grpc-go/tree/master/examples/helloworld )
- Have an instance of dashboard and gw running
- Create Api definition for the gRPC service:
        - Set an specific port to serve our service via gw (I set 12345)
        - Set Listen Path: /
        - In target-url set the ip:port where the gRPC server is running, it should contain h2c as protocol, in that way the gateway infers that for this API it should use h2c, as an example: `h2c://myservice:50051`
- Consume the gRPC service (I used grpcurl ): `grpcurl -plaintext -proto helloworld.proto -d '{"name": "sredny"}' tyk-gateway:12345 helloworld.Greeter/SayHello`
- Received response:
```
{
  "message": "Hello sredny"
}
```
- Consumed from client app as well
- Consume a REST API, to check that nothing is broken

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring or add test (improvements in base code or adds test coverage to functionality)

## Checklist

- [x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). If pulling from your own
      fork, don't request your `master`!
- [x] Make sure you are making a pull request against the **`master` branch** (left side). Also, you should start
      *your branch* off *our latest `master`*.
- [x] My change requires a change to the documentation.
  - [ ] If you've changed APIs, describe what needs to be updated in the documentation.
  - [ ] If new config option added, ensure that it can be set via ENV variable
- [x] I have updated the documentation accordingly.
- [ ] Modules and vendor dependencies have been updated; run `go mod tidy && go mod vendor`
- [ ] When updating library version must provide reason/explanation for this update.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] Check your code additions will not fail linting checks:
  - [ ] `go fmt -s`
  - [ ] `go vet`

PR #3380, opened 2020-11-04
TT-968 enabled h2c by default

sredxny:
## Description
now for all the http proxies (including the main router of gw) is activated h2c. However, this doesn't means we will consume the upstream via h2c as this should be activated via `target_url`. **This should be merged after** https://github.com/TykTechnologies/tyk/pull/3372

## Related Issue
https://tyktech.atlassian.net/browse/TT-968

## Motivation and Context
give solution to https://tyktech.atlassian.net/browse/TT-968

## How This Has Been Tested
- Run gw in http mode and dashboard
- Created api A without custom port
- consume api  A via postman, and via curl (Eg: `curl -v --http2-prior-knowledge http://tyk-gateway:8080/keyless/uuid`)
- Created api B with custom port
- consume api B via postman and curl
- Created and configured grpc service
- Consume grpc service via grpcurl (Eg: `grpcurl -plaintext -proto helloworld.proto -d '{"name": "sredny"}' tyk-gateway:12345 helloworld.Greeter/SayHello`)

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


matiasinsaurralde (approved):
LGTM, we need a rebase

gateway/api_loader.go
```diff
@@ -119,6 +119,9 @@ func processSpec(spec *APISpec, apisByListen map[string]int,
 		"prefix": "coprocess",
 	})
 
+	if strings.HasPrefix(spec.Proxy.TargetURL, "h2c://") {
+		spec.Proxy.TargetURL = strings.Replace(spec.Proxy.TargetURL, "h2c://", "http://", 1)
+	}
```
gernest, line 124: why are we replacing target scheme here?

gateway/api_loader.go
```diff
@@ -763,6 +766,9 @@ func loadApps(specs []*APISpec) {
 	shouldTrace := trace.IsEnabled()
 	for _, spec := range specs {
 		func() {
+			if strings.Contains(spec.Proxy.TargetURL, "h2c://") {
+				spec.Protocol = "h2c"
```
gernest, line 770: should we check if `spec.Protocol` is already set? If for instance `spec.Protocol=http` and target is `h2c` is that allowed? If not then spec is not valid we should emit an error or something.

gateway/proxy_muxer.go
```diff
@@ -406,14 +408,14 @@ func (m *proxyMux) serve() {
...
 			var h http.Handler
 			h = &handleWrapper{p.router}
-			if p.protocol == "h2c" {
-				// wrapping handler in h2c. This ensures all features including tracing work
-				// in h2c services.
-				h = &h2cWrapper{
-					w: h.(*handleWrapper),
-					h: h2c.NewHandler(h, &http2.Server{}),
-				}
+			// by default enabling h2c by wrapping handler in h2c. This ensures all features including tracing work
```
gernest, line 411: This is a very big assumptions. By default we provide http proxy, so h2c,tcp etc are special case. You explicitly set protocol to h2c inside loadApps why not check it here?. I see this will be a major source of headache.

gernest (changes requested):

gateway/proxy_muxer.go
sredxny: @gernest hi, can you elaborate regarding why it could be a major source of headache? currently by enabling h2c by default we're looking to simplify the process the most that we can

gernest: @sredxny 
By design `http2` is supposed to be served over TLS,  `h2c` is an edge case/optional that is not supported by major browsers.

The only use case for `h2c` is when using `gRPC` without tls, which is also an edge case. Why do we make edge cases default behaviour? Like I said before http/http2 is the only thing we should take for granted, we should let our customers to choose anything else

Please see this stackoverflow answer on why http2 should always be on TLS https://stackoverflow.com/a/46789195

>currently by enabling h2c by default we're looking to simplify the process the most that we can

What process are you simplifying? A process of choosing protocol to use? Why this  `protocol: "h2c"` is not simple?

buger: protocol field require you to have separate gateway port, while in general there is no reasons for this.

gateway/api_loader.go
buger: Because we treat h2c as alias to http. Probably should be followed by conned in source code.

gateway/proxy_muxer.go
gernest: So the issue here is to remove the need for separate port?

gateway/api_loader.go
gernest: I'm confused, from reading `h2c` is not `http` traffic please see this doc comment https://pkg.go.dev/golang.org/x/net/http2/h2c#NewHandler

gateway/proxy_muxer.go
buger: In general yes. The whole idea of h2c that it works though standard http protocol, and if special header found, it enable special mode which kinda emulate http2 by using http as transport.

gateway/api_loader.go
buger: This link actually has good description, saying that h2c is http1.1 compatible ible

gateway/api_loader.go
sredxny: @gernest  if http is set, and the target contains "h2c://" then it has major precedence, hence we set h2c as protocol. In the very beggining we did this in this way due that FE doesn't have a protocol option `h2c`.

gernest (approved):

gateway/proxy_muxer.go
```diff
@@ -362,7 +363,12 @@ func (m *proxyMux) swap(new *proxyMux) {
 			}
 		}
 	}
-
+	p := m.getProxy(config.Global().ListenPort)
+	if p != nil && p.router != nil {
+		// All APIs processed, now we can healthcheck
+		// Add a root message to check all is OK
+		p.router.HandleFunc("/"+config.Global().HealthCheckEndpointName, liveCheckHandler)
```
gernest, line 370: Why are we doing this twice?

We are already registering this here https://github.com/TykTechnologies/tyk/blob/8650e1803309405da9ecd814da5e8efbf43732d5/gateway/server.go#L416

gernest (changes requested):

gateway/proxy_muxer.go
sredxny: thanks, I made the change

sredxny (commented):
changes done

tbuchaillot (approved):
Looks good!

matiasinsaurralde (approved):
LGTM

PR #3384, opened 2020-11-11
Osin fork change to TykTechnologies (part of TT-998)

joshblakeley:
## Description

The osin library is used by Tyk for building it's internal OAuth2.0 server. The current import uses a fork in the lonelycode org that is very old and misses out on the updates made to the original openshift/osin lib in the intervening years. i.e. PKCE support OOTB for Auth code flows etc

Because we need some custom changes to support out data model in Tyk we have created a new openshift fork and changed what we needed to for Tyk compatibility. We then use the TykTechnologies/osin fork as our osin library going forward.

## Motivation and Context

Enhance current OAuth capabilities and work towards full adherence to the OAuth2.1 spec, in which feature such as PKCE are mandatory. Therefore giving the ability for Tyk users to stay at the most up to date security standards.

## How This Has Been Tested
All Oauth related unit tests passing with some minor changes.

## Types of changes

## Checklist

- [x] When updating library version must provide reason/explanation for this update.
  - [x] `go fmt -s`
  - [x] `go vet`


buger:
Looks like you need to update it from fresh master, and ensure that it use Go modues.

buger:
Can you provide instructions for the QA team on how to test it?

joshblakeley:
@buger how do i remove all the crap i've accidentally pulled in to the PR?

joshblakeley:
This is a *soft* change. We now have optional PKCE under the hood. We will open another PR after this is merged to configure mandatory PKCE in oauth settings as it will require UI work and there is not much demand for it yet so can go in 3.4

joshblakeley:
I broke this branch real good - might open a new one. Putting this in draft again for now

joshblakeley:
Ok fixed and rebased on master

joshblakeley:
@buger Now has full PKCE support added and new unit test for it too.

apidef/api_definitions.go
```diff
@@ -374,6 +374,7 @@ type APIDefinition struct {
 		AllowedAccessTypes     []osin.AccessRequestType    `bson:"allowed_access_types" json:"allowed_access_types"`
 		AllowedAuthorizeTypes  []osin.AuthorizeRequestType `bson:"allowed_authorize_types" json:"allowed_authorize_types"`
 		AuthorizeLoginRedirect string                      `bson:"auth_login_redirect" json:"auth_login_redirect"`
+		RequirePKCE            bool                        `bson:"require_pkce" json:"require_pkce"`
```
asoorm, line 377: You should not "Require" PKCE. This is entirely optional & a client's decision to optionally send a `code_challenge`.

gateway/api.go
```diff
@@ -1508,6 +1508,10 @@ func createOauthClient(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+		if apiSpec.Oauth2Meta.RequirePKCE {
+			newClient.ClientSecret = ""
+		}
+
```
asoorm, line 1514: Why are we forcing ClientSecret to be empty? What if the client is a confidential client - meaning that it can send the client secret in the authorization request?

gateway/oauth_manager_test.go
```diff
@@ -374,6 +376,87 @@ func TestAPIClientAuthorizeAuthCode(t *testing.T) {
...
+			},
+		}
+	})
+
+	var redirectURI string
+	// If separator is not set that means multiple redirect uris not supported
+	if config.Global().OauthRedirectUriSeparator == "" {
+		redirectURI = "http://client.oauth.com"
+
+		// If separator config is set that means multiple redirect uris are supported
```
asoorm, line 402: Does this test's output depend on who runs it?

gateway/api.go
joshblakeley: In that case that partilcular API would not have `RequirePKCE` set and would work without the code challenge

gateway/oauth_manager_test.go
```diff
@@ -374,6 +376,87 @@ func TestAPIClientAuthorizeAuthCode(t *testing.T) {
...
+
+		_, _ = ts.Run(t, test.TestCase{
+			Path:      "/APIID/tyk/oauth/authorize-client/",
+			AdminAuth: true,
+			Data:      param.Encode(),
+			Headers:   headers,
+			Method:    http.MethodPost,
+			Code:      http.StatusForbidden,
+		})
+	})
```
asoorm, line 433: I'm not sure we should enforce this. It's up to the client, not the server.

gateway/oauth_manager_test.go
joshblakeley: No it wont make any difference - this is basically the boilerplate code for intitialising an Oauth client for tests but with the client ID set to an empty string like it would be when doing this via the Gateway API.

gateway/oauth_manager_test.go
```diff
@@ -374,6 +376,87 @@ func TestAPIClientAuthorizeAuthCode(t *testing.T) {
...
+
+	t.Run("Client authorize token request with PKCE and code_challenge", func(t *testing.T) {
+
+		challenge := "12345678901234567890123456789012345678901234567890"
+
+		param := make(url.Values)
+		param.Set("response_type", "code")
+		param.Set("client_id", authClientID)
+		param.Set("key_rules", keyRules)
+		param.Set("code_challenge", challenge)
```
asoorm, line 443: Would like to see a test with different `code_challenge_method`s. e.g. Plain / S256
Would like to see examples of failed auth using incorrect code_verifier (plan & s256)

joshblakeley: Yep thats easy to add and is handled in the underlying lib

gateway/oauth_manager_test.go
joshblakeley: RFC says differently https://tools.ietf.org/html/rfc7636#section-4.4.1

asoorm (changes requested):
Thanks for this. Next week - lt's do a zoom, as I'm not quite sure how this works.

apidef/api_definitions.go
joshblakeley: The language can be changed sure - but in terms of Tyk someone enabling PKCE is the same as requiring it for clients accessing a certain API.

joshblakeley: I picked require because the RFC talked a lot about servers requiring the PKCE flow not just enabling it

joshblakeley: For it to be optional you simply dont use the RequirePKCE option and you can send optional code challenges

joshblakeley:
> Thanks for this. Next week - lt's do a zoom, as I'm not quite sure how this works.

Fantastic - pretty sure I have this at the point i can explain it all in a satisfactory way

gateway/api.go
asoorm: What happens if one client is a public client, and the other is a confidential one? How to let both public and confidential clients use this API?

If we force PKCE, then confidential clients would send their client_secret and code_challenge in the authorize request. We should also validate the client_secret right?

gateway/oauth_manager_test.go
asoorm: What if there are 2 gateways in the cluster, and the authorize request happens on one GW, and the token request happens on another. Does the code_challenge get stored against the authorize request in Redis or something?

gateway/oauth_manager_test.go
asoorm: So this is an an optional security thing then right? force PKCE.
In which case, should  we have another config option - `enable pkce`?

gateway/api.go
joshblakeley: Thats a limitation for sure - if you want strict enforcement of pkce then you can only expose that API to public clients. Why you would want both at the same time on the same API is worth diving into though.

gateway/oauth_manager_test.go
joshblakeley: PKCE is there by default so no need to enable. If the code challenge it present then the token call is validated against it for that client

gateway/oauth_manager_test.go
joshblakeley: The code challenge lives on the client so the same as current behaviour basically.

buger:
@joshblakeley @asoorm what was results of your call? 
Lets finally merge it.

joshblakeley:
This was "done" as of May but has fallen through the cracks. It needs a maintainer from one of the squads to take over if we are to get it tested and merged.

PR #3389, opened 2020-11-13
SQL support

buger:
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


apidef/api_definitions.go
```diff
@@ -75,6 +79,76 @@ const (
...
+func (j ObjectId) String() string {
+	return j.Hex()
+}
+
+func (j ObjectId) GetBSON() (interface{}, error) {
+	return bson.ObjectId(j), nil
+}
+
+func ObjectIdHex(hex string) ObjectId {
+	return ObjectId(bson.ObjectIdHex(hex))
```
vaske, line 129: in case that `hex` is not proper hex format this will panic, should we add prevent of that at this level or?

buger: But it is the same as do `bson.ObjectIdHex`. So it really depend on how you handle it on dash side.

furkansenharputlu (approved):

PR #3397, opened 2020-11-19
Enforce key ID usage when retrieving basic auth keys

matiasinsaurralde:
To be merged in combination with the dashboard PR [here](https://github.com/TykTechnologies/tyk-analytics/pull/2141).

In slaves nodes when we attempt to remove a key that is an username we do not know the orgId, then in this case will use the orgId of the context of the node. We need the orgId to generate the token and know what to remove, otherwise we would need to loop over all the apis and attempt to remove the key, which is not efficient.
This should be merged at the same time as TykTechnologies/tyk-analytics#2141

## Related Issue
https://tyktech.atlassian.net/browse/TT-984

## Motivation and Context
Handle properly the keys that are username and replicate correctly the delete/update operations to worker nodes

## How This Has Been Tested
- Ran MDCB env
- Create basic auth api
- create policy
- Create key, and use it in worker node. Then search by base64 key and remove it...the key is removed in slave as well
- Create key and use it in worker node. Then search it by hash and remove it, the key is removed in slaved node
- create key and use it in worker node. Then search it by username and remove it, the key is removed in slaved node
- Also tested the update operations in a similar way

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/api.go
+++ b/gateway/api.go
@@ -540,6 +540,13 @@ func handleGetDetail(sessionKey, apiID string, byHash bool) (interface{}, int) {
 		}
 	}
 
+	// If it's a basic auth key and a valid Base64 string, use it as the key ID:
+	if session.BasicAuthData.Password != "" {
+		if _, err := base64.StdEncoding.DecodeString(sessionKey); err == nil {
+			session.KeyID = sessionKey
+		}
+	}
+
 	log.WithFields(logrus.Fields{
 		"prefix": "api",
 		"key":    obfuscateKey(sessionKey),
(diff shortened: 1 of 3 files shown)


gateway/api.go
```diff
@@ -540,6 +540,13 @@ func handleGetDetail(sessionKey, apiID string, byHash bool) (interface{}, int) {
 		}
 	}
 
+	// If it's a basic auth key and a valid Base64 string, use it as the key ID:
+	if session.BasicAuthData.Password != "" {
+		if _, err := base64.StdEncoding.DecodeString(sessionKey); err == nil {
```
buger, line 545: Same as https://github.com/TykTechnologies/tyk/pull/3398/files

matiasinsaurralde:
Removing the ticket tag from this one and closing it. We should use #3400 which it's pointing to `master`.

PR #3419, opened 2020-12-02
[TT-311] Don't cache tls config  when we use certificate ID and redis is down.

### LINKED ISSUE #3292: Gateway fails to read certificate
- Branch/Version: Stable 3.0
- Environment: All

**Describe the bug**
Gateway fails to read a certificate that was functional in tyk gateway 2.9.4

**Reproduction steps**

1. Add a .pem containing cert and private key to dashboard
2. Add the cert id to tyk.conf at http_server_options.ssl_certificates and set use_ssl true

**Actual behavior**
Gateway fails with the error:
`TLS handshake error from xxx.xxx.xxx.xxx:xxxxx: tls: no certificates configured`

Gateway additionally reports:
`level=warning msg="Reconnecting storage: Redis is either down or ws not configured" prefix=pub-sub`
But the certificate is visible in redis.

This behavior is observed both on a fresh installation of 3.0 connecting to a fresh docdb and redis cluster as well as upgrading and keeping the pre-existing data.

**Expected behavior**
Gateway starts correctly and receives ssl traffic from dashboard

gernest:
## Description

This  makes sure we don't cache tls configuration when we are expecting certificates from redis (via certificate ID) and redis is not available yet. The resulting tls handshake will fail, it will eventually pass when redis is available.

When certificates are added to dashboard, we make sure they are eventually loaded to the gateway.

## Related Issue
Fixes #3292

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/cert.go
+++ b/gateway/cert.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/TykTechnologies/tyk/certs"
 	"github.com/TykTechnologies/tyk/config"
+	"github.com/TykTechnologies/tyk/storage"
 
 	"github.com/gorilla/mux"
 	cache "github.com/pmylund/go-cache"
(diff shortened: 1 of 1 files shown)


gateway/cert.go
```diff
@@ -271,38 +272,40 @@ var tlsConfigCache = cache.New(60*time.Second, 60*time.Minute)
...
-		certNameMap[certData.Name] = &cert
-	}
 
-	for _, cert := range CertificateManager.List(config.Global().HttpServerOptions.SSLCertificates, certs.CertificatePrivate) {
-		if cert != nil {
-			serverCerts = append(serverCerts, *cert)
-		}
-	}
+		// if we have certificates ID's we don't cache the config if redis is not available yet.
+		noCaching := len(config.Global().HttpServerOptions.SSLCertificates) > 0 && !storage.Connected()
```
buger, line 283: Should not depend just on `SSLCertificates` field. If you use mTLS or similar you also should not cache if redis not enabled. So just change to `storage.Connected()`

buger: Also add a comment with explanation why it is needed

buger:
Needs JIRA ticket and test added

sedkis:
For TT-311

sredxny:
@gernest hi, is this PR ready for review? I see it as a draft

gernest:
@sredxny this is ancient can't remember much. You can resurrect this if you want, or come up with better solution. I should probably close this.

PR #3424, opened 2020-12-09
Sync automation workflows automatically

alephnull:
# Description
These related PRs,

Add a new workflow triggered on branch delete that will help with cleanup of developer environments. This workflow was forgotten in the last commit.

Add a workflow (sync-automation.yml) that will sync the automation workflows from master to known release branches. The list of branches is maintained [pr.zsh](https://github.com/TykTechnologies/tyk-ci/blob/master/wf-gen/pr.zsh). The workflow is only triggerred for appropriate paths and will only work for for the FILES in `pr.zsh`.

# Testing
Syntax checks have passed. Most of the real functionality is external to the workflow.

The status of the integration workflows can be inspected in this PR itself in the _Checks_ tab. The branch is prefixed with `feature/` to test that it triggers the required workflows when it is deleted.

The automation sync workflow was tested on [tyk-pump](https://github.com/TykTechnologies/tyk-pump/actions?query=workflow%3A%22Sync+automation%22) and the results can be viewed there.

# Merging
Once this PR is merged to master it will sync the automation workflows to all branches in the `matrix->branch` node.

And it will continue to do so for future commits.


.github/workflow/sync-automation.yml
```diff
@@ -0,0 +1,45 @@
...
+
+name: Sync automation
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - .github/workflows/*
+      - integration/*
```
buger, line 16: Still I guess it should work with feature/* branches?

alephnull: These are the paths that the workflow will be triggered on. .`g/w/*` contains the workflows and `integration/*` contains some docker and terraform stuff needed to make the whole thing work.

alephnull: sync-automation is purely involved in syncing workflows to all the release branches, not directly related to the developer environments.

.github/workflow/sync-automation.yml
```diff
@@ -0,0 +1,45 @@
...
+  FILES: .github/workflows/int-image.yml .github/workflows/del-env.yml integration/terraform/outputs.tf integration/image/Dockerfile .github/workflow/sync-automation.yml
+  
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        branch: [ release-2.9, release-3, release-3-lts, release-3.0.2, release-3.0.2-update, release-3.0.3, release-3.1, release-3.1.0, release-3.1.1, release-3.1.2 ]
```
buger, line 28: Can we just support all release-* branches?

alephnull: We could but the tradeoff is:
- a more complex script (`grep`, `sed`, `awk`, etc.)
- cannot use the matrix build to do everything in parallel

With this implementation, the limit of the automation is very clear. Moreover, these workflows are not maintained by hand, they are generated from templates. While reviewing, I noticed that the path of sync-workflow is wrong, pushing a fix.

alephnull:
All PRs have been updated.

PR #3430, opened 2020-12-17
[TT-1258] Fix JWKs compatibility with previous TYk versions

buger:
Support public key in PEM format is found in x5c (old way)

Fix https://tyktech.atlassian.net/browse/TT-1258

## Motivation and Context
In 3.0.2 we broke clients who used Tyk specific non-RFC compatible JWKs

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/mw_jwt.go
+++ b/gateway/mw_jwt.go
@@ -1,6 +1,7 @@
 package gateway
 
 import (
+	"bytes"
 	"crypto/md5"
 	"crypto/x509"
 	"encoding/base64"
@@ -43,6 +44,21 @@ func (k *JWTMiddleware) EnabledForSpec() bool {
 
 var JWKCache *cache.Cache
 
+type JWK struct {
+	Alg string   `json:"alg"`
+	Kty string   `json:"kty"`
+	Use string   `json:"use"`
+	X5c []string `json:"x5c"`
+	N   string   `json:"n"`
+	E   string   `json:"e"`
+	KID string   `json:"kid"`
+	X5t string   `json:"x5t"`
+}
+
+type JWKs struct {
+	Keys []JWK `json:"keys"`
+}
+
 func parseJWK(buf []byte) (*jose.JSONWebKeySet, error) {
 	var j jose.JSONWebKeySet
 	err := json.Unmarshal(buf, &j)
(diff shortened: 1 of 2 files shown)


gateway/mw_jwt.go
```diff
@@ -52,13 +68,67 @@ func parseJWK(buf []byte) (*jose.JSONWebKeySet, error) {
...
 	if JWKCache == nil {
 		k.Logger().Debug("Creating JWK Cache")
 		JWKCache = cache.New(240*time.Second, 30*time.Second)
 	}
 
+	if key, err := k.legacyGetSecretFromURL(url, kid, keyType); err == nil {
+		return key, nil
+	} else {
+		k.Logger().WithError(err).Info("Legacy JWK qery")
+	}
```
gernest, line 130: I think this should be a fallback. When we fail to get certificates in the standard way. `parseJWK` will fail if we have keys inside `x5c` , I think this should happen after `parseJWK` returns an error.

 Also, we don't need to make 2 calls to the JWK endpoint just to receive the same data( when legacy fails, we have to try again with standard). We can call the endpoint once, read the body into a buffer and use the buffer for both standard and legacy mode.

gernest (changes requested):

gateway/mw_jwt.go
buger: Yes I thought about it. It required bigger rewrite. But at the same time we do caching so performance not really an issue. Any ideas on it without rewriting too much stuff?

Also note that once cached it will fail fast since just check if blob has prefix.

gernest: @buger 

something like this will do

```
		if jwkSet, err = parseJWK(buf); err != nil {
			k.Logger().WithError(err).Error("Failed to decode standard body JWK")
			k.Logger().Debug("Trying legacy JWK")
			return k.legacyGetSecretFromURL(url, kid, keyType)
		}
```

gernest: `legacyGetSecretFromURL` can accept `buf`  and this will avoid additional call with minimal refactoring.

buger: Done

gateway/mw_jwt_test.go
```diff
@@ -1402,20 +1402,20 @@ func TestJWTSessionRSAWithEncodedJWK(t *testing.T) {
...
-	})
-	t.Run("Direct JWK URL with bad jwk", func(t *testing.T) {
+	// t.Run("Direct JWK URL", func(t *testing.T) {
+	// 	spec.JWTSource = testHttpJWK
+	// 	LoadAPI(spec)
+	// 	flush()
+	// 	ts.Run(t, test.TestCase{
+	// 		Headers: authHeaders, Code: http.StatusOK,
+	// 	})
+	// })
```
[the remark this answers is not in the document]
buger: Done

gernest (approved):

PR #3434, opened 2020-12-23
[TT-1284] Load gateway health check to control port

furkansenharputlu:
With this PR, health check will be loaded to:
- control port, if `control_api_port` is set in `tyk.conf`
- listen port, if `control_api_port` is not set in `tyk.conf`

(Briefly it will be loaded in the same level with control api, only difference it will not require any authorization)

For example, you don't set `control_api_port` and you want to listen control api from `listen_port`. Although user sets listen path of one of apis to `/`, health check will be triggered when you call its endpoint e.g.`tyk-gateway:8181/hello`

Fixes https://github.com/TykTechnologies/tyk-operator/issues/205

### DIFF
--- a/gateway/gateway_test.go
+++ b/gateway/gateway_test.go
@@ -812,7 +812,7 @@ func TestListener(t *testing.T) {
 // Admin api located on separate port
 func TestControlListener(t *testing.T) {
 	ts := StartTest(TestConfig{
-		sepatateControlAPI: true,
+		SeparateControlAPI: true,
 	})
 	defer ts.Close()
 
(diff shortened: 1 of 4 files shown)


asoorm:
Thank you. This is great for my specific use-case. Maybe we should wait for @excieve review to confirm if it is flexible enough.

excieve (approved):
Looks good to me, thanks @furkansenharputlu! As I understand, this also fixes the bug with `/hello` being ignored in traversal when it's living on the main port with API listen path being set to `/`, right?

furkansenharputlu:
@excieve Yes, that’s right.

buger:
@furkansenharputlu does this second fix require linking second Jira issue too?

furkansenharputlu:
@buger Which Jira issue? The only one I know is TT-1284.

buger:
Ah, I miss-read it. Looks good. 
Would be good to have some test, which sets control API on the separate endpoint, and checks if all works as expected (before and after API reload).

furkansenharputlu:
Ok, I will add that case to the existing tests.

furkansenharputlu:
@buger I added the test cases.

asoorm (approved):

PR #3438, opened 2021-01-11
TT-1188 mdcb testing framework / Gw refactor 

sredxny:
## Description
refactor GW in order that we can run multiple instances as objects via code (not docker, not binaries), added method to run gw in testing mode as a worker node.  Created gw object that wrapps some information and methods specific for a gateway instance. This must be merged in tandem with:
- https://github.com/TykTechnologies/tyk-analytics/pull/2166
- https://github.com/TykTechnologies/tyk-sink/pull/109

## Related Issue
https://tyktech.atlassian.net/browse/TT-1188

## Motivation and Context
Created the proper structure so we can run integration tests for MDCB environment

## How This Has Been Tested
- Run Gw 
- Run GW instances from MDCB testing framework

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


gateway/server.go
```diff
@@ -1428,19 +1438,25 @@ func setupPortsWhitelist() {
 	config.SetGlobal(globalConf)
 }
 
-func startServer() {
+type Gateway struct {
+	DefaultProxyMux *proxyMux
+	Port            int
```
buger, line 1443: We also have a control port. Maybe just pass the whole configuration object instead?

gateway/server.go
```diff
@@ -1196,6 +1196,14 @@ func getGlobalStorageHandler(keyPrefix string, hashKeys bool) storage.Handler {
 	return &storage.RedisCluster{KeyPrefix: keyPrefix, HashKeys: hashKeys}
 }
 
+func NewGateway() Gateway {
```
tbuchaillot, line 1199: I think it would be better if we receive a config here and maintain that config in our Gateway struct.

gateway/server.go
tbuchaillot: If we have a config field inside the gateway struct, we should also change all the gateway receiver from `func (gw Gateway)` to a pointer receiver`func (gw *Gateway)`, just in case we modify the config in any gateway func.

tbuchaillot (changes requested):

matiasinsaurralde (approved):
Looks good

sredxny:
At this point 2 things are missing to be fixed:

1- Some tests are failing, and some other fails initializing their gateway, we got:

```
panic: html/template: pattern matches no files: `templates/error*` [recovered]
        panic: html/template: pattern matches no files: `templates/error*`
```

2- The function `TykTriggerEvent(CEventName, CPayload *C.char)` was modified to be as `func (gw *Gateway) TykTriggerEvent(CEventName, CPayload *C.char)` but now when we attempt to compile the gateway, we get: `gateway/coprocess_api.go:41:11: unrecognized Go type Gateway` I know this is related to the tag `// export` now we need to find a way to export that function correctly

buger:
For 1. see the recent Test which added for testing XML error templates. Essentially it creates tmp directory with mock error template. 

As for 2. maybe @matiasinsaurralde has some ideas? If not we can try to make exception, and make this function using global variables or similar.

gateway/server.go
sredxny: done

sredxny:
@matiasinsaurralde  check this commit https://github.com/TykTechnologies/tyk/pull/3438/commits/58617fa3a7b7f95bb9f7f6567f88f740299e8096

Basically there am doing:
- Declaring a global function that will be called by TykTriggerEvent
- When the gateway starts then set a value to `TykTriggerEvent` (set the function that belongs to the gateway instance)

At this point am still fixing tests and lint

matiasinsaurralde:
> @matiasinsaurralde check this commit [58617fa](https://github.com/TykTechnologies/tyk/commit/58617fa3a7b7f95bb9f7f6567f88f740299e8096)

@sredxny Great, LGTM!

TESTING.md
```diff
@@ -25,6 +25,7 @@ Main points of the test framework are:
 - All tests run HTTP requests though the full HTTP stack, same as user will do
 - Test definition logic separated from test runner.
 - Official mocks for the Dashboard, RPC, and Bundler
+- The majority of the time you will need a gateway instance in order to use the gateway functions and properties
```
padiazg, line 28: What about 
`Most of the time, you will need a gateway instance to use the gateway functions and properties`

padiazg (approved):
LGTM

TESTING.md
sredxny: makes sense, done!

tbuchaillot (approved):
LGTM

matiasinsaurralde (approved):

Issue #3100, opened 2020-05-27
Authorisation using certificate does not work when domain specified as regexp


buger:
**Branch/Environment/Version**
- Branch/Version: master

**Describe the bug**
If custom domain field using regexp, and we enabled Authorization using a certificate Tyk not always ask client for a certificate. 

**Reproduction steps**
Have a clean install, or ensure that there are no APIs which currently use mutual tls or auth token with a certificate.

1) Create an API with Auth token and certificate check, with domain containing a regexp, for example: `{?:example.com}`
2) Create a token based on certificate and try to access the API
3) It should say that token is not found

**Actual behavior**
It does not require a certificate

**Expected behavior**
It should understand domain as regexp and run certificate check for it

**Additional context**
There is a way to overcome the issue, if you create a dummy API on the domain without a "regexp", like `example.com` (can be even API with  `internal` checkbox), with Mutual TLS or Auth Token with certificate check, and some random listen path, so it will not conflict with your original API. This will trigger Tyk to request certificate for such Domain.

PR #3442, opened 2021-01-12
[TT-107] Handle client certificate for custom domains in regex format

furkansenharputlu:
For custom domains routing, `gorilla` has its own type of regex. While handling client certificate with custom domain, we should use the same regex to make it work properly.

Fixes https://github.com/TykTechnologies/tyk/issues/3100

### DIFF
--- a/gateway/cert.go
+++ b/gateway/cert.go
@@ -390,21 +390,17 @@ func getTLSConfigForClient(baseConfig *tls.Config, listenPort int) func(hello *t
 			}
 		}
 
-		matches := false
+		newConfig.ClientAuth = tls.NoClientCert
+
 		for key, clientAuth := range domainRequireCert {
 			req := http.Request{Host: hello.ServerName, URL: &url.URL{}}
 			if mux.NewRouter().Host(key).Match(&req, &mux.RouteMatch{}) {
 				newConfig.ClientAuth = clientAuth
-				matches = true
 				break
 			}
 		}
 
-		if !matches {
-			newConfig.ClientAuth = tls.NoClientCert
-		}
-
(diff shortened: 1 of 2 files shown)


buger:
We have a pretty good test suite for TLS, let's new test cases there.

gateway/cert.go
```diff
@@ -389,7 +390,20 @@ func getTLSConfigForClient(baseConfig *tls.Config, listenPort int) func(hello *t
 			}
 		}
 
-		newConfig.ClientAuth = domainRequireCert[hello.ServerName]
+		matches := false
+		for key, clientAuth := range domainRequireCert {
+			req := http.Request{Host: hello.ServerName, URL: &url.URL{}}
+			if mux.NewRouter().Host(key).Match(&req, &mux.RouteMatch{}) {
```
buger, line 396: Checked the library, was not able to find a better way to do such check...

Otherwise looks good, lets add more tests.

gateway/cert.go
```diff
@@ -389,7 +390,20 @@ func getTLSConfigForClient(baseConfig *tls.Config, listenPort int) func(hello *t
...
+				newConfig.ClientAuth = clientAuth
+				matches = true
+				break
+			}
+		}
+
+		if !matches {
```
buger, line 403: We can remove negative logic, and `matches` boolean var too, and just move this to line 393, e.g. always set to `tls.NoClientCert` by default.

buger (approved):
Looks good!

buger:
Some tests related to TLS are failing, but looks like it is not part of this PR.

spetrunin (approved):

Issue #3406, opened 2020-11-24
2.6.4/2.7.7 -> 2.8.5+: "Url Rewrite" w/ $tyk_context / slug stripping behavior broken


bitsofinfo:
# Issue

*API definition at bottom*

Trying to upgrade from 2.6.4 hybrid to 2.9.x and then on to 3.x

I am attempting this release by release. 

2.6.4 is what is running production fine

2.7.7 after upgrade works fine

2.8.5 the behavior changes and it no longer works

# Scenario

Inbound request to Tyk:
```
POST https://myapi-stage.frontend.com/xyz/4.2/person/joe.smith@testing.com/get-stuff
```

Slug is: `xyz/4.2/person/`

Middleware sets request header: `X-Zzz-Proxytourl: https://myapi-stage.backend.com/person`

Tyk should take the value from the request header `X-Zzz-Proxytourl` and use it in the **Url Rewrite** rule as follows
```
...
    {
    "path": "/(.+)",
    "method": "POST",
    "match_pattern": "/(.+)",
    "rewrite_to": "$tyk_context.headers_X_Zzz_Proxytourl/$1",
    "triggers": []
    },
...
```

Expected upstream/target:
```
POST https://myapi-stage.backend.com/person/joe.smith@testing.com/get-stuff
```

**It appears that in 2.8.x the slug stripping no longer works the same?**

*2.6.4/2.7.7*
```
time="Nov 24 15:34:05" level=debug msg="Stripping: /xyz/4.2/person" 
time="Nov 24 15:34:05" level=debug msg="Upstream Path is: /person/joe.smith@testing.com/get-stuff" 
```

*2.8.x?*
```
time="Nov 24 15:39:29" level=debug msg="Stripping: /xyz/4.2/person" 
time="Nov 24 15:39:29" level=debug msg="Upstream Path is: /person/xyz/4.2/person/joe.smith@testing.com/get-stuff" 
```

# 2.6.4 Works as expected 

time="Nov 24 15:37:45" level=debug msg="JSVM middleware execution took: (ns) 11821000" 
time="Nov 24 15:37:45" level=debug msg="Rewriter active" 
time="Nov 24 15:37:45" level=debug msg="/xyz/4.2/person/joe.smith@testing.com/get-stuff" 
time="Nov 24 15:37:45" level=debug msg="Inbound path: /xyz/4.2/person/joe.smith@testing.com/get-stuff" 
[...]

PR #3447, opened 2021-01-25
[TT-437]  Add test cases for white/blacklist endpoint including listen path

furkansenharputlu:
The issue described in https://tyktech.atlassian.net/browse/TT-437 is not valid. To handle that situation `$` can be used.
This PR adds that test case.

### DIFF
--- a/gateway/api_definition_test.go
+++ b/gateway/api_definition_test.go
@@ -115,7 +115,7 @@ func TestWhitelist(t *testing.T) {
 	t.Run("Simple Paths", func(t *testing.T) {
 		BuildAndLoadAPI(func(spec *APISpec) {
 			UpdateAPIVersion(spec, "v1", func(v *apidef.VersionInfo) {
-				v.Paths.WhiteList = []string{"/simple", "/regex/{id}/test"}
+				v.Paths.WhiteList = []string{"/simple", "pathWithoutSlash", "/regex/{id}/test"}
 				v.UseExtendedPaths = false
 			})
 
@@ -125,6 +125,7 @@ func TestWhitelist(t *testing.T) {
 		ts.Run(t, []test.TestCase{
 			// Should mock path
 			{Path: "/simple", Code: http.StatusOK},
+			{Path: "/pathWithoutSlash", Code: http.StatusOK},
 			{Path: "/regex/123/test", Code: http.StatusOK},
 			{Path: "/regex/123/differ", Code: http.StatusForbidden},
 			{Path: "/", Code: http.StatusForbidden},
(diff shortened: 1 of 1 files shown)


gateway/api_definition.go
```diff
@@ -1027,9 +1027,11 @@ func (a *APISpec) getURLStatus(stat URLStatus) RequestStatus {
 
 // URLAllowedAndIgnored checks if a url is allowed and ignored.
 func (a *APISpec) URLAllowedAndIgnored(r *http.Request, rxPaths []URLSpec, whiteListStatus bool) (RequestStatus, interface{}) {
+	path := strings.TrimPrefix(r.URL.Path, strings.TrimSuffix(a.Proxy.ListenPath, "/"))
```
furkansenharputlu, line 1030: ~~/apples~~/apple

spetrunin (approved):

buger:
It is a VERY dangerous change...

buger:
It is a breaking change for the customers who already have whitelisting enabled. And using this "hack". 
I believe the we need to support almost both modes (literally check match with and without listen path)

furkansenharputlu:
Bug fixing is always a breaking change for a user depending on wrong behavior. If you look at the JIRA ticket, there is a bug. https://www.loom.com/share/0cc6a3ad48ec4a10bb55bb1d8e3f019a

The problem is the matching check with listen path. We need to remove it to solve the problem.

What is your proposal to fix that?

buger:
Definiton of the bug that if smth was working and now it is not. 
But since this behavior was there for ages, and our customers 100% depend on it, it is not a bug anymore it is a default behavior :)

We had very similar change in 2.8, which broke behavior of a lot of SLA customers, and we had to emergently roll it back. 
If we need to merge this change we need to ensure that if you specify whitelist like "/apple" or like "/apples/apple" it still will work. 

And IMHO it looks as simple as changing this line:
```
if !rxPaths[i].Spec.MatchString(path)
```
to 
```
if !(rxPaths[i].Spec.MatchString(path) || rxPaths[i].Spec.MatchString(r.URL.Path))
```

cc @asoorm

buger:
Also very related long-staging issue https://github.com/TykTechnologies/tyk/issues/3406

furkansenharputlu:
@buger 

`if !(rxPaths[i].Spec.MatchString(path) || rxPaths[i].Spec.MatchString(r.URL.Path))`

This looks same to the existing one. ` rxPaths[i].Spec.MatchString(r.URL.Path)` already covers `rxPaths[i].Spec.MatchString(path)`, doesn't it?

So, the thing I understand what you say is that it is not an issue and the ticket can be closed. Is that right?

furkansenharputlu:
Or we can add a new bool check like `dont_match_listen_path`.

buger:
@furkansenharputlu also needs a test case when listen path is included into whitelist endpoint.

furkansenharputlu:
Do you mean apple without $ should match case?

buger:
When in the Whitelist you have smth like "/apples/apple". 
The main logic here that you can overcome the issue using two ways, either adding "$" at the end of URL, or by adding listen path as a prefix.

PR #3450, opened 2021-01-29
Add support for GraphQL engine v2 (TT-688)

pvormste:
This PR adds support for the GraphQL engine v2. This includes:
 - new GraphQL engine configuration in API Definition
 - a new version field in GraphQL configuration 
 - adapter for converting GraphQL configuration into engine config v2
 - a new API definition semantic validator
   - extensible via implementing the validation rule interface
   - added a rule to validate for duplicated data source names
 - semantic validation of API definition when adding or updating API definitions
   - added a test for this functionality
 - GraphQL middleware can distinguish between v1 and v2 engine configuration
 - reverse proxy can handle v1 and v2 engine requests
 - added hooks for v2 engine

## Types of changes


apidef/graphql_config_adapter.go
```diff
@@ -0,0 +1,284 @@
...
+
+		switch ds.Kind {
+		case GraphQLEngineDataSourceKindREST:
+			restConfig := GraphQLEngineDataSourceConfigREST{}
+			err = json.Unmarshal(ds.Config, &restConfig)
+			if err != nil {
+				return nil, err
+			}
+
+			factory := &rest_datasource.Factory{}
```
furkansenharputlu, line 94: `rest_datasource` is not a valid package call. Maybe `restDatasource`

apidef/graphql_config_adapter.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/jensneuse/graphql-go-tools/pkg/engine/datasource/graphql_datasource"
+	"github.com/jensneuse/graphql-go-tools/pkg/engine/datasource/httpclient"
+	"github.com/jensneuse/graphql-go-tools/pkg/engine/datasource/rest_datasource"
+	"github.com/jensneuse/graphql-go-tools/pkg/engine/plan"
+	"github.com/jensneuse/graphql-go-tools/pkg/graphql"
+)
+
+var ErrUnsupportedGraphQLConfigVersion = errors.New("provided version of GraphQL config is not supported for this operation")
+
+type GraphQLConfigAdapter struct {
```
furkansenharputlu, line 19: Shouldn't this struct and all of its methods inside of the graphql library? I could not see gateway specific thing inside it.

gateway/api_test.go
```diff
@@ -1554,3 +1558,80 @@ func TestRotateClientSecretHandler(t *testing.T) {
 		})
 	}
 }
+
+func TestHandleAddOrUpdateApi(t *testing.T) {
+	testFs := afero.NewMemMapFs()
+
+	t.Run("should return error when api definition json is invalid", func(t *testing.T) {
+		apiDefJson := []byte("{")
+		req, err := http.NewRequest("POST", "http://gateway", bytes.NewBuffer(apiDefJson))
```
furkansenharputlu, line 1567: better to use `http.MethodPOST`

gateway/api_test.go
```diff
@@ -1554,3 +1558,80 @@ func TestRotateClientSecretHandler(t *testing.T) {
...
+		require.NoError(t, err)
```
furkansenharputlu, line 1568: I think we should remove error check for built in functions to make tests clean.

gateway/mw_graphql.go
```diff
@@ -235,3 +264,43 @@ func (p postReceiveHttpHook) Execute(ctx datasource.HookContext, resp *http.Resp
 			},
 		).Debugf("%s.%s: postReceiveHttpHook executed", ctx.TypeName, ctx.FieldName)
 }
+
+type beforeFetchHookV2 struct {
```
furkansenharputlu, line 268: Why not `GraphQLMiddleware` implements `OnBeforeFetch`, `OnData `...

furkansenharputlu (changes requested):

apidef/api_definitions.go
```diff
@@ -551,6 +553,62 @@ type GraphQLConfig struct {
...
 	// GraphQLPlayground is the Playground specific configuration.
 	GraphQLPlayground GraphQLPlayground `bson:"playground" json:"playground"`
+	// EngineConfig holds the configuration for engine v2 and upwards
+	EngineConfig GraphQLEngineConfig `bson:"engine_config" json:"engine_config"`
+}
+
+type GraphQLConfigVersion string
+
+const (
+	GraphQLConfigVersionNone = ""
```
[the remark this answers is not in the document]
pvormste: Versions are not integers, they are strings ("1", "2", "2.1") and as a string it guarantees backward-compatibility

apidef/api_definitions.go
```diff
@@ -551,6 +553,62 @@ type GraphQLConfig struct {
...
+
+type GraphQLEngineDataSource struct {
+	Kind       GraphQLEngineDataSourceKind `bson:"kind" json:"kind"`
+	Name       string                      `bson:"name" json:"name"`
+	Internal   bool                        `bson:"internal" json:"internal"`
+	RootFields []GraphQLTypeFields         `bson:"root_fields" json:"root_fields"`
+	Config     json.RawMessage             `bson:"config" json:"config"`
+}
+
+type GraphQLTypeFields struct {
```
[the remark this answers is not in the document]
pvormste: The library already has a graphql package and putting apidef related things in another graphql package would be very confusing - besides the fact they are apidef-related (they should really live in this package)

apidef/graphql_config_adapter.go
pvormste: this package name is part of the library - I can only alias it in this file

apidef/api_definitions.go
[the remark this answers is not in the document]
furkansenharputlu: They will be inside `apidef` but in a separate package. There are very long names all starting with `GraphQL`, it is a little awkward.

apidef/graphql_config_adapter.go
pvormste: As an adapter, it should live where it adapts things - as it takes the apidef config and adapts it to the engine config the right place is the apidef package

apidef/api_definitions.go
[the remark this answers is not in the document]
furkansenharputlu: Ok, I missed they are for users. However, as I said they can be like `v1`, `v2`.

gateway/api_test.go
pvormste: it's not an assertion here - `require` ensures that no other side effect will break this test. This is for making tests clean.

apidef/graphql_config_adapter.go
furkansenharputlu: Then maybe, it is more than an adapter because I feel like `engineConfigV2DataSources` is not related to `apidef`

gateway/api_test.go
furkansenharputlu: Nope. 
```
apiDefJson := []byte("{")
req, err := http.NewRequest("POST", "http://gateway", bytes.NewBuffer(apiDefJson))
```

This is deterministic and as a tester, you know that it will not fail because you define everything and it is a built-in function.

furkansenharputlu: See some of the built-in go tests. They are ignoring errors because no need to check each error.

apidef/api_definitions.go
```diff
@@ -543,6 +543,8 @@ type GraphQLConfig struct {
 	Enabled bool `bson:"enabled" json:"enabled"`
 	// ExecutionMode is the mode to define how an api behaves.
 	ExecutionMode GraphQLExecutionMode `bson:"execution_mode" json:"execution_mode"`
+	// Version defines the version of the GraphQL config and engine to be used
+	Version string `bson:"version" json:"version"`
```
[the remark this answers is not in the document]
pvormste: yep, changed!

apidef/api_definitions.go
```diff
@@ -551,6 +553,62 @@ type GraphQLConfig struct {
 	TypeFieldConfigurations []datasource.TypeFieldConfiguration `bson:"type_field_configurations" json:"type_field_configurations"`
 	// GraphQLPlayground is the Playground specific configuration.
 	GraphQLPlayground GraphQLPlayground `bson:"playground" json:"playground"`
+	// EngineConfig holds the configuration for engine v2 and upwards
+	EngineConfig GraphQLEngineConfig `bson:"engine_config" json:"engine_config"`
```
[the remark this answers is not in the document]
pvormste: makes sense: changed!

apidef/graphql_config_adapter.go
```diff
@@ -0,0 +1,284 @@
...
+		return nil, err
+	}
+
+	conf.SetFieldConfigurations(fieldConfigs)
+	conf.SetDataSources(datsSources)
+
+	return &conf, nil
+}
+
+func (g *GraphQLConfigAdapter) engineConfigV2FieldConfigs() (planFieldConfigs plan.FieldConfigurations, err error) {
```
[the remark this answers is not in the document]
pvormste: yeah it does not - changed!

apidef/graphql_config_adapter.go
pvormste: changed!

gateway/api_test.go
pvormste: Done!

gateway/mw_graphql.go
pvormste: Good idea! Changed!

gateway/api_test.go
pvormste:
```
This is deterministic and as a tester, you know that it will not fail because you define everything and it is a built-in function.
```
Highly disagree - I cannot predict the future. So maybe the inner-working of this function will change in the future and my input may then return an error. It's a good habit to exclude any possible side-effect and I really think it's not worth it to discuss this simple line of code if a change won't make it significantly better.

apidef/api_definitions.go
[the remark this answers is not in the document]
pvormste: all other apidef related types are located in this package - it would be unexpected to have the GraphQL ones in another. Also, long names are not bad even though the Go Community don't like them

gateway/api_test.go
furkansenharputlu: We are talking about a Go built-in function. Overthinking makes things complex instead of easier. Also, not a simple line, I saw it is repetitive.

Side note: Yes you are right but this is not a blocker to merge this one. If you don't want to change, it is ok. I just wrote my idea.

apidef/api_definitions.go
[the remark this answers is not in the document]
furkansenharputlu: maybe you forgot

apidef/api_definitions.go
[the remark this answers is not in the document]
furkansenharputlu: any update on this?

apidef/graphql_config_adapter.go
```diff
@@ -0,0 +1,280 @@
...
+)
+
+var ErrUnsupportedGraphQLConfigVersion = errors.New("provided version of GraphQL config is not supported for this operation")
+
+type GraphQLConfigAdapter struct {
+	config     GraphQLConfig
+	httpClient *http.Client
+}
+
+func NewGraphQLConfigAdapter(config GraphQLConfig) GraphQLConfigAdapter {
```
spetrunin, line 24: What do u think about rename adapter to smth like `GraphQLConfigV2Adapter` as this whole adapter is responsible for creating v2 engine config

in this way, we could omit v2 prefixes for method names

apidef/graphql_config_adapter.go
```diff
@@ -0,0 +1,280 @@
...
+	return GraphQLConfigAdapter{config: config}
+}
+
+func (g *GraphQLConfigAdapter) EngineConfigV2() (*graphql.EngineV2Configuration, error) {
+	if g.config.Version != GraphQLConfigVersion2 {
```
spetrunin, line 29: we could do this check in `NewGraphQLConfigAdapter` and we will know immediately is adapter capable to handle config

apidef/graphql_config_adapter.go
```diff
@@ -0,0 +1,280 @@
...
+func (g *GraphQLConfigAdapter) engineConfigV2FieldConfigs() (planFieldConfigs plan.FieldConfigurations) {
+	for _, fc := range g.config.Engine.FieldConfigs {
+		planFieldConfig := plan.FieldConfiguration{
+			TypeName:              fc.TypeName,
+			FieldName:             fc.FieldName,
+			DisableDefaultMapping: fc.DisableDefaultMapping,
+			Path:                  fc.Path,
+		}
+
+		planFieldConfigs = append(planFieldConfigs, planFieldConfig)
```
spetrunin, line 61: should we preallocate this array as we know its side in advance?

apidef/graphql_config_adapter.go
```diff
@@ -0,0 +1,280 @@
...
+			RootNodes: []plan.TypeField{},
+		}
+
+		for _, typeField := range ds.RootFields {
+			planTypeField := plan.TypeField{
+				TypeName:   typeField.Type,
+				FieldNames: typeField.Fields,
+			}
+
+			planDataSource.RootNodes = append(planDataSource.RootNodes, planTypeField)
```
spetrunin, line 79: `RootNodes` also could be preallocated

apidef/graphql_config_adapter.go
```diff
@@ -0,0 +1,280 @@
...
+			planDataSource.Custom = graphqlDataSource.ConfigJson(graphqlDataSource.Configuration{
+				Fetch: graphqlDataSource.FetchConfiguration{
+					URL:    graphqlConfig.URL,
+					Method: graphqlConfig.Method,
+					Header: g.convertHttpHeadersToEngineV2Headers(graphqlConfig.Headers),
+				},
+			})
+		}
+
+		planDataSources = append(planDataSources, planDataSource)
```
spetrunin, line 129: `planDataSources` also could be preallocated

apidef/graphql_config_adapter.go
```diff
@@ -0,0 +1,280 @@
...
+				TypeName:   typeField.Type,
+				FieldNames: typeField.Fields,
+			}
+
+			planDataSource.RootNodes = append(planDataSource.RootNodes, planTypeField)
+		}
+
+		switch ds.Kind {
+		case GraphQLEngineDataSourceKindREST:
+			restConfig := GraphQLEngineDataSourceConfigREST{}
```
spetrunin, line 84: `var` could be more readable

apidef/graphql_config_adapter.go
```diff
@@ -0,0 +1,280 @@
...
+
+	err = g.determineChildNodes(&planDataSources)
+	return planDataSources, err
+}
+
+func (g *GraphQLConfigAdapter) SetHttpClient(httpClient *http.Client) {
+	g.httpClient = httpClient
+}
+
+func (g *GraphQLConfigAdapter) convertURLQueriesToEngineV2Queries(apiDefQueries map[string]string) []restDataSource.QueryConfiguration {
```
spetrunin, line 140: Is it required to have `convertURLQueriesToEngineV2Queries` and `convertHttpHeadersToEngineV2Headers` helpers as method of adapter? Could it be just helper functions?

apidef/graphql_config_adapter.go
```diff
@@ -0,0 +1,280 @@
...
+
+	engineV2Headers := make(http.Header)
+	for apiDefHeaderKey, apiDefHeaderValue := range apiDefHeaders {
+		engineV2Headers[apiDefHeaderKey] = []string{apiDefHeaderValue}
+	}
+
+	return engineV2Headers
+}
+
+func (g *GraphQLConfigAdapter) determineChildNodes(planDataSources *[]plan.DataSourceConfiguration) error {
```
spetrunin, line 171: why we have to pass `planDataSources` as an slice pointer?

spetrunin: slice is already a pointer type so we could modify its elements

https://play.golang.org/p/4HnX020UUxF

did I miss anything?

spetrunin (changes requested):
have a couple of questions, will continue reviewing tomorrow

apidef/api_definitions.go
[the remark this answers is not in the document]
pvormste: we are going without the "v" prefix at it seems unnecessary

apidef/graphql_config_adapter.go
pvormste: The adapter adapts a single input (GraphQLConfig) into different outputs (currently only v2, but could be more in the future)

apidef/graphql_config_adapter.go
pvormste: this is similar to the previous comment: As an adapter is for generating different outputs it could be, that one output will need a `GraphQLConfigVersion3`for example - so the check should live outside `NewGraphQLConfigAdapter`

apidef/api_definitions.go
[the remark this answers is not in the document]
pvormste: now its fixed!

apidef/api_definitions.go
```diff
@@ -551,6 +553,62 @@ type GraphQLConfig struct {
...
 	// GraphQLPlayground is the Playground specific configuration.
 	GraphQLPlayground GraphQLPlayground `bson:"playground" json:"playground"`
+	// Engine holds the configuration for engine v2 and upwards.
+	Engine GraphQLEngineConfig `bson:"engine" json:"engine"`
+}
+
+type GraphQLConfigVersion string
+
+const (
+	GraphQLConfigVersionNone = ""
```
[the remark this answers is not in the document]
pvormste: yes, they should - fixed!

apidef/graphql_config_adapter.go
pvormste: As it makes only sense that they exist when the `GraphQLAdapter`exists, I think it is better to couple them with the adapter. If you delete the adapter, you also delete the helpers then. Makes sense?

apidef/graphql_config_adapter.go
pvormste: changed!

apidef/graphql_config_adapter.go
pvormste: You did miss: ...nothing. Pointer was not necessary here!

apidef/graphql_config_adapter.go
spetrunin: Maybe, in this case, it will be better to have an adapters package and different implementations for the different kinds of configs?

apidef/graphql_config_adapter.go
pvormste: it's not in the hot path, so I wouldn't bother with it, what do you think? Same for others

apidef/api_definitions.go
```diff
@@ -551,6 +553,62 @@ type GraphQLConfig struct {
...
+type GraphQLTypeFields struct {
+	Type   string   `bson:"type" json:"type"`
+	Fields []string `bson:"fields" json:"fields"`
+}
+
+type GraphQLEngineDataSourceConfigREST struct {
+	URL     string            `bson:"url" json:"url"`
+	Method  string            `bson:"method" json:"method"`
+	Headers map[string]string `bson:"headers" json:"headers"`
+	Query   map[string]string `bson:"query" json:"query"`
```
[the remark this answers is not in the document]
pvormste: Good catch - changed it

apidef/graphql_config_adapter.go
pvormste: moved into own package

furkansenharputlu (approved):
Approving for now, but we will move adapter to the graphql library. There are things that are not related to the gateway. We created a ticket to handle it in future.

spetrunin (changes requested):

spetrunin (approved):

pvormste:
manually merged to `release-3`

PR #3454, opened 2021-02-06
Go 1.5 migration

buger:
Go 1.4 introduced TLS 1.3 by default, which we can't do since a lot of our clients depend on 1.2.
In order to disable 1.3 Go added new `MaxVersion` option to `tls.Config` structure.
For example one of 1.3 breaking "features" is the inability to explicitly set TLS ciphers (in 1.3 they should be picked automatically), and a lot of enterprises depend on it.

In order to make it non-breaking change I added new max SSL version attributes to all needed sections (everywhere where there is min version), and it is set to 1.2 by default.

New config fields:
`http_server_options.max_version`
`proxy_ssl_max_version`
Api definition fields:
`proxy.transport.ssl_max_version`

Additionally, fixed a bunch of tests (not related to 1.5 change from what I see).

All the vendoring removed as well!

As part of QA we need to ensure that all outbound and inbound Gateway calls use TLS 1.2, which includes:
- [ ] Uptime tests
- [ ] JSVM HTTP calls
- [ ] Batch API
- [ ] Reverse proxy
- [ ] Gateway -> Dashboard communication
- [ ] Gateway -> MDCB communication
- [ ] UDG GraphQL and JSON data sources

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`


certs/manager.go
```diff
@@ -476,6 +473,9 @@ func (c *CertificateManager) ListAllIds(prefix string) (out []string) {
 		}
 
 		keys := c.storage.GetKeys("raw-" + prefix + "*")
+
+		fmt.Println("KEYS!!!!!!!", keys, prefix)
```
joshblakeley, line 477: Println needs removing

PR #3456, opened 2021-02-09
[TD-10] Goreleaser for more repos

alephnull:
# Description
These related PRs add a goreleaser based workflow for the following distribution channels
- Ubuntu and Debian
- RHEL/OL
- tarballs
- docker hub
- devenv ECR
- AWS mktplace
for the following repos
- tyk
- tyk-analytics
- tyk-pump
- tyk-sink

## Pending
- [x] tykio/golang-cross
- [x] go version needed for gw, dash, pump and sink
- [x] AMIs for tyk and pump

# Testing
The status of the integration workflows can be inspected in this PR itself in the _Checks_ tab.

# Merging
Once this PR is merged, the following will be deprecated
- Buddy actions that build packages
- Buddy actions that build AMIs
- int-image.yml


install/post_install.sh
```diff
@@ -5,7 +5,7 @@ SYSTEMD="/lib/systemd/system"
 UPSTART="/etc/init"
 SYSV1="/etc/init.d"
 SYSV2="/etc/rc.d/init.d/"
-DIR="/opt/tyk-gateway/install"
+DIR="/opt/tyk/install"
```
excieve, line 8: If not required let's avoid install directory changes please as it might potentially break existing users/cause confusion.

Dockerfile.std
```diff
@@ -0,0 +1,21 @@
+# Generated by: tyk-ci/wf-gen
+# Generated on: Wed 17 Feb 17:51:15 UTC 2021
+
+# Generation commands:
+# ./pr.zsh -title use nfpm for std docker #2 -branch goreleaser/more -p
+# m4 -E -DxREPO=tyk
+
+FROM debian:buster-slim
+RUN apt-get update \
```
excieve, line 9: Doesn't this need Python and friends as in the existing image?

Dockerfile.std
```diff
@@ -0,0 +1,21 @@
+# Generated by: tyk-ci/wf-gen
+# Generated on: Wed 17 Feb 17:51:15 UTC 2021
+
+# Generation commands:
+# ./pr.zsh -title use nfpm for std docker #2 -branch goreleaser/more -p
+# m4 -E -DxREPO=tyk
+
+FROM debian:buster-slim
```
excieve, line 8: Could you please use `buster-20210208-slim` specifically to make sure `buster-slim` isn't cached at an earlier revision? Same for the rest of the images.

excieve (commented):
Looks great @alephnull! Just a concern about the installation directory change for the packages and Python in the standard container image.

Dockerfile.std
alephnull: Thanks!

excieve (approved):
All good! Directory structure and compilation changes will need a good QA pass as discussed, but happy to merge now and address those as they come up. Very excited to bring this change to our release process!

PR #3461, opened 2021-02-18
Remove idle subrouter from ChainObject

furkansenharputlu:
This PR removes the following idle fields from `ChainObject`:
- `Domain`
- `Index`
- `Subrouter`

They are causing confusion while reading the code.


buger:
Looks good!

PR #3462, opened 2021-02-18
[TT-657] Move CORS handler from chain to middleware to cover oauth2 endpoints

furkansenharputlu:
The reason for the move is that the oauth2 handlers don't have a chain in front of it but CORS is in the chain and it is the reason of why oauth endpoints don't support CORS. This PR moves it to routers middleware level to cover all endpoints in an API router.


buger (approved):
Looks solid!

spetrunin (approved):

PR #3467, opened 2021-02-19
Migrate to github actions for CI tests

furkansenharputlu:


.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: actions/setup-go@v1
+        with:
+          go-version: ${{ matrix.go }}
+
+      - name: start redis
+        uses: supercharge/redis-github-action@1.1.0
+        with:
+          redis-version: ${{ matrix.redis-version }}
+
+      - uses: actions/cache@v1
```
furkansenharputlu, line 36: question 1

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,52 @@
...
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.mod') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - name: go ${{ matrix.go }} tests
+        run: |
+          ../../ci-tests.sh
+
+      - name: upload coverage
```
furkansenharputlu, line 47: question 2

buger:
@alephnull can you review pls?

furkansenharputlu:
@buger @alephnull it is not ready for reviewing yet. I will complete it in this week.

furkansenharputlu:
Lots of stuff came in but I will finish in this sprint :)

furkansenharputlu:
@buger Why doesn't it trigger Github actions? It was triggering previously.

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,69 @@
...
+  TYK_DB_REDISHOST: localhost
+  TYK_GW_STORAGE_HOST: localhost
+
+jobs:
+  test:
+    name: Go ${{ matrix.go }} Redis ${{ matrix.redis-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
```
alephnull, line 15: I think it makes sense to run this on a range of mongo, redis and go versions. You can pick the currently supported versions as the required check in the branch protection rules.

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,69 @@
...
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.mod') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - name: Running Gateway Tests
+        run: |
+          go build -tags 'coprocess lua'
+          ./bin/ci-tests.sh
+
+      - name: Upload Coverage
```
alephnull, line 64: Can drop the upload as the UI doesn't provide a place to show this file. For coverage reports, see https://coveralls.io.

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,69 @@
...
+        uses: actions/cache@v1
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.mod') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - name: Running Gateway Tests
+        run: |
+          go build -tags 'coprocess lua'
```
alephnull, line 61: Is a build required for tests? There is already automation that tests two kinds of builds (goreleaser based and direct go cmd line) for all PRs and pushes to release branches.

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,69 @@
...
+          pip install google
+          pip install protobuf
+
+      - name: Start Redis
+        uses: supercharge/redis-github-action@1.1.0
+        with:
+          redis-version: ${{ matrix.redis-version }}
+
+      - name: Cache
+        uses: actions/cache@v1
```
alephnull, line 52: If the cache isn't working, try actions/cache@v2. Documentation at https://github.com/actions/cache
Also, go.mod is for the build, right? It will not help you cache test dependencies and at worst, it might do the wrong thing.

alephnull (changes requested):
Some minor points. Only blocker is the upload which is not useful and adds to our github charges.

It looks like it is catching test failures, this should help a lot when merged. https://github.com/TykTechnologies/tyk/runs/2126207908?check_suite_focus=true

.github/workflows/ci-tests.yml
furkansenharputlu: I have changed to this one: https://github.com/actions/cache/blob/main/examples.md#go---modules

.github/workflows/ci-tests.yml
furkansenharputlu: For the first iteration, we can go with it I think. If we want in future, we can extend it.

alephnull (approved):
Thinking about it, I don't see why this needs to be a separate workflow. Ideally, it should be part of the [build workflow](https://github.com/TykTechnologies/tyk/actions/runs/665110339) as a pre-requisite.
Adding this means adding it to the template at https://github.com/TykTechnologies/tyk-ci/blob/master/wf-gen/release.yml.m4. I am happy to help you do it or please log a task in the devops Jira to have this available in tyk, tyk-analytics, tyk-pump and tyk-sink. It would be great if all the tests can be run in the same way. Like `go test ./..` for example.

PR #3480, opened 2021-03-01
[TT-1461] Call letsencrypt.Manager.Watch once

furkansenharputlu:
The `Watch` function is called again and again. This PR calls it once and gets the channel.


buger:
👍🏻

PR #3484, opened 2021-03-04
[TT-1606] Separated analytics storage

tbuchaillot:
## Description

Added configuration options `enable_separate_analytics_store` and `analytics_store` to have a separated analytics storage.
Codewise, it works in the same way as `cache_storage` and `enable_separate_cache`.

**DOCS WIP**

## Related Issue

https://tyktech.atlassian.net/browse/TT-1606

## Motivation and Context

If Redis runs out of memory during a high workload (where analytics would probably have a big part on it), it will make the entire Tyk cluster unable to continue functioning.
To minimize the impact in those situations, the idea is to add an option to have a separated Redis / Redis cluster for analytics.

## How This Has Been Tested

- Checked if the whole analytics flow (gw - request - pump receive the record) is still working without any modification.
- Checked if we set `enable_separate_analytics_store:true` and configure `analytics_store` to another Redis, everything should work on the gateway side but the pump shouldn't process anything. If we change the pump config to this new Redis, it should purge all the records.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/storage/redis_cluster.go
+++ b/storage/redis_cluster.go
@@ -149,7 +149,7 @@ func ConnectToRedis(ctx context.Context, onConnect func()) {
 	tick := time.NewTicker(time.Second)
 	defer tick.Stop()
 	c := []RedisCluster{
-		{}, {IsCache: true},{IsAnalytics: true},
+		{}, {IsCache: true}, {IsAnalytics: true},
 	}
 	var ok bool
 	for _, v := range c {
--- a/storage/redis_cluster.go
+++ b/storage/redis_cluster.go
@@ -149,10 +149,7 @@ func ConnectToRedis(ctx context.Context, onConnect func()) {
 	tick := time.NewTicker(time.Second)
 	defer tick.Stop()
 	c := []RedisCluster{
-		{}, {IsCache: true},
-	}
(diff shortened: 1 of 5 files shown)


storage/redis_cluster.go
```diff
@@ -80,44 +82,55 @@ func WaitConnect(ctx context.Context) bool {
 	}
 }
 
-func singleton(cache bool) redis.UniversalClient {
+func singleton(cache, analytics bool) redis.UniversalClient {
```
buger, line 85: Are you sure we need boolean flag here?
How separate cache storage works then? Does it mean it currently broken?

tbuchaillot: @buger what do you mean? the cache flag is also passed to singleton func.

buger: Damn not sure how I read that. Indeed.

buger:
Tests are failing

tbuchaillot:
> Tests are failing

solved @buger

sredxny (approved):
LGTM

matiasinsaurralde (approved):
Looks good

PR #3486, opened 2021-03-10
[TT-1698] Work with subrouters inside processSpec

furkansenharputlu:
The master router is passed to the `processSpec` and while adding handlers to it always needs to put listen path in front. I realized that it is more than a redundancy. For example, if I want to add CORS middleware to subrouter, it wasn't matching with the listen path because they were living in the master router. This PR gets subrouter first and passes it to the `processSpec` function.

- I saw that there is a problem in graphql playground. So I created a PR to that repo as well: https://github.com/TykTechnologies/graphql-go-tools/pull/100

I created this PR as draft, after merging graphql PR, I will update `go.mod` and this PR will be ready.

### DIFF
--- a/gateway/api_loader.go
+++ b/gateway/api_loader.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"path"
 	"path/filepath"
 	"runtime/debug"
 	"sort"
@@ -25,11 +26,13 @@ import (
 	"github.com/TykTechnologies/tyk/trace"
 )
 
+const (
+	rateLimitEndpoint = "/tyk/rate-limits/"
+)
+
 type ChainObject struct {
-	ListenOn       string
 	ThisHandler    http.Handler
 	RateLimitChain http.Handler
-	RateLimitPath  string
 	Open           bool
 	Skip           bool
 }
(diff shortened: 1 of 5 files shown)


buger:
I see playground tests failing.
Is it because you need to update to latest mod version of graphql-go-tools?
Also its PR was merged, lets update and remove draft state.

furkansenharputlu:
👍

gateway/api_loader.go
```diff
@@ -679,7 +680,7 @@ func loadHTTPService(spec *APISpec, apisByListen map[string]int, gs *generalStor
 		router.Handle(chainObj.RateLimitPath, chainObj.RateLimitChain)
 	}
 
-	router.Handle(chainObj.ListenOn, chainObj.ThisHandler)
+	subrouter.NewRoute().Handler(chainObj.ThisHandler)
```
buger, line 683: Is it equal to `subrouter.Handle("/", chainObj.ThisHandler)`? 
And if not, what is the difference?

furkansenharputlu: It is not equal. The difference is `/` only matches `/`. However, the `NewRoute()` matches anything.
With `/` you will get `Not Found` for a simple httpbin get operation.

buger:
@furkansenharputlu some branches require manual merge

furkansenharputlu:
I will merge manually. Thanks!

furkansenharputlu:
merged manually to:
- `release-3.0.5`
- `release-3-lts`
- `release-3.1`

PR #3487, opened 2021-03-10
[TT-1574] Multiple analytics keys in slaves gateway

tbuchaillot:
## Description

Extends https://github.com/TykTechnologies/tyk/pull/3483
Modified the logic for rpc_analytics_purger so it can read from keys from multiple keys. It's coded in full backward-compatible mode, which means that it's going to look for `tyk-system-analytics` key and then for `tyk-system-analytics_X `where X goes from 0 to 9.

After reading, it will send the analytics records to sink through RPC. If tyk-sink has enable_multiple_analytics_keys = true, it will write the records in multiple keys and if it's false, it will write them in only one key.

## Related Issue

https://tyktech.atlassian.net/browse/TT-1574
## Motivation and Context

Reduce Redis overload and have a better analytics distribution across the cluster for slave gateways.

## Types of changes

## Checklist

- [] My change requires a change to the documentation.
- [X] All new and existing tests passed.
  - [X] `go fmt -s`
  - [X] `go vet`

### DIFF
--- a/rpc/rpc_analytics_purger.go
+++ b/rpc/rpc_analytics_purger.go
@@ -3,6 +3,7 @@ package rpc
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"time"
 
 	msgpack "gopkg.in/vmihailenco/msgpack.v2"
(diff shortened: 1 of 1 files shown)


sredxny:
@tbuchaillot  would be helpful to have information about how was this tested

PR #3488, opened 2021-03-11
Do not support feature/ branches for go.mod update in tyk-analytics

alephnull:
It does not appear to be used much and the errors on Slack are annoying.

This means that if you have a branch feature/something in this repo, the corresponding feature/something branch in tyk-analytics will _not_ have its go mod updated.


alephnull:
CI failing due to
```
 FAIL: TestGraphQL_HeadersInjection (1.15s)
```
which is unrelated to these changes.

excieve (approved):

PR #3489, opened 2021-03-11
[tt-1924] Go plugins per method and path

joshblakeley:
## Description

Allows a Tyk user to run a different Go function in every single possible part of their API if they should want to. Can be a high performance replacement for the JSVM virtual endpoints.

Think of it as having either as many native Go lambda style functions inside the Gateway available to you as you want, or if you want, unlike existing JS virtual endpoints, requests pass through to real upstreams and you just get the extra flexibility of this setup.

## Related Issue

Internal ticket viewable here https://tyktech.atlassian.net/browse/TT-1924

## Motivation and Context

Our virtual endpoints are really well loved by users but have some issues with high performance scenarios. Go plugins are the most native way to inject extra code into the API Gateway. Previous there were set hooks that a plugin function could sit and every request would route through it.

This PR makes it so that the Go plugins can be loaded per path and method using the usual API designer definitions.

Users can utilise this in a few different ways:

- A single Go plugin file shared object with multiple handlers that can be called from various APIs
- Many shared objects each referenced from different plugins.

Both cases are handled and users can mix and match depending on their complexity and code ownership.

## How This Has Been Tested

See the /test/goplugin folder for function examples.

In your API definition set up your per path/method goplugin functions as an array of config blocks just like any other API designer middleware.

Functions can all be in a single file or be pulled in from various files.

The same rules apply as with the existing go plugins and we recommend using the Go Plugin compiler container to help you create the shared objects that are guaranteed to work with a particular gateway version.

```
... 
"extended_paths":  {
            "go_plugin: [
                  {
                "plugin_path": "../test/goplugins/goplugins.so",
                "path": "/get",
                "method": "GET",
                "func_name": "MyPluginPerPathFoo"
              }
             ]
          }
...
```

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


gateway/mw_go_plugin.go
```diff
@@ -108,11 +128,29 @@ func (m *GoPluginMiddleware) EnabledForSpec() bool {
...
-
 	return true
 }
 
+func (m *GoPluginMiddleware) goPluginConfigFromRequest(r *http.Request) {
+	_, versionPaths, _, _ := m.Spec.Version(r)
+
+	found, perPathPerMethodGoPlugin := m.Spec.CheckSpecMatchesStatus(r, versionPaths, GoPlugin)
+	if found {
+		//fmt.Printf("%#v", m)
```
buger, line 139: remove debugging

gateway/mw_go_plugin.go
```diff
@@ -108,11 +128,29 @@ func (m *GoPluginMiddleware) EnabledForSpec() bool {
...
+		m.Meta = perPathPerMethodGoPlugin.(*GoPluginMiddleware).Meta
+		m.Path = perPathPerMethodGoPlugin.(*GoPluginMiddleware).Path
+		m.SymbolName = perPathPerMethodGoPlugin.(*GoPluginMiddleware).SymbolName
+		m.logger = perPathPerMethodGoPlugin.(*GoPluginMiddleware).logger
+	}
+}
 func (m *GoPluginMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Request, conf interface{}) (err error, respCode int) {
+
+	// is there a go plugin per path - we copy the handler etc from the urlspec if we find one
+	m.goPluginConfigFromRequest(r)
```
buger, line 152: I'm not sure if it is correct behavior. 
If you have both API level plugin, and endpoint plugin, it will ignore global plugin with this approach. 
I would rather attach this middleware two times to the request pipeline, with some parameter. One will run for global plugins, one for per-endpoint.

joshblakeley: This works because they are attached at different places in the pipeline like you say. This function is a noop for per api go plugins as it simply looks for the gplugin config in the extended paths part of the API definition and then copies it over when found

joshblakeley: All tests for global plugins still pass for example

joshblakeley: Actually i see what you mean I think - if we have both per path and per api on same definition only the per path will be executed

joshblakeley: I'll add a unit test for it

buger: > Actually i see what you mean I think - if we have both per path and per api on same definition only the per path will be executed

Yes this one. 
For example in the case of JSVM and Virtual Endpoint, they work in parallel. And VP is just a separate middleware, which has set of shared functions to avoid code duplication.

sredxny (approved):
LGTM! amazing job

PR #3496, opened 2021-03-17
[TT-1885] Fix grpc->h2c and h2c->grpc communication

buger:
For `grpc->h2c` you need to have API protocol set to https (or default) and targetURL to `htc://`
For `h2c->grpc` you need to have API protocol set to http and targetURL to https.

## Related Issue
Fix https://tyktech.atlassian.net/browse/TT-1885

## Motivation and Context
This is quite common flow when you put Tyk in k8s, and it handles TLS but with inner world it will talk with http/h2c.

## How This Has Been Tested
No tests added yet, only manual testing.

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`


gateway/api_loader.go
```diff
@@ -121,10 +121,6 @@ func processSpec(spec *APISpec, apisByListen map[string]int,
 		"prefix": "coprocess",
 	})
 
-	if strings.Contains(spec.Proxy.TargetURL, "h2c://") {
```
buger, line 124: This check moved to reverse proxy code https://github.com/TykTechnologies/tyk/pull/3496/files#diff-e6e07722257f7e41691e471185ad6d84fd56dc9e5459526ea32e9a5e8fa1a01bR1052

Otherwise, inside reverse proxy code while creating the transport we will not know that upstream is h2c.
Plus it does not work with load balancing, or service discovery, so this issue automatically got solved too.

gateway/api_loader.go
```diff
@@ -782,9 +778,6 @@ func loadApps(specs []*APISpec) {
 	shouldTrace := trace.IsEnabled()
 	for _, spec := range specs {
 		func() {
-			if strings.Contains(spec.Proxy.TargetURL, "h2c://") {
-				spec.Protocol = "h2c"
```
buger, line 786: h2c protocol was deprecated from the previous change (http protocol automatically handle it). And goal is that User->Tyk logic should not depend on Tyk->Upstream logic.

gateway/reverse_proxy.go
```diff
@@ -679,7 +681,10 @@ func httpTransport(timeOut float64, rw http.ResponseWriter, req *http.Request, p
 		http2.ConfigureTransport(transport)
 	}
 
-	if p.TykAPISpec.Protocol == "h2c" {
+	p.logger.Debug("Out request url: ", outReq.URL.String())
+
+	if outReq.URL.Scheme == "h2c" {
```
buger, line 686: On transport create we just check if our target is h2c. 
As mentioned above it now works with load balancing too.

gateway/reverse_proxy.go
```diff
@@ -924,28 +929,6 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 	}
 	var roundTripper *TykRoundTripper
 
-	p.TykAPISpec.Lock()
-
-	// create HTTP transport
-	createTransport := p.TykAPISpec.HTTPTransport == nil
```
buger, line 930: This part just moved below, so we will have access to outbound request.

gateway/reverse_proxy.go
```diff
@@ -1044,11 +1027,32 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
...
+	createTransport := p.TykAPISpec.HTTPTransport == nil
+
+	// Check if timeouts are set for this endpoint
+	if !createTransport && config.Global().MaxConnTime != 0 {
+		createTransport = time.Since(p.TykAPISpec.HTTPTransportCreated) > time.Duration(config.Global().MaxConnTime)*time.Second
+	}
+
+	if createTransport {
+		_, timeout := p.CheckHardTimeoutEnforced(p.TykAPISpec, req)
+		p.TykAPISpec.HTTPTransport = httpTransport(timeout, rw, req, outreq, p)
```
buger, line 1041: The only change to this code that we pass new `outreq` param to the function.

gateway/reverse_proxy.go
```diff
@@ -1044,11 +1027,32 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
...
+	}
+
+	roundTripper = p.TykAPISpec.HTTPTransport
+
 	if roundTripper.transport != nil {
 		roundTripper.transport.TLSClientConfig.Certificates = tlsCertificates
 	}
 	p.TykAPISpec.Unlock()
 
+	if outreq.URL.Scheme == "h2c" {
```
buger, line 1052: We still need to replace it with proper protocol which http stack will undertand.

sredxny (approved):
LGTM, in the near future I will cover this use case in unit testing (different PR probably)

sredxny:
Added Unit tes for the use case: User-> Tyk in HTTPS and Tyk->Upstream via H2c

`How was this tested?`
As this PR basically separates the protocols logic of user -> tyk and then tyk->upstream then the main focus was on this. The manual test was as:

- Ran the same tests in https://github.com/TykTechnologies/tyk/pull/3470 to check that nothing is broken
- Ran tyk GW in HTTPS (port 8081)
- Ran the Greeter gRPC  service to use it as upstream
- Register one gRPC service like this:
      - Listen Path: /helloworld.Greeter/
      - Strip listen path: false
      -  Target URL: h2c://localhost:50052
- Consume service via tyk with: `grpcurl --insecure -proto helloworld.proto -d ‘{“name”: “sredny"}’ localhost:8081 helloworld.Greeter/SayHello` and I got as answer:
```
{
  "message": "Hello sredny"
}
```

gateway/grpc_test.go
```diff
@@ -31,26 +31,31 @@ import (
 
 // For gRPC, we should be sure that HTTP/2 works with Tyk in H2C configuration also for insecure grpc over http.
 func TestHTTP2_h2C(t *testing.T) {
+
 	defer ResetTestConfig()
 	var port = 6666
 
-	EnablePort(port, "h2c")
+	EnablePort(port, "http")
```
sredxny, line 38: changed as the protocol `user->tyk` is not necessarily the same as `tyk->upstream`. Users that already were using this feature need to be aware of this change

buger:
Nice!

tbuchaillot (approved):
LGTM 🚀

sredxny:
manually cherry-picked into release-3-lts

buger:
@sredxny do not forget release-3

PR #3505, opened 2021-03-25
[TT-1542] Add fallback hash key functions to catch function changes

furkansenharputlu:
Added new `hash_key_function_fallback` array option to specify list of hash function which Gateway needs to check if key with default hash key function not found. 

This functionality needed only if you change default hashing algorithm, but your token information does not hold information about hashing function like custom keys, certificates, or basic auth tokens.

### DIFF
--- a/gateway/mw_basic_auth.go
+++ b/gateway/mw_basic_auth.go
@@ -178,7 +178,7 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
 	}
 
 	// Check if API key valid
-	keyName := generateToken(k.Spec.OrgID, username)
+	keyName := username
 	logger := k.Logger().WithField("key", obfuscateKey(keyName))
 	session, keyExists := k.CheckSessionAndIdentityForValidKey(&keyName, r)
 	if !keyExists {
--- a/cli/linter/schema.json
+++ b/cli/linter/schema.json
@@ -433,6 +433,22 @@
         "sha256"
       ]
     },
+    "hash_key_function_fallback": {
+      "type": [
+        "array",
+        "null"
+      ],
+      "items": {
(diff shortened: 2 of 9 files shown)


buger:
Having this code in mw_auth_token is not enough. 
Keys API will still be broken, as well as Dashboard keys screens. 
I think this logic should be embedded to `SessionDetail` function.

furkansenharputlu:
Is there any key API where we search by cert?

buger:
I do not think that only certificates affected. 
I think it is issue with custom keys and basic auth too. 

But needs be confirmed.

furkansenharputlu:
I will write a test for that scenario.

gateway/mw_basic_auth.go
```diff
@@ -178,7 +178,7 @@ func (k *BasicAuthKeyIsValid) ProcessRequest(w http.ResponseWriter, r *http.Requ
 	}
 
 	// Check if API key valid
-	keyName := generateToken(k.Spec.OrgID, username)
+	keyName := username
```
furkansenharputlu, line 181: @buger As hashed part is shrunk, collisions occur and the `TestBasicAuthCachedUserCollision` is failing. I think it is expected.

furkansenharputlu: A possible solution is to double the key to be cached:
```
	cacheKey := key
	if t.Spec.GlobalConfig.HashKeys {
		cacheKey = storage.HashStr(key `+key`)
	}
```

furkansenharputlu:
Merged manually to:
- `release-2.9`
- `release-3`
- `release-3-lts`
- `release-3.1`
- `release-3.2`

ilijabojanovic:
@buger @furkansenharputlu 

We need to merge this fix on release-3.0.5 branch

Issue #3511, opened 2021-03-30
[TT-2216] - Suport passing custom signature in a query parameter


gmaiztegi:
**Is your feature request related to a problem? Please describe.**
Tyk supports [custom signature validation](https://github.com/TykTechnologies/tyk/pull/2171), which is great, but I am currently managing a migration from Mashery that puts the signature in a query parameter, which is currently not supported.

**Describe the solution you'd like**
I would like to be able to pass the signature in a query parameter, just like we can pass authentication tokens there.

**Describe alternatives you've considered**
Not really.

**Additional context**
Working on a PR right now.

n2linux:
@gmaiztegi Thanks so much for creating this issue and working on a PR! 🥇 

We will get someone assigned to review everything and will be in touch.

PR #3512, opened 2021-03-30
[TT-2216] - Add support for custom signature in query parameters

gmaiztegi:
## Description
This PR adds `use_param` and `param` to the signature configuration, in order to be able to pass the signature as a query parameter (I would appreciate some help to know how to correctly document these new fields). These parameters behave very similar to those regarding the token.

## Related Issue

https://github.com/TykTechnologies/tyk/issues/3511

## Motivation and Context

https://github.com/TykTechnologies/tyk/commit/bde50bceb33dc7c98ff47cdf0d1f6f33dd240720 added support for a Mashery-style signature validation, but it only supports getting the signature from a header. Mashery supports passing the signature in a query parameter, so it would be great to support this feature and mimic this behaviour.

## How This Has Been Tested

I have to say I'm super new to the Golang community so I tested this only with the test added to this PR, so I would appreciate some feedback while I build the Docker image and test it inside Kubernetes.

## Types of changes

## Checklist

- [X] My change requires a change to the documentation.
  - [X] If you've changed APIs, describe what needs to be updated in the documentation.
  - [X] `go fmt -s`
  - [X] `go vet`


christtyk:
@buger can you review pls :pray:

apidef/api_definitions.go
```diff
@@ -477,6 +477,8 @@ type AuthConfig struct {
 type SignatureConfig struct {
 	Algorithm        string `mapstructure:"algorithm" bson:"algorithm" json:"algorithm"`
 	Header           string `mapstructure:"header" bson:"header" json:"header"`
+	UseParam         bool   `mapstructure:"use_param" bson:"use_param" json:"use_param"`
+	Param            string `mapstructure:"param" bson:"param" json:"param"`
```
buger, line 481: Lets make it consistent with another config options we have https://github.com/TykTechnologies/tyk/blob/76103a37c1a082c167d085bfee5d8fdde378d0ab/apidef/api_definitions.go#L468

Just name it ParamName  `param_name`.

buger (changes requested):

buger:
Added a small comment, otherwise looks solid! 
@chris-fewtrell pls help create Jira ticket and assign to next release.

apidef/api_definitions.go
gmaiztegi: As the header parameter is just `header` and not `header_name`, I was following that naming. But you are right to make it equal to the `param_name`.

furkansenharputlu (approved):

christtyk:
hey @gmaiztegi - this will probably get pushed out in 3.3 - will let you know. Want to drop me your address, we'd like to ping you some swag for the contribution! chris@tyk.io

buger:
Merged via https://github.com/TykTechnologies/tyk/pull/3662 (essentially it is your change, with fixed conflicts).

PR #3513, opened 2021-03-30
[TT-1942] mw_jwt: implement method to extract OAuth client ID from JWT tokens

matiasinsaurralde:
## Description

This PR modifies the JWT middleware to be able to extract OAuth Client ID from JWT tokens.
Custom claims are used by identity providers, this includes common claim names for IDPs that are fully tested in our DCR flow.

Once the OAuth Client ID is extracted it's added to the session object, this ensures that a proper analytics record is generated.

In the future we could take two approaches: (i) keep extending the list for other common claim names or (ii) make this configurable.

## How This Has Been Tested
- Set a DCR enabled API.
- Register a client through the developer portal.
- Generate a token using the IDP token endpoint.
- Send a request to Tyk.
- Explore the log browser, the analytics record should contain the OAuth Client ID.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/mw_jwt.go
+++ b/gateway/mw_jwt.go
@@ -16,11 +16,13 @@ import (
 	"time"
 
 	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/lonelycode/osin"
 	cache "github.com/pmylund/go-cache"
 	jose "github.com/square/go-jose"
 
 	"github.com/TykTechnologies/tyk/apidef"
 	"github.com/TykTechnologies/tyk/config"
+	"github.com/TykTechnologies/tyk/storage"
 	"github.com/TykTechnologies/tyk/user"
 )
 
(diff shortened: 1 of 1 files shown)


gateway/mw_jwt.go
```diff
@@ -547,6 +568,7 @@ func (k *JWTMiddleware) processCentralisedJWT(r *http.Request, token *jwt.Token)
 		}
 	}
 
+	session.OauthClientID = oauthClientID
```
buger, line 571: There is really easy final step we can do:
```
if client, err := k.Spec.OAuthManager.OsinServer.Storage.GetClient(session.OauthClientID); err == nil {
  if developer_id, found := client.MetaData['developer_id']; found {
    session.DeveloperId = developer_id
  }
}
```

sredxny (approved):
So far LGTM

tbuchaillot (commented):
+1 to @buger comments but overall LGTM. If we can add some unit tests to this it would be great.

gateway/mw_jwt.go
matiasinsaurralde: I've been able to get this working and I'm guessing the goal is to make it available inside the session metadata object.
I don't think this change would benefit the analytics side for now ([PR](https://github.com/TykTechnologies/tyk-analytics/pull/2245)) as we don't use this value on the analytics record itself, only OAuth client ID is used.

matiasinsaurralde: Solved

matiasinsaurralde:
Added a commit that properly handles the developer ID injection into session metadata in order to enable the logic that's implemented here:
https://github.com/TykTechnologies/tyk/blob/master/gateway/handler_success.go#L122

As expected the analytics record contains the developer ID:
Captura de Pantalla 2021-04-08 a la(s) 02 46 02

Issue #3516, opened 2021-04-05
plugin was built with a different version of package github.com/TykTechnologies/tyk/goplugin


zj611:
**errors in building goplogin**
i used the docker to build the goplugin, and then volume it to the tyk container. When run the container, the error occurs, as show below.

`"Could not load Go-plugin" error="plugin.Open("./middleware/bundles/plugins"): plugin was built with a different version of package github.com/TykTechnologies/tyk/goplugin" mwPath="./middleware/bundles/plugins.so" mwSymbolName=MyPluginHmacCheck"`

The building image and tyk image i used are ‘**tykio/tyk-plugin-compiler:v3.0.4**’ and '**tykio/tyk-gateway:v3.0.4**', respectively. Both are downloaded from the dockerhub officially.

My question is, is the version of package github.com/TykTechnologies/tyk/goplugin of **tykio/tyk-plugin-compiler:v3.0.4** different from that of **tykio/tyk-gateway:v3.0.4**？

zj611:
in other way of questioning,  as for custom plugin building, can i use the func from the  package of github.com/TykTechnologies/tyk/goplugin?

alephnull:
Hey @zj611 thanks for reporting this! Could you attach a log of how you are building the plugin with the commands used and the output from the commands? Also, is it possible to share your plugin code?
Did you have this plugin working in any version of the gateway?

zj611:
@alephnull Thanks for your reply!
In our teamwork, we plan to build a custom plugin(hmac or rsa) using GO. The problem we come across  for  plugin building is how to load secret key, such as hmac secret, from in-memory cache or redis DB. 

Firstly, we tried to use the "BaseMiddleware" in custom plugin code,  as show below. 
`type MyHTTPSignatureValidationMiddlewareHmac struct {
	BaseMiddleware
	lowercasePattern *regexp.Regexp
}`

The target we did this is to imitate the "mw_http_signature_validation.go", located in gateway package, and load the session from in-memory cache or redis DB. However, the errors occur like this.


`tyk-gateway_1  | time="Apr 08 10:08:55" level=error msg="Could not load Go-plugin" error="plugin.Open("/opt/tyk-gateway/middleware/bundles/plugins"): plugin was built with a different version of package github.com/TykTechnologies/tyk/goplugin" mwPath="/opt/tyk-gateway/middleware/bundles/plugins.so" mwSymbolName=MyPluginAuthCheck
[...]

PR #3522, opened 2021-04-08
releng: add test infra for goplugins

alephnull:
Fetches things from Hub, suitable for testing -rc tags.

## Motivation and Context
Goplugins depend on many things being in alignment.

## How This Has Been Tested
Requires a functional docker-compose.

Run it as
```shell
% ./test.zsh v2.9.5
....
```

Look for `msg="API Loaded" api_id= api_name="Goplugin test"` in the output. Control-C will stop everything.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/images/README.md
+++ b/images/README.md
@@ -1,25 +1,9 @@
-# build-env
-
-Docker environment used to build official images.
-
-This used be here but now lives at the repo attached to
-[tyk-build-env](https://hub.docker.com/r/tykio/tyk-build-env)
-
 # plugin-compiler
 
-The usecase is that you have a plugin (probably Go) that you require
-to be built.
-
-Navigate to where your plugin is and build using a docker volume to
-mount your code into the image. Since the vendor directory needs to be
-identical between the gateway build and the plugin build, this means
-that you should pull the version of this image corresponding to the
-gateway version you are using.
-
-This also implies that if your plugin has vendored modules that are
-[also used by Tyk
+Does not support go modules. If your plugin has vendored modules that
+are [also used by Tyk
 gateway](https://github.com/TykTechnologies/tyk/tree/master/vendor)
-then your module will be overridden by the version that Tyk uses. 
+then your module will be overridden by the version that Tyk uses.
 
 ``` shell
 cd ${GOPATH}/src/tyk-plugin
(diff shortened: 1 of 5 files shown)


alephnull:
Fixes #3516

images/plugin-compiler/testplugin/main.go
```diff
@@ -0,0 +1,239 @@
...
+	Headers   string `json:"body" msg:"body"`
+	Domain   string `json:"domain" msg:"domain"`
+	Resource   string `json:"resource" msg:"resource"`
+
+}
+func main() {}
+
+
+/*
+func CustomAuthPluginOAuthOkta(rw http.ResponseWriter, r *http.Request)  () {
```
buger, line 150: Feels like we can remove this dead code?

buger:
Will be great if instead of checking log output, it can run curl query, and check the response output (e.g. plugin can add some header  stuff inside, and we actually validate the full pipeline).

images/plugin-compiler/testplugin/main.go
letzya: it's not even okta, it looks like auth0

alephnull:
@buger Added some docs for end-to-end testing with curl. Lint errors fixed.

images/README.md
```diff
@@ -1,25 +1,9 @@
...
-that you should pull the version of this image corresponding to the
-gateway version you are using.
-
-This also implies that if your plugin has vendored modules that are
-[also used by Tyk
+Does not support go modules. If your plugin has vendored modules that
+are [also used by Tyk
 gateway](https://github.com/TykTechnologies/tyk/tree/master/vendor)
-then your module will be overridden by the version that Tyk uses. 
+then your module will be overridden by the version that Tyk uses.
```
excieve, line 6: This repo doesn't have a `vendor` dir anymore and uses Go modules exclusively :see_no_evil:

images/plugin-compiler/test.zsh
```diff
@@ -0,0 +1,25 @@
...
+TRAPINT() {
+    print Tearing down test env
+    docker-compose -f test.yml down
+}
+
+[[ -z $1 ]] && usage $0
+export tag=$1
+
+rm -v testplugin/*.so
+podman run --rm -v `pwd`/testplugin:/plugin-source tykio/tyk-plugin-compiler:${tag} testplugin.so
```
excieve, line 24: What if there's no `podman` on whichever machine this is executed on (e.g. macOS)? In fact `docker` is probably better as `podman-docker` compat packages in various distros would normally handle the aliasing.

images/plugin-compiler/test.zsh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env zsh
```
excieve, line 1: Can it be just bash for better compatibility?

excieve (commented):
Very nice! I'd suggest replacing podman with docker though for wider compatibility.

images/plugin-compiler/test.zsh
alephnull: tbh, I don't know how to implement `TRAPINT()` in bash. The script is just sugar over two straightforward commands.

gwotg: If it helps, think this will trap Enter, SIGINT or SIGTERM
`
trap '' EXIT
echo ' --- press ENTER to close --- '
read response
`

alephnull:
@excieve With the change in the documentation, is this PR mergeable?

excieve (approved):
Thanks @alephnull, looks great!

vesko-tyk (approved):

PR #3527, opened 2021-04-10
[CI] Post failures on PRs

alephnull:
Reduced the verbosity of ci-tests.sh so that output is compact.

The failure comment is from a flaky test that sometimes fails.


jlucktay (approved):
👏

excieve (approved):

PR #3546, opened 2021-05-06
Moving docs to code

buger:
Moving docs to code


furkansenharputlu:
@buger I am really happy for this new progress. However, in [GoDoc conventions](https://blog.golang.org/godoc), documented field or func should start with its name. Wouldn't it be better to obey these rules when we think they will be visible [here](https://pkg.go.dev/github.com/TykTechnologies/tyk/config) at the same time?

buger:
Let’s try to find solution. Go names are not needed in public docs and mean nothing to our users. How we can write some special so it will be understood in both cases? Maybe smth like. “ListenPort: ....“ and initial prefix gets removed while markdown generation?

furkansenharputlu:
`ListenPort sets the port that Tyk listens on.` This is the Go way. Maybe such a rule like remove first word and capitalize second word's first letter. 

# listen_port

Sets the port that Tyk listens on.


But this needs all the docs should be written in this format `X makes....`. Not sure it'd be the best solution.

buger:
Your solution require thinking about working of each sentence. And forces author to pick specific voice... It can be hard to maintain it. We need to find some "general" solution which works for any situation. For example:

```
ListenPort: Setting this value will change the port that Tyk listens on. Default: 8080.
```

Which turns to:

```
## listen_port

Setting this value will change the port that Tyk listens on. Default: 8080.
```

furkansenharputlu:
As far as I know `ListenPort: Setting this value will change the port that Tyk listens on. Default: 8080.` is not GoDoc compatible either. IMHO, I'd prefer the current version to it. Maybe, we should say this is not a GoDoc and the problem is solved :)

buger:
I like it. So do nothing :)

PR #3549, opened 2021-05-07
Feature/OAS

furkansenharputlu:
This PR is to implement some of OAS support for Tyk API definition. When it is finalized, any Tyk API will be able to be exported in the form of an OAS documentation.


apidef/oas/authentication.go
```diff
@@ -0,0 +1,262 @@
...
+	}
+
+	if a.JWT != nil {
+		a.JWT.ExtractTo(api)
+	}
+}
+
+type Token struct {
+	Enabled                 bool `bson:"enabled" json:"enabled"` // required
+	AuthSources             `bson:",inline" json:",inline"`
```
vaske, line 64: is this by intention to miss bson name for this one?

furkansenharputlu: Yes

gateway/api.go
```diff
@@ -871,6 +931,7 @@ func handleDeleteAPI(apiID string) (interface{}, int) {
 	}
 
 	os.Remove(defFilePath)
+	os.Remove(defOASFilePath)
```
vaske, line 934: I see report here 
```
Check failure on line 933 in gateway/api.go
```

vaske (changes requested):
Great work so far, there are some check failures, could you please address them?

vaske (approved):

buger:
I have a feeling that majority of this code should live in the gateway. 
We also want to have OSS Gateway run using OAS API definitions, how we can make it happen?

furkansenharputlu:
@buger Yes, the conversion functions should live in the gateway. However, storing and returning proper style for user should be handled on both dashboard users and gateway users. For OSS, we already have such a logic where we store OAS format of an api definition in json file in this PR.

gateway/api.go
```diff
@@ -825,8 +868,35 @@ func handleAddOrUpdateApi(apiID string, r *http.Request, fs afero.Fs) (interface
...
 	}
 
+	// api
+	err, errCode := writeToFile(fs, newDef, newDef.APIID)
+	if err != nil {
+		return apiError(err.Error()), errCode
+	}
+
+	// oas
+	err, errCode = writeToFile(fs, &oasDoc, newDef.APIID+"-oas")
```
furkansenharputlu, line 878: @buger like here

buger (approved):

furkansenharputlu:
Manually cherry-picked all commits to `release-4`

PR #3558, opened 2021-05-13
[TT-2479] Uptime tests sample cache

tbuchaillot:
## Description

Removing sample cache TTL because it brings a lot of problems with its default value and misconfiguration.
To simplify that, the dependency on `"github.com/pmylund/go-cache"` is removed and we just use a `sync.Map` with a new `HostSample` struct. 
Also, removed the `unHealthyList` and added that information in each sample.

## Related Issue

https://tyktech.atlassian.net/browse/TT-2479
## Motivation and Context

https://tyktech.atlassian.net/browse/TT-2479
## How This Has Been Tested

https://tyktech.atlassian.net/browse/TT-2479
## Types of changes

- [X] Refactoring or add test (improvements in base code or adds test coverage to functionality)

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/host_checker.go
+++ b/gateway/host_checker.go
@@ -290,8 +290,10 @@ func (h *HostUptimeChecker) CheckHost(toCheck HostData) {
 
 func (h *HostUptimeChecker) Init(workers, triggerLimit, timeout, sampleExpiration int, hostList map[string]HostData, failureCallback, upCallback, pingCallback func(HostHealthReport)) {
 	if sampleExpiration == 0 {
+		log.Debug("[HOST CHECKER] Config:SampleExpiration: ", 30)
 		h.sampleCache = cache.New(30*time.Second, 30*time.Second)
 	} else {
+		log.Debug("[HOST CHECKER] Config:SampleExpiration: ", sampleExpiration)
 		h.sampleCache = cache.New(time.Duration(sampleExpiration)*time.Second, time.Duration(sampleExpiration)*time.Second)
 	}
 
--- a/gateway/host_checker.go
+++ b/gateway/host_checker.go
@@ -168,15 +168,20 @@ func (h *HostUptimeChecker) HostReporter() {
 				sample.count = sample.count + 1
 			}
 
-			if sample.count == h.sampleTriggerLimit {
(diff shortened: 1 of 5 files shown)


matiasinsaurralde (approved):
LGTM

gateway/host_checker.go
```diff
@@ -51,6 +49,11 @@ type HostHealthReport struct {
 	IsTCPError   bool
 }
 
+type HostSample struct {
```
buger, line 52: Why you decided to add a struct, instead of just using counter as previously?
Why we need a new `reachedLimit` field?

tbuchaillot: @buger I did it this way so we can delete `unHealthyList` and avoid having 2 sources of the truth of which host is down / up.

If we don't have some kind of flag / list that determines that a host reached the trigger limit (which was basically what `unHealthyList` did), the "host up" will start decreasing the counter number even if it doesn't reach the trigger limit.
This can cause that a host which is responding 1 down, 1 up with a trigger limit 2 never reach the trigger limit and will never be removed from the round robin.

gateway/host_checker.go
```diff
@@ -131,42 +133,55 @@ func (h *HostUptimeChecker) HostReporter() {
...
+
+			//If a host fails, we check if it has failed already
+			if hostSample, found := h.samples.Load(failedHost.CheckURL); found {
+				sample = hostSample.(HostSample)
+				// we add THIS failure to the count
+				sample.count = sample.count + 1
 			}
 
-			if newVal >= h.sampleTriggerLimit {
+			if sample.count == h.sampleTriggerLimit {
```
buger, line 171: Application logic has changed here. 
Previously it was `sample.count < h.sampleTriggerLimit` or to be exact `else` statement of `newVal >= h.sampleTriggerLimit`

tbuchaillot: @buger  I thought it wasn't necessary but I forgot about the fallback call. Good catch! 
Updated.

buger:
Needs be back ported to master and 3.x release branches.

PR #3563, opened 2021-05-27
[TT-880] Send unique gw session id in node registration request

furkansenharputlu:
The idea is to send unique session id from gateway to prevent multiple node id allocation for one gateway.

Works with: https://github.com/TykTechnologies/tyk-analytics/pull/2331

### DIFF
--- a/headers/headers.go
+++ b/headers/headers.go
@@ -34,6 +34,7 @@ const (
 	XXSSProtection      = "X-XSS-Protection"
 	XFrameOptions       = "X-Frame-Options"
 	XTykNodeID          = "x-tyk-nodeid"
+	XTykSessionID       = "x-tyk-session-id"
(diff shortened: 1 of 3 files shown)


headers/headers.go
```diff
@@ -34,6 +34,7 @@ const (
 	XXSSProtection      = "X-XSS-Protection"
 	XFrameOptions       = "X-Frame-Options"
 	XTykNodeID          = "x-tyk-nodeid"
+	XTykSessionID       = "x-tyk-session-id"
```
vaske, line 37: do we need to add this to documentation?

vaske (approved):
Looks ok to me, maybe someone else have some other input on top of those changes.

komalsukhani (approved):

PR #3575, opened 2021-06-01
normalize GraphQL schema (TT-2581)

pvormste:
This PR adds normalization for GraphQL schemas during API loading.

Before, the following schema did not work with proper Query:

Schema:
```
type Query {
  me: String
}

extend type Query {
  you: String
}
```

Query:
```
{
  you
}
```

Now it works.

## Types of changes


buger:
The schema in the PR description looks valid to me. 
Why it is broken, and how it looks after normalization? 

Also I wonder if we can add a test which failed previousy, and now not fails thanks to this fix.

PR #3586, opened 2021-06-18
[TT-960] Add allowance scope in case no pol in key and per api

furkansenharputlu:
The problem is that when you add two APIs to a key without policy, they are using same redis key for quota. In policy case, it uses allowance scope and for each API, it stores in different redis key. 

This is how we generate redis key:
https://github.com/TykTechnologies/tyk/blob/047733bd92e150c110be0d3afa8c2a4106ea3699/gateway/session_manager.go#L257

Briefly, when two APIs key without policy and per api enabled `quotaScope` is `""`, that is the issue.

### DIFF
--- a/gateway/middleware.go
+++ b/gateway/middleware.go
@@ -350,7 +350,7 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 				}
 
 				idForScope := apiID
-				// check if we already have limit on API level specified when policy was created
+				// check if we don't have limit on API level specified when policy was created
 				if accessRights.Limit == nil || *accessRights.Limit == (user.APILimit{}) {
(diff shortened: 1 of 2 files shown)


furkansenharputlu:
My concern is that we have some cases depending on empty allowance scope:
https://github.com/TykTechnologies/tyk/blob/047733bd92e150c110be0d3afa8c2a4106ea3699/gateway/middleware.go#L603

https://github.com/TykTechnologies/tyk/blob/047733bd92e150c110be0d3afa8c2a4106ea3699/gateway/session_manager.go#L315

buger (approved):

vaske (approved):

furkansenharputlu:
Merged manually to: `release-3-lts`

PR #3591, opened 2021-06-21
TT-2687 added ability to receive cert id on remove key

sredxny:
## Description
On remove key, added ability to receive cert id so we can remove keys with certs.  This is a pre-requisite to run https://github.com/TykTechnologies/tyk-analytics/pull/2357 so we must merge them in tandem

## Related Issue
https://tyktech.atlassian.net/browse/TT-2687

## Motivation and Context
On remove certs, remove keys

## How This Has Been Tested
- Created 1 api with cert enabled
- Created via portal the key by pasting the cert
- Check that we have access
- Remove the cert via cert manager in dashboard
- Consume api... we should not have access
- Check in dashboard the key page, we should not see the key anymore
- Do the same but now in MDCB environment, once we remove the certificate then we should not have access to the api

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [ ] `go vet`


matiasinsaurralde (approved):

sredxny:
Added test coverage

tbuchaillot (approved):

gateway/api.go
```diff
@@ -979,6 +980,10 @@ func keyHandler(w http.ResponseWriter, r *http.Request) {
 		}
 
 	case http.MethodDelete:
+		if certID != "" {
+			keyName =  generateToken(orgID, certID)
```
buger, line 984: I'm not quite sure why it requires a special case. Why we can't just standard "key" based on certificate fingerprint from the dashboard side?

sredxny: @buger makes sense. We do not longer need this PR. Made some changes in dashboard to take advantage of key deletion using only orgId+certId

sredxny:
Closing this PR in favor to remove the keys using only the certID from the dashboard

PR #3611, opened 2021-07-03
Fix/lts go plugins

buger:
## Description
TBD

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


alephnull:
While this fix _might_ work, this implementation is not suitable on release-3-lts, for the following reasons, in order of increasing importance:
- goreleaser validation skipped
- upgrade tests failing
- `images/plugin-compiler/data/build.sh` masks incompatibilities between plugin versions and tyk versions. User should be made aware of incompatibilites
- the source templates in tyk-ci/wf-gen have not been updated with this fix

A compromise might be to commit this to a hotfix branch, say release-3.0.6 and use that branch leaving release-3-lts alone. The testplugin code is an improvement over the existing tests and should be used in all smoke tests.

buger:
@alephnull 
> goreleaser validation skipped

To make it work like 3.0.4, it require moving vendor folder to GOPATH. This action modify Git, and goreleaser by default does not allow "dirty git". So I do not think we can make it work without this flag. 

> images/plugin-compiler/data/build.sh masks incompatibilities between plugin versions and tyk versions. 
What kind of incompatibibilitise you mean? It works exactly the same as 3.0.4 version, except that I improved it, and instead of just copying vendor folders one on top of another, it first removes "duplicate" folders, and then does fresh copy. 
The only limitation here is that if plugin use the same package as gateway does, gateway version will be used, and we can cleary state it in docs.

> the source templates in tyk-ci/wf-gen have not been updated with this fix
As discussed you have some ideas on how to implement this fix the other way. So I suggest merge this fix, and work based on it. 

> upgrade tests failing
🤔

alephnull:
> > images/plugin-compiler/data/build.sh masks incompatibilities between plugin versions and tyk versions.
> What kind of incompatibibilitise you mean? It works exactly the same as 3.0.4 version, except that I improved it, and instead of just copying vendor folders one on top of another, it first removes "duplicate" folders, and then does fresh copy.
> The only limitation here is that if plugin use the same package as gateway does, gateway version will be used, and we can cleary state it in docs.

My reasoning that this is not the approach of least surprise for the user. Seeing this as good developer experience is a false economy. We should use what the user provided and break if that is a problem and get out of the way of the compiler in detecting these problems. Note that this is not a blocker for release, it is only a problem for including it in the releng code.

> > upgrade tests failing
> thinking

I'll see if something is missing from https://github.com/TykTechnologies/tyk/pull/3596 for this branch.

PR #3616, opened 2021-07-08
Replicate changes of  custom keys in slave nodes (deletion/updates)

sredxny:
## Description
When create custom key in slave node, we must ensure that the hash is the same as in the master node, otherwise we could face issues updating and deleting tokens. Also in this PR was solved an issue where we would end up having 2 keys in redis when we used the key (when we used the key then a new key and quota key were registered)

## Related Issue
- https://tyktech.atlassian.net/browse/TT-2820
- https://tyktech.atlassian.net/browse/TT-2639

## Motivation and Context
Fix CRUD of custom keys in slave nodes. Prevent the creation of 2 keys 

## How This Has Been Tested
Replicating the steps to reproduce the bug

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`


furkansenharputlu:
Any test? I know it is in MDCB env but we can mock things.

sredxny:
Fixed duplication of keys when we use custom keys...Currently the only issue is that the quota-key in slave nodes has a different keyId...checking

gateway/middleware.go
```diff
@@ -718,13 +721,25 @@ func (t BaseMiddleware) CheckSessionAndIdentityForValidKey(originalKey *string,
...
+	} else {
+		key = customKey
+	}
+	*/
 	if found {
 		// update value of originalKey, as for custom-keys it might get updated (the key is generated again using alias)
 		*originalKey = key
 
 		session := session.Clone()
 		session.SetKeyHash(keyHash)
```
buger, line 740: You may need to add keyHash generation here too.

matiasinsaurralde (commented):
Both issues (duplication and sync with slave) seems to be fixed with these changes.

I've found that the following scenario doesn't work:

1. Create a custom key named "mykey", the creation endpoint will return the session data.
2. Try to delete the key using the following dashboard API call: `DELETE /api/keys/mykey`.

IIRC this scenario was previously working and to me it looks like the friendlier way to delete a key through the Dashboard API, without having to use the `key_id`.

I've added https://github.com/TykTechnologies/tyk/pull/3616/commits/1a8ff37158ec6f59e1bf55b48ce45257a676aef5 in case it's useful as a workaround, feel free to revert it if this is not something we wan't to cover. The reason why this commit expands two scenarios is `keyHandler` is because dashboard deletion actually triggers two events, a `GET` is sent first, then `DELETE`.

The following scenario works fine:

1. Create a custom key named "mykey". From the session data take the value from `key_id`, e.g. `eyJvcmciOiI2MGU1NDA3YTIxYzkzNzcyNzI5MzQyZDkiLCJpZCI6ImJsYWJsYTY2NiIsImgiOiJtdXJtdXI2NCJ9`.
2. Call `DELETE /api/keys/eyJvcmciOiI2MGU1NDA3YTIxYzkzNzcyNzI5MzQyZDkiLCJpZCI6ImJsYWJsYTY2NiIsImgiOiJtdXJtdXI2NCJ9`.

Other tested scenarios:

- Deleting from the dashboard from the key list works.
- When listing isn't enabled, lookup + delete from the "Key Search" works fine.

@sredxny Regarding ID extractor issues, I don't think they're related with this PR, I have found similar errors on `master` tests.

sredxny:
@matiasinsaurralde It looks consistent. I ran some tests as well and its working even in MDCB, I tried deleting and updating keys (I searched by alias, key id and key hash) and its working.

Now, regarding the tests in python bundles, then can we consider safe to merge these changes?

matiasinsaurralde:
> @matiasinsaurralde It looks consistent. I ran some tests as well and its working even in MDCB, I tried deleting and updating keys (I searched by alias, key id and key hash) and its working.
> 
> Now, regarding the tests in python bundles, then can we consider safe to merge these changes?

Great. Yes, should be safe.

sredxny:
@matiasinsaurralde please check my latest changes, basically I reverted your changes and provided a valid orgID to `handleGetDetail`, before we were sending an empty org so the token generation returned a wrong key hence it was never found. Now, if we ensure to provide a valid OrgID then we can properly find our key.

I tested:
- Creating a custom key via dashboard api
- Use the key in slave node
- Remove the key via alias, via key id and via hash id using the dashboard api. I also tried this using the UI
- Try to consume the key in slave node...in all of them I get: access disallowed

cc:  @buger

matiasinsaurralde:
Got the following while trying the latest changes:

**When using “enable_hashed_keys_listing” as “false”:**
- Key lookup “mykey” doesn’t work.
- Key lookup when using key ID doesn’t work, e.g. eyJvcmciOiI2MGU1NDA3YTIxYzkzNzcyNzI5MzQyZDkiLCJpZCI6ImJsYWJsYTY2NiIsImgiOiJtdXJtdXI2NCJ9
- Key lookup when using key hash doesn’t work, e.g. 2df3fa315aa48799

**When using “enable_hashed_keys_listing” as “true”:**
- Delete key from list works fine.

**Deleting using `DELETE /api/keys/mykey` works fine**

matiasinsaurralde:
Did some investigation/testing with Python 3.6 + Ubuntu 18.04:
- Non-auth hooks on plugins work fine.
- ID extractor faces several issues and there were mixed results, Python plugins worked fine but ID extractor was ignoring the cache action, Go plugins returned an error (it was fixed by removing `sessionID = previousSession.KeyID` from `coprocess_id_extractor.go` as this value was empty). I did a few tests with `hash_keys` on/off, same results.

sredxny:
With the latest commit now the quota key is created correctly with the correct keyhash/keyid, now the issue that remains is that the api-key is not created, we should see something like: `quota-mykey` and  `apikey-mykey` but the api key is not created...

ctx/ctx.go
```diff
@@ -61,8 +61,8 @@ func ctxSetSession(r *http.Request, s *user.SessionState, token string, schedule
 
 	ctx := r.Context()
 	ctx = context.WithValue(ctx, SessionData, s)
+//	ctx = context.WithValue(ctx, AuthToken, s.KeyID)
```
sredxny, line 64: with this, many tests related to authentication fails...

sredxny: We need to make it dependant of s.KeyID for consistency.

matiasinsaurralde:
- Added potential fixes for client certificates and JWT code, seems key lookup was failing in these scenarios (found through automated tests). For client certificates I'm using a workaround, we first try to load the certificate in the same way we did before, then we use the contents from `keyID`.
- Would be great to have @sredxny's input on this one: https://github.com/TykTechnologies/tyk/pull/3616/commits/c1e12efba3d1734b29e20c83319d7a258b103a78. I feel that instead of doing this fallback, we should try to debug all possible functions that cause the root issue (empty key ID), I already fixed some of them.
- From what I've investigated, in a real scenario we always send the org ID but some of our tests don't. This is a workaround to include "default" orgs and temporarily fix tests. I will come up with a better solution over the next hours: https://github.com/TykTechnologies/tyk/pull/3616/commits/e0a40c16390d8d11e61561803758b1aa0208a9d8
- I find most tests working on two test machines (OS X and Linux), coprocess tests are broken even on `master`, we should ignore those for now.

gateway/mw_auth_key.go
```diff
@@ -102,19 +103,28 @@ func (k *AuthKey) ProcessRequest(_ http.ResponseWriter, r *http.Request, _ inter
...
 	}
 
 	if authConfig.UseCertificate {
 		certID := session.OrgID + certHash
-		if _, err := CertificateManager.GetRaw(certID); err != nil {
-			return k.reportInvalidKey(key, r, MsgNonExistentCert, ErrAuthCertNotFound)
+		_, err := CertificateManager.GetRaw(certID)
+		if err != nil {
+			// Try alternative approach:
+			id := storage.TokenID(session.KeyID)
```
buger, line 122: I do not see how it can be related to your changes. 
certID is always `session.OrgID + certHash`. 
It should have nothing to do wit  keyIDs or similar.

matiasinsaurralde: In previous changes we had the following:
```go
        session, keyExists = k.CheckSessionAndIdentityForValidKey(key, r)
        key = session.KeyID
        if !keyExists {
                // fallback to search by cert
                session, keyExists = k.CheckSessionAndIdentityForValidKey(certHash, r)
                certHash = session.KeyID
                if !keyExists {
                        return k.reportInvalidKey(key, r, MsgNonExistentKey, ErrAuthKeyNotFound)
                }
        }

        if authConfig.UseCertificate {
                certID := session.OrgID + certHash
                _, err := CertificateManager.GetRaw(certID)
                if err != nil {

```
We reviewed the tests too and they were broken, after adding this `GetRaw` workaround it was fixed.

sredxny: Just for the records, without this change the test `TestHashKeyFunctionChanged/client_certificate` fails...we need to check why it passed before

matiasinsaurralde:
Manually merged to:
- `release-3` (commit d0cf5e6a28350e45d36a5e39eeb9648a95c6bda9)
- `release-3-lts` (commit 222fb0dbd040760e16e4b21fa958c85958ea4a47)
- `release-3.0.7` (commit bea893448b94cfac1f789bf707850be5485ae26d)
- `release-3.2` (commit 5267f1c7c90c83646a99dd9daa6316b268b0c36b)
- `release-3.2.2` (commit 50b3f677899155a5e9c4d2f27c2395af1494c7df)

matiasinsaurralde:
Manually merged to:
- `release-3` (commit d0cf5e6a28350e45d36a5e39eeb9648a95c6bda9)
- `release-3-lts` (commit 222fb0dbd040760e16e4b21fa958c85958ea4a47)
- `release-3.0.7` (commit bea893448b94cfac1f789bf707850be5485ae26d)
- `release-3.2` (commit 5267f1c7c90c83646a99dd9daa6316b268b0c36b)
- `release-3.2.2` (commit 50b3f677899155a5e9c4d2f27c2395af1494c7df)

PR #3626, opened 2021-07-19
[POC] wasm middlewares

gernest:
```json
{
    "name": "API with wasm middleware",
    "api_id": "3",
    "org_id": "default",
    "use_keyless": true,
    "proxy": {
        "listen_path": "/wasm/",
        "target_url": "http://httpbin.org/",
        "strip_listen_path": true
    },
    "do_not_track": true,
    "custom_middleware": {
        "pre": [
            {
                "name": "replace test header",
                "path": "http_headers.wasm"
            }
        ],
        "driver": "wasm"
    }
}
```


matiasinsaurralde (commented):
Nice one!

wasm/v1/handler/http_request.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	return &Header{head: r.Request.Trailer}
+}
+
+func (r *Request) GetHttpRequestMetadata() common.HeaderMap {
+	if r.Request == nil {
+		return nil
+	}
+	return &Header{head: r.Request.Trailer}
+}
```
asoorm, line 45: This looks like a bug.

buger:
@gernest wasm package looks like a copy from another project. What was the original?

gernest:
@buger https://github.com/gernest/tt

gernest:
With exception of https://github.com/gernest/tt/blob/master/wasm/instance.go which Is a modified copy of https://github.com/mosn/proxy-wasm-go-host/blob/main/wasmer/instance.go

I will add proper attribution (Its under Apache). My code is under MIT.

gernest:
My project has different goals and release cadence. I had to copy over so as to have something tyk specific for this POC.

buger:
@gernest how extending it with custom function will look like? 
For example, exposing simple memory storage functions, SetKey, GetKey ?
Also how user can get access to user session object?

gernest:
@buger 
>  how extending it with custom function will look like?
easy, you can register go functions that will be available in the module as imports.

> For example, exposing simple memory storage functions, SetKey, GetKey ?
This is already described  by the ABI so it is doable(I didn't implement this on this branch)

> Also how user can get access to user session object?
Many ways, they will require serialization/deserialrization though. we can extend and add functions for manipulating sessions if need be or we can stick with the ABI and expose using abi primitives.

buger:
I do not think we need to close it

gernest:
My bad! Is there still interest in this?

excieve:
Sounds like a better way to do close to native speed plugins for most users. What is it missing to go from a PoC to a feature @gernest?

FYI @alephnull: that's the PoC I mentioned.

PR #3630, opened 2021-07-27
[TT-2932] fix upgrades for gw from 3.0.6

alephnull:
This is an alternate strategy to https://github.com/TykTechnologies/tyk-analytics/pull/2386

## Description
With this PR, we fallback to the old Buddy pipeline release process for 3.0.x releases. This PR will need to be applied to release-3.0.7 and release-3-lts.

## Related Issue
https://tyktech.atlassian.net/browse/TT-2932

## Motivation and Context
We backup and restore the systemd service file using rpm specific triggers. Since goreleaser and xenial are in opposition, using fpm can allow for LTS releases to continue supporting xenial _and_ work-around the upgrade bug.

While this works there are a lot of scripts that run at various points and it is not easy for a person without context to understand what is going on here. This is not good for LTS branches which prioritise stability.

## How This Has Been Tested
Tested locally using `tyk-gateway-3.0.7~273.3268ac0` from tyk-gateway-unstable repo. Upgraded from 3.0.6 from tyk-gateway repo.

### DIFF
--- a/bin/dist_build.sh
+++ b/bin/dist_build.sh
@@ -121,7 +121,7 @@ FPMCOMMON=(
     --before-install $TEMPLATEDIR/install/before_install.sh
     --after-install $TEMPLATEDIR/install/post_install.sh
     --after-remove $TEMPLATEDIR/install/post_remove.sh
-    --after-upgrade $TEMPLATEDIR/install/posttrans.sh
+    --posttrans $TEMPLATEDIR/install/posttrans.sh
 )
 
 cd $BUILDDIR
--- a/bin/dist_build.sh
+++ b/bin/dist_build.sh
@@ -121,7 +121,7 @@ FPMCOMMON=(
     --before-install $TEMPLATEDIR/install/before_install.sh
     --after-install $TEMPLATEDIR/install/post_install.sh
     --after-remove $TEMPLATEDIR/install/post_remove.sh
-    --before-upgrade $TEMPLATEDIR/install/backup.sh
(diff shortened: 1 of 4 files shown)


bin/dist_build.sh
```diff
@@ -121,11 +121,8 @@ FPMCOMMON=(
     --before-install $TEMPLATEDIR/install/before_install.sh
     --after-install $TEMPLATEDIR/install/post_install.sh
     --after-remove $TEMPLATEDIR/install/post_remove.sh
-)
-[ -z $PKGCONFLICTS ] || FPMCOMMON+=( --conflicts $PKGCONFLICTS )
-FPMRPM=(
-    --before-upgrade $TEMPLATEDIR/install/post_remove.sh
```
buger, line 127: Why this hooks was removed?

alephnull: Since it was never used and `--{before,after}-upgrade` is incompatible with the triggers as we use them. From `fpm -h`:
```
   --after-upgrade FILE          A script to be run after package upgrade. If not specified,
                                  --before-install, --after-install, --before-remove, and 
                                  --after-remove will behave in a backwards-compatible manner
                                  (they will not be upgrade-case aware).
                                  Currently only supports deb, rpm and pacman packages.
    --before-upgrade FILE         A script to be run before package upgrade. If not specified,
                                  --before-install, --after-install, --before-remove, and 
                                  --after-remove will behave in a backwards-compatible manner
                                  (they will not be upgrade-case aware).
                                  Currently only supports deb, rpm and pacman packages.
```

PR #3638, opened 2021-08-10
[TT-2820] Extend key duplicate on update test for long custom keys

furkansenharputlu:
The issue is fixed with https://github.com/TykTechnologies/tyk/pull/3616 but the test case is missing. This PR extends the test case for long custom key case. I pushed same commit to `release-2.9`, it is failing there. So, we make sure that that scenario is passing in `master` and `release-3`, `release-3.0.7`... branches.

The problem was occurring here before the fix https://github.com/TykTechnologies/tyk/pull/3616 :
https://github.com/TykTechnologies/tyk/blob/HEAD/storage/storage.go#L128-L130

### DIFF
--- a/gateway/api_test.go
+++ b/gateway/api_test.go
@@ -592,21 +592,49 @@ func TestKeyHandler_CheckKeysNotDuplicateOnUpdate(t *testing.T) {
 		spec.Auth.UseParam = true
 	})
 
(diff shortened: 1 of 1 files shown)


furkansenharputlu:
release-3-lts

PR #3650, opened 2021-08-23
Revert "[CI] Post failures on PRs (#3527)"

furkansenharputlu:
After the change, there are a lot of `gofmt`, `goimports` failures are merged to master. When there is a failure it should fail the tests immediately so owner of a PR should fix formatting. Also, when there is a test failure, we can see in the `Checks` session easily, the failure comment in the PR creates complexity.


alephnull:
Hey Furkan, in https://tyktech.atlassian.net/browse/TD-565 there is a need for

> specific feedback about the tests or build state should be communicated in the GH PR comment.

Also, if the ci-tests.sh are noisy, I think we should remove the noise from the tests or fix the tests rather than remove the notification but please do share your thoughts on the ticket above.

We were planning on extending this behaviour to tyk-anaytics in the near future so now would be a good time to discuss the best way of running tests.

PR #3659, opened 2021-09-02
Remove duplicated tags in case that we pass over duplicated items

vaske:
This PR brings logic to prevent duplicated tags to be passed over

### DIFF
--- a/gateway/util.go
+++ b/gateway/util.go
@@ -87,3 +87,27 @@ func greaterThanInt(first, second int) bool {
 
 	return first > second
 }
+// remove duplicate strings from string slice
+func removeDuplicateStr(strSlice []string) []string {
+	allKeys := make(map[string]bool)
+	list := []string{}
+	for _, item := range strSlice {
+		if _, value := allKeys[item]; !value {
+			allKeys[item] = true
+			list = append(list, item)
+		}
+	}
+	return list
+}
+// remove duplicate int values from int slice
+func removeDuplicateInt(intSlice []int) []int {
+	allKeys := make(map[int]bool)
+	list := []int{}
+	for _, item := range intSlice {
+		if _, value := allKeys[item]; !value {
+			allKeys[item] = true
+			list = append(list, item)
+		}
+	}
+	return list
+}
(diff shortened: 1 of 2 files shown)


gateway/util.go
```diff
@@ -87,3 +87,27 @@ func greaterThanInt(first, second int) bool {
...
+	for _, item := range strSlice {
+		if _, value := allKeys[item]; !value {
+			allKeys[item] = true
+			list = append(list, item)
+		}
+	}
+	return list
+}
+// remove duplicate int values from int slice
+func removeDuplicateInt(intSlice []int) []int {
```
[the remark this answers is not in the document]
vaske: No we do not, but since I added this unique I just added this one for integers we might use it sometimes, but yeah can remove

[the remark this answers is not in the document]
vaske: @buger removed, please check

gateway/util.go
```diff
@@ -87,3 +87,15 @@ func greaterThanInt(first, second int) bool {
 
 	return first > second
 }
+// remove duplicate strings from string slice
+func removeDuplicateStr(strSlice []string) []string {
+	allKeys := make(map[string]bool)
+	list := []string{}
+	for _, item := range strSlice {
+		if _, value := allKeys[item]; !value {
```
furkansenharputlu, line 95: value -> contains

gateway/util.go
```diff
@@ -87,3 +87,15 @@ func greaterThanInt(first, second int) bool {
 
 	return first > second
 }
+// remove duplicate strings from string slice
```
furkansenharputlu, line 90: I think we should apply GoDoc rule here:

```
// removeDuplicateStr removes duplicates values from string slice by respecting the order.
```

gateway/util.go
```diff
@@ -87,3 +87,15 @@ func greaterThanInt(first, second int) bool {
...
+func removeDuplicateStr(strSlice []string) []string {
```
furkansenharputlu, line 91: strSlice -> items

furkansenharputlu (approved):

PR #3677, opened 2021-09-20
[TT-3409] Plugin compiler with latest fixes

alephnull:
This PR is for release-3.0.7 only. This attempts to enforce a directory structure which is what the goplugins are most sensitive to. This is still a WIP and needs more testing. It is a puzzle as to how 3.0.6 managed to work without any of this. CentOS 7 support also needs to be verified.

## Related Issue
https://tyktech.atlassian.net/browse/TT-3409

## How This Has Been Tested
See _status_ tab.

The smoke tests from master have been ported to this branch.


buger:
Not needed anymore?

PR #3687, opened 2021-09-29
[TT-1557] Implement basic auth OAS conversion

furkansenharputlu:


vaske (approved):

apidef/oas/authentication_test.go
```diff
@@ -93,3 +93,15 @@ func TestSignature(t *testing.T) {
 
 	assert.Equal(t, emptySignature, resultSignature)
 }
+
+func TestBasic(t *testing.T) {
```
buger, line 97: Looks good! But I would have a test with all fields set.

furkansenharputlu: We have that test too but it is for entire apidef. So it is skipped for now as we didn’t implement all conversions.

furkansenharputlu: The final goal is to make that test green. Here the test: https://github.com/TykTechnologies/tyk/blob/ab13bdb0d200d9f38c067012be20e365bcc70e0e/apidef/oas/root_test.go#L27

PR #3696, opened 2021-10-06
[TT-1559] Implement oauth OAS conversion

furkansenharputlu:


apidef/oas/authentication_test.go
```diff
@@ -105,3 +105,15 @@ func TestBasic(t *testing.T) {
 
 	assert.Equal(t, emptyBasic, resultBasic)
 }
+
+func TestOAuth(t *testing.T) {
```
buger, line 109: I'll be frank, I do not believe into these tests that much because they totally do not test Fill and Extract functions, because all fields are empty. Using empty struct you can't really know if a new object is properly filled with data. 

Even function which you showed earlier which test global API object, it also operate with empty struct.

furkansenharputlu: @buger This test only checks whether there is a `panic` because there are too much `nil` cases. This test doesn't check conversion. You know I don't leave tests to chance :) Please see the `filled` case of the test that I shared in the previous PR: https://github.com/TykTechnologies/tyk/blob/ab13bdb0d200d9f38c067012be20e365bcc70e0e/apidef/oas/root_test.go#L27

PR #3704, opened 2021-10-15
[TT-3621] Use session.Certificate to search whether certificate existing

furkansenharputlu:
I think it is not necessary to build cert id from scratch again. We already have that in session. The problems occur while rebuilding it.


furkansenharputlu:
Merged manually to:
- `release-3-lts`
- `release-3.0.8`

PR #3705, opened 2021-10-15
Update api_definitions.go

komuw:
Fix APIDefinition.Auth deprecation comment so as to follow Golang standards[1][2]
This will enable linting tools (like godoc, statitcheck, golangci-lint etc) to be able to warn whenever it is used.

Although the inline form of deprecation is allowed by Go[3][4], various tools aren't able to pick it up.  
I tested both staticcheck and golangci-lint, & they are only able to warn of the deprecation if it is not inline.

1. https://github.com/golang/go/wiki/Deprecated
2. https://go.dev/blog/godoc
3. https://github.com/golang/go/issues/10909#issuecomment-136492606
4. https://github.com/golang/go/blob/go1.16.9/src/archive/zip/struct.go#L132-L133


Tyk-ITS:
Branch used: refs/pull/3705/merge
Commit: 69fb6bac24722cbad0fb712adfd3be36f2f3c0e2 
Triggered by: pull_request (@komuw)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2344150702)

titpetric:
I think this PR may be invalid, same-line deprecation comments are reported by golangci-lint, e.g.:

```
api_definitions.go:	VersionData                          VersionData            `bson:"version_data" json:"version_data"` // Deprecated. Use VersionDefinition instead.
```

This one is definitely reported with the linter. If you feel strongly about it, I'd reopen and address the complete codebase. The changes are comments-only so there's no testing, QA, but we need to be consistent when we address this. But there are other more valuable areas to focus on given it's a no-op change :)

Issue #3256, opened 2020-08-05
SSE flow broken in v.2.9.4-1


jeffrngu:
**Branch/Environment/Version**
- Environment: 
Tyk GW CE 2.9.4-1
SSE server
curl command

**Describe the bug**
The Server Send Event (SSE) support from Tyk GW CE is broken in version 2.9.4-1. This feature was working in version 2.8.1-1.
Always receive error message "internal error: 101 switching protocols response with non-writable body" when attempting to connect to SSE server via Tyk GW.

**Reproduction steps**

1/- Deploy an SSE server or use a test SSE server such as https://github.com/kljensen/golang-html5-sse-example
2/- Deploy Tyk GW CE 2.9.4-1
3/- Define a test API endpoint as open api (no auth required) which routes the request to the SSE server
Listenpath: /sse-test
Target: the SSE server
4/- From the Tyk container, run the following command to receive SSE update from the SSE server
curl -iL -H "Accept: text/event-stream" "http://localhost:8080/sse-test/events/"

**Actual behavior**
Connection upgrade failed and the following response is observed:

HTTP/1.1 500 Internal Server Error
Content-Type: application/json
X-Generator: tyk.io
Date: Tue, 04 Aug 2020 23:50:15 GMT
Content-Length: 90
Connection: close

{
    "error": "internal error: 101 switching protocols response with non-writable body"
}

**Expected behavior**
Connection is internally upgraded by Tyk GW and the curl command blocks on SSE update from the SSE server

**Screenshots/Video**
If applicable, add screenshots or video to help explain your problem.

**Logs (debug mode or log file):**
Log from console or from log file.

**Configuration (tyk config file):**
Attach tyk configuration file

**Additional context**

matiasinsaurralde:
Have you tried the settings suggested in [3214](https://github.com/TykTechnologies/tyk/issues/3214#issuecomment-668969402)?

jeffrngu:
@matiasinsaurralde yes, I've tried the above scenario with enable_websockets set to false and the SSE flow works. However, now SSE cannot work without WebSocket being disabled. Is there a plan to address this so that both can be enabled at the same time?

matiasinsaurralde:
@jeffrngu That's great. I will discuss the scenario with the team and reach you back.

Issue #3214, opened 2020-07-15
SSE stream fails to upgrade properly


jlucktay:
**Branch/Environment/Version**
- Branch/Version: [`latest` tag from tykio/tyk-gateway](https://hub.docker.com/layers/tykio/tyk-gateway/latest/images/sha256-83be102d8829fbe91cde86baeef8cf29ef56369aad2644d6adc60f1cea16375b?context=explore)
- Environment: [`tyk-demo`](https://github.com/TykTechnologies/tyk-demo)

**Describe the bug**
When an SSE stream is proxied through the gateway, it fails to upgrade properly.

**Reproduction steps**
1. clone [tyk-demo](https://github.com/TykTechnologies/tyk-demo) and run `up.sh`
2. import the API definition (JSON posted below) into the Tyk Dashboard
3. clone [this SSE PoC](https://github.com/kljensen/golang-html5-sse-example)
4. make one very small change and remove the leading slash from [the `/events/` path on L14 of `templates/index.html`](https://github.com/kljensen/golang-html5-sse-example/blob/d4eba81ddea2b4191f039adb6929086ca6b39c0f/templates/index.html#L14) so that it is `events/` instead
5. run `go run server.go`
6. browse to http://localhost:8080/ssepoc/

**Actual behaviour**
A `connection refused` error is given; full text is posted below.

Chrome DevTools shows the response from `events/` as such:
```
{
    "error": "internal error: 101 switching protocols response with non-writable body"
}
```

Sad path:
![10614 sad-path 2](https://user-images.githubusercontent.com/460028/87575279-9a6f5300-c6c7-11ea-9dfb-b5159afaa41f.gif)

**Expected behaviour**
A WebSocket should be established and timestamps should start appearing on the web page.

Happy path:
![10614 happy-path](https://user-images.githubusercontent.com/460028/87575326-a9ee9c00-c6c7-11ea-9ac8-fbc0493ff5c2.gif)

**Screenshots/Video**
See above.

**Logs (debug mode or log file):**
```
time="Jul 15 15:26:35" level=error msg="http: proxy error: dial tcp 127.0.0.1:8000: connect: connection refused" api_id=3b7cf65bc06a4ed65c5d609a365f5d78 api_name=ssepoc mw=ReverseProxy org_id=5e9d9544a1dcd60001d0ed20 prefix=proxy server_name="localhost:8000" user_id=-- user_ip=172.23.0.1 user_name=

[...]

PR #3717, opened 2021-11-01
[TT-1607] Fix SSE when web sockets are enabled

furkansenharputlu:
This PR fixes the SSE when `enable_websockets` is `true`. The problem is inside [IsUpgrade](https://github.com/TykTechnologies/tyk/blob/8ba3c49443a1d3dc2568c425a824fb7e3cb625b7/gateway/reverse_proxy.go#L1650) function, when websockets are enabled that function returns `true` and it tries to behave the response as upgrade response and it enters [handleUpgradeResponse](https://github.com/TykTechnologies/tyk/blob/master/gateway/reverse_proxy.go#L1226). If the reason of adding `"text/event-stream"` check inside `IsUpgrade` function is to consume incoming data immediately, I believe it does that [here](https://github.com/TykTechnologies/tyk/blob/master/gateway/reverse_proxy.go#L1356).

The problem was happened because of missing test, so this PR adds a test including websockets is enabled/disabled cases.

Fixes https://github.com/TykTechnologies/tyk/issues/3214
Fixes https://github.com/TykTechnologies/tyk/issues/3256

### DIFF
--- a/gateway/reverse_proxy.go
+++ b/gateway/reverse_proxy.go
@@ -1652,11 +1652,6 @@ func (p *ReverseProxy) IsUpgrade(req *http.Request) (bool, string) {
 		return false, ""
 	}
 
-	EncodeAccept := strings.ToLower(strings.TrimSpace(req.Header.Get(headers.Accept)))
-	if EncodeAccept == "text/event-stream" {
-		return true, ""
-	}
-
 	connection := strings.ToLower(strings.TrimSpace(req.Header.Get(headers.Connection)))
 	if connection != "upgrade" {
 		return false, ""
--- a/gateway/reverse_proxy_test.go
+++ b/gateway/reverse_proxy_test.go
@@ -12,9 +12,12 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"strings"
+	"sync"
 	"testing"
 	"text/template"
 	"time"
+
+	"github.com/stretchr/testify/assert"
 
 	"github.com/jensneuse/graphql-go-tools/pkg/execution/datasource"
 	"github.com/jensneuse/graphql-go-tools/pkg/graphql"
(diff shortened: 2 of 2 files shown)


buger (approved):
Looks good!

matiasinsaurralde:
I've tried a scenario using [gorilla/websockets](https://github.com/gorilla/websocket). With SSE integrated into the same WS server (based on [r3labs/sse/v2](https://github.com/r3labs/sse/v2)):
```go
// Copyright 2013 The Gorilla WebSocket Authors. All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

package main

import (
	"flag"
	"log"
	"net/http"
	"time"

	"github.com/r3labs/sse/v2"
)

var addr = flag.String("addr", ":8080", "http service address")

func serveHome(w http.ResponseWriter, r *http.Request) {
	log.Println(r.URL)
	if r.URL.Path != "/" {
		http.Error(w, "Not found", http.StatusNotFound)
		return
	}
	if r.Method != "GET" {
		http.Error(w, "Method not allowed", http.StatusMethodNotAllowed)
		return
	}
	http.ServeFile(w, r, "home.html")
}

func main() {
	flag.Parse()
	hub := newHub()
	go hub.run()
	http.HandleFunc("/", serveHome)

	// The WebSockets handler is attached here (under /ws):
	http.HandleFunc("/ws", func(w http.ResponseWriter, r *http.Request) {
		serveWs(hub, w, r)
	})

	// The SSE handler is here (under /events):
	sseServer := sse.New()
	sseServer.CreateStream("messages")
	go func() {
		for {
			sseServer.Publish("messages", &sse.Event{
				Data: []byte("ping"),
			})
			time.Sleep(1 * time.Second)
		}
	}()
	http.HandleFunc("/events", sseServer.HTTPHandler)
	err := http.ListenAndServe(*addr, nil)
	if err != nil {
		log.Fatal("ListenAndServe: ", err)
	}
}
```
The API definition configured as:
```json
{
    "name": "Tyk Test API",
    "api_id": "3",
    "org_id": "default",
    "definition": {
        "location": "",
        "key": ""
    },
    "use_keyless": true,
    "auth": {
        "auth_header_name": ""
    },
    "version_data": {
        "not_versioned": true,
        "versions": {
            "Default": {
                "name": "Default",
                "expires": "3000-01-02 15:04",
                "use_extended_paths": true,
                "extended_paths": {
                    "ignored": [],
                    "white_list": [],
                    "black_list": []
                }
            }
        }
    },
    "proxy": {
        "listen_path": "/quickstart/",
        "target_url": "http://localhost:8080/",
        "strip_listen_path": true
    },
    "do_not_track": true
}
```
And an SSE client that access the SSE endpoint through Tyk (running in 8081):
```go
package main

import (
	"fmt"

	"github.com/r3labs/sse/v2"
)

func main() {
	client := sse.NewClient("http://localhost:8081/quickstart/events")

	client.Subscribe("messages", func(msg *sse.Event) {
		fmt.Println("got msg:", string(msg.Data))
	})
}
```
When testing the same scenario with `master`, SSE doesn't work. When trying the `fix-sse` branch, both WS and SSE work fine, video [here](https://www.loom.com/share/848d55a494c74af3acbcf743474214fd) 👍

matiasinsaurralde (approved):
Looking good, did a quick test

furkansenharputlu:
@matiasinsaurralde Thanks for verifying!

furkansenharputlu:
Manuall merged to:
- `release-3-lts`
- `release-4`

buger:
@furkansenharputlu what is the Jira task ID?

PR #3719, opened 2021-11-03
TT-3648 started to process oauth clients events

sredxny:
## Description
Now gateway process any signal of oauth client event from mdcb, it will process any: create, update and delete operation. It will allow slaves datacenters to store locally the oauth clients so it will not hit mdcb everytime that an operation related to auth is performed. 

## Related Issue
https://tyktech.atlassian.net/browse/TT-3648

## Motivation and Context
Make slave datacenter more resilient to mdcb outages

## How This Has Been Tested
- Ran the whole mdcb environment. Using the branches of sink and dashboard: TT-3646-feature/emit-oauth-events and 3647/feature-process-oauth-events
- Create an api with ouath enabled via dashboard
- via dashboard created an oauth client for that api and after some seconds we should see that client in the slave redis
- via dashboard update the oauth client and after some seconds the changes should be visible in the slave redis
- via dashboard delete the oauth client and after some seconds we should not see the oauth client in the slave redis neither the register in the index key for oauth clients
- Created client
- Shut down sink
- generated access token in worker node

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`


gateway/rpc_storage_handler.go
```diff
@@ -817,13 +820,67 @@ func (gw *Gateway) getSessionAndCreate(keyName string, r *RPCStorageHandler, isH
 	}
 }
 
+func (gw *Gateway) ProcessOauthClientsOps(clients map[string]string) {
```
tbuchaillot, line 823: It looks good! can we add a func description?

tyk-gateway.pid
```diff
@@ -1 +1 @@
-43267
\ No newline at end of file
+69425
```
tbuchaillot, line 1: I think we should discard this

tbuchaillot (approved):
Looks good! it would be great if we add some code comments 👍

tyk-gateway.pid
sredxny: I added this file into the gitignore file, it was introduced by mdcb testing framework and we dont need to keep track of changes on this file

gateway/rpc_storage_handler.go
sredxny: done

gateway/rpc_storage_handler.go
```diff
@@ -817,13 +820,69 @@ func (gw *Gateway) getSessionAndCreate(keyName string, r *RPCStorageHandler, isH
...
+		if err != nil {
+			log.Error("Could not get oath storage for api")
+			return
+		}
+		switch action {
+		case OauthClientAdded:
+			// on add: pull from rpc and save it in local redis
+			client, err := store.GetClient(oauthClientId)
+			if err != nil {
+				log.WithError(err).Error("Could not retrieve new oauth client information")
```
matiasinsaurralde, line 843: Should we `return` after this fails or the expected behavior is to continue?

gateway/rpc_storage_handler.go
```diff
@@ -817,13 +820,69 @@ func (gw *Gateway) getSessionAndCreate(keyName string, r *RPCStorageHandler, isH
...
+				log.WithError(err).Error("Could not retrieve oauth client information")
+			}
+			err = store.DeleteClient(oauthClientId, orgID, false)
+
+			if err != nil {
+				log.WithError(err).Error("Could not delete oauth client")
+			}
+			client, err = store.GetClient(oauthClientId)
+			if err != nil {
+				log.WithError(err).Error("Could not retrieve oauth client information")
```
matiasinsaurralde, line 868: Same question as the previous one, for all the above conditions.

matiasinsaurralde (approved):
Looks good, have added a few comments

gateway/rpc_storage_handler.go
sredxny: Thank you for pointing this out, indeed it should be a "continue" statement, so other events are processed as well. I just uploaded the change

PR #3721, opened 2021-11-04
[TT-3807] Omit declared but empty slice, map and pointer fields from OAS api

furkansenharputlu:
This PR fixes omitting of empty but declared slice, map and pointer cases in OAS structure. (`[]`, `{}`)

`json:"something, omitempty"` doesn't help in case of maps, arrays and objects because FE sends them declared but as empty. We would like to remove them as they don't have any value inside. We set them `nil` by manually checking with `shouldOmit` in the code.

I customized the built-int `reflect.Value.IsZero` as `IsZero`. The built-in function accepts slice, map and pointer fields having 0 length as not zero. In OAS, we would like them to be counted as empty so I separated slice, map and pointer to different cases.

### DIFF
--- a/apidef/oas/authentication.go
+++ b/apidef/oas/authentication.go
@@ -1,8 +1,6 @@
 package oas
 
 import (
-	"reflect"
-
 	"github.com/lonelycode/osin"
 
 	"github.com/TykTechnologies/tyk/apidef"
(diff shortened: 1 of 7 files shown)


vaske (approved):

jeffy-mathew (approved):

furkansenharputlu:
Manually merged to `release-4`

PR #3736, opened 2021-11-12
set golangci-lint issue exit code to 0

jeffy-mathew:
golangci-lint fails and returns non zero exit code, refer - https://github.com/TykTechnologies/tyk/runs/4179268017?check_suite_focus=true. 

It need not to be exited with a non zero exit code on that stage, instead let sonarcloud analyse the lint output


alephnull:
I don't think that there isn't much point in running a linter if the exit code is being ignored. Especially at this stage when the code is about to merged to master or a release branch. Alternatives to this PR in order of effort are:
- remove golangci-lint, just use sonarcloud
- move the golangci-lint to a pre-commit hook so that it runs locally and the person can decide what they want to do about it
- use the `--exclude*` options of golangci-lint to explicitly ignore non-issues

lghiur (approved):

lghiur:
@alephnull
1.  if there were linting errors, the process would exit (thing which we don't want). 
2. Sonarclould linter is pretty weak and that's why we want to use golangci-lint

alephnull:
So if I understand correctly, with this PR, we want,
- Sonarcloud webUI to use the golangci-lint report
- lint failures should no longer block merges
I'm happy to defer to the needs of the squads. I'm fine with whatever you decide.

alephnull (approved):

PR #3744, opened 2021-11-16
[TT-3880] Deprecate JWTScopeToPolicyMapping and JWTScopeToPolicyMapping

jeffy-mathew:
This PR intends to deprecate `JWTScopeToPolicyMapping` and `JWTScopeToPolicyMapping` and separate JWT and OIDC scopes. It also ensures backward compatibility


furkansenharputlu:
I think we need to add detailed backward compatibility tests. Creating an old way api, and see that it works, setting new fields on it and see that it works…

apidef/api_definitions.go
```diff
@@ -802,6 +813,14 @@ func (a *APIDefinition) EncodeForDB() {
 	if a.Auth.AuthHeaderName == "" {
 		a.Auth = a.AuthConfigs["authToken"]
 	}
+	// JWTScopeToPolicyMapping and JWTScopeClaimName are deprecated and following code ensures backward compatibility
+	if a.UseOpenID && a.Scopes.OIDC.ScopeToPolicy != nil {
```
furkansenharputlu, line 817: What happens if scopeToPolicy is empty but not nil. It might better if we check claim name is “” or not.

furkansenharputlu (approved):

vaske (approved):

PR #3749, opened 2021-11-22
[TT-3916] Goplugins for release-4

alephnull:
## Description
This was never sync'd to release-4.0 or master

## Related Issue
https://tyktech.atlassian.net/browse/TT-3916

## How This Has Been Tested
See the status in the _Checks_ tab.

# Merging
This PR is is for master, release-4.0 and release-4

PR #3750, opened 2021-11-22
[TT-3668] Build go plugin from dynamically generated directory 

jeffy-mathew:
## Description
While building multiple go plugin using tyk-plugin-compiler, the build path for plugin remains the same which causes Go to return error `plugin already loaded`. This PR makes it such that the build path will be a dynamically generated directory inside GOPATH.


matiasinsaurralde (approved):
Looks good to me, this is a replacement for the `pluginpath` approach we were previously using.
Randomizing the build path shouldn't affect the plugin compile flow.

vaske (approved):
ship it 🚀

furkansenharputlu (approved):

alephnull (commented):
Please merge this PR _after_ https://github.com/TykTechnologies/tyk/pull/3749

-For complete context, https://github.com/TykTechnologies/tyk/commit/b638cf3aa4f3c4c3953344de69fda962c01832bb#diff-eab9ce931973d7e6a000ccbd3dd3037310f38c30328dcd4846ac9449c10cc7df had a similar fix using `-pluginpath`, this was removed in https://github.com/TykTechnologies/tyk/commit/415abf8df69c508d23b881c1f07bb789c2e95ab0#diff-eab9ce931973d7e6a000ccbd3dd3037310f38c30328dcd4846ac9449c10cc7df as it didn't work for the latest 1.12 (still needed for LTS)

Does the smoke-test in https://github.com/TykTechnologies/tyk/tree/master/integration/smoke-tests/plugin-compiler pass with your change currently?

buger:
Pls check my Jira comment first.

alephnull (changes requested):
The smoke tests are run in automation by https://github.com/TykTechnologies/tyk/blob/master/.github/workflows/release.yml#L320-L347

The design is for each test to have its own directory and for each test to be independent of the others. The current implementation adds its test in `integration/smoke-tests/plugin-compiler` but I feel it should be in `integration/smoke-tests/plugin-aliasing`.

The reasons for asking for this is to separate the implementation of tests so that they are clear. 

Please also add some further text to https://github.com/TykTechnologies/tyk/blob/master/integration/smoke-tests/README.md describing the behaviour that is being tested. Right now it talks about OAuth2 and doesn't really make clear that the behaviour being tested is that of one plugin file being loaded in multiple APIs.

alephnull (approved):

PR #3751, opened 2021-11-23
TT-1799 - in case of error in signing method show proper error message

vaske:
Show proper error message in case of incorrect signing method


gateway/mw_jwt_test.go
```diff
@@ -382,6 +382,25 @@ func TestJWTSessionRSABearer(t *testing.T) {
...
+func TestJWTSessionFailRSA_WrongJWT_Signature(t *testing.T) {
+	ts := StartTest(nil)
+	defer ts.Close()
+
+	//default values, same as before (keeps backward compatibility)
+	ts.prepareGenericJWTSession(t.Name(), RSASign, KID, false)
+	authHeaders := map[string]string{"authorization": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"}
+
+	t.Run("Request with invalid JWT signature", func(t *testing.T) {
+		ts.Run(t, test.TestCase{
```
furkansenharputlu, line 395: linter complains here, maybe we should do this: `_, _ = ts.Run....`

gateway/mw_jwt_test.go
```diff
@@ -382,6 +382,25 @@ func TestJWTSessionRSABearer(t *testing.T) {
...
 }
 
+
+func TestJWTSessionFailRSA_WrongJWT_Signature(t *testing.T) {
+	ts := StartTest(nil)
+	defer ts.Close()
+
+	//default values, same as before (keeps backward compatibility)
+	ts.prepareGenericJWTSession(t.Name(), RSASign, KID, false)
+	authHeaders := map[string]string{"authorization": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"}
```
furkansenharputlu, line 392: can we move this to a constant? like: `invaliSigToken`

gateway/mw_jwt.go
```diff
@@ -792,7 +792,10 @@ func (k *JWTMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Request, _
 	k.reportLoginFailure(tykId, r)
 	if err != nil {
 		logger.WithError(err).Error("JWT validation error")
-		return errors.New("Key not authorized:Unexpected signing method."), http.StatusForbidden
+		errorDetails := strings.Split(err.Error(), ":")
+		if errorDetails[0] == "Unexpected signing method" {
+			return errors.New("Key not authorized:" + errorDetails[0] + "."), http.StatusForbidden
```
furkansenharputlu, line 797: errors should start lower case and shouldn't have `.` at the end

gateway/mw_jwt_test.go
vaske: will check

gateway/mw_jwt_test.go
vaske: yes

gateway/mw_jwt.go
vaske: I do agree but here it's a bit weird, since we use this middleware and return errors like that to API response, I wish that we refectory all of it and add proper messages we are about to return to users but this is at the moment big refactor thing

gateway/mw_jwt.go
```diff
@@ -792,7 +792,10 @@ func (k *JWTMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Request, _
 	k.reportLoginFailure(tykId, r)
 	if err != nil {
 		logger.WithError(err).Error("JWT validation error")
-		return errors.New("Key not authorized:Unexpected signing method."), http.StatusForbidden
+		errorDetails := strings.Split(err.Error(), ":")
+		if errorDetails[0] == "Unexpected signing method" {
```
furkansenharputlu, line 796: can we move this to a constant? Like here: https://github.com/TykTechnologies/tyk/blob/0ed167f12f96fb684d369df4eb72823d30b74943/gateway/handler_error.go#L31

gateway/mw_jwt_test.go
furkansenharputlu: It would be better to add the constant inside the test instead of global.

gateway/mw_jwt.go
vaske: we want here generic message, but yes we can use constant, btw do we have general approach about how we want to use text constants? maybe in future we want to have full language support with translations etc..

gateway/mw_jwt_test.go
vaske: it's weird since it's not complaining on other parts with same code

gateway/mw_jwt_test.go
```diff
@@ -45,6 +46,8 @@ jQIDAQAB!!!!
 -----END PUBLIC KEY-----
 `
 
+const invalidSignToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c"
```
furkansenharputlu, line 49: `invalidSignToken` is specific to `TestJWTSessionFailRSA_WrongJWT_Signature`. We can move it to the test.

gateway/mw_jwt_test.go
```diff
@@ -382,6 +383,24 @@ func TestJWTSessionRSABearer(t *testing.T) {
...
+
+	//default values, same as before (keeps backward compatibility)
+	ts.prepareGenericJWTSession(t.Name(), RSASign, KID, false)
+	authHeaders := map[string]string{"authorization": invalidSignToken}
+
+	t.Run("Request with invalid JWT signature", func(t *testing.T) {
+		_, _ = ts.Run(t, test.TestCase{
+			Headers:   authHeaders,
+			Code:      http.StatusForbidden,
+			BodyMatch: `Key not authorized: Unexpected signing method`,
```
jeffy-mathew, line 399: can we reuse constant `MsgKeyNotAuthorizedUnexpectedSigningMethod`?

jeffy-mathew (approved):

lghiur (approved):

PR #3755, opened 2021-11-26
[TT-3970] Override with deprecated auth when jwt or auth token is enabled

furkansenharputlu:
This is an old hidden bug appeared with OAS.

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -874,20 +874,20 @@ func (a *APIDefinition) DecodeFromDB() {
 	}
 
 	// Auth is deprecated so this code tries to maintain backward compatibility
-	makeCompatible := func(authType string) {
+	makeCompatible := func(authType string, enabled bool) {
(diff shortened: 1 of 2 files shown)


apidef/api_definitions.go
```diff
@@ -875,20 +875,20 @@ func (a *APIDefinition) DecodeFromDB() {
...
-	makeCompatible := func(authType string) {
+	makeCompatible := func(authType string, enabled bool) {
 		if a.AuthConfigs == nil {
 			a.AuthConfigs = make(map[string]AuthConfig)
 		}
 
 		_, ok := a.AuthConfigs[authType]
 
-		if !ok {
+		if !ok && enabled {
```
vaske, line 885: do we need to cover case if it's not ok and enabled is false? or w are good like this?

vaske (approved):
LGTM, I left one minor comment but beside of that all is good

jeffy-mathew (approved):

PR #3759, opened 2021-11-30
Add full XTykAPIGateway conversion test

furkansenharputlu:
Filling old API definition and converting to OAS and reconverting test is skipped because all conversions are not ready yet. However, the reverse of it is possible. We can fill OAS and convert it to old API definition and reconverting to OAS. This PR implements that test and fix bugs which cause the test to fail.

### DIFF
--- a/apidef/oas/authentication.go
+++ b/apidef/oas/authentication.go
@@ -1,6 +1,8 @@
 package oas
 
 import (
+	"sort"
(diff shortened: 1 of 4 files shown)


jeffy-mathew (approved):

vaske (approved):
LGTM

PR #3763, opened 2021-12-01
TT-3906 - additional logic in case of non existing cert

vaske:
this is workaround missing cert

### DIFF
--- a/gateway/mw_auth_key.go
+++ b/gateway/mw_auth_key.go
@@ -96,7 +96,7 @@ func (k *AuthKey) ProcessRequest(_ http.ResponseWriter, r *http.Request, _ inter
 		key = stripBearer(key)
 	} else if authConfig.UseCertificate && key == "" && r.TLS != nil && len(r.TLS.PeerCertificates) > 0 {
 		log.Debug("Trying to find key by client certificate")
-		certHash = certs.HexSHA256(r.TLS.PeerCertificates[0].Raw)
+		certHash = k.Spec.OrgID + certs.HexSHA256(r.TLS.PeerCertificates[0].Raw)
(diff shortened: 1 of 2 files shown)


gateway/mw_auth_key.go
```diff
@@ -91,11 +91,12 @@ func (k *AuthKey) ProcessRequest(_ http.ResponseWriter, r *http.Request, _ inter
 
 	keyExists := false
 	var session user.SessionState
+	updateSession := false
 	if key != "" {
 		key = stripBearer(key)
 	} else if authConfig.UseCertificate && key == "" && r.TLS != nil && len(r.TLS.PeerCertificates) > 0 {
 		log.Debug("Trying to find key by client certificate")
-		certHash = certs.HexSHA256(r.TLS.PeerCertificates[0].Raw)
+		certHash = k.Spec.OrgID + certs.HexSHA256(r.TLS.PeerCertificates[0].Raw)
```
vaske, line 99: we have to concatenate orgId + hash since all cert hashes are build like that

gateway/mw_auth_key.go
```diff
@@ -107,22 +108,31 @@ func (k *AuthKey) ProcessRequest(_ http.ResponseWriter, r *http.Request, _ inter
 	if !keyExists {
 		// fallback to search by cert
 		session, keyExists = k.CheckSessionAndIdentityForValidKey(certHash, r)
-		certHash = session.KeyID
```
vaske, line 110: this override is not necessary here otherwise in case of mTLS it will be always diff than `session.Certificate`

buger (approved):
👍🏻

jeffy-mathew (approved):

vaske:
merged manually to 3lts

PR #3789, opened 2021-12-16
[TT-4163]  Implement new api expiration feature

furkansenharputlu:
This PR implements new expiration feature. It adds a flag named `expiration` to api definition. Previous expiration feature works only when versioning is enabled. However, this one will work no matter new versioning is enabled or not.

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -480,6 +480,8 @@ type Scopes struct {
 type APIDefinition struct {
 	Id                  ObjectId      `bson:"_id,omitempty" json:"id,omitempty" gorm:"primaryKey;column:_id"`
 	Name                string        `bson:"name" json:"name"`
+	Expiration          string        `bson:"expiration" json:"expiration"`
+	ExpirationTs        time.Time     `bson:"-" json:"-"`
 	Slug                string        `bson:"slug" json:"slug"`
 	ListenPort          int           `bson:"listen_port" json:"listen_port"`
 	Protocol            string        `bson:"protocol" json:"protocol"`
--- a/gateway/api_definition.go
+++ b/gateway/api_definition.go
@@ -263,13 +263,15 @@ func (a APIDefinitionLoader) MakeSpec(def *apidef.APIDefinition, logger *logrus.
 		logger = logrus.NewEntry(log)
 	}
 
-	if t, err := time.Parse(expiredTimeFormat, def.Expiration); err != nil {
+	// new expiration feature
+	if t, err := time.Parse(apidef.ExpirationTimeFormat, def.Expiration); err != nil {
 		logger.WithError(err).WithField("name", def.Name).WithField("Expiration", def.Expiration).
 			Error("Could not parse expiration date for API")
 	} else {
 		def.ExpirationTs = t
 	}
 
+	// Deprecated
 	// parse version expiration time stamps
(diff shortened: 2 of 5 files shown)


gateway/api_definition.go
```diff
@@ -262,6 +263,15 @@ func (a APIDefinitionLoader) MakeSpec(def *apidef.APIDefinition, logger *logrus.
 		logger = logrus.NewEntry(log)
 	}
 
+	// new expiration feature
+	if t, err := time.Parse(apidef.ExpirationTimeFormat, def.Expiration); err != nil {
+		logger.WithError(err).WithField("name", def.Name).WithField("Expiration", def.Expiration).
```
vaske, line 268: code scanning is complain here to receive unsanitized user input, I think we an leave with it for now

apidef/api_definitions.go
```diff
@@ -480,6 +480,8 @@ type Scopes struct {
 type APIDefinition struct {
 	Id                  ObjectId      `bson:"_id,omitempty" json:"id,omitempty" gorm:"primaryKey;column:_id"`
 	Name                string        `bson:"name" json:"name"`
+	Expiration          string        `bson:"expiration" json:"expiration"`
+	ExpirationTs        time.Time     `bson:"-" json:"-"`
```
jeffy-mathew, line 484: can we use `ExpirationTs` as `time.Time` pointer, so that we can exclude `Expiration`?

gateway/api_definition.go
furkansenharputlu: An example that I found to solve it like this:
`if (!Pattern.matches("[A-Za-z0-9_]+", username)) `

I think it makes the code complex. Is there any other simpler solution that you know?

apidef/api_definitions.go
furkansenharputlu: When we use `time.Time`, it is marshalled as:
`"expiration": "2021-12-20T12:52:29.035661+03:00"`

but we are using a special format in old expiration as:
`"expiration": "2021-12-20 12:52"`

If we don't do that, we will break backward compatibility

gateway/api_definition.go
lghiur: why does it makes complex @furkansenharputlu ?

gateway/api_definition_test.go
```diff
@@ -1037,3 +1038,37 @@ func TestAPIDefinitionLoader_Template(t *testing.T) {
...
+		spec.UseKeylessAccess = true
+		spec.VersionData.NotVersioned = true
+		spec.VersionDefinition.Enabled = true
+	})[0]
+
+	for _, versioned := range []bool{false, true} {
+		api.VersionDefinition.Enabled = versioned
+
+		t.Run(fmt.Sprintf("versioning=%v", versioned), func(t *testing.T) {
+			t.Run("not expired", func(t *testing.T) {
```
lghiur, line 1057: @furkansenharputlu I still think we should provide more details for the tests, I would like to be able to read the description to understand what's being tested

gateway/api_definition.go
furkansenharputlu: We are adding extra check if api name is in this format. It is unnecessary. However, if you tell me an important reason that I miss, I can try to add that.

vaske (approved):

jeffy-mathew (approved):

PR #3794, opened 2021-12-20
[TT-4035] Release 4.0 plugin compiler workaround

jeffy-mathew:
This PR intends to fix - `plugin.Open("plugin"): plugin was built with a different version of package` for gw dependencies. 
This issue won't occur locally if a replace directive is used in go.mod to replace the gw dependency with a local copy.

Since the gw is built in our release pipeline in github actions workspace directory `__w/tyk/tyk`, we can clone the corresponding gw version into `__w/tyk/tyk` and use a replace directive for gw while building the plugin and it'll be fixed. 

Related issue - https://github.com/golang/go/issues/31354

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -21,6 +21,7 @@ builds:
 
     flags:
       - -tags=goplugin
+      - -trimpath
     ldflags:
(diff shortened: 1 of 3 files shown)


alephnull (approved):
Looks good as long as the tests pass.

PR #3807, opened 2021-12-24
TT-4091 - OAS Caching Paths support

vaske:
This PR:
- Adds `disabled` flag to cache plugin
- Implements migration from old cache string array to new advanced cache config
- Implements OAS conversion for advanced cache config

### DIFF
--- a/apidef/migration.go
+++ b/apidef/migration.go
@@ -206,3 +206,37 @@ func (a *APIDefinition) Migrate() (versions []APIDefinition, err error) {
 
 	return versions, nil
 }
+
+func (a *APIDefinition) MigrateCachePlugin() (err error) {
+	if a.VersionDefinition.Enabled || len(a.VersionDefinition.Versions) != 0 {
+		return errors.New("not migratable - new versioning is enabled")
+	}
+
+	if a.VersionData.NotVersioned && len(a.VersionData.Versions) > 1 {
+		return errors.New("not migratable - if not versioned, there should be 1 version info in versions map")
+	}
+
+	for vName, vInfo := range a.VersionData.Versions {
+		var updated = false
+		if vInfo.UseExtendedPaths && len(vInfo.ExtendedPaths.Cached) > 0 {
+			var methods = []CacheMeta{}
+			for _, cache := range vInfo.ExtendedPaths.Cached {
+				newCache := CacheMeta{
+					Path:     cache,
+					Disabled: false,
+				}
+				methods = append(methods, newCache)
+			}
+			vInfo.ExtendedPaths.AdvanceCacheConfig = methods
+			updated = true
+		}
+
+		if updated {
+			delete(a.VersionData.Versions, vName)
+			a.VersionData.Versions[vName] = vInfo
+		}
+
+	}
+
+	return
+}
--- a/apidef/migration.go
+++ b/apidef/migration.go
@@ -201,19 +201,21 @@ func (a *APIDefinition) Migrate() (versions []APIDefinition, err error) {
 	}
 
 	a.MigrateEndpointMeta()
+	a.MigrateCachePlugin()
 	for i := 0; i < len(versions); i++ {
 		versions[i].MigrateEndpointMeta()
+		a.MigrateCachePlugin()
 	}
 
 	return versions, nil
 }
 
-func (a *APIDefinition) MigrateCachePlugin() (err error) {
+func (a *APIDefinition) MigrateCachePlugin() {
 	vInfo := a.VersionData.Versions[""]
 	list := vInfo.ExtendedPaths.Cached
 
 	if vInfo.UseExtendedPaths && len(list) > 0 {
-		var advCacheMethods = []CacheMeta{}
+		var advCacheMethods []CacheMeta
 		for _, cache := range list {
 			newGetMethodCache := CacheMeta{
 				Path:   cache,
@@ -236,6 +238,4 @@ func (a *APIDefinition) MigrateCachePlugin() (err error) {
 	}
 
 	a.VersionData.Versions[""] = vInfo
-
-	return nil
 }
--- a/apidef/migration.go
+++ b/apidef/migration.go
@@ -2,6 +2,7 @@ package apidef
 
 import (
 	"errors"
+	"net/http"
 	"net/url"
 	"sort"
 	"strings"
@@ -214,11 +215,22 @@ func (a *APIDefinition) MigrateCachePlugin() (err error) {
 	if vInfo.UseExtendedPaths && len(list) > 0 {
 		var advCacheMethods = []CacheMeta{}
 		for _, cache := range list {
-			newCache := CacheMeta{
+			newGetMethodCache := CacheMeta{
 				Path:     cache,
 				Disabled: false,
+				Method:   http.MethodGet,
 			}
-			advCacheMethods = append(advCacheMethods, newCache)
+			newHeadMethodCache := CacheMeta{
+				Path:     cache,
+				Disabled: false,
+				Method:   http.MethodHead,
+			}
+			newOptionsMethodCache := CacheMeta{
+				Path:     cache,
+				Disabled: false,
+				Method:   http.MethodOptions,
+			}
+			advCacheMethods = append(advCacheMethods, newGetMethodCache, newHeadMethodCache, newOptionsMethodCache)
 		}
 		vInfo.ExtendedPaths.AdvanceCacheConfig = advCacheMethods
 	}
--- a/apidef/migration.go
+++ b/apidef/migration.go
@@ -216,24 +216,25 @@ func (a *APIDefinition) MigrateCachePlugin() (err error) {
 		var advCacheMethods = []CacheMeta{}
 		for _, cache := range list {
 			newGetMethodCache := CacheMeta{
-				Path:     cache,
-				Disabled: false,
-				Method:   http.MethodGet,
+				Path:   cache,
+				Method: http.MethodGet,
 			}
 			newHeadMethodCache := CacheMeta{
-				Path:     cache,
-				Disabled: false,
-				Method:   http.MethodHead,
+				Path:   cache,
+				Method: http.MethodHead,
 			}
 			newOptionsMethodCache := CacheMeta{
-				Path:     cache,
-				Disabled: false,
-				Method:   http.MethodOptions,
+				Path:   cache,
+				Method: http.MethodOptions,
 			}
 			advCacheMethods = append(advCacheMethods, newGetMethodCache, newHeadMethodCache, newOptionsMethodCache)
 		}
+
 		vInfo.ExtendedPaths.AdvanceCacheConfig = advCacheMethods
+		// reset cache to empty
+		vInfo.ExtendedPaths.Cached = nil
 	}
+
 	a.VersionData.Versions[""] = vInfo
 
 	return nil
--- a/apidef/migration.go
+++ b/apidef/migration.go
@@ -208,35 +208,21 @@ func (a *APIDefinition) Migrate() (versions []APIDefinition, err error) {
 }
 
 func (a *APIDefinition) MigrateCachePlugin() (err error) {
-	if a.VersionDefinition.Enabled || len(a.VersionDefinition.Versions) != 0 {
-		return errors.New("not migratable - new versioning is enabled")
-	}
+	vInfo := a.VersionData.Versions[""]
+	list := vInfo.ExtendedPaths.Cached
 
-	if a.VersionData.NotVersioned && len(a.VersionData.Versions) > 1 {
-		return errors.New("not migratable - if not versioned, there should be 1 version info in versions map")
-	}
-
-	for vName, vInfo := range a.VersionData.Versions {
-		var updated = false
-		if vInfo.UseExtendedPaths && len(vInfo.ExtendedPaths.Cached) > 0 {
-			var methods = []CacheMeta{}
-			for _, cache := range vInfo.ExtendedPaths.Cached {
-				newCache := CacheMeta{
-					Path:     cache,
-					Disabled: false,
-				}
-				methods = append(methods, newCache)
+	if vInfo.UseExtendedPaths && len(list) > 0 {
+		var advCacheMethods = []CacheMeta{}
+		for _, cache := range list {
+			newCache := CacheMeta{
+				Path:     cache,
+				Disabled: false,
 			}
-			vInfo.ExtendedPaths.AdvanceCacheConfig = methods
-			updated = true
+			advCacheMethods = append(advCacheMethods, newCache)
 		}
-
-		if updated {
-			delete(a.VersionData.Versions, vName)
-			a.VersionData.Versions[vName] = vInfo
-		}
-
+		vInfo.ExtendedPaths.AdvanceCacheConfig = advCacheMethods
 	}
+	a.VersionData.Versions[""] = vInfo
 
-	return
+	return nil
 }
--- a/apidef/migration_test.go
+++ b/apidef/migration_test.go
@@ -324,7 +324,7 @@ func TestAPIDefinition_MigrateEndpointMeta(t *testing.T) {
 }
 
 func TestAPIDefinition_MigrateCache(t *testing.T) {
-	name := "Default"
+	name := ""
 
 	versionInfo := VersionInfo{
 		Name:             name,
@@ -357,8 +357,22 @@ func TestAPIDefinition_MigrateCache(t *testing.T) {
 		path1 := defaultVersionInfo.ExtendedPaths.AdvanceCacheConfig[0]
 		path1Cache := defaultVersionInfo.ExtendedPaths.Cached[0]
 
+		expectedCachedDefault := []string{"test"}
+		cacheItem := CacheMeta{
+			Method:        "",
+			Disabled:      false,
+			Path:          "test",
+			CacheKeyRegex: "",
+		}
+		expectedAdvCacheMethods := []CacheMeta{
+			cacheItem,
+		}
+
 		assert.Equal(t, len(defaultVersionInfo.ExtendedPaths.AdvanceCacheConfig), len(defaultVersionInfo.ExtendedPaths.Cached))
 		assert.Equal(t, path1.Path, path1Cache)
+
+		assert.Equal(t, expectedCachedDefault, defaultVersionInfo.ExtendedPaths.Cached)
+		assert.Equal(t, expectedAdvCacheMethods, defaultVersionInfo.ExtendedPaths.AdvanceCacheConfig)
 	}
 
 	t.Run("check migration of old data to new", func(t *testing.T) {
--- a/apidef/migration_test.go
+++ b/apidef/migration_test.go
@@ -323,37 +323,23 @@ func TestAPIDefinition_MigrateEndpointMeta(t *testing.T) {
 	assert.Equal(t, expectedMockResponse, api.VersionData.Versions[""].ExtendedPaths.MockResponse)
 }
 
-func TestAPIDefinition_MigrateCache(t *testing.T) {
-
+func TestAPIDefinition_MigrateCachePlugin(t *testing.T) {
 	versionInfo := VersionInfo{
-		Name:             "",
 		UseExtendedPaths: true,
 		ExtendedPaths: ExtendedPathsSet{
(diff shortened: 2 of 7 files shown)


apidef/migration_test.go
```diff
@@ -205,3 +205,48 @@ func TestAPIDefinition_MigrateVersioning_StripPath(t *testing.T) {
...
+		}
+	}
+
+	check := func(t *testing.T, base APIDefinition) {
+		err := base.MigrateCachePlugin()
+		assert.NoError(t, err)
+
+		defaultVersionInfo := base.VersionData.Versions[base.VersionData.DefaultVersion]
+		path1 := defaultVersionInfo.ExtendedPaths.AdvanceCacheConfig[0]
+		path1Cache := defaultVersionInfo.ExtendedPaths.Cached[0]
```
furkansenharputlu, line 241: Not just one field check. Instead, we should write an `expectedAdvCache` and compare it with actual.

apidef/migration.go
```diff
@@ -78,3 +78,37 @@ func (a *APIDefinition) MigrateVersioning() (versions []APIDefinition, err error
...
 	return
 }
+
+func (a *APIDefinition) MigrateCachePlugin() (err error) {
+	if a.VersionDefinition.Enabled || len(a.VersionDefinition.Versions) != 0 {
+		return errors.New("not migratable - new versioning is enabled")
+	}
+
+	if a.VersionData.NotVersioned && len(a.VersionData.Versions) > 1 {
+		return errors.New("not migratable - if not versioned, there should be 1 version info in versions map")
```
furkansenharputlu, line 88: This part is already implemented in versioning function. No need to call here again

apidef/migration.go
```diff
@@ -78,3 +78,37 @@ func (a *APIDefinition) MigrateVersioning() (versions []APIDefinition, err error
...
+	}
+
+	for vName, vInfo := range a.VersionData.Versions {
+		var updated = false
+		if vInfo.UseExtendedPaths && len(vInfo.ExtendedPaths.Cached) > 0 {
+			var methods = []CacheMeta{}
```
furkansenharputlu, line 94: `methods` can be named as `advCacheMetas`

apidef/migration.go
```diff
@@ -78,3 +78,37 @@ func (a *APIDefinition) MigrateVersioning() (versions []APIDefinition, err error
...
+func (a *APIDefinition) MigrateCachePlugin() (err error) {
+	if a.VersionDefinition.Enabled || len(a.VersionDefinition.Versions) != 0 {
+		return errors.New("not migratable - new versioning is enabled")
+	}
+
+	if a.VersionData.NotVersioned && len(a.VersionData.Versions) > 1 {
+		return errors.New("not migratable - if not versioned, there should be 1 version info in versions map")
+	}
+
+	for vName, vInfo := range a.VersionData.Versions {
```
furkansenharputlu, line 91: We don't need to iterate over versions. As we will call this function after migrating versioning, every definition will have just one vInfo like `a.VersionData.Versions[""]`. See https://github.com/TykTechnologies/tyk/pull/3805/files#diff-e1d9b55a26f9d6225d56d6f0161959217308e5ad4d6934e7d7df4595d9c2a130R104

apidef/migration.go
```diff
@@ -78,3 +78,37 @@ func (a *APIDefinition) MigrateVersioning() (versions []APIDefinition, err error
...
+		var updated = false
+		if vInfo.UseExtendedPaths && len(vInfo.ExtendedPaths.Cached) > 0 {
+			var methods = []CacheMeta{}
+			for _, cache := range vInfo.ExtendedPaths.Cached {
+				newCache := CacheMeta{
+					Path:     cache,
```
furkansenharputlu, line 97: I know in old cache there is no `method` but how will we write it in OAS format if method is empty? We need to find a solution here.

apidef/migration.go
```diff
@@ -78,3 +78,37 @@ func (a *APIDefinition) MigrateVersioning() (versions []APIDefinition, err error
...
+				methods = append(methods, newCache)
+			}
+			vInfo.ExtendedPaths.AdvanceCacheConfig = methods
+			updated = true
+		}
+
+		if updated {
+			delete(a.VersionData.Versions, vName)
+			a.VersionData.Versions[vName] = vInfo
+		}
```
furkansenharputlu, line 109: No need this part

apidef/migration.go
furkansenharputlu: As cache works only for the safe methods, `GET`,`HEAD`, `OPTIONS`. We need to add 3 advance cache for all.

https://github.com/TykTechnologies/tyk/blob/f603dec31b91ae9c7221c990b078d82aa04bfece/gateway/mw_redis_cache.go#L377-L379

apidef/oas/middleware.go
```diff
@@ -429,3 +432,76 @@ func (a *Allowance) ExtractTo(endpointMeta *apidef.EndPointMeta) {
...
+	}
+}
+
+type CacheMiddleware struct {
+	Enabled            bool   `bson:"enabled" json:"enabled"`
+	CacheByRegex       string `bson:"cacheByRegex,omitempty" json:"cacheByRegex,omitempty"`
+	CacheResponseCodes []int  `bson:"cacheResponseCodes,omitempty" json:"cacheResponseCodes,omitempty"`
+}
+
+func (a *CacheMiddleware) Fill(cm apidef.CacheMeta) {
```
furkansenharputlu, line 497: Where is this called?

apidef/oas/middleware.go
```diff
@@ -429,3 +432,76 @@ func (a *Allowance) ExtractTo(endpointMeta *apidef.EndPointMeta) {
...
+	a.Enabled = !cm.Disabled
+	a.CacheByRegex = cm.CacheKeyRegex
+	a.CacheResponseCodes = cm.CacheOnlyResponseCodes
+}
+
+func (a *CacheMiddleware) ExtractTo(endpointMeta *apidef.CacheMeta) {
```
furkansenharputlu, line 503: where is this called?

vaske: This is not called but I made it due to similar semantic we do everywhere

apidef/oas/middleware.go
vaske: on line 452 in same file.

apidef/migration.go
vaske: from where we can get info about method?

furkansenharputlu: As I said above, you don’t get info about method. We need to write it to 3 safe methods.

apidef/migration_test.go
```diff
@@ -315,3 +315,79 @@ func TestAPIDefinition_MigrateEndpointMeta(t *testing.T) {
...
+			cacheItemHead,
+			cacheItemOptions,
+		}
+
+		// for every cache item we have to have 3 new AdvanceCacheConfig items
+		// since we do support caching only for safe methods GET, OPTIONS and HEAD
+		assert.Equal(t, len(defaultVersionInfo.ExtendedPaths.AdvanceCacheConfig), len(defaultVersionInfo.ExtendedPaths.Cached)*3)
+		assert.Equal(t, path1.Path, path1Cache)
+
+		assert.Equal(t, expectedCachedDefault, defaultVersionInfo.ExtendedPaths.Cached)
```
furkansenharputlu, line 384: After migration the old one should be empty.

apidef/oas/middleware.go
```diff
@@ -179,6 +179,7 @@ func (ps Paths) Fill(ep apidef.ExtendedPathsSet) {
 	ps.fillAllowance(ep.BlackList, block)
 	ps.fillAllowance(ep.Ignored, ignoreAuthentication)
 	ps.fillMockResponse(ep.MockResponse)
+	ps.fillCache(ep)
```
furkansenharputlu, line 182: Can we make `ep.AdvanceCacheConfig` like others?

apidef/oas/middleware.go
```diff
@@ -475,3 +478,76 @@ type Header struct {
 	Name  string `bson:"name" json:"name"`
 	Value string `bson:"value" json:"value"`
 }
+
+func (ps Paths) fillCache(ep apidef.ExtendedPathsSet) {
+	advanceCacheConfig := &ep.AdvanceCacheConfig
+
+	for _, advCache := range *advanceCacheConfig {
```
furkansenharputlu, line 485: we can take `cacheMetas []CacheMeta` as param. So we can get rid of 483.
`for _, cm := range cacheMetas {`

apidef/oas/middleware.go
```diff
@@ -475,3 +478,76 @@ type Header struct {
...
+		if ps[advCache.Path] != nil {
+			plugins := ps[advCache.Path].getMethod(advCache.Method)
+			plugins.fillCache(advCache)
+
+		}
+	}
+}
+
+func (p *Plugins) fillCache(cacheMeta apidef.CacheMeta) {
```
furkansenharputlu, line 494: We can get rid of this one if we apply above logic. That approach is used in other plugins too, it is good for consistency.

apidef/oas/middleware.go
```diff
@@ -475,3 +478,76 @@ type Header struct {
...
+		newCache.CacheKeyRegex = p.Cache.CacheByRegex
+		newCache.Method = method
+		newCache.Path = path
+
+		advanceCacheConfig = append(advanceCacheConfig, newCache)
+		ep.AdvanceCacheConfig = advanceCacheConfig
+		return
+	}
+
+	founded := false
```
furkansenharputlu, line 517: Why did we implement this part?

furkansenharputlu: Can we do the same with below ? Let's keep consistency.
```
func (p *Plugins) extractMockResponseTo(ep *apidef.ExtendedPathsSet, path string, method string) {
	if p.MockResponse == nil {
		return
	}

	mockMeta := apidef.MockResponseMeta{Path: path, Method: method}
	p.MockResponse.ExtractTo(&mockMeta)
	ep.MockResponse = append(ep.MockResponse, mockMeta)
}
```

apidef/oas/middleware.go
```diff
@@ -475,3 +478,76 @@ type Header struct {
...
+		newCache.CacheKeyRegex = p.Cache.CacheByRegex
+		newCache.Method = method
+		newCache.Path = path
+
+		advanceCacheConfig = append(advanceCacheConfig, newCache)
+		ep.AdvanceCacheConfig = advanceCacheConfig
+	}
+}
+
+type CacheMiddleware struct {
```
furkansenharputlu, line 537: Let's rename it as `CachePlugin`, cache middleware is the one in the global level.

gateway/mw_version_check_test.go
```diff
@@ -463,3 +464,60 @@ func TestOldVersioning_Expires(t *testing.T) {
 		})
 	})
 }
+
+func TestOld_Cache(t *testing.T) {
```
furkansenharputlu, line 468: `TestOldCache`

gateway/mw_version_check_test.go
```diff
@@ -463,3 +464,60 @@ func TestOldVersioning_Expires(t *testing.T) {
...
+
+	ts.Gw.LoadAPI(api())
+	_, _ = ts.Run(t, test.TestCase{Path: "/test", BodyMatch: "default", Code: http.StatusOK})
+
+	response = "updated"
+
+	check(t, api(), []test.TestCase{
+		{Path: "/test", BodyMatch: "default", Code: http.StatusOK},
+		{Path: "/anything", BodyMatch: "updated", Code: http.StatusOK},
+	})
```
furkansenharputlu, line 522: This part should be `check` and we need to extract `t.Run("migration"` outside and call `check` inside migration again.

apidef/migration_test.go
```diff
@@ -315,3 +315,79 @@ func TestAPIDefinition_MigrateEndpointMeta(t *testing.T) {
...
+				NotVersioned:   false,
+				DefaultVersion: name,
+				Versions: map[string]VersionInfo{
+					name: versionInfo,
+				},
+			},
+		}
+	}
+
+	check := func(t *testing.T, base APIDefinition) {
```
furkansenharputlu, line 345: separate check function is not necessary here because you call it once

apidef/migration_test.go
```diff
@@ -315,3 +315,79 @@ func TestAPIDefinition_MigrateEndpointMeta(t *testing.T) {
...
+		// for every cache item we have to have 3 new AdvanceCacheConfig items
+		// since we do support caching only for safe methods GET, OPTIONS and HEAD
+		assert.Equal(t, len(defaultVersionInfo.ExtendedPaths.AdvanceCacheConfig), len(defaultVersionInfo.ExtendedPaths.Cached)*3)
+		assert.Equal(t, path1.Path, path1Cache)
+
+		assert.Equal(t, expectedCachedDefault, defaultVersionInfo.ExtendedPaths.Cached)
+		assert.Equal(t, expectedAdvCacheMethods, defaultVersionInfo.ExtendedPaths.AdvanceCacheConfig)
+	}
+
+	t.Run("check migration of old data to new", func(t *testing.T) {
```
furkansenharputlu, line 388: why did we put it inside a sub-test? This is an unnecessary layer. If we'd test more than one thing we'd add but this layer is the main one.

apidef/migration_test.go
```diff
@@ -315,3 +315,79 @@ func TestAPIDefinition_MigrateEndpointMeta(t *testing.T) {
...
+			Path:          "test",
+			CacheKeyRegex: "",
+			Method:        http.MethodHead,
+		}
+		cacheItemOptions := CacheMeta{
+			Disabled:      false,
+			Path:          "test",
+			CacheKeyRegex: "",
+			Method:        http.MethodOptions,
+		}
```
furkansenharputlu, line 372: Let's get rid of redundancy:
```
cacheItemHead:=cacheItemGet
cacheItemHead.Method = http.MethodHead

cacheItemOptions:=cacheItemGet
cacheItemOptions.Method = http.MethodOptions
```

apidef/migration_test.go
```diff
@@ -315,3 +315,79 @@ func TestAPIDefinition_MigrateEndpointMeta(t *testing.T) {
...
+		expectedAdvCacheMethods := []CacheMeta{
+			cacheItemGet,
+			cacheItemHead,
+			cacheItemOptions,
+		}
+
+		// for every cache item we have to have 3 new AdvanceCacheConfig items
+		// since we do support caching only for safe methods GET, OPTIONS and HEAD
+		assert.Equal(t, len(defaultVersionInfo.ExtendedPaths.AdvanceCacheConfig), len(defaultVersionInfo.ExtendedPaths.Cached)*3)
+		assert.Equal(t, path1.Path, path1Cache)
```
furkansenharputlu, line 382: In the below, we already check, these can be removed I think
```
assert.Equal(t, expectedAdvCacheMethods, defaultVersionInfo.ExtendedPaths.AdvanceCacheConfig)
```

apidef/migration_test.go
```diff
@@ -315,3 +315,79 @@ func TestAPIDefinition_MigrateEndpointMeta(t *testing.T) {
 	assert.Equal(t, expectedWhitelist, api.VersionData.Versions[""].ExtendedPaths.Ignored)
 	assert.Equal(t, expectedMockResponse, api.VersionData.Versions[""].ExtendedPaths.MockResponse)
 }
+
+func TestAPIDefinition_MigrateCache(t *testing.T) {
+	name := ""
```
furkansenharputlu, line 320: ?

apidef/migration.go
```diff
@@ -188,3 +189,34 @@ func (a *APIDefinition) Migrate() (versions []APIDefinition, err error) {
...
+				Disabled: false,
+				Method:   http.MethodHead,
+			}
+			newOptionsMethodCache := CacheMeta{
+				Path:     cache,
+				Disabled: false,
+				Method:   http.MethodOptions,
+			}
+			advCacheMethods = append(advCacheMethods, newGetMethodCache, newHeadMethodCache, newOptionsMethodCache)
+		}
```
furkansenharputlu, line 216: let's skip a line after `for`

apidef/migration.go
```diff
@@ -188,3 +189,34 @@ func (a *APIDefinition) Migrate() (versions []APIDefinition, err error) {
...
+		vInfo.ExtendedPaths.AdvanceCacheConfig = advCacheMethods
+	}
```
furkansenharputlu, line 218: let's skip a line after `if`

apidef/migration_test.go
```diff
@@ -315,3 +315,79 @@ func TestAPIDefinition_MigrateEndpointMeta(t *testing.T) {
...
+
+		assert.Equal(t, expectedCachedDefault, defaultVersionInfo.ExtendedPaths.Cached)
+		assert.Equal(t, expectedAdvCacheMethods, defaultVersionInfo.ExtendedPaths.AdvanceCacheConfig)
+	}
+
+	t.Run("check migration of old data to new", func(t *testing.T) {
+		base := old()
+		check(t, base)
+	})
+
```
furkansenharputlu, line 392: unnecessary empty line

apidef/migration.go
```diff
@@ -188,3 +189,34 @@ func (a *APIDefinition) Migrate() (versions []APIDefinition, err error) {
...
+func (a *APIDefinition) MigrateCachePlugin() (err error) {
+	vInfo := a.VersionData.Versions[""]
+	list := vInfo.ExtendedPaths.Cached
+
+	if vInfo.UseExtendedPaths && len(list) > 0 {
+		var advCacheMethods = []CacheMeta{}
+		for _, cache := range list {
+			newGetMethodCache := CacheMeta{
+				Path:     cache,
+				Disabled: false,
```
furkansenharputlu, line 202: `Disabled`s are already `false`. Why would we add?

apidef/oas/middleware.go
```diff
@@ -475,3 +478,76 @@ type Header struct {
...
+	CacheResponseCodes []int  `bson:"cacheResponseCodes,omitempty" json:"cacheResponseCodes,omitempty"`
+}
+
+func (a *CacheMiddleware) Fill(cm apidef.CacheMeta) {
+	a.Enabled = !cm.Disabled
+	a.CacheByRegex = cm.CacheKeyRegex
+	a.CacheResponseCodes = cm.CacheOnlyResponseCodes
+}
+
+func (a *CacheMiddleware) ExtractTo(endpointMeta *apidef.CacheMeta) {
```
furkansenharputlu, line 549: `endpointMeta` -> `cm`

apidef/migration_test.go
vaske: yeah but I applied here logic the more the merrier :)

apidef/migration_test.go
vaske: leftover will put inside then exposed here

apidef/migration.go
vaske: is that by some convention?

gateway/mw_version_check_test.go
```diff
@@ -463,3 +464,57 @@ func TestOldVersioning_Expires(t *testing.T) {
...
+	check := func(t *testing.T, response string) {
+
+		_, _ = ts.Run(t, []test.TestCase{
+			{Path: "/test", BodyMatch: "default", Code: http.StatusOK},
+			{Path: "/anything", BodyMatch: response, Code: http.StatusOK},
+		}...)
+
+	}
+
+	ts.Gw.LoadAPI(api())
```
furkansenharputlu, line 510: I think load should be done inside check function. Right now,  migrated api is not loaded. Check function should get api param.

apidef/oas/middleware.go
```diff
@@ -521,3 +524,55 @@ func (mt *MethodTransform) ExtractTo(meta *apidef.MethodTransformMeta) {
...
+}
+
+func (p *Plugins) extractCacheTo(ep *apidef.ExtendedPathsSet, path string, method string) {
+
+	if p.Cache == nil {
+		return
+	}
+
+	newCacheMeta := apidef.CacheMeta{
+		CacheOnlyResponseCodes: p.Cache.CacheResponseCodes,
```
furkansenharputlu, line 552: ExtractTo function already fills this fields. Just method and path should be defined outside.

furkansenharputlu (approved):
Added 2 small comments. Overall LGTM!

jeffy-mathew (approved):

PR #3813, opened 2021-12-27
[TT-4228] Fix strip listen path which removes forwarding slash in path

furkansenharputlu:
Strip listen path removes forwarding slash in endpoint if listen path has a trailing slash like:
`listen_path` -> `/listen/`
`path` -> `/listen/get`
stripping result: `get`

It should respect the forwarding slash in the endpoint and the result should be `/get`

### DIFF
--- a/gateway/handler_success.go
+++ b/gateway/handler_success.go
@@ -305,12 +305,7 @@ func (s *SuccessHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) *http
 	defer s.Base().UpdateRequestSession(r)
 
 	// Make sure we get the correct target URL
-	if s.Spec.Proxy.StripListenPath {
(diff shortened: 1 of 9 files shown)


gateway/handler_success.go
```diff
@@ -305,12 +305,7 @@ func (s *SuccessHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) *http
 	defer s.Base().UpdateRequestSession(r)
 
 	// Make sure we get the correct target URL
-	if s.Spec.Proxy.StripListenPath {
-		log.Debug("Stripping: ", s.Spec.Proxy.ListenPath)
```
buger, line 309: We lost debug output here, and it is quite useful. Lets put it back somewhere.

buger (approved):
👍🏻

jeffy-mathew (approved):

PR #3816, opened 2021-12-28
TD-664/Aws-cf-templates

ermirizio:
M4 template sync for TD-664 branch AWS CF templates


gwotg:
Hey @ermirizio can we check the filing test? Loop me in if I can help

ermirizio:
I Added the playground files here to make goreleaser build successful, however not sure if this is desirable/makes sense  https://github.com/TykTechnologies/tyk/tree/aws/TD-664/templates/playground

gwotg (approved):
Thanks @ermirizio! LGTM!

PR #3817, opened 2021-12-28
TD-664/Aws-cf-templates for release-4

ermirizio:
M4 template sync for TD-664 branch AWS CF templates


gwotg (approved):
Thanks @ermirizio! LGTM

PR #3823, opened 2021-12-30
TT-4066 include mdcb test framework in lts

sredxny:
## Description
Added the required changes to execute the mdcb testing framework in dashboard v-release-4-lts

## Related Issue
https://tyktech.atlassian.net/browse/TT-4066

## Motivation and Context
Give solution to https://tyktech.atlassian.net/browse/TT-4066

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


gateway/testutil.go
```diff
@@ -1006,7 +1006,7 @@ func (s *Test) BootstrapGw(ctx context.Context, cancelFn context.CancelFunc, gen
 	s.MockHandle = MockHandle
 
 	var err error
-	gwConfig.Storage.Database = rand.Intn(15)
+	gwConfig.Storage.Database = 1
```
sredxny, line 1009: this is required as we have some tests that depends on what other use-cases do in redis. An example of that situation is https://github.com/TykTechnologies/tyk/blob/master/gateway/rpc_test.go#L221 that basically depends on a backup populated by the previous use-case. So, as now we do not use a singleton connection but one singleton per gateway it lead us to randomly failing tests (because they check in different databases)

furkansenharputlu (approved):

jeffy-mathew (approved):

tbuchaillot (approved):

PR #3825, opened 2022-01-03
[TT-4141] Add auto-generated api ids to newly created APIs in migration

furkansenharputlu:
This PR auto-generates sub-version API ids in migration.

### DIFF
--- a/apidef/migration.go
+++ b/apidef/migration.go
@@ -5,6 +5,8 @@ import (
 	"net/url"
 	"sort"
 	"strings"
+
+	uuid "github.com/satori/go.uuid"
 )
 
 func (a *APIDefinition) MigrateVersioning() (versions []APIDefinition, err error) {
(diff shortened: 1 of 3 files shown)


apidef/migration.go
```diff
@@ -13,30 +15,45 @@ func (a *APIDefinition) MigrateVersioning() (versions []APIDefinition, err error
 	}
 
 	if a.VersionData.NotVersioned && len(a.VersionData.Versions) > 1 {
-		return nil, errors.New("not migratable - if not versioned, there should be 1 version info in versions map")
+		return nil, errors.New("not migratable - if not versioned, there should be just one version info in versions map")
```
lghiur, line 18: can you change `not migratable` to `This API can not be migrated because it doesn't have any version. At least one version is required so that the migration can happen.`

apidef/migration.go
```diff
@@ -13,30 +15,45 @@ func (a *APIDefinition) MigrateVersioning() (versions []APIDefinition, err error
...
 	}
 
 	if !a.VersionData.NotVersioned {
+		if _, ok := a.VersionData.Versions[a.VersionData.DefaultVersion]; !ok {
+			return nil, errors.New("not migratable - if versioned, default version should match one of the versions")
```
lghiur, line 23: same sa above maybe we can reformulate

apidef/migration_test.go
```diff
@@ -131,7 +137,7 @@ func TestAPIDefinition_MigrateVersioning(t *testing.T) {
 
 		t.Run("multiple versions", func(t *testing.T) {
 			versions, err = versionDisabledBase.MigrateVersioning()
-			assert.EqualError(t, err, "not migratable - if not versioned, there should be 1 version info in versions map")
+			assert.EqualError(t, err, "not migratable - if not versioned, there should be just one version info in versions map")
```
lghiur, line 140: same as above

lghiur (approved):

apidef/migration.go
```diff
@@ -13,30 +15,45 @@ func (a *APIDefinition) MigrateVersioning() (versions []APIDefinition, err error
...
 
 			newAPI := *a
-			newAPI.APIID = ""
+
+			newID := uuid.NewV4()
+			apiID := strings.Replace(newID.String(), "-", "", -1)
+
+			newAPI.APIID = apiID
 			newAPI.Id = ""
+			newAPI.Name += "-" + url.QueryEscape(vName)
```
jeffy-mathew, line 38: For better readability can we use
```suggestion
			fmt.Sprintf("%s-%s", newAPI.Name, url.QueryEscape(vName)
```

jeffy-mathew (approved):

apidef/migration.go
furkansenharputlu: If the name wouldn’t have a value already I would agree. However, we are doing an append so I think it is more readable that we  add a suffix instead of replacement.

PR #3828, opened 2022-01-04
Adding healthcheck info and lock into gw struct

tbuchaillot:
## Description

Adding Healthcheck info and lock into gateway struct to add consistency and avoid overlapping values in MDCB testing framework

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


gateway/server.go
```diff
@@ -179,6 +180,9 @@ type Gateway struct {
 	// RedisController keeps track of redis connection and singleton
 	RedisController *storage.RedisController
 	hostDetails     hostDetails
+
+	healthCheckInfo atomic.Value
+	healthCheckLock sync.Mutex
```
furkansenharputlu, line 185: if `healthCheckInfo` is atomic, why do we need a mutex to guard it?

gateway/server.go
```diff
@@ -82,6 +82,7 @@ var (
 const appName = "tyk-gateway"
 
 type Gateway struct {
+
```
furkansenharputlu, line 85: I think it is a leftover

gateway/server.go
tbuchaillot: good point @furkansenharputlu , probably old code. Removed the mutex 👍

furkansenharputlu (approved):

matiasinsaurralde (approved):

jeffy-mathew (approved):

PR #3840, opened 2022-01-12
git checkout . after fix vendor in release

jeffy-mathew:
go release throws dirty git state after `go mod tidy`
https://github.com/TykTechnologies/tyk/runs/4787073278?check_suite_focus=true

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -128,6 +128,7 @@ jobs:
           rm -rf vendor
           mkdir -p /go/src/github.com/TykTechnologies/tyk
           cp -r ./* /go/src/github.com/TykTechnologies/tyk
+          git checkout .
 
       - uses: goreleaser/goreleaser-action@v2
         with:


alephnull (commented):
Does this mean that `go mod tidy` cannot be used on this branch?

Per https://github.com/TykTechnologies/tyk/actions/runs/1686808607 I guess we need the same thing on release-4-lts as well

jeffy-mathew:
> Does this mean that `go mod tidy` cannot be used on this branch?

Nop, we're not using go modules during build since we use GOPATH for building gw. So this change doesn't really make a difference 
Also, this is added to `release-3.2.3` [here](https://github.com/TykTechnologies/tyk/blob/release-3.2.3/.github/workflows/release.yml#L125)

furkansenharputlu (approved):

alephnull (approved):
tbh, my head hurts a little :crossed_fingers:

PR #3854, opened 2022-01-20
Add matrix job for go1.15, go1.15-el7

jeffy-mathew:
- Add matrix job for go1.15, go1.15-el7
- Remove darwin support
- Update filename format to have golang cross version in package name
- Provide target os versions on package cloud action


alephnull:
This PR needs to target `release-4.0` first. `master` is in a significantly different state.

PR #3855, opened 2022-01-20
Add matrix job for go1.15, go1.15-el7

jeffy-mathew:
- Add matrix job for go1.15, go1.15-el7
- Remove darwin support
- Update filename format to have golang cross version in package name
- Provide target os versions on package cloud action


alephnull:
```
diff --git a/.goreleaser.yml b/.goreleaser.yml
index a4d6206a..d9aae692 100644
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -216,7 +216,6 @@ nfpms:
     maintainer: "Tyk <info@tyk.io>"
     description: Tyk API Gateway
     package_name: tyk-gateway
-    file_name_template: "{{ .PackageName }}-{{.Env.GOLANG_CROSS}}_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}{{ if .Mips }}_{{ .Mips }}{{ end }}"
     builds:
 
       - std-linux
```
and tagged v4.0.0-rc22 (locally) for QA of plugin-compiler

.goreleaser.yml
```diff
@@ -117,6 +109,7 @@ dockers:
     - "policies"
     - "coprocess"
     - "tyk.conf.example"
+  skip_push: {{ if .Env.SKIP_DOCKER_PUSH == 'true' }}true{{ else }}false{{ end }}
```
alephnull, line 112: Why not have `{{ .Env.SKIP_DOCKER_PUSH }}`?

.goreleaser.yml
```diff
@@ -224,7 +222,7 @@ nfpms:
     homepage: "https://tyk.io"
     maintainer: "Tyk <info@tyk.io>"
     description: Tyk API Gateway
-    package_name: tyk-gateway
+    package_name: tyk-gateway\
```
alephnull, line 225: That `\` seems wrong.

.goreleaser.yml
```diff
@@ -182,25 +177,28 @@ dockers:
...
   extra_files:
     - "images/hybrid/"
+  skip_push: {{ if .Env.SKIP_DOCKER_PUSH == 'true' }}true{{ else }}false{{ end }}
 
 # Build plugin-compiler multiarch
 - ids:
     - std
   image_templates:
-    - "tykio/tyk-plugin-compiler:{{ .Tag }}"
+    - "tykio/tyk-plugin-compiler{{ if .Env.SKIP_DOCKER_PUSH == 'true' }}-el7{{ end }}:{{ .Tag }}"
```
alephnull, line 186: I think this if is the only one required, the remaining could just use the env var directly?

alephnull (changes requested):
Asking for some changes as these will finally need to be templated and have an interest in keeping the templates as readable as possible.

.goreleaser.yml
jeffy-mathew: yea, we could use that. tried this since env var comes as string

.goreleaser.yml
jeffy-mathew: env var would be `1.15` or `1.15-el7`. that would change the plugin compiler name to `tykio/tyk-plugin-compiler-1.15` and `tykio/tyk-plugin-compiler-1.15-el7`?

alephnull: Missed this earlier. It will have to be `"tykio/tyk-plugin-compiler:{{ .Tag }}{{ if .Env.SKIP_DOCKER_PUSH == 'true' }}-el7{{ end }}"`
We do not have a different repo for the el7 plugin compiler. We just need the tag to indicate it.

alephnull: Also, that variable name is looking weird. Perhaps it could be called `EL7_BUILD` instead of `SKIP_DOCKER_PUSH`? I can make this change.

alephnull:
All tests need to pass. The AMI build failure as well, if possible.

.github/workflows/release.yml
```diff
@@ -116,23 +116,27 @@ jobs:
...
           go mod vendor
           echo "Moving vendor"
-          mv -f vendor/* $GOPATH/src
+          cp -r -f vendor/* $GOPATH/src
           rm -rf vendor
           mkdir -p /go/src/github.com/TykTechnologies/tyk
           cp -r ./* /go/src/github.com/TykTechnologies/tyk
-          git checkout .
+          echo "Making a source tarball for plugin compiler build"
+          find . -name "*.go" -o -name "*.mod" -o -name "*.sum" | tar -czf src.tgz -T -
```
alephnull, line 133: Is the complete gw code needed?

images/plugin-compiler/Dockerfile
```diff
@@ -20,9 +18,9 @@ COPY images/plugin-compiler/data/build.sh /build.sh
 # RUN cd $TYK_GW_PATH && go mod tidy && go mod vendor
 RUN chmod +x /build.sh
 
-RUN cd $TYK_GW_PATH &&  \
-    git clone --depth 1 --branch $TYK_GW_TAG  https://github.com/TykTechnologies/tyk . && \
-    go mod vendor
+COPY src.tgz .
```
alephnull, line 21: go.mod is already in `extra_files`, is this not sufficient to create a vendor folder?

alephnull:
@asutosh To test this locally, I will have to create a tarball which is likely to unexpected things, based on what is the state of my local checkout.

images/plugin-compiler/Dockerfile
asutosh: No, it doesn't seem to work. With just the go.mod, the vendor command doesn't just fetch any vendor dependencies.

asutosh:
> @asutosh To test this locally, I will have to create a tarball which is likely to unexpected things, based on what is the state of my local checkout.

Yes, we'll have to emulate the steps in the fix vendor stage including the tarball creation for getting this to work locally. To avoid the state of local checkout affecting the build, we may have to check the git tree status, and also we may have to probably exclude everything from the `.gitignore` list.

.goreleaser.yml
```diff
@@ -183,24 +174,51 @@ dockers:
...
     - "--label=org.opencontainers.image.created={{.Date}}"
     - "--label=org.opencontainers.image.title=tyk-plugin-compiler"
     - "--label=org.opencontainers.image.revision={{.FullCommit}}"
     - "--label=org.opencontainers.image.version={{.Version}}"
     - "--build-arg=TYK_GW_TAG={{ .Tag }}"
+    - "--build-arg=GOLANG_CROSS={{ .Env.GOLANG_CROSS }}"
   goarch: amd64
   goos: linux
   dockerfile: images/plugin-compiler/Dockerfile
   extra_files:
```
buger, line 194: Why we need this section here, if we copy tar.gz? 
Having such a long list of files and dirs can cause a lot of human errors.

alephnull: This way we don't have a tar.gz making local builds simpler. And we'll have to maintain this list, either implicitly in a tarball or explicitly here.

.goreleaser.yml
```diff
@@ -290,6 +291,7 @@ nfpms:
     bindir: "/opt/tyk-gateway"
     overrides:
       rpm:
+        file_name_template: '{{ .PackageName }}_{{ replace .Version "-" "~" }}-1.{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}{{ if .Mips }}_{{ .Mips }}{{ end }}'
```
asutosh: @alephnull the common `file_name_template` is for debian and we override that for rpm. Or should we avoid the common `file_name_template` altogether and add both in `overrides?

alephnull:
```
[Invalid workflow file : .github/workflows/release.yml#L398](https://github.com/TykTechnologies/tyk/actions/runs/1806714425/workflow)
You have an error in your yaml syntax on line 398
```

PR #3880, opened 2022-02-08
Update release engineering templates on 3-lts

ermirizio:
Sync m4 templates


alephnull:
This appears to be older version of the templates. We should use the latest changes from release-4.0 right?

PR #3910, opened 2022-02-28
[TT-4655] Federation crashes gateway during performance tests

buraksezer:
PR for [TT-4655](https://tyktech.atlassian.net/browse/TT-4655)

Basically, Tyk gateway creates a `memconn.Provider` to handle subgraphs for every request. it also creates a new HTTP client with a fresh `http.Transport`. 

It repeats this flow for every request and the dialer implementation in the Go HTTP library allocates new buffers to dial the upstream server. This creates a massive memory leak in the runtime.  I implemented a simple caching approach to overcome that problem. Now it uses a global `http.Client` instance and the dialer function gets a `memconn.Provider` for the subgraph. Before the fix, Tyk gateway had almost 12GB of memory on my local setup. With the fix, it’s stable at around 140 MB.

Users can rename the subgraph or delete it completely. If a subgraph is renamed, a new `memconn.Provider` will be created and cached.

From memory profiler:

Screen Shot 2022-02-25 at 12 20 51


Tyk-ITS:
Branch used: refs/pull/3910/merge
Commit:  
Triggered by: pull_request (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1910542598)

Tyk-ITS:
Branch used: refs/pull/3910/merge
Commit: cf1a136b5d27c17a88d10f30ceafdcf941ed0e9e 
Triggered by: pull_request (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1910544710)

Tyk-ITS:
Branch used: refs/pull/3910/merge
Commit: 56490610add4c31382cf905f742239c92fed55b9 
Triggered by: pull_request (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1910878030)

gateway/reverse_proxy.go
```diff
@@ -789,33 +789,69 @@ const (
...
+	_, ok := memconnProvidersCache[r.Host]
+	if ok {
+		return nil
+	}
+	// use separate provider for each request
+	provider := &memconn.Provider{}
 	// start in mem listener
 	lis, err := provider.Listen(inMemNetworkType, inMemNetworkName)
+	if err != nil {
+		memconnProvidersMtx.Unlock()
```
titpetric: .Unlock is already being defered

gateway/reverse_proxy.go
```diff
@@ -789,33 +789,69 @@ const (
 	inMemNetworkType = "memu"
 )
 
-func handleInMemoryLoop(handler http.Handler, r *http.Request) (resp *http.Response, err error) {
-	// use separate provider for each request
-	provider := memconn.Provider{}
+var (
+	memconnProvidersMtx sync.Mutex
```
titpetric: Did you consider using a RWMutex? I'm asuming the providers are mostly read-only after they are created.

gateway/reverse_proxy.go
```diff
@@ -785,73 +790,126 @@ func (rt *TykRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	listener net.Listener
+	provider *memconn.Provider
+	lastUsed int64
+}
 
-func createMemconnProviderIfNeeded(handler http.Handler, r *http.Request) error {
-	memconnProvidersMtx.Lock()
-	defer memconnProvidersMtx.Unlock()
-	_, ok := memconnProvidersCache[r.Host]
+var memConnProviders = &struct {
```
titpetric, line 800: Is using &struct some particular convention that's being followed? Anonymous structs don't have function receivers, and pointer semantics don't make any sense for globals

gateway/reverse_proxy_test.go
```diff
@@ -897,6 +898,96 @@ func TestGraphQL_InternalDataSource(t *testing.T) {
...
+
+	memConnProviders.mtx.Lock()
+	require.Contains(t, memConnProviders.m, tykSubgraphAccounts.Name)
+	require.Contains(t, memConnProviders.m, tykSubgraphReviews.Name)
+	memConnProviders.mtx.Unlock()
+
+	<-time.After(2 * time.Second)
+	// Remove memconn.Provider structs from the cache, if they are idle for
+	// 1 second.
+	cleanIdleMemConnProvidersEagerly(-time.Second)
```
titpetric: I know the semantics of "time.Add(-dur)..." are a constant nuissance, would it be good if we stuck to absolute (positive) values, and only convert duration just before the final time.Now().Add(-dur)?

gateway/reverse_proxy.go
```diff
@@ -785,73 +790,126 @@ func (rt *TykRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
...
+}
+
+// cleanIdleMemConnProvidersEagerly deletes unused memconn.Provider instances and closes
+// its listener.
+func cleanIdleMemConnProvidersEagerly(maxIdleDuration time.Duration) {
+	memConnProviders.mtx.Lock()
+	defer memConnProviders.mtx.Unlock()
+
+	for host, mp := range memConnProviders.m {
+		if time.Now().Add(maxIdleDuration).Unix() > mp.lastUsed {
```
titpetric: So, technically we're substracting maxIdleDuration here, which should be passed as an absolute/positive value. I'd prefer (-maxIdleDuration) here (semantics)

gateway/reverse_proxy.go
```diff
@@ -803,38 +798,27 @@ type memConnProvider struct {
...
 
-// cleanIdleMemConnProvidersEagerly deletes unused memconn.Provider instances and closes
+// cleanIdleMemConnProviders deletes unused memconn.Provider instances and closes
 // its listener.
-func cleanIdleMemConnProvidersEagerly(maxIdleDuration time.Duration) {
-	memConnProviders.mtx.Lock()
-	defer memConnProviders.mtx.Unlock()
-
+func cleanIdleMemConnProviders(maxIdleDuration, idleCheckInterval time.Duration) {
+	if memConnProviders.lastIdleCheck > time.Now().Add(-idleCheckInterval).Unix() {
```
titpetric, line 811: I'm imagining the outside loop to include the idleCheckInterval e.g.

```go
go func() {
    for {
       time.Sleep(checkIdleMemConnInterval)
       cleanIdleMemConnProviders(maxIdleMemConnDuration)
    }
}()```

Comment (titpetric) [reply gateway/reverse_proxy.go:811]: I'd honestly recommend a sync.Pool for this. It seems we can just drop half of this code, and let Pool take care of allocating (and freeing) the memConnProviders as the load rises/falls.

Comment (buraksezer) [reply gateway/reverse_proxy.go:811]: > I'd honestly recommend a sync.Pool for this. It seems we can just drop half of this code, and let Pool take care of allocating (and freeing) the *memConnProvider-s as the load rises/falls.

We discussed this topic extensively on Slack. Here is my opinion:

> I think `sync.Pool` would be cool but the current implementation requires a running HTTP server in the provider. I couldn’t find any way to clean the goroutines and stop listeners.

**Comment by buger** [comment]
Overall looks good! Once @titpetric will be approve can be merged.

**Comment by titpetric** [comment]
Burak: i think the time/duration handling can be improved. Start by
avoiding the time.After call in the tests, and rather pass the timestamp(s)
from the test. Instead of doing the time.Add math and updating last used
timestamps in fields, we can log an expire value
(time.Now().Add(expireDuration)) and then just check against time.Now() >
ExpireVal. The unix conversions can probablly be all dropped, and we can
use a time.Time instead of int64 unixtimes?

**Comment by titpetric** [comment]
Additionally: we never free memory on an idle service after the benchmarks,
because the cleanup is request driven. This shouldn't be an issue, but a
continous goroutine which cleans these up is my preferred method. We can
use the sync.Once construct to start it, if there is not a cleaner
entrypoint in the code for it.

**Comment by Tyk-ITS** [comment]
Branch used: refs/pull/3910/merge
Commit: 31ff2c8a15e6fd90b28e2c64fb3cbabf93c38457 
Triggered by: pull_request (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1944078510)

**Comment by Tyk-ITS** [comment]
Branch used: refs/pull/3910/merge
Commit: 7a5361d6211f7ee2fe529025629763f6983dbe15 
Triggered by: pull_request (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1944483605)

**Comment by buraksezer** [comment]
@titpetric and @buger Can you please review the latest version? Now I use a background goroutine to clean idle items in the map. It's triggered by `initialiseSystem` and the lifecycle is controlled by the gateway's context. I also removed the int64 timestamps and started using `time.Time` to calculate expiration time.

**Comment by titpetric** [comment]
@buraksezer please squash so it's easier to review the final code.

File: gateway/reverse_proxy.go [hunk 7]
```diff
@@ -794,31 +794,46 @@ const (
...
 			_ = mp.listener.Close()
 		}
 	}
-	memConnProviders.lastIdleCheck = time.Now().Unix()
+}
+
+// cleanIdleMemConnProviders checks memconn.Provider instances periodically and
+// deletes idle ones.
+func cleanIdleMemConnProviders(ctx context.Context) {
+	timer := time.NewTimer(checkIdleMemConnInterval)
```
titpetric, line 825: NewTicker would be better, removes the need for timer.Reset lower

buraksezer:
> @buraksezer please squash so it's easier to review the final code.

Merging into this repository already requires a squash merge and I'm not sure how to squash the commits without creating a new branch. Also, the checks can take a few hours because of unreliable tests. Could you please see the changes on **Files changed** tab?

Tyk-ITS:
Branch used: refs/pull/3910/merge
Commit: f56ecccf30edfc7f0b69d1e2145860e7910c11bb 
Triggered by: pull_request (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1944990776)

Tyk-ITS:
Branch used: refs/pull/3910/merge
Commit: e2ceaec93a1465800aa03274393df50f3aeae705 
Triggered by: pull_request (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1944996281)

gateway/reverse_proxy.go
```diff
@@ -785,37 +785,130 @@ func (rt *TykRoundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
...
+			return
+		case <-ticker.C:
+			cleanIdleMemConnProvidersEagerly(time.Now())
+		}
+	}
+}
+
+// getMemConnProvider return the cached memconn.Provider, if it's available in the cache.
+func getMemConnProvider(addr string) (*memconn.Provider, error) {
+	host, _, err := net.SplitHostPort(addr)
```
titpetric, line 840: Do we want to ignore ':port' ?

titpetric (approved):

Tyk-ITS:
Branch used: refs/pull/3910/merge
Commit: 93b418efe93fa5f3dd87f0b83566c083a2880646 
Triggered by: pull_request (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1945605375)

Issue #3858, opened 2022-01-21
Transform Response configuration doesn't work when rewrite rules are configured


LowCostCustoms:
**Branch/Environment/Version**
- Branch/Version: v3.2.2
- Environment: On-prem

**Describe the bug**
When `transform_response` and `url_rewrites` are used together, transform rules are applied improperly.

**Reproduction steps**
1. Apply the following Tyk configuration:

```json
{
  "name": "my-api",
  "api_id": "my-api",
  "org_id": "my-org",
  "proxy": {
    "listen_path": "/",
    "target_url": "",
    "strip_listen_path": true,
    "disable_strip_slash": true
  },
  "version_data": {
    "not_versioned": true,
    "versions": {
      "Default": {
        "name": "Default",
        "use_extended_paths": true,
        "extended_paths": {
          "cache": [".*"],
          "url_rewrites": [
            {
              "path": "^/$",
              "method": "GET",
              "match_pattern": "^/$",
              "rewrite_to": "http://upstream:8080/api"
            },
            {
              "path": "^/([^/]+)$",
              "method": "GET",
              "match_pattern": "^/([^/]+)$",
              "rewrite_to": "http://upstream:8080/api/$1"
            },
            {
              "path": "^/([^/]+)/([^/]+)$",
              "method": "GET",
              "match_pattern": "^/([^/]+)/([^/]+)$",
              "rewrite_to": "http://upstream:8080/api/$1/records/$2"
            }
          ],
          "transform_response": [
            {
              "path": "^/$",
              "method": "GET",
              "template_data": {
                "input_type": "json",
                "template_mode": "file",
                "template_source": "./templates/template-0.tmpl"
              }
            },
            {
              "path": "^/([^/]+)$",
              "method": "GET",
              "template_data": {
                "input_type": "json",
                "template_mode": "file",
                "template_source": "./templates/template-1.tmpl"
              }
            },
            {
              "path": "^/([^/]+)/([^/]+)$",
              "method": "GET",
              "template_data": {
                "input_type": "json",
                "template_mode": "file",
                "template_source": "./templates/template-2.tmpl"
              }
            }
          ]
        }
      }
    }
  },
  "response_processors": [
    {
      "name": "response_body_transform",
      "options": {}
    }
  ],
  "use_keyless": true,
}
```
2. Make a `GET` request against `http://localhost:8080/first/second` and ensure the response wasn't updated.

**Actual behavior**
Request path is not matched and response body transform rules aren't applied.

[...]

PR #3925, opened 2022-03-08
Fix transform response context containing wrong rewrite path

PatrickTaibel:
## Description
Changes the path context variable set by the url rewrite middleware to the correct value.

## Related Issue

- #3858

## Motivation and Context
When using regex to match paths within the url rewrite middleware, the path provided to response transform middlewares is unusable. It should get the path before the url rewriting takes place but instead it is provided with the specified path value of the url rewrite middleware (which includes all specified regex symbols).To use it like that, one would need to build a regex for matching another regex pattern.

## How This Has Been Tested
See below for the relevant parts of the config. In general, it's simply a url_rewrites section combined with a transform_response_headers section using regex for matching paths. After applying the fix this config works flawless.

extended_paths:
```json
"extended_paths": {
    "transform_response_headers": [
      {
        "delete_headers": null,
        "add_headers": {
          "Server": "mock-server",
          "x-completely-new": "Hello World",
        },
        "path": "^/test-1/override",
        "method": "GET",
        "act_on": false
      },
      {
        "delete_headers": [
          "x-completely-new",
          "Server"
        ],
        "add_headers": null,
        "path": "^/test-1/delete",
        "method": "GET",
        "act_on": false
      },
      {
        "delete_headers": [
          "Server"
        ],
        "add_headers": {
          "Content-Type": "text/plain"
        },
        "path": "^/test-1/all",
        "method": "GET",
        "act_on": false
      }
    ],
    "url_rewrites": [
      {
        "path": "^/test-1",
        "method": "GET",
        "match_pattern": "^/API-SLUG/test-1/(.*)$",
        "rewrite_to": "http://mock-api/headers.php",
        "triggers": null
      }
    ]
  }
```
response_processors:
```json
"response_processors":[{"name": "header_injector"}]
```

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


buger:
Appologies we missed this PR somehow! 

If you can add test case for this will be great.
You can take inspiration from the examples which I posted to your another PRs (you rock!), plus this is how to initialize API with a complex Extended paths in tests: https://github.com/TykTechnologies/tyk/blob/master/gateway/mw_validate_json_test.go#L35

Thanks!

PR #3940, opened 2022-03-15
[TT-4869] Implement base of paths and operations

furkansenharputlu:
This PR implements `operationId` usage to match plugins for path and methods. See the example Tyk OAS doc:
```
{
  "components": {},
  "info": null,
  "openapi": "",
  "paths": {
    "/user": {
      "get": {
        "operationId": "userGET",
        "responses": null
      },
      "post": {
        "operationId": "userPOST",
        "responses": null
      }
    }
  },
  "x-tyk-api-gateway": {
    "info": {
      "name": "",
      "state": {
        "active": false
      }
    },
    "upstream": {
      "url": ""
    },
    "server": {
      "listenPath": {
        "value": ""
      }
    },
    "middleware": {
      "operations": {
        "userGET": {
          "allow": {
            "enabled": true
          }
        }
      }
    }
  }
}
```

### DIFF
--- a/apidef/oas/oas.go
+++ b/apidef/oas/oas.go
@@ -21,6 +21,7 @@ func (s *OAS) Fill(api apidef.APIDefinition) {
 	}
 
 	xTykAPIGateway.Fill(api)
+	s.fillPathsAndOperations(api.VersionData.Versions[""].ExtendedPaths)
 	s.fillSecurity(api)
 
 	if ShouldOmit(xTykAPIGateway) {
@@ -41,6 +42,16 @@ func (s *OAS) ExtractTo(api *apidef.APIDefinition) {
 
 	if s.GetTykExtension() != nil {
 		s.GetTykExtension().ExtractTo(api)
+	}
+
+	var ep apidef.ExtendedPathsSet
+	s.extractPathsAndOperations(&ep)
+
+	api.VersionData.Versions = map[string]apidef.VersionInfo{
+		"": {
+			UseExtendedPaths: true,
+			ExtendedPaths:    ep,
+		},
 	}
 }
 
@@ -188,4 +199,20 @@ func (s *OAS) getTykSecurityScheme(name string) interface{} {
 	}
 
 	return securitySchemes[name]
+}
+
+func (s *OAS) getTykMiddleware() (middleware *Middleware) {
+	if s.GetTykExtension() != nil {
+		middleware = s.GetTykExtension().Middleware
+	}
+
+	return
+}
+
+func (s *OAS) getTykOperations() (operations Operations) {
+	if s.getTykMiddleware() != nil {
+		operations = s.getTykMiddleware().Operations
+	}
+
+	return
 }
--- a/apidef/oas/oas.go
+++ b/apidef/oas/oas.go
@@ -8,6 +8,7 @@ import (
 )
 
 const ExtensionTykAPIGateway = "x-tyk-api-gateway"
+const Main = ""
 
 type OAS struct {
 	openapi3.T
@@ -21,7 +22,7 @@ func (s *OAS) Fill(api apidef.APIDefinition) {
 	}
 
 	xTykAPIGateway.Fill(api)
-	s.fillPathsAndOperations(api.VersionData.Versions[""].ExtendedPaths)
+	s.fillPathsAndOperations(api.VersionData.Versions[Main].ExtendedPaths)
 	s.fillSecurity(api)
 
 	if ShouldOmit(xTykAPIGateway) {
@@ -48,7 +49,7 @@ func (s *OAS) ExtractTo(api *apidef.APIDefinition) {
 	s.extractPathsAndOperations(&ep)
 
 	api.VersionData.Versions = map[string]apidef.VersionInfo{
-		"": {
+		Main: {
 			UseExtendedPaths: true,
 			ExtendedPaths:    ep,
 		},
--- a/apidef/oas/operation.go
+++ b/apidef/oas/operation.go
@@ -44,12 +44,14 @@ func (s *OAS) extractPathsAndOperations(ep *apidef.ExtendedPathsSet) {
 	}
 
 	for id, tykOp := range tykOperations {
+	found:
 		for path, pathItem := range s.Paths {
 			for method, operation := range pathItem.Operations() {
 				if id == operation.OperationID {
 					tykOp.extractAllowanceTo(ep, path, method, allow)
 					tykOp.extractAllowanceTo(ep, path, method, block)
 					tykOp.extractAllowanceTo(ep, path, method, ignoreAuthentication)
+					break found
 				}
 			}
 		}
--- a/apidef/oas/operation.go
+++ b/apidef/oas/operation.go
@@ -67,23 +67,11 @@ func (s *OAS) fillAllowance(endpointMetas []apidef.EndPointMeta, typ AllowanceTy
 
 		switch typ {
 		case block:
-			if operation.Block == nil {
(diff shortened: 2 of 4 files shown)


Tyk-ITS:
Branch used: refs/pull/3940/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1987128749)

Tyk-ITS:
Branch used: refs/pull/3940/merge
Commit: baf8d4888206fecd0afe237938203cf6d855017a 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1987131566)

apidef/oas/operation.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+}
+
+func (s *OAS) fillAllowance(endpointMetas []apidef.EndPointMeta, typ AllowanceType) {
+	for _, em := range endpointMetas {
+		operationID := s.getOperationID(em.Path, em.Method)
+		operation := s.GetTykExtension().getOperation(operationID)
+
+		var allowance *Allowance
```
jeffy-mathew, line 64: don't we have to assign allowance to OAS object field?

furkansenharputlu: As it is a pointer we set it like this for example `allowance = operation.Block` and when you change it, `operation.Block` changes.

jeffy-mathew (approved):

apidef/oas/oas.go
```diff
@@ -21,6 +21,7 @@ func (s *OAS) Fill(api apidef.APIDefinition) {
 	}
 
 	xTykAPIGateway.Fill(api)
+	s.fillPathsAndOperations(api.VersionData.Versions[""].ExtendedPaths)
```
titpetric, line 24: Reading this bit of code, it isn't clear to me why "" was chosen and what value it represents. Can we do a constant for the keyname, so it becomes Versions[???]

apidef/oas/operation.go
```diff
@@ -0,0 +1,157 @@
...
+
+const (
+	allow                AllowanceType = 0
+	block                AllowanceType = 1
+	ignoreAuthentication AllowanceType = 2
+)
+
+type AllowanceType int
+
+func (s *OAS) fillPathsAndOperations(ep apidef.ExtendedPathsSet) {
```
titpetric, line 26: I'd rather do *apidef.ExtendedPathSet here so it's clear that it's fields are being modified within the functions (even if whitelist, blacklist and ignored fields are pointers)

apidef/oas/operation.go
```diff
@@ -0,0 +1,157 @@
...
+func (s *OAS) extractPathsAndOperations(ep *apidef.ExtendedPathsSet) {
+	tykOperations := s.getTykOperations()
+	if len(tykOperations) == 0 {
+		return
+	}
+
+	for id, tykOp := range tykOperations {
+		for path, pathItem := range s.Paths {
+			for method, operation := range pathItem.Operations() {
+				if id == operation.OperationID {
```
titpetric, line 49: Do we even need to loop over `tykOperations` or can we just do `tykOp, ok := tykOperations[operation.OperationID]` instead of `id == operation.OperationID`?

apidef/oas/operation.go
```diff
@@ -0,0 +1,157 @@
...
+			allowance = operation.IgnoreAuthentication
+		default:
+			if operation.Allow == nil {
+				operation.Allow = &Allowance{}
+			}
+
+			allowance = operation.Allow
+		}
+
+		allowance.Fill(em)
```
titpetric, line 87: A field value of `em` (apidef.EndPointMeta) is being modified, but not passed as a pointer value. Should it be 'fillAllowance' responsibility to populate .Block, .IgnoreAuthentication, .Allow? Or should we have something more like:

```go
switch typ {
case block:
    allowance = operation.Block.Fill(em)
case ignoreAuthentication:
    allowance = operation.IgnoreAuthentication.Fill(em)
...
```
Fill() would return *Allowance, and could work on a nil value for the receiver to create a new Allowance object as needed

apidef/oas/operation_test.go
```diff
@@ -0,0 +1,66 @@
...
+			Post: &openapi3.Operation{
+				OperationID: existingOperationId,
+			},
+		},
+	}
+	convertedOAS.SetTykExtension(&XTykAPIGateway{Middleware: &Middleware{Operations: Operations{}}})
+	convertedOAS.fillPathsAndOperations(ep)
+
+	assert.Equal(t, oas.getTykOperations(), convertedOAS.getTykOperations())
+
```
titpetric, line 53: I'd like to see a failing test case here, particularly a non-existant OperationID (e.g. what error is expected if i pass "xxx" which doesn't have a match in .Paths).

titpetric (changes requested):

apidef/oas/operation.go
furkansenharputlu: Actually, nothing is modified inside it

apidef/oas/operation_test.go
furkansenharputlu: I think it should be handled in the schema validation at the first place.

apidef/oas/oas.go
furkansenharputlu: correct.

Tyk-ITS:
Branch used: refs/pull/3940/merge
Commit: dc48b9a8ccb5d54ab71d686e0e9cccdd6aa98e83 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1993491897)

apidef/oas/operation.go
furkansenharputlu: The idea of putting `tykOperations` to top was because it is source of truth. Actually, I forgot to add break statements there to making this approach meaningful. I will append them so that way will make sense.

furkansenharputlu: The idea of putting `tykOperations` to top was because it is source of truth. Actually, I forgot to add break statement there to making this approach meaningful. I will append it so that way will make sense. Otherwise, `tykOperations[operation.OperationID]` causes unnecessary iterations.

furkansenharputlu: The idea of putting `tykOperations` to top was because it is source of truth. Actually, I forgot to add break statement there to making this approach meaningful. I will append it so that way will make sense. Otherwise, `tykOperations[operation.OperationID]` causes unnecessary iterations.

titpetric: If tykOperations is local to the function, we can also delete(tykOperations, operation.OperationID) - unless we particularly want to overwrite conflicts? Should we raise a warning in case of duplicated operationIDs? The code would be more efficient, if we optimize for uniqueness of operationIDs...

apidef/oas/operation_test.go
titpetric: There are still dark code paths we can follow here - duplicate operationIds for one, and we can test for that behavior (should the first operationId take effect, or the last one, or... what do the tests say). Currently the tests are only testing for the "success" case, but not for failure.

Tyk-ITS:
Branch used: refs/pull/3940/merge
Commit: a97c221d5fd29249996dadf711849a538620572d 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1998700140)

Tyk-ITS:
Branch used: refs/pull/3940/merge
Commit: 30a1c0dff69931280b9e9a95c2b01eb135cfcf57 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/1998733483)

titpetric (approved):

Issue #3976, opened 2022-03-31
Request body handling can cause high memory usage


PatrickTaibel:
**Branch/Environment/Version**
- Branch/Version: master
- Environment: On-prem

**Describe the bug**
Tyk API Gateway reads any request body without limitation into memory. Without some kind of load balancer or proxy in front to limit the request body size this can easily lead to out of memory conditions with large incoming requests. Especially as it seems that the request body is duplicated in memory multiple times.

**Reproduction steps**

1. Add any API endpoint.
2. Do a POST request with either a large file as Multipart Upload or generate a request with a large body (e.g. 3 GB)

**Actual behavior**
Tyk API Gateway copies the full request body multiple times into memory. It not only stores the request body in full once but depending on the used middlewares multiple times. In my test setup I could easily exceed 32 GB of memory with a file upload of a 3 GB ISO.

**Expected behavior**
- There should be at least a configurable request body size limit available to restrict incoming requests and prevent out-of-memory situations (I will create a pull request for such a feature).
- In general, request body handling could be improved:
  - Is it really necessary to read the whole body first when handling a request? In case no middleware needs to read it, the body could be forwarded without fully reading it into memory.
  - Caching/Reusing the request body should work without duplicating the memory usage multiple times. At the moment the request body reader gets replaced by a re-readable reader. Every middleware that accesses the body (e.g. JavaScript middlewares) create a duplicate copy while reading which needs to be cleaned up by garbage collection later.

andyo-tyk:
Thank you for this suggestion and the PR.

Sorry for the delay reviewing the PR - we will pick this up again shortly.

andyo-tyk:
Hi,

I'm pleased to confirm we'll be including this functionality in the upcoming Tyk 5.1 release.

Thank you for your patience and support for Tyk!

PR #3978, opened 2022-03-31
[TT-5186] Add a global request body size limit

PatrickTaibel:
## Description
Adds a global request body size limit to the config. For incoming requests the Content-Length will be checked against the limit. If it's set, the limit enabled and the Content-Length exceeds the limit a status 413 (Request Entity Too Large) will be returned and the further handling aborted. Additionally, a `http.MaxBytesReader` is applied to the request body to force any reads to be limited to the global server limit (e.g. when the Content-Length is not set or manipulated).

## Related Issue
https://github.com/TykTechnologies/tyk/issues/3976

## Motivation and Context
Tyk API Gateway copies the whole request into memory at the beginning of the request handling. Large requests could fill up memory if they are not blocked. At the moment this would need to be done by an additional load balancer or proxy in front.

## How This Has Been Tested
As long as the limit is not enabled (which is the default case) there is no change in behavior. Therefore, no existing installations are affected by the change. In case it's enabled only requests larger then the configured limit are affected.

## Types of changes

## Checklist

  - [x] `go fmt -s`
  - [x] `go vet`

### DIFF
--- a/cli/linter/schema.json
+++ b/cli/linter/schema.json
@@ -607,6 +607,9 @@
           "items": {
             "type": "string"
           }
+        },
+        "max_request_body_size": {
+          "type": "integer"
         }
       }
     },
--- a/config/config.go
+++ b/config/config.go
@@ -406,6 +406,9 @@ type HttpServerOptionsConfig struct {
 
 	// Custom SSL ciphers. See list of ciphers here https://tyk.io/docs/basic-config-and-security/security/tls-and-ssl/#specify-tls-cipher-suites-for-tyk-gateway--tyk-dashboard
 	Ciphers []string `json:"ssl_ciphers"`
+
+	// Maximum accepted request body size in byte.
+	MaxRequestBodySize int64 `json:"max_request_body_size"`
 }
 
 type AuthOverrideConf struct {
--- a/gateway/proxy_muxer.go
+++ b/gateway/proxy_muxer.go
@@ -29,7 +29,8 @@ import (
 
 // handleWrapper's only purpose is to allow router to be dynamically replaced
 type handleWrapper struct {
-	router *mux.Router
+	router             *mux.Router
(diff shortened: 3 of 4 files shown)


buger:
Nice! 

Clean and simple! 

Would also want add simple test to cover it? 

The test can look like this (have not tested):

```go
func TestRequestBodyLimit(t *testing.T) {
	ts := StartTest(nil)
	defer ts.Close()

	globalConf := ts.Gw.GetConfig()
	globalConf. MaxRequestBodySize = 1
	ts.Gw.SetConfig(globalConf)

	ts.Gw.BuildAndLoadAPI(func(spec *APISpec) {
		spec.UseKeylessAccess = false
		spec.Auth.UseParam = true
	})

	_, _ = ts.Run(t, test.TestCase{
		Path: "/sample/", Method: "POST", Data: "test", Code: http.StatusRequestEntityTooLarge,
	})
}
```

PatrickTaibel:
I've added a small test for this. Thank you very much for the pretty much finished test snippet!
I've changed the way to set the config though. With your way it seems to not restart/reinitialize the http handler and won't reread the HttpServerOptions. Therefore, I've set the config function in the `StartTest` call for this.

titpetric:
@PatrickTaibel I'm noticing this is used as the base of the following PR: https://github.com/TykTechnologies/tyk/pull/3979 - can you please include those changes here, and close that PR?

PatrickTaibel:
@titpetric Those two PRs are independent of each other (and both branches are directly based on master). This one adds a global serverwide hard limit on the request body size. The other PR fixes enabling the request size limit middleware which is API specific.
In my opinion, those are two different things. But if it's easier to handle on your side, I can for sure combine both changes here.

titpetric:
My mistake, it seemed like a patch of this one. Will do housekeeping there too, thanks :)

PR #3991, opened 2022-04-06
Run go mod tidy so that goreleaser can run.

asutosh:
## Description
Run `go mod tidy` to update go.mod so that goreleaser doesn't complain
about dirty git state while releasing.


buger:
Fixed in branch

PR #3998, opened 2022-04-07
Fix plugin build script

asutosh:
## Description
Update plugin compiler smoke tests to align with  the new changes in plugin compiler  build where the plugin name has version and platform info in the name, 
Also fix the plugin compiler build script - there seems to be an issue when there's an internal dependency in the `go.mod` file(eg: `github.com/TykTechnologies/tyk/certs`) - it causes the build to fail as it causes the the source tree to be deleted while building the plugin.

### DIFF
--- a/ci/tests/plugin-compiler/test.sh
+++ b/ci/tests/plugin-compiler/test.sh
@@ -25,6 +25,7 @@ trap "$compose down" EXIT
 
 rm -fv testplugin/*.so || true
 docker run --rm -v `pwd`/testplugin:/plugin-source tykio/tyk-plugin-compiler:${tag} testplugin.so
+mv testplugin/testplugin*.so testplugin/testplugin.so
 $compose up -d
 sleep 2 # Wait for init
 curl -vvv http://localhost:8080/goplugin/headers
--- a/ci/images/plugin-compiler/data/build.sh
+++ b/ci/images/plugin-compiler/data/build.sh
@@ -63,9 +63,10 @@ yes | cp -rf $PLUGIN_BUILD_PATH/vendor/* $GOPATH/src || true \
 # We can't just copy Tyk dependencies on top of plugin dependencies, since different package versions have different file structures
 # First we need to find which deps GW already has, remove this folders, and after copy fresh versions from GW
 
-# github.com and rest of packages have different nesting levels, so have to handle it separately
-ls -d $TYK_GW_PATH/vendor/github.com/*/* | sed "s|$TYK_GW_PATH/vendor|$GOPATH/src|g" | xargs -d '\n' rm -rf
-ls -d $TYK_GW_PATH/vendor/*/* | sed "s|$TYK_GW_PATH/vendor|$GOPATH/src|g" | grep -v github | xargs -d '\n' rm -rf
+# github.com and rest of packages have different nesting levels, so have to handle it separately - also exlude the tyk repo so that
+# it doesn't get deleted if we have any deps internal to the tyk module
+find "$TYK_GW_PATH/vendor/github.com" -mindepth 2 -maxdepth 2  -type d -not -path "$TYK_GW_PATH/vendor/github.com/TykTechnologies/tyk" | sed "s|$TYK_GW_PATH/vendor|$GOPATH/src|g" | xargs rm -rf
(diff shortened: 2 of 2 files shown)


Tyk-ITS:
Branch used: refs/pull/3998/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2110323559)

ci/tests/plugin-compiler/test.sh
```diff
@@ -25,6 +25,7 @@ trap "$compose down" EXIT
 
 rm -fv testplugin/*.so || true
 docker run --rm -v `pwd`/testplugin:/plugin-source tykio/tyk-plugin-compiler:${tag} testplugin.so
+mv testplugin/testplugin*.so testplugin/testplugin.so
```
buger, line 28: Why we need this on master?
Plugin compiler for master generate output files with architecture and version in file name. 
Master gateway knows how to load such files. 
So this line should be only in release-4-lts branches I guess, which not yet support such plugin names?
And even release-4 (upcoming release-4.1) should support it just fine too... 

**UPDATE**
On the other thought I realized that in this test we load plugin via file, which kinda force you change file name in API definition every time. I raise this as a bug for this functionality here https://tyktech.atlassian.net/browse/TT-4354. 
Now this line does makes sense.

buger, line 28: Why we need this on master?
Plugin compiler for master generate output files with architecture and version in file name. 
Master gateway knows how to load such files. 
So this line should be only in release-4-lts branches I guess, which not yet support such plugin names?
And even release-4 (upcoming release-4.1) should support it just fine too... 

**UPDATE**
On the other thought I realized that in this test we load plugin via file, which kinda force you change file name in API definition every time. I raise this as a bug for this functionality here https://tyktech.atlassian.net/browse/TT-4354. 
Now this line does makes sense.

alephnull:
Needs a rebase.

alephnull (changes requested):
Since we depends on GOPATH, and its status is somewhat nebulous 1.16 onwards, please consider adding
```
[ -z $GOPATH ] && { echo Please set GOPATH; exit 1 }
```
to the top of the script. As a belt and braces thing, perhaps also explicitly set GOPATH via ENV in the image.

blagerweij:
We've used a different approach than the proposed `go mod vendor` solution:

```
cd $TYK_SOURCE_PATH
go list -m -f '{{ if not .Main }}{{ .Path }} {{ .Version }}{{ end }}' all > dependencies.txt

# for any dependency also present in Tyk, change the module version to Tyk's version
cd $PLUGIN_SOURCE_PATH

# remove go module information
rm -rf go.mod go.sum

# initialize fresh module
go mod init mymodule

# add graphql-go-tools pinned version
go mod edit -replace github.com/jensneuse/graphql-go-tools=github.com/TykTechnologies/graphql-go-tools@$GOTOOLS_REF

# add Tyk version
go get github.com/TykTechnologies/tyk@$TYK_REF

# scan and update go dependencies
go mod tidy

# collect all plugin dependencies
go list -m -f '{{ if not .Main }}{{ .Path }} {{ .Version }}{{ end }}' all > dependencies.txt

# for any shared dependency, pin the version to Tyk gateway version
awk 'NR==FNR{seen[$1]=$2; next} seen[$1] && seen[$1] != $2' $PLUGIN_SOURCE_PATH/dependencies.txt $TYK_SOURCE_PATH/dependencies.txt | while read PKG VER; do
  go mod edit -replace=$PKG=$PKG@$VER
done
```
This uses Go modules 'replace' behaviour to fix versions for any dependency which is shared with Tyk gateway.

By tweaking the Docker images for plugin-compiler a bit, we can ensure that listing all dependencies for Tyk gateway is done ahead of time (during the creation of the Docker image), since these dependencies are fixed for a specific version of Tyk

alephnull:
Hey @asutosh would you follow up on this?

alephnull:
https://tyktech.atlassian.net/browse/TT-6402 seems to indicate that GOOS and GOARCH cannot be overridden on the command line.

asutosh:
Thanks @blagerweij ! 
Will follow up on your suggested changes.

PR #3999, opened 2022-04-07
[TD-959]: Add automation for running go mod tidy in release branches.

asutosh:
## Description
Add a GHA  workflow that will update go.mod and go.sum after running `go mod tidy`
on the current release branches.

## Related Issue
TD-959

## Motivation and Context
More details are in the associated issue.

## How This Has Been Tested
Can be tested once it's approved and merged to the appropriate release branches, and a commit is pushed to these
branches.

### DIFF
--- a/.github/workflows/go-mod-tidy.yml
+++ b/.github/workflows/go-mod-tidy.yml
@@ -1,0 +1,38 @@
+# This workflow will run on push to release-* branches, and runs a
+# go mod tidy on checked out source tree. If it causes go.mod or
+# go.sum to change - it would commit those changes.
+# This should take care of cherry-picked commits causing goreleaser
+# to fail with dirty git state errors.
+
+name: Tidy up go dependencies.
+
+on:
+  push:
+    branches:
+      - release-**
+
+jobs:
+  tidy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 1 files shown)


Tyk-ITS:
Branch used: refs/pull/3999/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2110384832)

ermirizio (approved):
Changes looks good, we have discussed with @asutosh arround raising awareness of this workflow which will make commits into release branches automatically. This is important to realize when tagging. Just make sure devs understand this workflow could eventually "patch" a release branch when porting. So at the time of tag lets make sure we tag over the new patched commit if that situation happens.
Some automation can be built arround this but, may better leave it as it is for now until we experience a lil bit more into this

alephnull (commented):
I have a philosophical objection to such automation. This is better served by a pre-commit hook.

ermirizio:
@alephnull we did consider the pre-hook option, but this won't work in case the release but (/release command) is used to port changes to a release branch. Unless there's a way to configure the bot to call the pre-commit hook. Also I'm concerned that hooks are not enforceable ( on non enterprise gh orgs) , so we'll need to trust on ppl goodwill and memory.
I would say lets create the pre-commit web-hook script as main tool but leave the workflow as fallback mechanism

alephnull:
My point is that it is too late to run such basic things after cherry picking a commit to a release branch.

PR #4097, opened 2022-06-01
TT-5017 Bugfix APIs with same domain and different mTLS rules

tbuchaillot:
When client does the HTTPS call, process of initiating connection call TLS Handshake. Before sending any HTTP data, client need to validate server certificates. In case of Mutual TLS, there is additional step which require server to validate client certificate too.

By default client should not send any certificates, so if server needs to explicitly tell client if certificate is required or not.
During the initial handshake server should respond to the client with the flag that indicated need for client certificate.
This flag can have the following values: "not need", "ask but not verify" (do not validate certificate on TLS handshake, and just forward it to HTTP level, so maybe some application logic can use this data), "ask and verify" (to not let request go to HTTP level, if certificate is not in the list of allowed certs, during TLS handshake)

We see the log line from client which says that we trying to make a call with client cert, and after you see log from the server saying that request has not any cert. This specific bug which we see happen because bug in the logic which decide which flag to send, and sends "not needed" flag, so even if you try to set "Certificate" propery when you make the request from client side, client after getting from server such flag, just do not send this certificate.

Why not always send "ask but not verify"? Because Web browsers. When users try to access such APIs, start showing you [following popup]( https://d3i71xaburhd42.cloudfront.net/04471d3bfde6cc2fce10d8e2ebf7c6b0364c6efa/7-Figure2-1.png. ) on each API use. E.g. they will ask you to specify the certificate. So we should be very careful, that we send the right certificate to the right API.
So the goal of the Tyk is based on the current domain (the only information available during the handshake), decide which API user trying to access, and if it has mTLS or not.

And things can get quite complex, what if you have multiple APIs on the same domain, but one has mTLS and another not? What if their domain is just empty? So we apply quite complex logic to resolve all these conflicts and do the right choice.

-  If it detected that with the given domain in mind, there is no APIs matching, send a "not needed flag". 

- If there is only ONE match, then validate the client on TLS level (the safest approach), and the "ask and verify" flag.

- And finally, if there is a conflict, and there are multiple APIs matching the given domain, and one of them uses mTLS and one not, send "ask but not verify", so we will still have a client certificate, but after it will jump to specific HTTP level middleware, and based on listen path, it will pinpoint the API we need, and will perform certificate check on HTTP level.

In 3.0.X we implemented a fix, which fixed mTLS for the domains which contain regexp.

But seems like it introduced a bug, so when you have multiple APIs with same domain, and diffrent mTLS rules, depending on API load order, it can send "not needed" flag. E.g. for the given domain we pick the first API in the list, without mTLS and not the one with mTLS.

This fix itself, it detects if API even have domain with regexp, and if not it rollbacks the logic we had previously (and it is actually also faster), and if it has regexp apply some specific logic (and we also fixed the bug there, to ensure that even if there were multiple matches, pick the "flag" value with the most priority).

## Related Issue

https://tyktech.atlassian.net/browse/TT-5017
## Motivation and Context

https://tyktech.atlassian.net/browse/TT-5017
## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


Tyk-ITS:
Branch used: refs/pull/4097/merge
Commit: a97dd1230c33140de57ff1b969bf3299e6f0f53f 
Triggered by: pull_request (@tbuchaillot)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2421178410)

Tyk-ITS:
Branch used: refs/pull/4097/merge
Commit:  
Triggered by: pull_request (@tbuchaillot)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2421159813)

Tyk-ITS:
Branch used: refs/pull/4097/merge
Commit: bb78023f1672e9af6cddca5f525f3aeef6567d5a 
Triggered by: pull_request (@tbuchaillot)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2429690015)

Tyk-ITS:
Branch used: refs/pull/4097/merge
Commit: f24314153af916fe94ac4a9be0fff971e1ec1592 
Triggered by: pull_request (@tbuchaillot)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2433942323)

sredxny (approved):
lgtm.

@buger @ilijabojanovic once we have information on how to replicate we will need to write unit tests for those use-cases

sredxny:
@buger could you review please? so we can add it to tyk 4.1

Tyk-ITS:
Branch used: refs/pull/4097/merge
Commit: 62aa91fd0afaaa4e80627eb0faf55621a2d1a5bc 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2448412393)

buger (approved):

Tyk-ITS:
Branch used: refs/pull/4097/merge
Commit: ebbfa9ffafb05079eddc63970f7103579a19cdd3 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2455730546)

tbuchaillot:
Cherry picked to `release-4-lts`

PR #4103, opened 2022-06-02
bugfix: Downgrade protobuf version for python 3.7

ermirizio:
Protobuf version major update introduces breaking changes against python 3.7 we need to downgrade in order to fix GW container not starting.

More info: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates


buger:
👍🏻

buger (approved):

PR #4129, opened 2022-06-14
Cleanup oas gateway tags

furkansenharputlu:


Tyk-ITS:
Branch used: refs/pull/4129/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2496607451)

apidef/oas/server.go
```diff
@@ -114,17 +118,14 @@ type GatewayTags struct {
 
 func (gt *GatewayTags) Fill(api apidef.APIDefinition) {
 	gt.Enabled = !api.TagsDisabled
-	if api.Tags == nil {
-		api.Tags = []string{}
-	}
 	gt.Tags = api.Tags
+	if gt.Tags == nil {
```
titpetric, line 122: We don't need to allocate, removed both cases

furkansenharputlu: yes, but schema fails when it is null. I don't know maybe we should change schema

apidef/oas/server.go
```diff
@@ -58,8 +59,11 @@ func (s *Server) ExtractTo(api *apidef.APIDefinition) {
 	if s.ClientCertificates != nil {
 		s.ClientCertificates.ExtractTo(api)
 	}
+
 	if s.GatewayTags != nil {
 		s.GatewayTags.ExtractTo(api)
+	} else {
+		api.TagsDisabled = true
```
titpetric, line 66: The nil check should be the responsibiity of ExtractTo(), as the responsibility to fill apidef is there

gateway/api_test.go
```diff
@@ -2038,10 +2038,6 @@ func TestOAS(t *testing.T) {
 				Value: "/oas-api/",
 				Strip: false,
 			},
-			GatewayTags: &oas.GatewayTags{
```
titpetric, line 2041: Removed in PR

gateway/api_test.go
```diff
@@ -2193,76 +2189,6 @@ func TestOAS(t *testing.T) {
 				})
 			})
 
-			t.Run("with oas and gateway tags enabled", func(t *testing.T) {
```
titpetric, line 2196: I'd like to keep this one but move it accordingly. It tests acceptance of inputs and outputs via API and provides a contract of implementation.

gateway/api_test.go
```diff
@@ -2461,10 +2387,6 @@ func TestOAS(t *testing.T) {
 			fillPaths(&apiInOAS)
 			tykExt := apiInOAS.GetTykExtension()
 			tykExt.Info.Name = "patched-oas-api"
-			tykExt.Server.GatewayTags = &oas.GatewayTags{
```
titpetric, line 2464: Removed in PR

gateway/api_test.go
```diff
@@ -2482,16 +2404,11 @@ func TestOAS(t *testing.T) {
 
 			tykExt := apiInOAS.GetTykExtension()
 			delete(apiInOAS.ExtensionProps.Extensions, oas.ExtensionTykAPIGateway)
-			tykExt.Server.GatewayTags = &oas.GatewayTags{
```
titpetric, line 2485: Removed in PR

apidef/oas/server.go
furkansenharputlu: it is also outside in your PR, the only difference here we set `true` in `else` case because of reverse logic bool.

gateway/api_test.go
```diff
@@ -2518,14 +2435,7 @@ func TestOAS(t *testing.T) {
 
 			expectedTykExt.Server.ListenPath.Value = listenPath
 			expectedTykExt.Upstream.URL = upstreamURL
-			expectedTykExt.Server.CustomDomain = &oas.Domain{
-				Enabled: true,
-				Name:    customDomain,
-			}
-			expectedTykExt.Server.GatewayTags = &oas.GatewayTags{
```
titpetric, line 2525: Removed in PR but, styling wise:

- tags should be defined multi-line, so they are friendly to read (one field per line)
- tags should always define full name keys e.g. none of {true, customDomain} (not the case here)

Ideally we'd run a LSC change that expands all structs in the codebase... one day.

furkansenharputlu: Think of it calling a constructor function. When it has short amount of fields to set, it is ok to set in one line like a constructor.

Tyk-ITS:
Branch used: refs/pull/4129/merge
Commit: ba23ec121417b7607637e5a16434f6d3ce510850 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2496757855)

apidef/oas/oas_test.go
```diff
@@ -33,6 +29,7 @@ func TestOAS(t *testing.T) {
 		var resultOAS OAS
 		resultOAS.Fill(convertedAPI)
 
+		emptyOASPaths.Extensions = nil
```
titpetric, line 32: Shared value here with next test(s)

titpetric: Cleaned this up in an additional commit in the PR.

apidef/oas/root_test.go
```diff
@@ -16,7 +16,6 @@ import (
 func TestXTykAPIGateway(t *testing.T) {
 	t.Run("empty", func(t *testing.T) {
 		var emptyXTykAPIGateway XTykAPIGateway
-		Fill(t, &emptyXTykAPIGateway.Server.GatewayTags, 0)
```
titpetric, line 19: Removed in PR

apidef/oas/security_test.go
```diff
@@ -418,19 +418,13 @@ func TestOAS_OAuth(t *testing.T) {
...
-					Tags:    []string{},
-				},
 			},
 		},
 	}
 
 	var api apidef.APIDefinition
 	oas.ExtractTo(&api)
 
-	assert.False(t, api.TagsDisabled)
```
titpetric, line 432: This tests a default assertion. Unless we change the default, and it's not likely, the test verifies it.

apidef/oas/server.go
titpetric: Implementing as-is, leaving for code-improvement sprints.

titpetric: @furkansenharputlu actually no, the default gateway behavior is to have DisableTags=false (zero value)?

apidef/oas/server.go
titpetric: it's not null, it's cleared based on len(tags)==0 :+1:

PR #3873, opened 2022-02-07
TT-4354 Go plugins to support multiple Tyk versions

sredxny:
## Description
In order to support plugins with multiple tyk versions then was added a logic on how to load them using information such as: tyk version, operative system and architecture in which its running. The load Go Plugin logic will find first if there's an exact match with the provided path, if that file doesn't exist then it will build a path+pluginName like this:

1- Gather information like: os, tyk version and architecture
2- Obtain the clean name of the plugin
3- Obtain the directory where the plugins are stored
4-  Build a new plugin name in the format: `{plugin-dir}/{plugin-name}_{GW-version}_{OS}_{arch}.so`
5- Load the plugin using the new name

## Related Issue
https://tyktech.atlassian.net/browse/TT-4354

## Motivation and Context
Allow users to upload multiple version of a go plugin, and load it based on where its running

## How This Has Been Tested
- created go plugin following the instructions available in https://tyk.io/docs/plugins/supported-languages/golang/#plugin-development-flow
- Created api and added a plugin like:
```
 "custom_middleware": {
      "pre": [],
      "post": [
        {
          "name": "AddFooBarHeader",
          "path": "./mygoplugin/mygoplugin.so",
          "require_session": false,
          "raw_body_only": false
        }
      ],
      "post_key_auth": [],
      "auth_check": {
        "name": "",
        "path": "",
        "require_session": false,
        "raw_body_only": false
      },
      "response": [],
      "driver": "goplugin",
      "id_extractor": {
        "extract_from": "",
        "extract_with": "",
        "extractor_config": {}
      }
    },
```
- Name the plugin: `mygoplugin_v4.1.0_darwin_amd64.so`
- Curl the api: `curl -i http://tyk-gateway:8081/goplugin/get`
- In the result I can see the header Foo: Bar

**Testing the Plugin compiler**
Now the plugin compiler have the ability to generate the plugins with the name format set as explained above, in order to test it we need to create the plugin and check that the name is set as expected. The plugin compiler have the ability to receive by params the OS and Arch as 3rd and 4th param. To test, lets do:

- Create the plugin directory
- `go mod init tyk_plugin`
- `go mod edit -replace github.com/jensneuse/graphql-go-tools=github.com/TykTechnologies/graphql-go-tools@f818861b88dc
`
- `go get github.com/TykTechnologies/tyk@af8883a`
- Write the plugin following the instructions of https://tyk.io/docs/3-lts/plugins/supported-languages/golang/#write-the-plugin
[...]

PR #4160, opened 2022-06-28
use $CGO_CGOENABLED  instead of $CGO_CGOENABLE that doesnt exist

sredxny:
## Description
In the plugin compiler, in the command to compile a plugin use the proper value to set for CGO_CGOENABLE var. It's an outcome for the comment in https://github.com/TykTechnologies/tyk/pull/3873#discussion_r908340190

At this point I think as well we should remove it as officially the plugin compiler will only work for linux OS and not windows/darwin etc... so it sounds to me that we should not support this param.

## Related Issue
- To be created

## Motivation and Context
Be consistent

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/ci/images/plugin-compiler/data/build.sh
+++ b/ci/images/plugin-compiler/data/build.sh
@@ -73,5 +73,5 @@ rm -rf $TYK_GW_PATH/vendor
 
 rm /go/src/modules.txt
 
-GO111MODULE=off CGO_ENABLE=$CGO_ENABLE GOOS=$GOOS GOARCH=$GOARCH  go build -buildmode=plugin -o $plugin_name \
+GO111MODULE=off CGO_ENABLE=$CGO_CGOENABLED GOOS=$GOOS GOARCH=$GOARCH  go build -buildmode=plugin -o $plugin_name \
     && mv $plugin_name $PLUGIN_SOURCE_PATH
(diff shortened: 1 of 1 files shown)


Tyk-ITS:
Branch used: refs/pull/4160/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2577629278)

titpetric:
I don't see why we couldn't extend support to arm64 targets for example. Given goos and goarch arguments and a cross build dockerfile base, unofficially at the very least, we do enable people doing that.

You made a quick find/replace, but you missed the variable name, it's CGO_ENABLED; currently it's unset (unless it's in the Dockerfile), setting it to =1 would likely change behavior? I'd appreciate a bit of an investigation, or a short think. I think, we shouldn't set CGO_ENABLED and just remove the non functional bits here, keep behavior.

titpetric:
Hate to say it, but still a typo (missing underscore). Next one will cost you a michelada 📃😅

Check out the rest of the first comment, maybe we don't need any changes at all, other than a cleanup/removal. I'd prefer that over this

titpetric:
Closing as stale/incorrect multiple times, if you need it file a ticket.

Tyk-ITS:
Branch used: refs/pull/4160/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2932688781)

PR #4162, opened 2022-06-30
[TT-5396] default basicauth hash fix

titpetric:
Due to some hash type fragmentation in code over multiple packages, hash functions were only validated in part. As the storage package only implement a subset without bcrypt, the validation defaulted to murmur32 on invalid values. The user package did not define a full set of valid HashType values nor validation. This commit adds that.

## Description

1. added full user.HashType values for keys we support and validation of known HashTypes,
2. validated input config to fall back to bcrypt,
3. adjusted checkPassword to reflect bcrypt as a fallback

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


Tyk-ITS:
Branch used: refs/pull/4162/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2589146210)

gateway/api.go
```diff
@@ -375,12 +373,16 @@ func (gw *Gateway) basicAuthHashAlgo() string {
...
 
 	// Use `basic_auth_hash_key_function` if set;
-	if config.BasicAuthHashKeyFunction != "" {
-		return config.BasicAuthHashKeyFunction
+	algo := config.BasicAuthHashKeyFunction
+
+	// If hash function name is empty/invalid
+	if ok := user.IsHashType(algo); !ok {
+		// set default basic auth hash to bcrypt
+		return "bcrypt"
```
jeffy-mathew, line 381: shall we use the const `HashBCrypt` here?

gateway/mw_basic_auth.go
```diff
@@ -217,27 +217,33 @@ var errUnauthorized = errors.New("Unauthorized")
...
+		fallthrough
+	case user.HashMurmur32:
+		fallthrough
+	case user.HashMurmur64:
+		fallthrough
+	case user.HashMurmur128:
 		// Verify we have a valid Hash value
 		hashAlgo := string(session.BasicAuthData.Hash)
-		basicAuthHashAlgo := storage.HashAlgo(hashAlgo)
-		if basicAuthHashAlgo != hashAlgo {
```
tbuchaillot, line 232: @titpetric why this validation is removed?

gateway/mw_basic_auth.go
```diff
@@ -217,27 +217,33 @@ var errUnauthorized = errors.New("Unauthorized")
...
 		}
-	default:
+
+	case user.HashSha256:
+		fallthrough
+	case user.HashMurmur32:
+		fallthrough
+	case user.HashMurmur64:
+		fallthrough
+	case user.HashMurmur128:
```
tbuchaillot, line 231: @titpetric wouldn't it be better to have all those types in one case instead of using `fallthrough`?

gateway/mw_basic_auth.go
titpetric: At this point, any hash algo is validated and supported, so HashAlgo() was removed. HashAlgo also had a fallback to murmur32, while we want to keep bcrypt for this behavior.

gateway/mw_basic_auth.go
titpetric: I preferred vertical readability. Same, but go fmt has made this a bit longer :)

gateway/mw_basic_auth.go
jeffy-mathew: https://github.com/TykTechnologies/tyk/blob/master/gateway/mw_basic_auth.go#L291 - not sure if this is with the right context, but do we really need to have this check here?

gateway/api.go
titpetric: I'd leave it for now - no strong argument pro/contra, would be `string(user.HashBCrypt)`, as outside of the user package `string` is used, even tho a `user.HashType` type exists; The responsibility of password hashing could be it's own package on itself (internal/hash), rather than having a storage/user/gateway fragmented implementation.

gateway/mw_basic_auth.go
titpetric: At first glance, no, but it wasn't the code I touched; Seems like it's comparing a murmur64 hash to a bcrypt hash? Uhhhhhhhhhhhhhhhhhh :dagger:

tbuchaillot: > At this point, any hash algo is validated and supported, so HashAlgo() was removed. HashAlgo also had a fallback to murmur32, while we want to keep bcrypt for this behavior.

All the new/updated Sessions are going to have a hash algo validated/supported. We don't know about the existing ones.  
A misconfiguration there now is going to lead to hashing the password - `storage.HashStr(plainPassword, hashAlgo)` then failing instead of failing before hashing.
It's a super side scenario, but I don't think it would hurt to maintain the check. What do you think?

titpetric: Existing ones already define "bcrypt" as the cipher type (it was not configurable before), so no difference. We keep the default behavior if invalid configuration. A misconfiguration will lead to "bcrypt" as the default, before reaching this code here... api.go:L379 already implements what you're suggesting, I think :)

gateway/api.go
titpetric: Added it. :)

gateway/mw_basic_auth.go
titpetric: I'm going to resolve this as okay (backwards compatible). The above has already passed QA as well.

gateway/mw_basic_auth.go
titpetric: I've compacted the case statement in one line

tbuchaillot (approved):

PR #4185, opened 2022-07-14
[TT-6057] fix validate request OAS middleware with hostname matching.

jeffy-mathew:
[changelog]
fixed: validate request OAS middleware hostname check.

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`


Tyk-ITS:
Branch used: refs/pull/4185/merge
Commit:  
Triggered by: pull_request (@jeffy-mathew)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2670553874)

gateway/mw_oas_validate_request.go
```diff
@@ -52,7 +53,15 @@ func (k *ValidateRequest) ProcessRequest(w http.ResponseWriter, r *http.Request,
 		return nil, http.StatusOK
 	}
 
-	router, err := gorillamux.NewRouter(&k.Spec.OAS.T)
+	oasSpec := k.Spec.OAS.T
+	apiURL := getAPIURL(*k.Spec.APIDefinition, k.Gw.GetConfig())
+	u, err := url.Parse(apiURL)
```
titpetric, line 58: This err needs a check

titpetric (approved):

gateway/mw_oas_validate_request.go
jeffy-mathew: changed this part to be simpler, keeping relative URL to the listen path will solve the problem

gateway/mw_oas_validate_request.go
```diff
@@ -52,7 +52,13 @@ func (k *ValidateRequest) ProcessRequest(w http.ResponseWriter, r *http.Request,
 		return nil, http.StatusOK
 	}
 
-	router, err := gorillamux.NewRouter(&k.Spec.OAS.T)
+	// replacing servers object to just have listen path so that router.FindRoute(r) will not fail with strict hostname check
+	oasSpec := k.Spec.OAS.T
+	oasSpec.Servers = openapi3.Servers{
+		{URL: k.Spec.Proxy.ListenPath},
```
buger, line 58: Do we really need URL line here. Can be it just emty servers object?
I can envision issues when ListenPath contains regexp.

gateway/mw_oas_validate_request.go
```diff
@@ -52,7 +52,13 @@ func (k *ValidateRequest) ProcessRequest(w http.ResponseWriter, r *http.Request,
...
+	}
+
+	router, err := gorillamux.NewRouter(&oasSpec)
```
titpetric, line 61: Can we create an oas.NewRouter(spec) func. Since we will fork apidef, having the openapi dependency imported there may avoid some problems :)

gateway/mw_oas_validate_request.go
jeffy-mathew: adding empty servers won't work since we're proxying behind a listen path. 
kin/openapi3 internally use gorilla/mux router matching. 
When OAS servers object is empty, the routes registered will be what's in the OAS doc. eg `/pet`, `/store/order`. 
However, when the listen path is say `petstore` and not `/`, the request url will be `/petstore/pet`, `/petstore/store/order` etc. 
This won't match with the registered routes from OAS spec.

gateway/mw_oas_validate_request.go
jeffy-mathew: oas.Newrouter to just wrap this call?

titpetric: Just to invoid the import of openapi3 into gateway. We will need to match/update the dependency with apidef, just having the openapi3 import under apidef oas.

An option would just be to add an oas.PrepareHost(*OAS, string) *OAS, so we don't have a mux dependency in apidef;

This avoids syncing the openapi3 dependency versions from apidef into go.mod for consumers (gw, dash...)

gateway/mw_oas_validate_request.go
buger: I see. Can you try to add a test with listen path containing regexp? (we have examples of it in our tests).

Tyk-ITS:
Branch used: refs/pull/4185/merge
Commit: ecc27c8efd1ae5f0db54a20e68440cde5d4abcdc 
Triggered by: pull_request (@jeffy-mathew)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2676010017)

gateway/mw_oas_validate_request.go
titpetric: Figured out we have a few imports already, so we are going to address this separately. :)

gateway/mw_oas_validate_request.go
jeffy-mathew: added tests with listen path having regexp

PR #4176, opened 2022-07-08
[TT-5262] add dialCtxFn in gw to test.

jeffy-mathew:
fix: upstream mTLS flaky tests.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Refactoring or add test (improvements in base code or adds test coverage to functionality)

## Checklist

- [ ] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). If pulling from your own
      fork, don't request your `master`!
- [ ] Make sure you are making a pull request against the **`master` branch** (left side). Also, you should start
      *your branch* off *our latest `master`*.
- [ ] My change requires a change to the documentation.
  - [ ] If you've changed APIs, describe what needs to be updated in the documentation.
  - [ ] If new config option added, ensure that it can be set via ENV variable
- [ ] I have updated the documentation accordingly.
- [ ] Modules and vendor dependencies have been updated; run `go mod tidy && go mod vendor`
- [ ] When updating library version must provide reason/explanation for this update.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] Check your code additions will not fail linting checks:
  - [ ] `go fmt -s`
  - [ ] `go vet`

PR #4189, opened 2022-07-15
Fix/inject dialer conf to storage

titpetric:
## Description

- Copy `NewRedisClusterPool` to `NewRedisClusterPoolWithOptions`, add new parameter to allow modifying redis options,
- Extend `RedisController` with a new `SetConstructor` method - defaulting to `NewRedisClusterPool`;
- Hook in testutil.go / newGateway function to set the redis dialer in the redis.UniversalOptions;
- Don't create *Test in HMAC test utilities functions, two *Test instances contribute/enable flaky tests;

## Related Issue

https://tyktech.atlassian.net/browse/TT-3973
https://tyktech.atlassian.net/browse/TT-5228

## Motivation and Context

This fixes most `TestHMAC*` tests. I've only had failures in TestHMACAuthSessionPass and TestHMACAuthSessionSHA512Pass after the first commit, and they happened rarely. Hopefully we can get a complete fix together along with #4176. 

Second commit fixes the remaining flakyness for TestHmac*Pass; two *Test instances have been alive during a test, and the tests don't like that. No sir.

## How This Has Been Tested

```
$ gotest -run='TestHMACAuth.+Pass' -count=100 .
ok  	github.com/TykTechnologies/tyk/gateway	22.040s
$ gotest -run='TestHMAC.+' -count=100 .
ok  	github.com/TykTechnologies/tyk/gateway	93.279s
$ gotest -run='TestHMAC.+' -count=100 -race .
ok  	github.com/TykTechnologies/tyk/gateway	323.257s
```

image

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/mw_http_signature_validation_test.go
+++ b/gateway/mw_http_signature_validation_test.go
@@ -125,10 +125,7 @@ func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
 	}
 }
 
-func testPrepareHMACAuthSessionPass(tb testing.TB, hashFn func() hash.Hash, eventWG *sync.WaitGroup, withHeader bool, isBench bool) (string, *APISpec, *http.Request, string) {
-	ts := StartTest(nil)
-	defer ts.Close()
-
+func testPrepareHMACAuthSessionPass(tb testing.TB, ts *Test, hashFn func() hash.Hash, eventWG *sync.WaitGroup, withHeader bool, isBench bool) (string, *APISpec, *http.Request, string) {
 	spec := ts.Gw.LoadSampleAPI(hmacAuthDef)
 
 	session := createHMACAuthSession()
@@ -244,12 +241,15 @@ func testPrepareRSAAuthSessionPass(tb testing.TB, eventWG *sync.WaitGroup, priva
 
 func TestHMACAuthSessionPass(t *testing.T) {
 	test.Racy(t) // TODO: TT-3973
+
 	// Should not receive an AuthFailure event
 	var eventWG sync.WaitGroup
 	eventWG.Add(1)
+
 	ts := StartTest(nil)
 	defer ts.Close()
-	encodedString, spec, req, sessionKey := testPrepareHMACAuthSessionPass(t, sha1.New, &eventWG, false, false)
+
+	encodedString, spec, req, sessionKey := testPrepareHMACAuthSessionPass(t, ts, sha1.New, &eventWG, false, false)
 
 	recorder := httptest.NewRecorder()
 	req.Header.Set("Authorization", fmt.Sprintf("Signature keyId=\"%s\",algorithm=\"hmac-sha1\",signature=\"%s\"", sessionKey, encodedString))
@@ -269,12 +269,15 @@ func TestHMACAuthSessionPass(t *testing.T) {
 
 func TestHMACAuthSessionSHA512Pass(t *testing.T) {
 	test.Flaky(t) // TODO: TT-3973
+
 	// Should not receive an AuthFailure event
 	var eventWG sync.WaitGroup
 	eventWG.Add(1)
+
 	ts := StartTest(nil)
 	defer ts.Close()
-	encodedString, spec, req, sessionKey := testPrepareHMACAuthSessionPass(t, sha512.New, &eventWG, false, false)
+
+	encodedString, spec, req, sessionKey := testPrepareHMACAuthSessionPass(t, ts, sha512.New, &eventWG, false, false)
 
 	recorder := httptest.NewRecorder()
 	req.Header.Set("Authorization", fmt.Sprintf("Signature keyId=\"%s\",algorithm=\"hmac-sha512\",signature=\"%s\"", sessionKey, encodedString))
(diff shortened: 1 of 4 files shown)


Tyk-ITS:
Branch used: refs/pull/4189/merge
Commit: 887e94581d2c500267bf9ccb906e63a049c28c95 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2677553241)

Tyk-ITS:
Branch used: refs/pull/4189/merge
Commit: d23dec4eed8978193d42769846fff46ee7c9b87c 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2677562700)

Tyk-ITS:
Branch used: refs/pull/4189/merge
Commit: c7fcf324f50d5874e65d2778e8e65a17e3d0a390 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2677882267)

gateway/testutil.go
```diff
@@ -1030,6 +1030,14 @@ func (s *Test) newGateway(genConf func(globalConf *config.Config)) *Gateway {
 	gw := NewGateway(gwConfig, s.ctx)
 	gw.setTestMode(true)
 
+	// This overrides the default dialer for redis to always use 127.0.0.1 in tests;
```
buger, line 1033: I wonder if using 127.0.0.1 works inside github actions?

storage/redis_cluster.go
```diff
@@ -97,6 +97,69 @@ func NewRedisClusterPool(isCache, isAnalytics bool, conf config.Config) redis.Un
 	return client
 }
 
+func NewRedisClusterPoolWithOptions(isCache, isAnalytics bool, conf config.Config, cb func(*redis.UniversalOptions)) redis.UniversalClient {
```
buger, line 100: Why we can't just modify original function implementation? 
I do not think that copypaste here justified: it has only one line difference. 
If you want make this function to keep same signature, just add this last `cb` argument as optional e.g. with `...`

gateway/mw_http_signature_validation_test.go
```diff
@@ -125,10 +125,7 @@ func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) bool {
 	}
 }
 
-func testPrepareHMACAuthSessionPass(tb testing.TB, hashFn func() hash.Hash, eventWG *sync.WaitGroup, withHeader bool, isBench bool) (string, *APISpec, *http.Request, string) {
-	ts := StartTest(nil)
-	defer ts.Close()
-
+func testPrepareHMACAuthSessionPass(tb testing.TB, ts *Test, hashFn func() hash.Hash, eventWG *sync.WaitGroup, withHeader bool, isBench bool) (string, *APISpec, *http.Request, string) {
```
buger, line 128: It seems like that new `ts` argument never get used?

gateway/testutil.go
```diff
@@ -1030,6 +1030,14 @@ func (s *Test) newGateway(genConf func(globalConf *config.Config)) *Gateway {
...
+	// It also avoids dns lookups, the dns mock.
+	gw.RedisController.SetConstructor(func(cache bool, analytics bool, config config.Config) redis.UniversalClient {
+		return storage.NewRedisClusterPoolWithOptions(cache, analytics, config, func(cfg *redis.UniversalOptions) {
+			cfg.Dialer = test.LocalDialer()
```
buger, line 1037: So this is the actual race fix?

buger:
How all this changes are even related to HMAC tests, it was all about DNS mock race? I do not see any hmac specifc fixes here, or I'm missing smth?

gateway/testutil.go
titpetric: @buger passing tests say yes, this would apply

gateway/mw_http_signature_validation_test.go
titpetric: Sure is! First line of function, one down :arrow_down: :)

gateway/testutil.go
titpetric: No, the `-StartTest` in testPrepareHMACAuthSessionPass(); the local dialer only dials `127.0.0.1`; gateway by default doesn't even pass the tests on macs because `localhost` is ipv6 or something and doesn't work. I guess this fixes tests on macs, or maybe just jeffy's machine.

titpetric:
@buger you're missing absolutely nothing, straight on the point. This is dns mock removal shrapnel + Gateway concurrency setup/teardown issues (unadressed but avoided/avoidable by removing StartTest parallelism) + some weird redis `localhost` issue (the default host, but it's hitting the resolver). I suspect IPv6 as noted in a comment above, LocalDialer with 127.0.0.1 avoids the dns lookup and forces IPv4 and the CI and Jeffys laptop is happy.

We could avoid the local dialer probablly/maybe if we set the magic incantation of a tyk env `TYK_STORAGE_HOST=127.0.0.1` (or similar) which I suppose would be a workaround to run the tests more stably. No need now :)

storage/redis_cluster.go
titpetric: I remembered an alternative implementation here while checking out code coverage, now invoke the longer constructor from the shorter one, just wrapping all the implementation. I consider this concern resolved :)

Tyk-ITS:
Branch used: refs/pull/4189/merge
Commit: a0371f35a8d9b335c7fad3bcc9bea05e576dc999 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2679479091)

titpetric:
After some thought, I've decided against injecting the local dialer into storage. While it may solve some peculiar test case failures, it is valid to configure the redis host to point to a non-localhost instance, or to configure it to 127.0.0.1 via config/env if somebody is hitting the weird localhost/ipv6 issues. The HMAC code has been moved to #4199, dropping this draft PR as N/A.

PR #4217, opened 2022-08-03
TT-6162 on setting session lifetime, consider the sessionlifetime set at api level

sredxny:
## Description
In the handler of adding a new key (used by worker gateways in a MDCB environment) then consider that some apis has the session lifetime set at api level, in those cases we should use this value. If we're in the situation where a key has multiple access rights, and each api has defined different session lifetimes then the bigger one will be used.

**UPDATE**
While developing I found out a situation in which we were not consistent in the behavior of the lifetime calculation, it happens on-premise that if you have 2 apis with different session lifetime set then when you create a key and apply those 2 apis then any of the lifetimes is set (sometimes the biggest one, sometimes the smallest one). Why does this happens? it happens because while applying policies we always overwritte the sessionLifetime, so the lifetime of the last api deifnition processed will be the one saved. How to replicate this? follow the next instructions:

- Create 2 apis with SessionLifetime set, use a different one for each Api. (api A and api B)
- Go and create a key, applying apis: first select the api A and then the api B
- Check the TTL set.
- Go and create a new key but applying api B first and then api A
- Check the TTL set. it's different than the first api key.

As a user I would expect to have always the same TTL set no matter how did I selected the APIS. So, if we leave this in this way we would end up to have different behaviors in the worker nodes and controller nodes, while worker nodes will always set the bigger TTL, the controller node will just choose any of the sessionLifetimes configured. In this PR I took care over those situations and normalized how it's calculated (it should be the same in any environment and always choose the bigger one.)

## Related Issue
https://tyktech.atlassian.net/browse/TT-6162

## Motivation and Context
Give solution to https://tyktech.atlassian.net/browse/TT-6162 and be consistent in MDCB environments

## How This Has Been Tested
- Tested manually running an MDCB environment: created an api with session lifetime different than zero.
- Then created a key via dashboard for that API
- Then check in the worker node, when the key is synchronized that the TTL is different than zero and consistent with the value set via api def.
- Also added unit test

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/api.go
+++ b/gateway/api.go
@@ -245,12 +245,48 @@ func (gw *Gateway) applyPoliciesAndSave(keyName string, session *user.SessionSta
 		return err
 	}
 
-	lifetime := session.Lifetime(spec.GetSessionLifetimeRespectsKeyExpiration(), spec.SessionLifetime, gw.GetConfig().ForceGlobalSessionLifetime, gw.GetConfig().GlobalSessionLifetime)
+	// calculate lifetime considering access rights
+	lifetime := gw.ApplyLifetime(session, spec)
 	if err := gw.GlobalSessionManager.UpdateSession(keyName, session, lifetime, isHashed); err != nil {
 		return err
 	}
 
 	return nil
+}
+
+func (gw *Gateway) GetApiSpecsFromAccessRights(sess *user.SessionState) []*APISpec {
+	var apis []*APISpec
+	if sess != nil && len(sess.AccessRights) > 0 {
+		for apiID := range sess.AccessRights {
+			spec := gw.getApiSpec(apiID)
+			if spec != nil {
+				apis = append(apis, spec)
+			}
+		}
+	}
+
+	return apis
+}
+
+// ApplyLifetime calculates the lifetime for the key. It considers the access rights and the bigger lifetime will be used
+func (gw *Gateway) ApplyLifetime(sess *user.SessionState, specs ...*APISpec) int64 {
+	var lifetime int64
+
+	if len(sess.AccessRights) > 0 {
+		specs = gw.GetApiSpecsFromAccessRights(sess)
+	}
+
+	for _, spec := range specs {
+		if spec != nil {
+			sessionLifeTime := sess.Lifetime(spec.GetSessionLifetimeRespectsKeyExpiration(), spec.SessionLifetime, gw.GetConfig().ForceGlobalSessionLifetime, gw.GetConfig().GlobalSessionLifetime)
+			// uses the greater lifetime
+			if sessionLifeTime > lifetime {
+				lifetime = sessionLifeTime
+			}
+		}
+	}
+
+	return lifetime
 }
 
 func resetAPILimits(accessRights map[string]user.AccessDefinition) {
@@ -274,7 +310,6 @@ func (gw *Gateway) doAddOrUpdate(keyName string, newSession *user.SessionState, 
 		// reset API-level limit to empty APILimit if any has a zero-value
 		resetAPILimits(newSession.AccessRights)
 		// We have a specific list of access rules, only add / update those
-
 		for apiId := range newSession.AccessRights {
 			apiSpec := gw.getApiSpec(apiId)
 			if apiSpec == nil {
@@ -315,13 +350,13 @@ func (gw *Gateway) doAddOrUpdate(keyName string, newSession *user.SessionState, 
 		log.Warning("No API Access Rights set, adding key to ALL.")
 		gw.apisMu.RLock()
 		defer gw.apisMu.RUnlock()
+
 		for _, spec := range gw.apisByID {
 			if !dontReset {
 				gw.GlobalSessionManager.ResetQuota(keyName, newSession, isHashed)
 				newSession.QuotaRenews = time.Now().Unix() + newSession.QuotaRenewalRate
 			}
 			gw.checkAndApplyTrialPeriod(keyName, newSession, isHashed)
-
 			// apply polices (if any) and save key
 			if err := gw.applyPoliciesAndSave(keyName, newSession, spec, isHashed); err != nil {
 				return err
@@ -672,7 +707,7 @@ func (gw *Gateway) handleGetAllKeys(filter string) (interface{}, int) {
 	return sessionsObj, http.StatusOK
 }
 
-func (gw *Gateway) handleAddKey(keyName, hashedName, sessionString, apiID string, orgId string) {
+func (gw *Gateway) handleAddKey(keyName, hashedName, sessionString, orgId string) {
 	sess := &user.SessionState{}
 	json.Unmarshal([]byte(sessionString), sess)
 	sess.LastUpdated = strconv.Itoa(int(time.Now().Unix()))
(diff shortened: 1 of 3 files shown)


sredxny:
@buger I would like you to validate the use-case in which we have multiple access rights with different session_lifetime set, so far we decided to use the bigger one, but is that right? or instead use the smaller one? 

This is related to the question posted in: https://stackoverflow.com/c/tyk-technologies/questions/2532

Tyk-ITS:
Branch used: refs/pull/4217/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2792841010)

gateway/api.go
```diff
@@ -3268,3 +3271,21 @@ func updateOASServers(spec *APISpec, conf config.Config, apiDef *apidef.APIDefin
 	newAPIURL := getAPIURL(*apiDef, conf)
 	oasObj.UpdateServers(newAPIURL, oldAPIURL)
 }
+
+// GetSessionLifetime defines the TTL of the key
+func GetSessionLifetime(sess *user.SessionState, gw *Gateway) int64 {
+	var lifetime int64
+	if sess != nil && len(sess.AccessRights) > 0 {
+		for apiID, _ := range sess.AccessRights {
```
tbuchaillot, line 3279: @sredxny please address linter comment here

gateway/api.go
```diff
@@ -3268,3 +3271,21 @@ func updateOASServers(spec *APISpec, conf config.Config, apiDef *apidef.APIDefin
 	newAPIURL := getAPIURL(*apiDef, conf)
 	oasObj.UpdateServers(newAPIURL, oldAPIURL)
 }
+
+// GetSessionLifetime defines the TTL of the key
+func GetSessionLifetime(sess *user.SessionState, gw *Gateway) int64 {
```
tbuchaillot, line 3276: @sredxny we already have a `session.Lifetime()`. Don't you think this would cause confusion?

gateway/api.go
```diff
@@ -681,11 +681,14 @@ func (gw *Gateway) handleAddKey(keyName, hashedName, sessionString, apiID string
 		return
 	}
 
+	lifetime := GetSessionLifetime(sess, gw)
+
 	var err error
 	if gw.GetConfig().HashKeys {
```
tbuchaillot, line 687: Maybe it could be cleaner if we create a key var for hashed/unhashed and we just call `gw.GlobalSessionManager.UpdateSession(keyVAR, sess, lifetime, gw.GetConfig().HashKeys)`?

gateway/api.go
```diff
@@ -3268,3 +3271,21 @@ func updateOASServers(spec *APISpec, conf config.Config, apiDef *apidef.APIDefin
 	newAPIURL := getAPIURL(*apiDef, conf)
 	oasObj.UpdateServers(newAPIURL, oldAPIURL)
 }
+
+// GetSessionLifetime defines the TTL of the key
+func GetSessionLifetime(sess *user.SessionState, gw *Gateway) int64 {
+	var lifetime int64
+	if sess != nil && len(sess.AccessRights) > 0 {
```
tbuchaillot, line 3278: we need to check if we are doing this in other parts of the gw code - maybe we can reutilize

tbuchaillot (changes requested):
@sredxny some comments.

buger (approved):
Looks good, but fix Tomas suggestions, or answer them.

gateway/api.go
sredxny: yes, will rename it

gateway/api.go
```diff
@@ -3268,3 +3271,21 @@ func updateOASServers(spec *APISpec, conf config.Config, apiDef *apidef.APIDefin
...
+			if spec != nil {
+				sessionLifeTime := sess.Lifetime(spec.GetSessionLifetimeRespectsKeyExpiration(), spec.SessionLifetime, gw.GetConfig().ForceGlobalSessionLifetime, gw.GetConfig().GlobalSessionLifetime)
+				// uses the greater lifetime
+				if sessionLifeTime > lifetime {
+					lifetime = sessionLifeTime
+				}
+			}
+		}
+	}
+	return lifetime
```
tbuchaillot, line 3290: this should returns -1 if the lifetime value is 0

gateway/api.go
sredxny: done! normalized the behavior

gateway/api.go
sredxny: @tbuchaillot zero is fine. The redis handler considers 0 as never expiry key

Tyk-ITS:
Branch used: refs/pull/4217/merge
Commit: db4c87659822214156341870f7c9d853f8df4988 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2799020958)

Tyk-ITS:
Branch used: refs/pull/4217/merge
Commit: 808922c2749e95b11734c3d05bba407fd970fece 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2799218924)

Tyk-ITS:
Branch used: refs/pull/4217/merge
Commit: 3fd2ce0cad85121b5641bffa45b606f885461bb5 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2799453369)

gateway/api.go
sredxny: Accordingly to the documentation: `// Zero expiration means the key has no expiration time.`

gateway/api.go
```diff
@@ -245,14 +245,51 @@ func (gw *Gateway) applyPoliciesAndSave(keyName string, session *user.SessionSta
...
+				apis = append(apis, spec)
+			}
+		}
+	}
+
+	return apis
+}
+
+// ApplyLifetime calculates the lifetime for the key. It considers the access rights and the bigger lifetime will be used
+func (gw *Gateway) ApplyLifetime(sess *user.SessionState, specs ...*APISpec) int64 {
```
tbuchaillot, line 273: @sredxny what are the risks of moving this logic inside `session.Lifetime`? it makes more sense if this is applied every time we try to get the session Lifetime..

sredxny: @tbuchaillot I would not do so, I tought about it as well. But this logic depends too much on apidefinitions and sometimes in a collection of apidefinitions, that's why I call `gw.GetApiSpecsFromAccessRights(sess)` so, I cannot just move this into the `user` package  because I would get a cycling import issue, so it will need to be refactored and receive all the values (only primitive data types and session), it will make the code less readable because everytime that you want to call the applyLifetime function then you will need to prepare the data to be sent. 
For sure it can be done, but I dont think the result will look beauty

tbuchaillot: @sredxny alright, thanks for explaining! Please create a ticket for that refactor so someone can have a look deeply.

tbuchaillot (approved):

Tyk-ITS:
Branch used: refs/pull/4217/merge
Commit: 0b81bc12112fd3b953bb865a8c0b0b6d66df3d3e 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2804139444)

Tyk-ITS:
Commit: fce0346e82fa13330aada2256bc23f6870143809 TT-6162 on setting session lifetime, consider the sessionlifetime set at api level (#4217)

* on setting session lifetime, consider the sessionlifetime set at api level

* normalize behaviour of session lifetime. It will always use the bigger lifetime found

* simplify handleAddKey
Triggered by: push (@sredxny)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2804908855)

sredxny:
cherry picked to `release-4.1` branch

sredxny:
manually cherry picked into release-4-lts and confirmed fix

PR #4211, opened 2022-07-29
Feature/redis storage drivers

titpetric:
## Description

WIP: Adding redis 7.0+ support, storage... *waves at everything*.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring or add test (improvements in base code or adds test coverage to functionality)

## Checklist

- [ ] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). If pulling from your own
      fork, don't request your `master`!
- [ ] Make sure you are making a pull request against the **`master` branch** (left side). Also, you should start
      *your branch* off *our latest `master`*.
- [ ] My change requires a change to the documentation.
  - [ ] If you've changed APIs, describe what needs to be updated in the documentation.
  - [ ] If new config option added, ensure that it can be set via ENV variable
- [ ] I have updated the documentation accordingly.
- [ ] Modules and vendor dependencies have been updated; run `go mod tidy && go mod vendor`
- [ ] When updating library version must provide reason/explanation for this update.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] Check your code additions will not fail linting checks:
  - [ ] `go fmt -s`
  - [ ] `go vet`

PR #4250, opened 2022-08-22
[TT-6144] Validate APIDefinition before use

titpetric:
## Description

There are scenarios where gateway doesn't tolerate some API inputs, notably from internal API endpoints which are implemented in `FromDashboard` and `FromRPC`. It's possible that an API Definition will be nil for a variety of reasons, the most important root cause of which is "never trust user input".

It's possible to be more strict here, rather than check for `nil`, we can use `ShouldOmit`, or even more strict, by creating a JSON schema and validating the API Definition playloads against it. All of this is out of scope, the commit fixes currently discovered errors in tests.

This is blocking for PR #4211 as a test was/is throwing DecodeFromDB errors there.

## Related Issue

https://tyktech.atlassian.net/browse/TT-6144

## Types of changes

## Checklist

  - [ ] `go fmt -s`
  - [ ] `go vet`

### DIFF
--- a/gateway/api_definition.go
+++ b/gateway/api_definition.go
@@ -408,7 +408,7 @@ func (f *nestedApiDefinitionList) validate() error {
 }
 
 type nestedApiDefinition struct {
-	*apidef.APIDefinition `json:"api_definition"`
+	*apidef.APIDefinition `json:"api_definition,inline"`
 	OAS                   *oas.OAS `json:"oas"`
 }
 
--- a/gateway/api_definition.go
+++ b/gateway/api_definition.go
@@ -398,12 +398,29 @@ type nestedApiDefinitionList struct {
 	Nonce   string
 }
 
+func (f *nestedApiDefinitionList) validate() error {
+	for idx, v := range f.Message {
+		if err := v.validate(); err != nil {
+			return fmt.Errorf("error validating api definition payload index=%d, %w", idx, err)
+		}
+	}
+	return nil
+}
+
 type nestedApiDefinition struct {
 	*apidef.APIDefinition `json:"api_definition"`
 	OAS                   *oas.OAS `json:"oas"`
 }
 
+func (n nestedApiDefinition) validate() error {
+	if n.APIDefinition == nil {
+		return errors.New("empty api definition")
+	}
+	return nil
+}
+
 func (f *nestedApiDefinitionList) set(defs []*apidef.APIDefinition) {
+	f.Message = make([]nestedApiDefinition, 0, len(defs))
 	for _, def := range defs {
 		f.Message = append(f.Message, nestedApiDefinition{APIDefinition: def})
 	}
@@ -454,16 +471,16 @@ func (f *nestedApiDefinitionList) filter(enabled bool, tags ...string) []nestedA
 // FromDashboardService will connect and download ApiDefintions from a Tyk Dashboard instance.
 func (a APIDefinitionLoader) FromDashboardService(endpoint string) ([]*APISpec, error) {
 	// Get the definitions
-	log.Debug("Calling: ", endpoint)
 	newRequest, err := http.NewRequest("GET", endpoint, nil)
 	if err != nil {
-		log.Error("Failed to create request: ", err)
+		return nil, fmt.Errorf("failed to create request: %w", err)
 	}
 
 	gwConfig := a.Gw.GetConfig()
 
-	newRequest.Header.Set("authorization", gwConfig.NodeSecret)
 	log.Debug("Using: NodeID: ", a.Gw.GetNodeID())
+
+	newRequest.Header.Set("authorization", gwConfig.NodeSecret)
 	newRequest.Header.Set(headers.XTykNodeID, a.Gw.GetNodeID())
 
 	a.Gw.ServiceNonceMutex.RLock()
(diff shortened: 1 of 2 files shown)


Tyk-ITS:
Branch used: refs/pull/4250/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2904675134)

gateway/api_definition.go
```diff
@@ -399,7 +399,7 @@ type nestedApiDefinitionList struct {
 }
 
 type nestedApiDefinition struct {
-	*apidef.APIDefinition `json:"api_definition,inline"`
+	*apidef.APIDefinition `json:"api_definition"`
```
furkansenharputlu, line 402: Please don’t remove it because it is the same structure with dashboard APIDefinition. Soon, we should use a common structure to prevent such changes.

gateway/api_definition.go
```diff
@@ -409,9 +409,21 @@ func (f *nestedApiDefinitionList) set(defs []*apidef.APIDefinition) {
 	}
 }
 
+func (f *nestedApiDefinitionList) all() []nestedApiDefinition {
+	result := make([]nestedApiDefinition, 0, len(f.Message))
+	for _, v := range f.Message {
+		// Clear nil APIDefinitions avoiding nil panics.
+		if v.APIDefinition == nil {
```
furkansenharputlu, line 416: We discussed this previously. Why would an API definition comes as nil? We should handle that part. In here if we want to do something, logging that api definition is invalid instead of returning valid ones.

furkansenharputlu (changes requested):
Added comments

gateway/api_definition.go
titpetric: https://github.com/golang/go/issues/6213 - `,inline` is not implemented on go stdlib json package, so this changes nothing. The value stayed embedded and read from `api_definition` field, as defined in the tag.

gateway/api_definition.go
titpetric: I agree, returning an error from such an API call (FromRPC, FromDashboard) makes sense, rather than panic (and instead of just skipping the nil value). I'll extend the PR accordingly. I don't think just logging a skipped nil is enough.

The error is also triggered from tests (`gateway.TestSyncAPISpecsRPCFailure_CheckGlobals`). If MDCB triggers a panic, which it did before and is logged in the JIRA ticket, that has it's own it's own owners who can investigate further. At least one case was that incompatible versions of MDCB and GW have been run.

furkansenharputlu: 👍

gateway/api_definition.go
furkansenharputlu: This is not a topic for this pr.

Tyk-ITS:
Branch used: refs/pull/4250/merge
Commit: d78bcc5c5dfda20016c9e8b9436408ef29909ac7 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2905487335)

gateway/api_definition.go
titpetric: It's like modifying a comment, it has no effect on the PR, the json package does not implement this. Do you have an idea of a test case that could fail?

furkansenharputlu: See it raises unrelated discussion to the actual issue. That’s why.

titpetric: The boyscout rule is the only way that this will realistically get merged, along with changes in the relevant code area. My first comment answers any possible concern you may have in removing `,inline`, I'd appreciate if you just click resolve conversation in this case, there isn't a conversation we can have about this, "this isn't a topic for this pr" also isn't helpful.

gateway/api_definition.go
```diff
@@ -441,16 +471,16 @@ func (f *nestedApiDefinitionList) filter(enabled bool, tags ...string) []nestedA
...
-		log.Error("Failed to create request: ", err)
+		return nil, fmt.Errorf("failed to create request: %w", err)
 	}
 
 	gwConfig := a.Gw.GetConfig()
 
-	newRequest.Header.Set("authorization", gwConfig.NodeSecret)
 	log.Debug("Using: NodeID: ", a.Gw.GetNodeID())
+
+	newRequest.Header.Set("authorization", gwConfig.NodeSecret)
```
jeffy-mathew, line 483: we have `headers.Authorization` defined and can be used here

gateway/api_definition.go
furkansenharputlu: If a change is made, it should be made in one ticket and all occurrences. This will create inconsistency in the code. The following ones should also be moved to one place and used in all places. Also, you are saying, removing it will not affect anything.

- https://github.com/TykTechnologies/tyk-analytics/blob/master/dashboard/model_api_definition.go#L57-L58
- https://github.com/TykTechnologies/tyk-sink/blob/ed4c1b9c0349bbe0e968ec15a53184132bce47fb/storage/store/store.go#L32-L33

Tyk-ITS:
Branch used: refs/pull/4250/merge
Commit: fca8f06fc3d25e1868880a5a1f6fac85264e9631 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2932141273)

Tyk-ITS:
Branch used: refs/pull/4250/merge
Commit: 4cbb4a07bed668926044df126c158e386ef11314 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2932199197)

gateway/api_definition.go
titpetric: Done.

jeffy-mathew (approved):

PR #4271, opened 2022-08-26
[TT-6400] Field based permissions whitelisting - Define better struct for RestrictedTypes

buraksezer:
I dropped [the first implementation](https://github.com/TykTechnologies/tyk/commit/25327c3fd6ab0590b12c3e9493fa40b4a1231b51) to prevent breaking the current API and GUI. I realized that tyk-analytics has the same `RestrictedTypes` field in various structs and business logic.

See the following:

https://github.com/TykTechnologies/tyk-analytics/blob/master/dashboard/api_admin_methods.go#L242
https://github.com/TykTechnologies/tyk-analytics/blob/6afe1ee6f5fcd3a3582287bf080a1d64fcf54b25/graphql/types.graphql#L7

This PR only(for [TT-6400](https://tyktech.atlassian.net/browse/TT-6400)) adds a new field called `AllowedFields` and has no impact on the API. I'm going to implement the actual business logic after merging this one. I plan to add a field named `enableAllowList` and modify the business logic in `CheckGraphqlRequestFieldAllowance` in `mw_graphql_granular_access.go` file. Then, probably we need to add a similar business logic to tyk-analytics but I'm not very familiar with that codebase. 

Related tickets on Jira:

https://tyktech.atlassian.net/browse/TT-5665
https://tyktech.atlassian.net/browse/TT-6110

### DIFF
--- a/gateway/middleware.go
+++ b/gateway/middleware.go
@@ -407,6 +407,14 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 							}
 						}
 
+						for _, t := range v.AllowedTypes {
+							for ri, rt := range r.AllowedTypes {
+								if t.Name == rt.Name {
+									r.AllowedTypes[ri].Fields = intersection(rt.Fields, t.Fields)
+								}
+							}
+						}
+
 						mergeFieldLimits := func(res *user.FieldLimits, new user.FieldLimits) {
 							if greaterThanInt(new.MaxQueryDepth, res.MaxQueryDepth) {
 								res.MaxQueryDepth = new.MaxQueryDepth
--- a/gateway/policy.go
+++ b/gateway/policy.go
@@ -23,6 +23,7 @@ type DBAccessDefinition struct {
 	Versions          []string                     `json:"versions"`
 	AllowedURLs       []user.AccessSpec            `bson:"allowed_urls" json:"allowed_urls"` // mapped string MUST be a valid regex
 	RestrictedTypes   []graphql.Type               `json:"restricted_types"`
+	AllowedTypes      []graphql.Type               `json:"allowed_types"`
 	FieldAccessRights []user.FieldAccessDefinition `json:"field_access_rights"`
 	Limit             *user.APILimit               `json:"limit"`
 }
@@ -34,6 +35,7 @@ func (d *DBAccessDefinition) ToRegularAD() user.AccessDefinition {
 		Versions:          d.Versions,
 		AllowedURLs:       d.AllowedURLs,
 		RestrictedTypes:   d.RestrictedTypes,
+		AllowedTypes:      d.AllowedTypes,
 		FieldAccessRights: d.FieldAccessRights,
 	}
 
--- a/gateway/policy_test.go
+++ b/gateway/policy_test.go
@@ -302,6 +302,26 @@ func (s *Test) TestPrepareApplyPolicies() (*BaseMiddleware, []testApplyPoliciesD
 					},
 				}},
 		},
+		"allowed-types1": {
+			ID: "allowed_types_1",
+			AccessRights: map[string]user.AccessDefinition{
+				"a": {
+					AllowedTypes: []graphql.Type{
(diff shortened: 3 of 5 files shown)


Tyk-ITS:
Branch used: refs/pull/4271/merge
Commit:  
Triggered by: pull_request (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2933397380)

buger:
It does makes sense! But can we have unit tests there? See `TestGraphQL_RestrictedTypes` as example. 

As for the dashboard, it should automatically sync this changes when PR will be merged, it use same data structures. 
And you should be able to set this fields via RESt API after this PR gets merged.
But we will need to add UI for that feature too.

gateway/middleware.go
```diff
@@ -407,6 +407,14 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 							}
 						}
 
+						for _, t := range v.AllowedTypes {
```
Aenimus, line 410: Is it possible we could improve the naming of these iterators? I don't see a reason why they couldn't be explicitly clear what the represent.

gateway/middleware.go
```diff
@@ -407,6 +407,14 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
...
+							for ri, rt := range r.AllowedTypes {
+								if t.Name == rt.Name {
```
Aenimus, line 412: Can the name match multiple times? So we cant break after the first match?
And similarly, would it be worth using a map or something where the keys are the name so we don't have to nest loops?

kofoworola (approved):

gateway/middleware.go
buraksezer: Golang authors promote short variable names https://github.com/golang/go/wiki/CodeReviewComments#variable-names

Aenimus: I think I have never disagreed with something more in life. Rather than looking at it and instantly knowing what means what, I have to keep looking back and forth to remind myself what this letter means or what that letter means. I understand that this isn't necessarily your philosophy, but it goes against every convention of Uncle Bob's Clean Code.

gateway/middleware.go
buraksezer: I think redundant names should be filtered somewhere else. For example, we can run a sanitizer before saving the configuration to filter the redundant names, if this is a problem. 

> And similarly, would it be worth using a map or something where the keys are the name so we don't have to nest loops?

Two levels of nested loops are just acceptable, I think.

Aenimus (approved):

Tyk-ITS:
Branch used: refs/pull/4271/merge
Commit: 7cf78989429ada08dcfc82542bd7463cdca5ab4e 
Triggered by: pull_request (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2963584190)

buger (approved):

buger:
Approved as well

buraksezer:
> It does makes sense! But can we have unit tests there? See `TestGraphQL_RestrictedTypes` as example.

I just added a new test case for merging allowed fields for the same GraphQL API. We'll implement the actual business logic in [TT-5665](https://tyktech.atlassian.net/browse/TT-5665).

Tyk-ITS:
Branch used: refs/pull/4271/merge
Commit: cfc938202c5a2b673f4b56e15647d134c195843b 
Triggered by: pull_request (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2963704451)

Tyk-ITS:
Commit: 2c9a1f5d4ea7905e218f946734b0ae7f102bb0f5 [TT-6400] Field based permissions whitelisting - Define better struct for RestrictedTypes (#4271)

* [TT-6400] Field based permissions whitelisting

* [TT-6400] add test cases for merging allowed fields for the same GraphQL API
Triggered by: push (@buraksezer)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/2963840409)

PR #4303, opened 2022-09-14
[TT-6587] Implement OAS mock response

furkansenharputlu:


buger:
Branch used: refs/pull/4303/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/3053118115)

buger:
Branch used: refs/pull/4303/merge
Commit: ac7126321b30102e290e2f61e7e4bf718f5b67ee 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/3053370079)

buger:
Branch used: refs/pull/4303/merge
Commit: 06314e7316e81d1cdcb2ed249dd1ca1f3b4a6378 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/3053440923)

gateway/mock_response.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (p *ReverseProxy) mockResponse(r *http.Request, route *routers.Route) (res *http.Response, err error) {
+	res = &http.Response{Header: http.Header{}}
+
+	var code int
+	var contentType string
+	var body []byte
+	var headers map[string]string
+
+	tykMockRespOp := p.TykAPISpec.OAS.GetTykExtension().Middleware.Operations[route.Operation.OperationID].MockResponse
```
buger, line 21: Should it actually check if map record exists or not?

gateway/mock_response.go
```diff
@@ -0,0 +1,50 @@
...
+	"bytes"
+	"io"
+	"net/http"
+
+	"github.com/TykTechnologies/tyk/apidef/oas"
+	header "github.com/TykTechnologies/tyk/headers"
+	"github.com/getkin/kin-openapi/routers"
+)
+
+func (p *ReverseProxy) mockResponse(r *http.Request, route *routers.Route) (res *http.Response, err error) {
```
buger, line 13: Why it is implemented as ReverseProxy code? Imho it should be middleware which sits in the end of the chain. See how it works in old mock plugin, you can write to response inside middleware, and then respond with `mwStatusRespond` which will stop MW chain and it will not reach reverse proxy.

```
    v.DoMockReply(w, meta)
    return nil, mwStatusRespond
```

gateway/reverse_proxy.go
```diff
@@ -923,6 +923,17 @@ func (p *ReverseProxy) handleOutboundRequest(roundTripper *TykRoundTripper, outr
 		latency = time.Since(begin)
 	}()
 
+	if p.TykAPISpec.HasMock {
```
buger, line 926: Same as above, imho should be standard middleware.

gateway/mock_response.go
furkansenharputlu: The idea here was not like to create it as a middleware. When you put it as middleware and return the response chains don't work. This way you can manipulate even response, add global headers etc.

gateway/mock_response.go
furkansenharputlu: It has the check in the outer function so here, it is safe to access to the map.

gateway/mock_response.go
buger: I see, hm. Have you considered then fix response middleware flow, and make it run when we respond with `mwStatusRespond` ?

buger: To be frank looking at `mwStatusRespond` current implementation, it is really bad, in a sense that it litelarly skips everything,probably including analytics. E.g. most likely you won't see legacy mock requests in log browser, and bunch of other stuff. And actually passing it though reverse proxy does makes sense. 
But I wonder if we still can make it work as middleware, but if `mwStatusRespond` is set, it will actually call reverse proxy, it will set some flag to request context or similar, which will automatically skip all middlewares except reverse proxy handler  (can be added inside `createMiddleware`, which will not require modifying all the Middlewares manually). And then this section:
```
			// Special code, bypasses all other execution
			if errCode != mwStatusRespond {
				// No error, carry on...
				meta["bypass"] = "1"
				h.ServeHTTP(w, r)
			} else {
				mw.Base().UpdateRequestSession(r)
			}
```
Will turn into smth like:
```
			// Special code, bypasses all other execution
			if errCode == mwStatusRespond {
                             setCtxValue(req, Response, ResponseProxyTypeObject)
                        }

                        // No error, carry on...
                        meta["bypass"] = "1"
                        h.ServeHTTP(w, r)
```

And after inside `SuccessHandler.ServeHTTP` you will have smth like:
``` 
        if val, found := getCtxValue(req, Response); found {
          resp = val
        } else {
	  resp := s.Proxy.ServeHTTP(w, r)
        }
```
Or maybe easier add it to `ReverseProxy.WrappedServeHTTP`, which skips all the HTTP call stuff, but it use response object from mock/mw.

But obviously it can look more complex as it looks...

furkansenharputlu: What is the harm of the current code ? Mocking the outbound request?

buger:
Branch used: refs/pull/4303/merge
Commit: 76eadf251526d83aece83f1adb399e280a9b5cdd 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/3060238373)

gateway/reverse_proxy.go
```diff
@@ -923,6 +923,17 @@ func (p *ReverseProxy) handleOutboundRequest(roundTripper *TykRoundTripper, outr
...
+		route, _, _ := p.TykAPISpec.OASRouter.FindRoute(outreq)
+		if route != nil {
```
titpetric, line 928: I suggest to move all these checks into mockResponse (without nesting ifs), add the validation on top of that function and return nil errors if appropriate/expected. Then we can unit test some things.

gateway/reverse_proxy.go
furkansenharputlu: HasMock does loops and checks every paths whether there is a mock. Why would it check for a normal request by looping everytime?

gateway/mock_response.go
furkansenharputlu: This is just a read operation. When a new update comes, it is not updating one field, a fresh api is loaded from scratch. And the global apis variable is protected with mutexes.

gateway/mock_response.go
```diff
@@ -0,0 +1,50 @@
...
+
+	res.Body = io.NopCloser(bytes.NewBuffer(body))
+
+	return res, nil
+}
+
+func mockFromConfig(tykMockRespOp *oas.MockResponse) (int, []byte, map[string]string) {
+	code := tykMockRespOp.Code
+	if code == 0 {
+		code = http.StatusOK
```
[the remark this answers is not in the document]
furkansenharputlu: It appear just once.

gateway/reverse_proxy.go
```diff
@@ -1289,6 +1300,12 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 			"org_id":      p.TykAPISpec.OrgID,
 			"api_id":      p.TykAPISpec.APIID,
 		}).Error("http: proxy error: ", err)
+
+		if strings.HasPrefix(err.Error(), "mock:") {
```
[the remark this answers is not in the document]
furkansenharputlu: The idea is to separate the error messages regarding the mock response. It is not just one error type that we can separate with `error.Is`. All mock errors have a common thing in that they have the prefix `mock:`. That's why it checks prefix.

Tyk-ITS:
Branch used: refs/pull/4303/merge
Commit: 0ca8775d35839ce9a42cb87b14bfa1b220dccaed 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/3149869457)

gateway/reverse_proxy.go
furkansenharputlu: Done

buger (approved):

Tyk-ITS:
Branch used: refs/pull/4303/merge
Commit: e8fb997fdfba14b85bebb18c1d99641f5410037d 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/3150007969)

jeffy-mathew (approved):

Tyk-ITS:
Commit: 91a5d5d9b845d1edeceea47d9233cebfa747bf8d Implement OAS mock response (#4303)
Triggered by: push (@furkansenharputlu)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/3150154387)

PR #4304, opened 2022-09-14
Tt 6335/kingfisher patch

ermirizio:
Kingfisher is running Gateway 3.1.2 on Gold SLA and experiencing the issue reported and fixed in https://tyktech.atlassian.net/browse/TT-1849.

They are not able to upgrade their gateway at this time and, to improve the reputation of Tyk and protect the revenue from Kingfisher, we should release a patch for GW 3.1.2 containing this fix.

This will allow them to continue with their operation and reinforce their confidence to upgrade to Tyk 4.x in due course.

This is the Slack discussion to-date: https://tyktech.slack.com/archives/C02AS4BP5HN/p1659006203972299
--------------------------------

We are currently having multiple issues with drifted files not only from CI but also from TYK code itself, like missing files related to https://tyktech.atlassian.net/browse/TT-2120


kolavcic (approved):
LGTM

Issue #4447, opened 2022-11-29
cannot have go-plugin bundles with support for v4.0.x and >=v4.1 


patriziobrunops:
I need to publish bundles containing go-plugins that support tyk-gateway v4.0 and >=v4.1. I'm currently running tyk-gateway 4.0.9 and I plan to update to v4.2.4. I have dozens of API definitions that use go-plugins for authorization. To prevent downtime, I need a plugin bundle that is compatible with all versions of Tyk.

Tyk >=v4.1 already supports plugin versioning in such a way that if the plain plugin name is not available, the versioned one is going to be used. v4.0.9 doesn't support such a behaviour and if the plain plugin name is not available, it simply fails.

**What is the current behavior?**
Currently, Tyk >=v4.1 supports plugin bundles compatible with multiple versions. So, once updated to v4.2.4 I'll be able to run such bundles. But in order to do that I need to allow some downtime when I update the gateway. The steps to update are as follows:
- build a new plugin bundle with plugins compatible with v4.2.4
- update the gateway to v4.2.4, during this phase my api-definition will be unable to run the authorization plugin, blocking my users
- change all of my api-definitions to use the new bundle

**What is the expected behavior?**
Build a bundle with plugins for v4.0.9 and v4.2.4 and when I update the gateway there's no downtime.

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
all

andyo-tyk:
Hi @patriziobrunops,

Sorry, closing this ticket was missed: we applied a solution for this in 4.3.2 (as discussed in your PR which you linked to this issue).

This is included from Tyk 4.3.2 onwards (and so is supported in current Tyk 5.x versions).

PR #4448, opened 2022-11-29
[FIX] attempt to load go-plugins from non-versioned paths causes panic

patriziobruno:
## Description
Enable Tyk to load plugin bundles that contain plugins compatible with tyk-gateway <=v4.1

Let's say that my plugin name is `super-authorization-plugin`. The solution I devised is:
- build a plugin for v4.0.9 named `super-authorization-plugin.so`
- build a plugin for v4.2.4 named `super-authorization-plugin_v4.2.4_linux_amd64.so`
- put both plugins in the same bundle with a manifest that looks as follows:
  ```json
  {
    "file_list": [
      "super-authorization-plugin_v4.0.9_linux_amd64.so",
      "super-authorization-plugin.so"
    ],
    "custom_middleware": {
      "auth_check": {
        "name": "DojoTykEnvoyAuthorize",
        "path": "super-authorization-plugin.so"
      },
      "driver": "goplugin"
    }
  }
  ```
- **when tyk v4.0.9 tries to load the plugin, it will find `super-authorization-plugin.so` and load it correctly**
- **when tyk v4.2.4 tries to load the plugin, it will find `super-authorization-plugin_v4.0.9_linux_amd64.so` load it correctly**, if it doesn't find it, it'll fall-back to `super-authorization-plugin.so`

## Related Issue
#4447 

## Motivation and Context

Update gateways with a high number of api-definitions using go-plugins from v4.0.x to >=v4.1.x causes down time. Because v4.0 doesn't support bundles containing multiple versions of the same plugin built for different gateway versions, and >=v4.1 is not able to load the correct plugin when a bundle contains plugins compatible with v4.0.

## How This Has Been Tested
Installed tyk-hybrid-gateway using helm on a GKE cluster, changing the container image to one built from this PR.

## Types of changes

## Checklist

### DIFF
--- a/gateway/mw_go_plugin.go
+++ b/gateway/mw_go_plugin.go
@@ -137,48 +137,26 @@ func (m *GoPluginMiddleware) loadPlugin() bool {
 
 	// try to load plugin
 	var err error
-	versionedPath := false
+	versionedPath := m.getPluginNameFromTykVersion(VERSION)
 
-	if !FileExist(m.Path) {
-		// if the exact name doesn't exist then try to load it using tyk version
-		m.Path = m.getPluginNameFromTykVersion(VERSION)
+	if FileExist(versionedPath) {
+		// Always prefer the versioned name
+		m.Path = versionedPath
 		m.logger = log.WithFields(logrus.Fields{
 			"versionedMwPath": m.Path,
 		})
-		versionedPath = true
 	}
 
 	defer func() {
 		if e := recover(); e != nil {
 			err = fmt.Errorf("%v", e)
 			m.logger.WithError(err).Error("Recovered from panic while loading Go-plugin")
-			m.Path = m.getPluginNameFromTykVersion(VERSION)
-			m.logger = log.WithFields(logrus.Fields{
-				"versionedMwPath": m.Path,
-			})
-
-			if m.handler, err = goplugin.GetHandler(m.Path, m.SymbolName); err != nil {
-				m.logger.WithError(err).Error("Could not load Go-plugin")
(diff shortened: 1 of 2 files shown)


sredxny:
@patriziobruno thank you for your contribution. Indeed we already do that, and it behaves like:

go-plugin-different-tyk-versions

Currently we have an issue with responsePlugins (they're not working with the new naming convention) and we will work in that in our closer sprint, as well as reorganizing precedence of loading goplugins based on naming.

patriziobruno:
> @patriziobruno thank you for your contribution. Indeed we already do that, and it behaves like:
> 
> go-plugin-different-tyk-versions
> 
> Currently we have an issue with responsePlugins (they're not working with the new naming convention) and we will work in that in our closer sprint, as well as reorganizing precedence of loading goplugins based on naming.

@sredxny thanks for getting back to me. I'm aware of the problems, but I probably explained it poorly. The problem is that versions >= 4.1 try to load the unversioned path first and often times end up `panic`king. If it tried to load the versioned path first, this wouldn't happen. Also, I'm using response plugins, and as you mentioned it's currently not supported. This PR addresses both problems and simplifies the plugin loading code.

sredxny:
@patriziobruno ahhh now I understand. Well, good new: we will be working to fix that precedence in the current sprint.

sredxny:
this has been implemented, tested and will be released in tyk v4.3.2.

patriziobruno:
@sredxny thanks! Is a backport to 4.2 and 4.0 to be expected?

PR #4501, opened 2022-12-07
[TT-6419] Querying for __typename in any graph causes a 502 error

buraksezer:
This PR fixes an edge case in proxy-only mode. When you run the following query:

```graphql
query { __typename }
```

The GW panics and returns 502 for the request because the recent version of graphql-go-tools can handle this query without hitting the upstream. It maintains the type names internally. 

See https://github.com/TykTechnologies/graphql-go-tools/pull/323 for further info.

PR #4521, opened 2022-12-12
Merging to release-4-lts: [TT-6419] Querying for __typename in any graph causes a 502 error (#4501)

buger:
[TT-6419] Querying for __typename in any graph causes a 502 error (#4501)

This PR fixes an edge case in proxy-only mode. When you run the
following query:

```graphql
query { __typename }
```

The GW panics and returns 502 for the request because the recent version
of graphql-go-tools can handle this query without hitting the upstream.
It maintains the type names internally.

See https://github.com/TykTechnologies/graphql-go-tools/pull/323 for
further info.

### DIFF
--- a/gateway/mw_graphql_test.go
+++ b/gateway/mw_graphql_test.go
@@ -1,6 +1,7 @@
 package gateway
 
 import (
+	"bytes"
 	"encoding/json"
 	"net/http"
 	"strings"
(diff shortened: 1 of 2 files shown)


Tyk-ITS:
Branch used: refs/pull/4521/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/3674963527)

Issue #4545, opened 2022-12-16
[TT-7393] Issue Redis Commands from Plugins


stewartlord:
**Is your feature request related to a problem? Please describe.**
Unable to directly issue commands to Redis from plugins

**Describe the solution you'd like**
Direct access to the Redis Universal Client

**Describe alternatives you've considered**
Attempted to use the `go:linkname` hack, but that fails to build due to symbols being unavailable.

**Additional context**
Evaluating Tyk at GitHub

andyo-tyk:
Thank you for your contribution to Tyk!

Per the discussion on your PR, we're taking storage management in a different direction, so I'll close this as "won't do".

We really appreciate you taking the time to make the suggestion and welcome any further ideas you have.

Thank you again.

PR #4546, opened 2022-12-16
[TT-7393] Add public function to directly access the Redis Universal Client

stewartlord:
## Description

Expose direct access to the underlying Redis Universal Client via `GetUniversalClient`

## Related Issue

https://github.com/TykTechnologies/tyk/issues/4545

## Motivation and Context

This is needed for plugin developers (in some cases) to issue commands that are not supported by the storage abstraction layer or have otherwise had their behavior changed. For example, it is not possible to issue a `SETEX` command. Similarly, the `MGET` command behaves differently and doesn't work in cases where some keys are expected to be missing.

## Types of changes

## Checklist

### DIFF
--- a/storage/redis_cluster.go
+++ b/storage/redis_cluster.go
@@ -138,6 +138,10 @@ func (r *RedisCluster) singleton() redis.UniversalClient {
 	return r.RedisController.singleton(r.IsCache, r.IsAnalytics)
 }
 
+func (r *RedisCluster) GetUniversalClient() redis.UniversalClient {
+	return r.singleton()
+}
+
 func (r *RedisCluster) hashKey(in string) string {
 
 	if !r.HashKeys {


titpetric:
We have work in progress under storage where we will be adding new drivers, which don't expose a RedisUniversalClient. Adding this function would be problematic, as it would become part of our API and we would have to support it indefinitely.

stewartlord:
> We have work in progress under storage where we will be adding new drivers, which don't expose a RedisUniversalClient. Adding this function would be problematic, as it would become part of our API and we would have to support it indefinitely.

I thought I added this to Redis specific code. Naturally it wouldn't belong on a generic interface. From my understanding this is below the storage abstraction layer.

Am I mistaken about that?

titpetric:
@stewartlord Hi. The underlying singleton() is accessible only in the storage package scope, while your addition would expose it as a public symbol one can use by importing the storage package. Already in-flight storage driver reworking will make it impossible to satisfy redis.UniversalClient as we're focusing on storage drivers improvements to support other kinds of storage drivers as well. The storage interface is intended to only provide a subset of redis functionality and some of our own on top of that required for gateway. This may be less than what you need for custom-behaviour plugins.

Particularly for SetEx however, we implement that behaviour with the underlying 'Set' (already exposed as SetKey() / SetRawKey()). Redis docs: https://pkg.go.dev/github.com/go-redis/redis/v8#Client.Set. I'm curious to learn more about what you're doing, if there are use case details you can share.

furkansenharputlu (approved):

stewartlord:
Hi @titpetric, sorry for the late reply. Thank you for considering this change. If it doesn't align with your future goals that's understandable. I appreciate the tip on `Set` 🙏

PR #4617, opened 2023-01-06
[TD-1413]: Add triggering of release workflow on tag push.

asutosh:
## Description
This seemed to have been mistakenly removed as part of an earlier PR, reverting.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,12 +16,10 @@ name: Release
 
 on:
   pull_request:
+  push:
     branches:
       - master
       - release-**
-      - integration/**
-      - feature/**
-      - perf/**
     tags:
       - 'v*'
 


Tyk-ITS:
Branch used: refs/pull/4617/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/3853191360)

alephnull (approved):

furkansenharputlu:
@asutosh It was on purpose actually. Why would it run twice for both pull request and push?

asutosh:
@furkansenharputlu All the packages and docker images are actually pushed on tag push with a `v` suffix. So whenever we cut a release - including triggering smoke tests with an `rc` release - it relies on this workflow to trigger. Otherwise, the release process wouldn't get triggered.

furkansenharputlu:
Then, should we remove pull request ? I just don’t like running them twice.

asutosh:
The problem is that it might add more process to the overall feedback loop. Let's say for example - during a PR if there is a change that affects the release pipeline(it's very common: even just renaming a package directory could cause the plugin compiler build to fail - as it also has to be added to the plugin compiler image using a directive in the release config).
So if the release workflow isn't run during the PR - it could be caught only after a merge and then requires another PR to fix it. With the workflow running for PRs, it's evident when the release pipeline fails and fixes can be included in the same PR.
@alephnull any thoughts?

ermirizio (approved):
Agreed , this process needs to run both on PR & Push

Tyk-ITS:
Branch used: refs/pull/4617/merge
Commit: 5ba833077a27f8c9e93f9c50af45c0ffa6ca7d34 
Triggered by: pull_request (@ermirizio)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/3873512306)

PR #4705, opened 2023-01-26
[TT-7873] remove `/tyk/{listenPath}/hello` path from swagger.yml

ls-michal-dabrowski:
## Description

Looks like `/tyk/{listenPath}/hello` doesn't exist and there is no way to enable it (I looked through all the docs I could find), it is only present in the `swagger.yml`. I also checked the source code a bit, but as I'm not super familiar with Go I can't say I'm 100% correct.

## Motivation and Context

Align `swagger.yml` with the actual implementation.

## Types of changes

## Checklist


lghiur:
Thank you @ls-michal-dabrowski for the contribution. Indeed the documented endpoint is not being implemented/used, we'll merge this issue in the following releases

Issue #4108, opened 2022-06-04
Tyk gateway reload api for policy list


cuttingedge1109:
**Branch/Environment/Version**
- Branch/Version: v4.1rc10
- Environment: On-prem

**Describe the bug**
I use tyk gateway OSS and manage(create/update/delete) resources(api and policy) via management apis.
When I create or delete new resource, I have to reload tyk process by consuming reload api.

When there is no policy left, reload is not working. i.e. The deleted policy is retrieved by list api even though there is no policy in policy path and also tyk log says `Policies found (0 total):`

**Reproduction steps**

1. Add a new policy if there is no policy yet in gateway
2. Delete all policies.
3. Reload
3. List policy
4. Then you can see one policy left which was deleted at last time

**Actual behavior**
The last deleted policy is retrieved by list api.

**Expected behavior**
After reload, list policy should return `[]`.

**Logs (debug mode or log file):**
```
tyk-gateway-docker-tyk-gateway-1  | time="Jun 04 14:13:16" level=debug msg="Primary instance set, I am master" prefix=host-check-mgr
tyk-gateway-docker-tyk-gateway-1  | time="Jun 04 14:13:26" level=debug msg="Primary instance set, I am master" prefix=host-check-mgr
tyk-gateway-docker-tyk-gateway-1  | time="Jun 04 14:13:34" level=debug msg="Deleting policy for: 17ca0ccc-b906-3915-8e41-6488143546c5"
tyk-gateway-docker-tyk-gateway-1  | time="Jun 04 14:13:34" level=info msg="Reload URL Structure - Scheduled" prefix=api
tyk-gateway-docker-tyk-gateway-1  | time="Jun 04 14:13:34" level=info msg="Reload queued" prefix=main
tyk-gateway-docker-tyk-gateway-1  | time="Jun 04 14:13:34" level=debug msg="Deleting API definition for: 2a25d1f9-b95b-3b0a-af5b-e59f19f29dd2"
tyk-gateway-docker-tyk-gateway-1  | time="Jun 04 14:13:34" level=info msg="Reload URL Structure - Scheduled" prefix=api
tyk-gateway-docker-tyk-gateway-1  | time="Jun 04 14:13:34" level=info msg="Reload queued" prefix=main
tyk-gateway-docker-tyk-gateway-1  | time="Jun 04 14:13:34" level=info msg="reload: initiating" prefix=main
[...]

PR #4855, opened 2023-03-11
[TT-5630]  [4108] Fixing policiesByID when last policy is deleted

SDkie:
The `if condition` is preventing overriding the `policiesByID` when the last policy is deleted. 

## Description

When any policy is deleted the map is reloaded, But when the last item is deleted, the map is not replaced with the empty map. So the `handleGetPolicyList` (`GET /tyk/policies`) returns the last deleted item

## Related Issue

https://github.com/TykTechnologies/tyk/issues/4108

## How This Has Been Tested

Tested the issue locally on my machine

## Types of changes

## Checklist

### DIFF
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -547,9 +547,7 @@ func (gw *Gateway) syncPolicies() (count int, err error) {
 
 	gw.policiesMu.Lock()
 	defer gw.policiesMu.Unlock()
-	if len(pols) > 0 {
-		gw.policiesByID = pols
-	}
+	gw.policiesByID = pols
 
 	return len(pols), err
 }


titpetric (approved):

buger:
Overall looks good! But i wonder if it was protection over situation when policies had issue during loading (like dashboard unavailable or RPC layer). I think it should check for the error instead. E.g. it should not set policies if there were error during their load. But in order to achieve it we also need to extend `gw.LoadPoliciesFromDashboard` to return error. 

Thanks!

SDkie:
@buger: how about checking the `pols` if it's empty or nil? As `pols` is nil when there is error fetching the policies.

I have made this change in the PR for your review.

SDkie:
@titpetric: Let me know what you think about the new changes. 
cc @buger

Issue #4831, opened 2023-03-04
[TT-8297] Enforced Timeout do not work as expected


ignashkins:
**Branch/Environment/Version**
- Branch/Version: master
- Environment: all

**Describe the bug**
[This code](https://github.com/TykTechnologies/tyk/blob/master/gateway/reverse_proxy.go#L1266) does not provide individual timeouts for each endpoint.
If MaxConnTime == 0, the transport below when using other endpoints will not be created and the timeout remains the same.

If MaxConnTime is set greater than 0, my guess is that the number of connections will increase, which will affect processing speed.

And it is not clear how it will work with http2...

**Reproduction steps**
1. Add API
2. Add hard_timeouts to extended_paths for multiple endpoints.
3. Execute http requests.

**Actual behavior**
The timeout will be chosen depending on the first called endpoint

**Expected behavior**
A different timeout will be worked for each endpoint

**Configuration (tyk config file):**
tyk.example.conf

SDkie:
If I add timeout for only `GET` method, I see that timeout is also added for other methods like `PUT` and `POST`
<img width="1207" alt="Screenshot 2023-03-12 at 4 19 53 PM" src="https://user-images.githubusercontent.com/1321321/224539862-6300264a-de94-481e-af13-e9db25f98b59.png">

SDkie:
@ignashkins is correct [this](https://github.com/TykTechnologies/tyk/blob/master/gateway/reverse_proxy.go#L1266) code is preventing re-creation of `HTTPTransport`. Because of which the timeout of the first called endpoint is used for all the matching endpoints called later. 

Actually the code structure is allowing to have only one HTTPTransport for all the matching endpoints. But each endpoint might have a different timeout.

andyo-tyk:
Hi @ignashkins 
I just spotted we hadn't replied on this issue - I'm sorry for the delay.

We are actively looking at this one at the moment, aiming to address it in an upcoming patch.
I will update you here when I have a confirmed release version.

Thanks for supporting Tyk!

andyo-tyk:
Hi @ignashkins 

I can confirm that we will be fixing this issue in the Tyk 5.0.7 patch, which is planned for October.

Thanks for supporting Tyk!

ignashkins:
@andyo-tyk 

Good day!  
Is there a solution to the problem at the moment?  
Thank you.

PR #4856, opened 2023-03-12
[#4831] Update HTTPTransport timeout based on Endpoint

SDkie:
Fixing #4831 Issue

## Description

Each `TykAPISpec` can have only one `HTTPTransport` therefore only one Timeout for all endpoints. This code looks for timeout information based on endpoint and updates the timeout in HTTPTransport. But this code adds overhead of looking for timeout information for each request
	
## Related Issue

#4831

## How This Has Been Tested

Ran the go tests and tested the fix manually

## Types of changes

## Checklist


titpetric:
I think this needs a better fix. Updating the transport timeout essentialy leaves us to pick the http transport based on the timeout value, rather than by apiid (would allocate more), or per-request - what this PR adds; I'm also not encouraged by the unprotected timeout write that applies to the whole api spec, and not just the request. Since every request is processed in a goroutine, concurrency needs to be taken into account when applying a timeout value, and ideally request lived or request-class shared transport needs to be used.

Something like a map[time.Duration]*http.Transport + a RWMutex to control map access should be good here.

SDkie:
@titpetric: Thanks for the feedback. This is just my first attempt to fix the issue, as I am new to the repo. I will try to update the PR based on the feedback I receive here.

PR #4913, opened 2023-03-26
[TT-8290] Fix policy acl-ratelimit partition combination

furkansenharputlu:
The rate and acl(access rights) partitioned policies merging for a key doesn't work correctly. This PR fixes it and adds a test case for it.

### DIFF
--- a/gateway/middleware.go
+++ b/gateway/middleware.go
@@ -530,9 +530,7 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 					ar.Limit.QuotaRenews = r.Limit.QuotaRenews
 				}
 
-				if !usePartitions || policy.Partitions.Acl {
-					rights[k] = ar
-				}
+				rights[k] = ar
 			}
 
 			// Master policy case
@@ -608,6 +606,11 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 
 	// If some APIs had only ACL partitions, inherit rest from session level
 	for k, v := range rights {
+		if !didACL[k] {
+			delete(rights, k)
+			continue
+		}
+
 		if !didRateLimit[k] {
 			v.Limit.Rate = session.Rate
 			v.Limit.Per = session.Per
--- a/gateway/policy_test.go
+++ b/gateway/policy_test.go
@@ -134,6 +134,21 @@ func (s *Test) TestPrepareApplyPolicies() (*BaseMiddleware, []testApplyPoliciesD
 			Rate:       4,
 			Per:        4,
 		},
+		"rate-for-a": {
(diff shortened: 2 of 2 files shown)


Tyk-ITS:
Branch used: refs/pull/4913/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/4526726732)

gateway/middleware.go
```diff
@@ -608,6 +606,11 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 
 	// If some APIs had only ACL partitions, inherit rest from session level
 	for k, v := range rights {
+		if !didACL[k] {
+			delete(rights, k)
```
buger, line 610: So, this code part assumes that if there were no policies which defined ACL, we should inherit it from the session object. In old logic it was do nothing. But in new logic, you just remove it, which can broke case when policies not used or no policy with ACL?

buger: I started writing comment for the first line too:

>I think it may introduce a new bug, when you added Quota for specific API, but not added ACL. 
I think the proper fix will be first run this code for all ACL/non-partition policies, to understand which APIs are allowed.

But then realised that this line attempts to fix it. And on line 664 https://github.com/TykTechnologies/tyk/pull/4913/files#diff-703054910891a4db633eca0f42ed779d6b4fa75cd9b3aa4c503e681364201c1bR664 we handle case when session rights are inherited. Which also probably another bug, since we we had a session with own ACL, and policy with customer rate/quotas partitions, it will loose this per API info. But I guess we can treat it as edge case.

buger:
Can you also add the following test, I have a feeling we can have issue with AllowanceScope..
Test case when there is standard ACL policy, but multiple partitioned, e.g separate for `a` and `b`

```
{
			Partitions:   user.PolicyPartitions{RateLimit: true},
 			AccessRights: map[string]user.AccessDefinition{"a": {}},
 			Rate:         4,
},
{
			Partitions:   user.PolicyPartitions{RateLimit: true},
 			AccessRights: map[string]user.AccessDefinition{"b": {}},
 			Rate:         2,
}
```

buger (approved):
Approving this PR, but also asked to double check another potential issue (if it will be, separate PR/issue will be raised)

Tyk-ITS:
Branch used: refs/pull/4913/merge
Commit: fe3815e499ae6411bcf4065bdf6facd9e7d21dc5 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/4541679588)

Tyk-ITS:
Branch used: refs/pull/4913/merge
Commit: c71a43f75830a11b7f12a3065718de60600ed2aa 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/4541693180)

jeffy-mathew (approved):

Issue #4469, opened 2022-12-01
[TT-6024]If middleware Go-Plugin not loaded successfully, request still allowed upstream


domsolutions:
**Branch/Environment/Version**
- Branch/Version: master
- Environment: all

**Describe the bug**
If a go plugin is compiled with the wrong version of Go or incorrect packages which don't match tyk gateway then the go plugin does not load but API is still available and allows requests upstream. We use go plugins for authentication so this is a big security concern. This scenario has happened multiple times where the provided plugin compiler image has either the wrong version of Go or wrong packages as vendor folder has not been updated in the image.

**Reproduction steps**

1. Add api... 
2. Add middleware   
```json
 "custom_middleware": {
    "pre": [
        {
            "path" : "/home/dominic/go/src/github.com/TykTechnologies/tyk/foobar-plugin.so",
            "name" : "AddFooBarHeader"
        }
    ],
        "post": [],
        "post_key_auth": [],
        "auth_check": {
        "name": "",
            "path": "",
            "require_session": false,
            "raw_body_only": false
    },
    "response": [],
        "driver": "goplugin",
        "id_extractor": {
        "extract_from": "",
            "extract_with": "",
            "extractor_config": {}
    }
}
```
Where `foobar-plugin.so` is compiled with different version of tyk that plugin is loaded into. Will see this error when tyk starts;

```sh
[Dec 01 12:05:50] ERROR Could not load Go-plugin error=plugin.Open("/home/dominic/go/src/github.com/TykTechnologies/tyk/foobar-plugin"): plugin was built with a different version of package github.com/google/uuid mwPath=/home/dominic/go/src/github.com/TykTechnologies/tyk/foobar-plugin.so mwSymbolName=AddFooBarHeader
```
But request is still allowed upstream.

Can an option be added to api def to make API unavailable if go-plugin is not loaded successfully? i.e. 

```json
        {
            "path" : "/home/dominic/go/src/github.com/TykTechnologies/tyk/foobar-plugin.so",
            "name" : "AddFooBarHeader",
            "fail_api_if_not_loaded" : true
        }
```

sod-lol:
can i try this?

furkansenharputlu:
Thanks for reporting! This issue is fixed and it will be released with `4.0.13` and `5.0.1` versions.

domsolutions:
Awesome, thanks for that.

PR #4922, opened 2023-03-30
[TT-6024] Return error when GoPlugin handler is nil

furkansenharputlu:
When GoPlugin load fails during API load, its handler is set as `nil`. The gateway should stop and return error when the API is called instead of skipping to other middlewares.

Fixes https://github.com/TykTechnologies/tyk/issues/4469

PR #4932, opened 2023-03-31
Merging to release-4-lts: [TT-6024] Return error when GoPlugin handler is nil (#4922)

buger:
[TT-6024] Return error when GoPlugin handler is nil (#4922)

When GoPlugin load fails during API load, its handler is set as `nil`.
The gateway should stop and return error when the API is called instead
of skipping to other middlewares.

Fixes https://github.com/TykTechnologies/tyk/issues/4469


Tyk-ITS:
Branch used: release-4-lts
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/4574300473)

PR #5175, opened 2023-06-19
Tt 851 process multivalue headers on response plugins

sredxny:
## Description

Updated the code in `gateway/coprocess.go` to work with headers represented as Key => []Values
Is good to mention that now the object passed to the middleware contains the old format header (one key - one value) and the new format header (the mutivalue headers: one key- multiple values). So:

- If a new header is added in the old field, then this will be added as well in the new field
- If a header is removed from the old field, then it will be removed as well in the new field
- If a value of a header is updated then it will overwritte all the values for that header in the new field.

In the end, the new field is the one being used to put all the headers back in the response, because after calling the middleware a synchronization process of headers is executed just to ensure that all the changes are replicated in the new field, therefore, all the values for every header are transmitted to the end user.

## Related Issue

TT-851

## Types of changes

## Checklist


Tyk-ITS:
Branch used: refs/pull/5175/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/5314397651)

gateway/coprocess.go
```diff
@@ -582,6 +597,38 @@ func (h *CustomMiddlewareResponseHook) HandleResponse(rw http.ResponseWriter, re
 	return nil
 }
 
+func syncHeadersAndMultiValueHeaders(headers map[string]string, multiValueHeaders []*coprocess.Header) []*coprocess.Header {
+	updatedMultiValueHeaders := []*coprocess.Header{}
+
+	for k, v := range headers {
+		found := false
+		for _, header := range multiValueHeaders {
```
tbuchaillot, line 605: @sredxny why is this second loop/lookup necessary? We are appending to multiValueHeaders[k] the headers value.
Wouldn't something like
```
for k,v := headers{
 currentHeader := multiValueHeaders[k]
 currentHeader.Value = append(currentHeader.Value, []string{v}...) 
}
```
Do the same here?

gateway/coprocess.go
```diff
@@ -582,6 +597,38 @@ func (h *CustomMiddlewareResponseHook) HandleResponse(rw http.ResponseWriter, re
 	return nil
 }
 
+func syncHeadersAndMultiValueHeaders(headers map[string]string, multiValueHeaders []*coprocess.Header) []*coprocess.Header {
```
tbuchaillot, line 600: Please add a comment on the function here

tbuchaillot (approved):
@sredxny small comments - it looks good overall!

gateway/coprocess_test.go
```diff
@@ -137,3 +139,103 @@ func Test_shouldAddConfigData(t *testing.T) {
...
+					Values: []string{"value1"},
+				},
+				{
+					Key:    "Header2",
+					Values: []string{"value2"},
+				},
+			},
+		},
+		{
+			name: "removing a header",
```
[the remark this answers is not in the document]
sredxny: it cannot be done, basically because our code doesn't remove headers it just receive 2 structures that currently store headers: header (map of string-string) and multivalueheaders that is a map of string key and value an array of strings, so, it know the status before executing the plugin middleware, then the plugin can do anything: remove, add or update headers and then the gateway reads the result of that operation in which the header or headers could have changed, so, at this point we synchronize the keys, we basically check what was removed, added or updated from the univalue struct and replicate those changes in the multivalue struct. So, we dont remove headers, we just replicate changes

gateway/coprocess.go
sredxny: I see the point, however `multivalueHeaders` is not a map, but a pointer of `coprocess.Header` so we cannot do things as:
`multiValueHeaders[k]` but instead we must do things as `multivalueHeaders[any-position].key`

tbuchaillot (approved):

mativm02 (approved):

PR #5213, opened 2023-06-22
[TT-6402]: Remove native arm64 plugin compiler image

asutosh:
## Description

Since binaries for all architectures are built using golang cross - there's no point in maintaining a native plugin compiler image as any plugin that is built using the native image wouldn't get loaded on the packaged gateway binary built with the golang-cross build image.

## Related Issue

TT-6402


Tyk-ITS:
Branch used: refs/pull/5213/merge
Commit: 9180248aaa1632d9611bc75dcc318807ef7f7601 
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/5346649431)

PR #5218, opened 2023-06-23
Bug redis cache orphaned connections

domsolutions:
…reassigned so defer res.Body.Close() does not close the connection leading to orphaned connections which eventually leads to max file descriptor limit being hit and requests not able to be sent from tyk with error; 'cannot assign requested address'

We noticed this error where our APIs use redis cache, which causes this block of code to run as `withCache` is `true`;

```golang
	inres := new(http.Response)
	if withCache {
		*inres = *res // includes shallow copies of maps, but okay

		if !upgrade {
```

This bug is also on `v2.9.4.7` where we noticed this issue.

## Types of changes

## Checklist

### DIFF
--- a/gateway/reverse_proxy.go
+++ b/gateway/reverse_proxy.go
@@ -38,7 +38,6 @@ import (
 	"github.com/TykTechnologies/graphql-go-tools/pkg/subscription"
 
 	"github.com/akutz/memconn"
-	"github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/http/httpguts"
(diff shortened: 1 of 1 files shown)


buger:
Hi! Can you explain a bit more this fix? It looks to me the same form Golang point of view.

domsolutions:
> Hi! Can you explain a bit more this fix? It looks to me the same form Golang point of view.

Hi Leo,

So in the original code;

```golang
                       defer res.Body.Close()

			// Buffer body data
			var bodyBuffer bytes.Buffer
			bodyBuffer2 := new(bytes.Buffer)

			p.CopyResponse(&bodyBuffer, res.Body)
			*bodyBuffer2 = bodyBuffer

			// Create new ReadClosers so we can split output
			res.Body = ioutil.NopCloser(&bodyBuffer)
			inres.Body = ioutil.NopCloser(bodyBuffer2)
```
`res.Body` gets reassigned to a new `io.ReaderCloser` so when `defer res.Body.Close()` runs it will close the new reader and not the network one.

PR #5174, opened 2023-06-19
TT-851 added new field in response object for multivalue headers

sredxny:
## Description

Added new field in response object to handle multivalue headers, this was added in the file: `coprocess/proto/coprocess_response_object.proto` all the other changes are automated changes for triggering the `update-bindings.sh` script

Note: merge this PR before https://github.com/TykTechnologies/tyk/pull/5173

## Related Issue

https://tyktech.atlassian.net/browse/TT-851

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring or add test (improvements in base code or adds test coverage to functionality)

## Checklist

- [ ] I ensured that the documentation is up to date
- [ ] I explained why this PR updates go.mod in detail with reasoning why it's required
- [ ] I would like a code coverage CI quality gate exception and have explained why

PR #5173, opened 2023-06-19
TT-851 update proto binding script

sredxny:
## Description

In order to fix the issue reported in TT-851 it was necessary to modify the structure of how the headers are stored and passed to the response middlewares, it happens that one header can have multiple values (Eg: Set-Cookie header) and the gateway currently considers only one value per header.
So, this PR contains:

- Update of the protobuf files and added a new field to handle the Headers, this is a repeated value instead a string and its name is `multivalue_headers`.
- Update of the update-bindings.sh script
- Updated the bindings for go, python, ruby
- Updated the code in gateway/coprocess.go to work with headers represented as Key => []Values

Is good to mention that now the object passed to the middleware contains the old format header (one key - one value) and the new format header (the mutivalue headers: one key- multiple values). So:

- If a new header is added in the old field, then this will be added as well in the new field
- If a header is removed from the old field, then it will be removed as well in the new field
- If a value of a header is updated then it will overwritte all the values for that header in the new field.

In the end, the new field is the one being used to put all the headers back in the response, because after calling the middleware a synchronization process of headers is executed just to ensure that all the changes are replicated in the new field, therefore, all the values for every header are transmitted to the end user.

## Related Issue

https://tyktech.atlassian.net/browse/TT-851

These are the changes of https://github.com/TykTechnologies/tyk/pull/5140 but segmented into multiple smaller PRs.

Children PRs:
- https://github.com/TykTechnologies/tyk/pull/5174
- https://github.com/TykTechnologies/tyk/pull/5175

## How This Has Been Tested

- Follow replication steps in the jira ticket

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring or add test (improvements in base code or adds test coverage to functionality)

## Checklist

- [ ] I ensured that the documentation is up to date
- [ ] I explained why this PR updates go.mod in detail with reasoning why it's required
- [ ] I would like a code coverage CI quality gate exception and have explained why

PR #5140, opened 2023-06-14
TT-851 fix/response coprocess headers are truncated

sredxny:
## Description

In order to fix the issue reported in TT-851 it was necessary to modify the structure of how the headers are stored and passed to the response middlewares, it happens that one header can have multiple values (Eg: Set-Cookie header) and the gateway currently considers only one value per header.
So, this PR contains:

- Update of the protobuf files and added a new field to handle the Headers, this is a repeated value instead a string and its name is `multivalue_headers`.
- Update of the update-bindings.sh script
- Updated the bindings for go, python, ruby
- Updated the code in gateway/coprocess.go to work with headers represented as Key => []Values

Is good to mention that now the object passed to the middleware contains the old format header (one key - one value) and the new format header (the mutivalue headers: one key- multiple values). So:

- If a new header is added in the old field, then this will be added as well in the new field
- If a header is removed from the old field, then it will be removed as well in the new field
- If a value of a header is updated then it will overwritte all the values for that header in the new field.

In the end, the new field is the one being used to put all the headers back in the response, because after calling the middleware a synchronization process of headers is executed just to ensure that all the changes are replicated in the new field, therefore, all the values for every header are transmitted to the end user.

## Related Issue

https://tyktech.atlassian.net/browse/TT-851

## Motivation and Context

Give solution to https://tyktech.atlassian.net/browse/TT-851

## How This Has Been Tested

The same as reported in the ticket

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring or add test (improvements in base code or adds test coverage to functionality)

## Checklist

- [ ] I ensured that the documentation is up to date
- [ ] I explained why this PR updates go.mod in detail with reasoning why it's required
- [ ] I would like a code coverage CI quality gate exception and have explained why

PR #5221, opened 2023-06-23
Merging to release-5.1: TT-851 update proto binding script (#5173)

buger:
TT-851 update proto binding script (#5173)

## Description

In order to fix the issue reported in TT-851 it was necessary to modify
the structure of how the headers are stored and passed to the response
middlewares, it happens that one header can have multiple values (Eg:
Set-Cookie header) and the gateway currently considers only one value
per header.
So, this PR contains:

- Update of the protobuf files and added a new field to handle the
Headers, this is a repeated value instead a string and its name is
`multivalue_headers`.
- Update of the update-bindings.sh script
- Updated the bindings for go, python, ruby
- Updated the code in gateway/coprocess.go to work with headers
represented as Key => []Values

Is good to mention that now the object passed to the middleware contains
the old format header (one key - one value) and the new format header
(the mutivalue headers: one key- multiple values). So:

- If a new header is added in the old field, then this will be added as
well in the new field
- If a header is removed from the old field, then it will be removed as
well in the new field
- If a value of a header is updated then it will overwritte all the
values for that header in the new field.

In the end, the new field is the one being used to put all the headers
back in the response, because after calling the middleware a
synchronization process of headers is executed just to ensure that all
the changes are replicated in the new field, therefore, all the values
for every header are transmitted to the end user.

## Related Issue

https://tyktech.atlassian.net/browse/TT-851

These are the changes of
https://github.com/TykTechnologies/tyk/pull/5140 but segmented into
multiple smaller PRs.

Children PRs:
- https://github.com/TykTechnologies/tyk/pull/5174
- https://github.com/TykTechnologies/tyk/pull/5175

## How This Has Been Tested

- Follow replication steps in the jira ticket

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing
functionality to change)
- [ ] Refactoring or add test (improvements in base code or adds test
coverage to functionality)

## Checklist

- [ ] I explained why this PR updates go.mod in detail with reasoning
why it's required
- [ ] I would like a code coverage CI quality gate exception and have
explained why


Tyk-ITS:
Branch used: refs/pull/5221/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/5361152406)

PR #5348, opened 2023-07-24
Call OIDC workflow

ermirizio:
testing OIDC workflow call


alephnull:
@ermirizio Is this PR required now that TD-1809 is done?

alephnull:
@ermirizio ping

Issue #5463, opened 2023-08-24
Add HTTP proxy protocol handling


ldmonster:
**Is your feature request related to a problem? Please describe.**
We already have an TCP handling of ha-proxy-protocol, but still not support HTTP/HTTPS.
In our case, we want to parse ha-proxy protocol in tyk, not any service before or after.

**Describe the solution you'd like**
We already have a solution for ourselves, but we want to tyk use it, because we want to use new stable versions of tyk with this feature.
We add new GatewayConfig toggle, to enable all ha-proxy-protocol handling on HTTP/HTTPS/h2c.

**Describe alternatives you've considered**
We already trying to use nginx before or after, but it looks like overhead, when we can do it on tyk

**Additional context**
This feature already requested before
https://github.com/TykTechnologies/tyk/issues/2300#issuecomment-560207754
Our branch
https://github.com/ldmonster/tyk/tree/feature/add-proxy-protocol

andyo-tyk:
Hi @ldmonster,

Thanks for getting in touch with this. I've taken a look at the original ticket and can see that it was auto-closed by a bot that we were running a while back, because the bot detected it as stale.

We're very open to community PRs, which are a lot easier to inspect than your fork of the Tyk codebase. Would you consider raising a PR, then we'll be able to take a look more promptly?

Thank you for supporting Tyk!

ldmonster:
> Hey folks! any update on supporting Proxy Protocol?

We have PR

https://github.com/TykTechnologies/tyk/pull/5464

OlegPS:
Hi @andyo-tyk , could you please take a look?

ldmonster:
@andyo-tyk PTAL

andyo-tyk:
Hi @OlegPS @ldmonster,

Sorry for the slow reply on this, not sure how I missed it. We'll take a look at this, please bear with me.

PrettyWood:
Any update on this @andyo-tyk. No pressure just to know if we can help and if yes how :)

andyo-tyk:
Hi,

Sorry for the lack of progress with this - it's failed to make the cut for the upcoming release (again 😬)... but still in sight and aiming to get this into the product when we can (hopefully Q2).

Again, I'm sorry this is taking so long - thank you for your patience.

PR #5464, opened 2023-08-24
[TT-9985] add proxy protocol for http

ldmonster:
Ha proxy protocol handling added

## Description

* When creating HTTPS listener - create default http listener and upgrade it. Because of flow we need. 
Usually, flow looks like proxyproto.Listener -> tls.Listener -> net.Listener. Because of it - we have an early tls handshake and fail. Now, we have a middle wrap, like tls.Listener -> proxyproto.Listener -> net.Listener , and it works smooth.
* Create helper functions to prevent repeating code
* Add parameters to setRouter func to create proxy with "useProxyProtocol"
* Add gateway config toggle named EnableProxyProtocolHTTP (HTTP because it toggle all kinds of HTTP traffic)

When i trying to upgrade listener in generateListener func - it throws error like "port is already in use". Because of it - listener upgrading now before HTTP server start.

## Related Issue

https://github.com/TykTechnologies/tyk/issues/5463

## Motivation and Context

With this feature we can use ha-proxy-protocol in our projects, with easy toggle from config file.

## How This Has Been Tested

Tested in our infrastructure using ha-proxy to wrap request with tls, and sending to tyk.
Result: we have a new header "x-forwarded-for" with correct IP on the board.

## Types of changes

## Checklist


gateway/proxy_muxer.go
```diff
@@ -439,43 +441,21 @@ func (m *proxyMux) serve(gw *Gateway) {
...
-				Handler:      handler,
+		case "https", "h2c":
+			mainLog.Warning("Starting HTTPS server on:", p.listener.Addr().String())
+			p.SetupServer(gw)
+			listener, err := UpgradeHTTPListenerToHTTPS(p.port, p.getListener(), gw)
+			if err != nil {
+				mainLog.WithError(err).Error("Can't start listener")
+				continue
 			}
-			if gw.ConnectionWatcher != nil {
```
titpetric, line 475: This likely shouldn't have been removed, it's use case is to track the connection state

gateway/api_loader.go
```diff
@@ -757,7 +757,7 @@ func (gw *Gateway) loadHTTPService(spec *APISpec, apisByListen map[string]int, g
 	router := muxer.router(port, spec.Protocol, gwConfig)
 	if router == nil {
 		router = mux.NewRouter()
-		muxer.setRouter(port, spec.Protocol, router, gwConfig)
+		muxer.setRouter(port, spec.Protocol, spec.EnableProxyProtocol, router, gwConfig)
```
titpetric, line 760: This seems wrong, EnableProxyProtocol is available from gwConfig, no additional parameters required.

config/config.go
```diff
@@ -1058,6 +1058,11 @@ type Config struct {
 	// Skip TLS verification for JWT JWKs url validation
 	JWTSSLInsecureSkipVerify bool `json:"jwt_ssl_insecure_skip_verify"`
 
+	// Enable proxy protocol for all HTTP/HTTPS/h2c
+	// Handle request with and without proxy protocol if true
+	// Handle only requests without proxy protocol if false
+	EnableProxyProtocolHTTP bool `json:"enable_proxy_protocol_http"`
```
titpetric, line 1064: This should come with a schema change in cli/linter; make sure you rebase and `test lint` is passing.

ldmonster: linter schema updated

gateway/api_loader.go
ldmonster: parameter removed

gateway/proxy_muxer.go
ldmonster: check returned

PR #5523, opened 2023-09-04
TT9860 - Decouple target response/dialer timeouts

domsolutions:
## Description

Currently the target dailer timeout (timeout to establish a connection) can't be changed without changing the target response timeout. This PR decouples the settings by adding a new setting in tyk.conf `proxy_dialer_timeout`

## Types of changes

## Checklist

### DIFF
--- a/cli/linter/schema.json
+++ b/cli/linter/schema.json
@@ -920,7 +920,10 @@
       }
     },
     "proxy_default_timeout": {
-      "type": "integer"
+      "type": "number"
+    },
+    "proxy_dialer_timeout": {
+      "type": "number"
     },
     "proxy_enable_http2": {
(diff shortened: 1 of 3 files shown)


buger:
Hi! Can you explain the actual scenario/use-case when this will be required?

Thanks!

domsolutions:
Hi @buger 

Say from checking analytics logs, it's determined the average RTT from GW to target is 100s. So we want to set the response timeout to 101 seconds, extra second to be on safe side. But we don't want to to wait 101 seconds to try and establish a connection to the target as if a connection can't be established say in 30 seconds, we can be confident there's an issue - no need to hang around for 101 seconds wasting resources.

PR #5526, opened 2023-09-04
[TT-9964] Prevent new idle connections to be generated.

lghiur:
https://tyktech.atlassian.net/browse/TT-9964

### DIFF
--- a/gateway/api_definition.go
+++ b/gateway/api_definition.go
@@ -275,7 +275,6 @@ func (s *APISpec) Release() {
 		// Prevent new idle connections to be generated.
 		s.HTTPTransport.transport.DisableKeepAlives = true
 		s.HTTPTransport.transport.CloseIdleConnections()
-		s.HTTPTransport = nil
 	}
 }
 
--- a/gateway/api_definition.go
+++ b/gateway/api_definition.go
@@ -270,6 +270,12 @@ func (s *APISpec) Release() {
 	if s.JSVM.VM != nil {
 		s.JSVM.DeInit()
 	}
+
+	if s.HTTPTransport != nil {
+		// Prevent new idle connections to be generated.
+		s.HTTPTransport.transport.DisableKeepAlives = true
+		s.HTTPTransport.transport.CloseIdleConnections()
+	}
 }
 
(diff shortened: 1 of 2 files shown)


jeffy-mathew (approved):

Tyk-ITS:
Branch used: refs/pull/5526/merge
Commit: dfb8af45c96c859038af3bf06e196f1d2d39068c 
Triggered by: pull_request (@jeffy-mathew)
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/6072866288)

gateway/api_definition.go
```diff
@@ -270,6 +270,12 @@ func (s *APISpec) Release() {
 	if s.JSVM.VM != nil {
 		s.JSVM.DeInit()
 	}
+
+	if s.HTTPTransport != nil {
+		// Prevent new idle connections to be generated.
+		s.HTTPTransport.transport.DisableKeepAlives = true
+		s.HTTPTransport.transport.CloseIdleConnections()
```
titpetric, line 277:
```suggestion
		s.HTTPTransport.transport.CloseIdleConnections()
		s.HTTPTransport = nil
```

gateway/api_definition.go
```diff
@@ -270,6 +270,13 @@ func (s *APISpec) Release() {
...
+		s.HTTPTransport = nil
```
jeffy-mathew, line 278: `curSpec.Release()` is called in a loop during API reload.
it's looping through old spec and releasing.
the specs are not replaced yet, and during the switchover if a request comes in, it would try to access the http transport, and this might end up in a nil pointer dereference? 
cc: @titpetric @buger

jeffy-mathew: `if s.HTTPTransport != nil {` is inside spec.Release(). 
what I'm concerned is about where HTTPTransport is actually being used in `(p *ReverseProxy) WrappedServeHTTP` 
https://github.com/TykTechnologies/tyk/blob/master/gateway/reverse_proxy.go#L1243
looking at https://github.com/TykTechnologies/tyk/blob/master/gateway/reverse_proxy.go#L1353 we can see it won't end up in nil pointer dereference, but might end up creating a new transport.

PR #5536, opened 2023-09-05
[TT-6446] Use integer values for timeouts in seconds

titpetric:
This splits up the float64/int changes over the reverse proxy into a separate PR for review.

### DIFF
--- a/gateway/reverse_proxy.go
+++ b/gateway/reverse_proxy.go
@@ -387,15 +387,15 @@ type ReverseProxy struct {
 
 var idleConnTimeout = 90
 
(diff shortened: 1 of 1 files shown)


jeffy-mathew (approved):

Tyk-ITS:
Branch used: refs/pull/5536/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/6081382849)

titpetric:
Dropping this, it's likely reasonable to set a dialer timeout smaller than a second right now, the change would be possibly changing behaviour in some edge case scenarios. Most values only allow `int` configuration, however, DefaultProxyTimeout allows something like `0.2`, technically. Not touching it right now.

PR #5541, opened 2023-09-06
[TT-6446] Minimal approach to setting a context-enforced timeout

titpetric:
This changes the default proxy timeout to not be based on HardTimeout configured from an APISpec.

- If a hard timeout is configured, a context with timeout is created for the outgoing request,
- The transport timeout is configured to [proxy_default_timeout](https://tyk.io/docs/tyk-oss-gateway/configuration/#proxy_default_timeout) or 30 secs by default.

https://tyktech.atlassian.net/browse/TT-6446

### DIFF
--- a/gateway/reverse_proxy.go
+++ b/gateway/reverse_proxy.go
@@ -538,21 +538,31 @@ func (p *ReverseProxy) ServeHTTPForCache(rw http.ResponseWriter, req *http.Reque
 	return resp
 }
 
-func (p *ReverseProxy) CheckHardTimeoutEnforced(spec *APISpec, req *http.Request) (bool, float64) {
+func (*ReverseProxy) CheckHardTimeoutEnforced(spec *APISpec, req *http.Request) (bool, float64) {
+	return checkHardTimeoutEnforced(spec, req)
+}
+
+func checkHardTimeoutEnforced(spec *APISpec, req *http.Request) (bool, float64) {
+	var defaultTimeout float64 = 30
+	if spec.GlobalConfig.ProxyDefaultTimeout > 0 {
+		defaultTimeout = spec.GlobalConfig.ProxyDefaultTimeout
+	}
+
 	if !spec.EnforcedTimeoutEnabled {
-		return false, spec.GlobalConfig.ProxyDefaultTimeout
+		return false, defaultTimeout
 	}
 
 	vInfo, _ := spec.Version(req)
 	versionPaths := spec.RxPaths[vInfo.Name]
 	found, meta := spec.CheckSpecMatchesStatus(req, versionPaths, HardTimeout)
 	if found {
-		intMeta := meta.(*int)
-		p.logger.Debug("HARD TIMEOUT ENFORCED: ", *intMeta)
-		return true, float64(*intMeta)
+		intMeta, ok := meta.(*int)
+		if ok && *intMeta > 0 {
+			return true, float64(*intMeta)
+		}
 	}
 
-	return false, spec.GlobalConfig.ProxyDefaultTimeout
+	return false, defaultTimeout
 }
 
 func (p *ReverseProxy) CheckHeaderInRemoveList(hdr string, spec *APISpec, req *http.Request) bool {
@@ -1349,6 +1359,8 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 
 	p.TykAPISpec.Lock()
 
+	_, enforcedTimeout := p.CheckHardTimeoutEnforced(p.TykAPISpec, outreq)
+
 	// create HTTP transport
 	createTransport := p.TykAPISpec.HTTPTransport == nil
 
@@ -1366,8 +1378,7 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 			oldTransport.DisableKeepAlives = true
 		}
 
-		_, timeout := p.CheckHardTimeoutEnforced(p.TykAPISpec, req)
-		p.TykAPISpec.HTTPTransport = p.httpTransport(timeout, rw, req, outreq)
+		p.TykAPISpec.HTTPTransport = p.httpTransport(enforcedTimeout, rw, req, outreq)
 		p.TykAPISpec.HTTPTransportCreated = time.Now()
 
 		if oldTransport != nil {
@@ -1409,6 +1420,12 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 		upstreamLatency time.Duration
 		err             error
 	)
+
+	// limit request time with context timeout
+	timeoutContext, cancel := context.WithTimeout(outreq.Context(), time.Duration(enforcedTimeout)*time.Second)
+	defer cancel()
+
+	outreq = outreq.WithContext(timeoutContext)
 
 	if breakerEnforced {
 		if !breakerConf.CB.Ready() {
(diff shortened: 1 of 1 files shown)


gateway/reverse_proxy.go
```diff
@@ -1349,6 +1367,16 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 
 	p.TykAPISpec.Lock()
 
+	isTimeoutEnforced, enforcedTimeout := checkHardTimeoutEnforced(p.TykAPISpec, outreq)
+
+	// limit request time with context timeout
+	if isTimeoutEnforced {
+		timeoutContext, cancel := context.WithTimeout(outreq.Context(), time.Duration(enforcedTimeout)*time.Second)
```
tbuchaillot, line 1374: @titpetric why are you using `outreq.Context` here instead or `req.Context()`?

titpetric: L1308 already modifies it; as i get it, it's used to issue the request.

gateway/reverse_proxy.go
```diff
@@ -538,21 +538,39 @@ func (p *ReverseProxy) ServeHTTPForCache(rw http.ResponseWriter, req *http.Reque
...
 
 	vInfo, _ := spec.Version(req)
 	versionPaths := spec.RxPaths[vInfo.Name]
 	found, meta := spec.CheckSpecMatchesStatus(req, versionPaths, HardTimeout)
 	if found {
-		intMeta := meta.(*int)
-		p.logger.Debug("HARD TIMEOUT ENFORCED: ", *intMeta)
-		return true, float64(*intMeta)
+		intMeta, ok := meta.(*int)
+		if ok && *intMeta > 0 {
```
furkansenharputlu, line 568: why did we remove debug log?

gateway/reverse_proxy.go
```diff
@@ -1349,6 +1367,16 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 
 	p.TykAPISpec.Lock()
 
+	isTimeoutEnforced, enforcedTimeout := checkHardTimeoutEnforced(p.TykAPISpec, outreq)
+
+	// limit request time with context timeout
+	if isTimeoutEnforced {
```
furkansenharputlu, line 1373: instead of this what if we call `if spec.EnforcedTimeoutEnabled {` and inside of `checkHardTimeoutEnforced` we just do

```
vInfo, _ := spec.Version(req)
	versionPaths := spec.RxPaths[vInfo.Name]
	found, meta := spec.CheckSpecMatchesStatus(req, versionPaths, HardTimeout)
	if found {
		intMeta := meta.(*int)
		p.logger.Debug("HARD TIMEOUT ENFORCED: ", *intMeta)
		return true, float64(*intMeta)
	}
```

gateway/reverse_proxy.go
```diff
@@ -538,21 +538,39 @@ func (p *ReverseProxy) ServeHTTPForCache(rw http.ResponseWriter, req *http.Reque
...
+
+// CheckHardTimeoutEnforced checks APISpec versions for a fine grained timeout
+// value. The value is defined in seconds, but we're using float64 to enable
+// sub-second durations for tests. Changing to int would break that behaviour.
 func (p *ReverseProxy) CheckHardTimeoutEnforced(spec *APISpec, req *http.Request) (bool, float64) {
+	return checkHardTimeoutEnforced(spec, req)
+}
+
+// Returns if the request should have a hard timeout enforced and the timeout value configured.
+func checkHardTimeoutEnforced(spec *APISpec, req *http.Request) (bool, float64) {
```
furkansenharputlu, line 558: why another function?

gateway/reverse_proxy.go
```diff
@@ -538,21 +538,39 @@ func (p *ReverseProxy) ServeHTTPForCache(rw http.ResponseWriter, req *http.Reque
 	return resp
 }
 
+// This returns the expected 30 second default timeout in case
+// the ProxyDefaultTimeout value is undefined.
+func proxyDefaultTimeout(spec *APISpec) float64 {
```
furkansenharputlu, line 543: can we apply rules of GoDoc, starting func name `proxyDefaultTimeout returns...`

gateway/reverse_proxy.go
```diff
@@ -538,21 +538,39 @@ func (p *ReverseProxy) ServeHTTPForCache(rw http.ResponseWriter, req *http.Reque
 	return resp
 }
 
+// This returns the expected 30 second default timeout in case
+// the ProxyDefaultTimeout value is undefined.
```
furkansenharputlu, line 542:
```suggestion
// proxyDefaultTimeout returns the default timeout of 30 seconds
// if the global ProxyDefaultTimeout value is undefined.
```

furkansenharputlu: added suggestion for GoDoc rule starting with func name and also some improvements

gateway/reverse_proxy.go
```diff
@@ -538,21 +538,39 @@ func (p *ReverseProxy) ServeHTTPForCache(rw http.ResponseWriter, req *http.Reque
...
+func proxyDefaultTimeout(spec *APISpec) float64 {
+	if spec.GlobalConfig.ProxyDefaultTimeout > 0 {
+		return spec.GlobalConfig.ProxyDefaultTimeout
+	}
+	return 30
```
furkansenharputlu, line 547: Can we put this one into a constant? `const defaultProxyTimeoutSeconds = 30.0`

gateway/reverse_proxy.go
titpetric: godoc applies to public functions

gateway/reverse_proxy.go
titpetric: because it does not need *ReverseProxy

gateway/reverse_proxy.go
titpetric: Seems subjective. What's the benefit?

gateway/reverse_proxy.go
furkansenharputlu: Reducing two if check to one

gateway/reverse_proxy.go
furkansenharputlu: Reducing two if check to one

gateway/reverse_proxy.go
furkansenharputlu: Then let’s remove it, why would we duplicate?

titpetric: It's removing an exposed function, can we guarantee the removal is non-breaking?

gateway/reverse_proxy.go
titpetric: it's not even the same value, if no hard timeout or invalid timeout is configured it will return false explicitly even if .EnforcedTimeoutEnabled is true

gateway/reverse_proxy.go
titpetric: re-added

gateway/reverse_proxy.go
titpetric: Reverted to single func due to the logger that's carried over *ReverseProxy.

gateway/reverse_proxy.go
titpetric: I'm rejecting this suggestion, as it just isn't good.

WrappedServeHTTP is SLOC a billion, CheckHardTimeouts func is SLOC ~20.
Function currently has locality of behaviour, splitting the condition out wouldn't.
Early returns are best, as they are easiest to understand for the reader.

It's a bad api to begin with, as the logger is passed via *ReverseProxy, the function is attached to the wrong scope or rather should get the logger explicitly and not even be aware of *ReverseProxy.

furkansenharputlu (approved):
Approving for it to go.

Tyk-ITS:
Branch used: refs/pull/5541/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/6121870387)

PR #5730, opened 2023-11-05
[TT-10109] Fix policy lookup map distortion

furkansenharputlu:
https://tyktech.atlassian.net/browse/TT-10109

The policy lookup map is distorted by `ApplyPolicies` function so it ends up wrong path base permission values in the session object.

See how it gets the original policy object and changes values inside it: https://github.com/TykTechnologies/tyk/blob/6c6b1535921543d2e2f34d65bbba7d67baffb547/gateway/middleware.go#L493

PR #5743, opened 2023-11-06
Merging to release-5.2: [TT-10109] Fix policy lookup map distortion (#5730)

buger:
[TT-10109] Fix policy lookup map distortion (#5730)

https://tyktech.atlassian.net/browse/TT-10109

The policy lookup map is distorted by `ApplyPolicies` function so it
ends up wrong path base permission values in the session object.

See how it gets the original policy object and changes values inside it:
https://github.com/TykTechnologies/tyk/blob/6c6b1535921543d2e2f34d65bbba7d67baffb547/gateway/middleware.go#L493

### DIFF
--- a/gateway/middleware.go
+++ b/gateway/middleware.go
@@ -492,6 +492,8 @@ func (t BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 				if !usePartitions || policy.Partitions.Acl {
 					didACL[k] = true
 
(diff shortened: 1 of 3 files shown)


buger:
Branch used: refs/pull/5743/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/6774538248)

Issue #5920, opened 2024-01-04
Optimize rate limit using Lua script


cuongtd1301:
**Is your feature request related to a problem? Please describe.**
I have the same issue with this https://github.com/TykTechnologies/tyk/issues/1876 
I see rate limiting as an important feature that should be implemented as a separate function instead of using a shared function (SetRollingWindow, GetRollingWindow) with other functionalities, which can result in resource waste

**Describe the solution you'd like**
After researching, I have decided to optimize the rate limit feature by using a Lua script for Redis. For more details, please refer to the pull request https://github.com/TykTechnologies/tyk/pull/5921

**Additional context**
I have built a sample API to compare the rate limit before and after the improvement. With a per value of 300, rate value of 60, and a server firing a load with 5 worker pools and 20k requests:

With the original rate limit, it takes 60 seconds to complete. The CPU usage increases by 60%, and the sorted set reaches 2MB with nearly 20k keys.

![image](https://github.com/TykTechnologies/tyk/assets/43245619/2620555e-a29a-4785-8340-1c0a239ba47c)
![image](https://github.com/TykTechnologies/tyk/assets/43245619/9a0741a7-8ab8-4467-840b-4af50d030ea9)

With the improved rate limit, it only takes 6 seconds to complete. The CPU usage peaks at 6%, and the sorted set size is reduced to 1KB with 60 keys.
![image](https://github.com/TykTechnologies/tyk/assets/43245619/53de4787-d087-42e9-b869-04a68c9c7d1b)
![image](https://github.com/TykTechnologies/tyk/assets/43245619/8edd9675-1cb4-41bd-863a-6340fb18cc60)

titpetric:
@cuongtd1301 hello, i've closed your PR with more detail; If you use the recently available fixed window rate limiter, you should be in the best case performance bracket, as it only needs to maintain a couple of redis keys. Note that it doesn't currently support spike arrest, so the expected behaviour is to block any overflow of `rate` for the remainder of `per` (rate limit window duration).

I'm leaving the issue open in case you'd want to rerun any tests or provide additional feedback. The sliding log behaviour is known to be inefficent, however, code structure since you've filed the issue has changed significantly, and fixed window rate limiting is available, as well as a sliding window rate limiter if you build your own gateway as it's currently only available when built with a `dev` tag; sliding window most matches sliding log by having spike arrest behaviour while having fixed resource usage on redis side.

andyo-tyk:
[...]

Issue #1876, opened 2018-08-21
Rate limit don't work under high QPS 


buxingzhe:
**Do you want to request a *feature* or report a *bug*?**

**What is the current behavior?**
Tyk uses a rolling window algorithm based on Redis commands like ZREMRANGEBYSCORE, ZRANGE, ZADD, EXPIRE. It always adds the new request track record first before rate limiting check.

**What is the expected behavior?**
When QPS is so high that it reaches the rate limit, then the incoming requests are rejected but these rejected requests are still counted(i.e. recorded) like normal requests in Redis. The problem is, from that point, all requests are rejected and can't recovering under high QPS.
The expected behavior should not counted the rejected requests when rate limit has reached.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem**
1. Set rate limit as 5 requests/second
2. Send 10 requests/second against gateway continuously.
3. Gateway will reach rate limit and after that time, it won't service any request and always complains too many requests.

**Which versions of Tyk affected by this issue? Did this work in previous versions of Tyk?**
Tyk 2.3+

buger:
Hello! Thank you for the report.  You should also try using our in-memory rate limiter (which is actually turned on by default now), by ensuring that both `enable_sentinel_rate_limiter` and `enable_non_transactional_rate_limiter` are set to false

buxingzhe:
Actually, we have tried the in-memory rate limiter, but it didn't fix the issue either.

buxingzhe:
It seems working now with in-memory rate limiter. But we deploy gateway as a cluster, does in-memory rate limiter check the rate globally?

buger:
Yes, it works across gateway cluster, since while it in-memory, multiple gateway instances constantly sync rate limits via Redis pub-sub mechanism.

buxingzhe:
That's great, does it work in Tyk 2.3? Currently, we use Tyk 2.3

buger:
DRL supported it from the start, so it should. But anyway, I would recommend updating to recent versions, since 2.3 version is not officially supported anymore.

Cheers.

wlf061:
If we use DRL,  should we set the db_app_conf_options.node_is_segmented  to ture and set all nodes' db_app_conf_options.tags as the same?

buger:
@wlf061 if you have setup with multiple gateway clusters, connected to the same redis, yes.

utkarsh079:
[...]

PR #5921, opened 2024-01-04
optimize rate limit using Lua script

cuongtd1301:
## Description

I see rate limiting as an important feature that should be implemented as a separate function instead of using a shared function (SetRollingWindow, GetRollingWindow) with other functionalities, which can result in resource waste
So I write func RunScript using Lua script for rate limit

## Related Issue

https://github.com/TykTechnologies/tyk/issues/5920

## Motivation and Context

I have the same issue with this [Rate limit don't work under high QPS](https://github.com/TykTechnologies/tyk/issues/1876)

## How This Has Been Tested

I have built a sample API to compare the rate limit before and after the improvement. With a per value of 300, rate value of 60, and a server firing a load with 5 worker pools and 20k requests:

With the original rate limit, it takes 60 seconds to complete. The CPU usage increases by 60%, and the sorted set reaches 2MB with nearly 20k keys.

With the improved rate limit, it only takes 6 seconds to complete. The CPU usage peaks at 6%, and the sorted set size is reduced to 1KB with 60 keys.

**Origin rate limit**

image
image

**Lua script rate limit**

image
image

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

### DIFF
--- a/certs/manager_test.go
+++ b/certs/manager_test.go
@@ -118,6 +118,10 @@ func (s *dummyStorage) RemoveSortedSetRange(string, string, string) error {
 	panic("implement me")
 }
 
+func (s *dummyStorage) RunScript(script string, keys []string, args ...interface{}) interface{} {
+	panic("implement me")
+}
+
 func newDummyStorage() *dummyStorage {
 	return &dummyStorage{
 		data:      make(map[string]string),
--- a/config/config.go
+++ b/config/config.go
@@ -721,6 +721,9 @@ type Config struct {
 	// This is used as part of the RPC / Hybrid back-end configuration in a Tyk Enterprise installation and isn’t used anywhere else.
 	AuthOverride AuthOverrideConf `json:"auth_override"`
 
+	// Rate limit using Lua script
+	EnableRedisScriptLimiter bool `json:"enable_redis_script_limiter"`
+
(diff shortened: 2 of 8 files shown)


titpetric:
It's unfortunate that a sliding log implementation is in use here, but recent versions allow you to consider using a fixed window rate limiter behaviour and this performance/resource usage penalty for sliding log is now clearly stated in the docs; Another behaviour to consider is spike arrest, which blocks traffic according to an ongoing rate (requests over the rate will count against the rate and block), and rate limit smoothing (ability to increase limits to allow more traffic gradually).

https://tyk.io/docs/getting-started/key-concepts/rate-limiting/#fixed-window-rate-limiter
https://tyk.io/docs/getting-started/key-concepts/rate-limiting/#rate-limit-smoothing

Per endpoint rate limits are scheduled for the next release.

Unfortunately your PR cannot be accepted, mostly due to this feature drift on our side, as well as how you approached the change. Our storage implementation is a particular area that's not suited for rate limits, as it's focused on providing an interface for several other processes. If you check into internal/rate or in general anywhere around rate limits today, you may find several positive changes during this time that try to address your root issue. I'm happy to revisit any feedback you may have against >=5.4 state (5.5 getting released soon).

Particularly sliding log is used to provide a buffered access-log like functionality in the analytics area and are mainly keeping it around for compatibility at this point.

PR #5925, opened 2024-01-05
[TT-10750] Extract and update sliding log functionality, update ratelimit

titpetric:
Implements an internal/rate.SlidingLog object to replace the sliding window functions inside the storage package. It extends the storage functions with an optimized counting method that avoids retrieving the sliding log from redis on a rate limiter decision. Rate limiter updated to use the optimized count functions.

The optimization is down to replacing ZRANGE with ZCARD, resulting in better memory use due to less data being requested from redis. It decreases the memory pressure on gateway.

| Benchmark                 | Iterations | Time/Iteration   | Memory/Iteration | Allocations/Iteration |
|---------------------------|------------|------------------|-------------------|------------------------|
| get_count_pipelined       | 294        | 40,607,040 ns/op | 1,617,069 B/op   | 43,032 allocs/op      |
| get_count_not_pipelined   | 316        | 44,422,147 ns/op | 1,264,799 B/op   | 33,025 allocs/op      |
| get_pipelined             | 76         | 155,023,361 ns/op| 22,227,962 B/op  | 545,534 allocs/op     |
| get_not_pipelined         | 75         | 150,580,911 ns/op| 21,883,555 B/op  | 535,527 allocs/op     |

Let's focus on summarizing the differences between get_count_pipelined and get_pipelined benchmarks:

- get_count_pipelined is approximately 51% faster than get_pipelined in terms of execution time per iteration.
- get_count_pipelined consumes approximately 86% less memory per iteration compared to get_pipelined.
- get_count_pipelined involves approximately 87% fewer allocations per iteration compared to get_pipelined.

These observations highlight the significant performance advantages of get_count_pipelined over get_pipelined in terms of execution time, memory consumption, and allocations per iteration.

Ultimately the sliding log functions are expensive for Redis as well, due to the time complexity of the operations, and the amount of log data being stored in redis. Consider the following:

- having shorter limit windows mean less data gets stored (60 req/second is better than 3600 req/minute)
- when going over the rate limit, requests still get logged to redis and continue without reducing load on redis

Different rate limiter implementations will be necessary to optimize scaling behaviour.

### DIFF
--- a/gateway/session_manager.go
+++ b/gateway/session_manager.go
@@ -1,6 +1,7 @@
 package gateway
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
@@ -10,6 +11,7 @@ import (
 	"github.com/TykTechnologies/leakybucket/memorycache"
 
 	"github.com/TykTechnologies/tyk/config"
+	"github.com/TykTechnologies/tyk/internal/rate"
 	"github.com/TykTechnologies/tyk/storage"
 	"github.com/TykTechnologies/tyk/user"
 )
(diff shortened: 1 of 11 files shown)


buger:
**API tests result - postgres15-murmur64 env: success** :white_check_mark: 
Branch used: refs/pull/5925/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/7421017134)

buger:
**API tests result - postgres15-sha256 env: success** :white_check_mark: 
Branch used: refs/pull/5925/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/7421017134)

buger:
**API tests result - mongo44-sha256 env: success** :white_check_mark: 
Branch used: refs/pull/5925/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/7421017134)

buger:
**API tests result - mongo44-murmur64 env: success** :white_check_mark: 
Branch used: refs/pull/5925/merge
Commit:  
[Execution page](https://github.com/TykTechnologies/tyk/actions/runs/7421017134)

internal/rate/sliding_log.go
```diff
@@ -0,0 +1,144 @@
...
+		res = pipe.ZCard(ctx, keyName)
+
+		element := redis.Z{
+			Score: float64(now.UnixNano()),
+		}
+
+		element.Member = strconv.Itoa(int(now.UnixNano()))
+
+		pipe.ZAdd(ctx, keyName, &element)
+		pipe.Expire(ctx, keyName, time.Duration(per)*time.Second)
```
jeffy-mathew, line 52: do we need this expiry?

internal/rate/sliding_log.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	if err := r.pipeliner(ctx, pipeFn); err != nil {
+		return 0, err
+	}
+
+	return res.Result()
+}
+
+// Set will append to a sorted set in redis and return the contents of the window as a slice.
```
jeffy-mathew, line 85: `Set` also trims the set, hall we add that to the godoc

internal/rate/sliding_log.go
```diff
@@ -0,0 +1,144 @@
...
+		res = pipe.ZRange(ctx, keyName, 0, -1)
+
+		element := redis.Z{
+			Score: float64(now.UnixNano()),
+		}
+
+		element.Member = strconv.Itoa(int(now.UnixNano()))
+
+		pipe.ZAdd(ctx, keyName, &element)
+		pipe.Expire(ctx, keyName, time.Duration(per)*time.Second)
```
jeffy-mathew, line 103: do we need this expiry?

gateway/session_manager.go
```diff
@@ -38,45 +40,59 @@ type SessionLimiter struct {
...
-		rate = currentSession.Rate
+		cost = currentSession.Rate
 	}
 
 	log.Debug("[RATELIMIT] Inbound raw key is: ", key)
 	log.Debug("[RATELIMIT] Rate limiter key is: ", rateLimiterKey)
 	pipeline := globalConf.EnableNonTransactionalRateLimiter
 
-	var ratePerPeriodNow int
+	ratelimit, err := rate.NewSlidingLog(store, pipeline)
```
jeffy-mathew, line 69: shall we maintain this as a singleton member in `SessionLimiter`

internal/rate/sliding_log.go
titpetric: Yes, but not on every request. It's preexisting and can be optimized further

internal/rate/sliding_log.go
```diff
@@ -0,0 +1,144 @@
...
+func NewSlidingLog(cluster storage.Handler, pipeline bool) (*SlidingLog, error) {
+	conn := new(redis.UniversalClient)
+	if err := cluster.As(conn); err != nil {
+		return nil, err
+	}
+
+	return NewSlidingLogRedis(*conn, pipeline), nil
+}
+
+func NewSlidingLogRedis(conn redis.UniversalClient, pipeline bool) *SlidingLog {
```
sredxny, line 27: should wemake this function only pkg scoped? and expose only NewSlidingLog?

titpetric: @sredxny I'm more inclined to remove the default one. I already implemented it as an interface so i can fully drop the storage package import (a lot of that package needs to be kicked out).

gateway/session_manager.go
titpetric: @jeffy-mathew actually no, as the SessionLimiter is not request scoped. We can't ensure a redis.UniversalClient will live forever, redis can go offline and this is where we need to handle the error. Depending on the situation that may mean that the gateway chokes on incoming requests as they each wait on the connection timeouts, if a singleton would be in use.

Added a constructor benchmark, it's 24B/op; essentially 4 pointers. I don't think they even escape to heap but did not verify.

jeffy-mathew (approved):

PR #6151, opened 2024-03-15
Fix/concurrent map access

buger:
Apply policies at the moment modify Session MetaData field directly. However if you look at implementation of other fileds like AccessRights or Tags, it creates a new objects, and after assign them to sessions.

This change should remove concurrency issues from MetaData map access, during session cloning, because technically now we never modify MetaData object, just replace it with another map.

It also avoids using mutexes so performance impact is minimal.

### DIFF
--- a/gateway/middleware.go
+++ b/gateway/middleware.go
@@ -381,10 +381,15 @@ func (t *BaseMiddleware) clearSession(session *user.SessionState) {
 // ApplyPolicies will check if any policies are loaded. If any are, it
 // will overwrite the session state to use the policy values.
 func (t *BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
+	session.Mu.Lock()
+	defer session.Mu.Unlock()
+	
 	rights := make(map[string]user.AccessDefinition)
 	tags := make(map[string]bool)
-	if session.MetaData == nil {
-		session.MetaData = make(map[string]interface{})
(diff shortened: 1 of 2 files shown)


gateway/middleware.go
```diff
@@ -682,6 +684,8 @@ func (t *BaseMiddleware) ApplyPolicies(session *user.SessionState) error {
 		session.Tags = appendIfMissing(session.Tags, tag)
 	}
 
+	session.MetaData = metadata
```
titpetric, line 687: Maps are pointers, this still leaves a data race over session.MetaData. Read at L386, Write here;

Suggestion:

- *user.SessionState needs to hold a (rw)mutex,
- protect access to underlying maps with a getter/setter,
- TT-5486 also diagnoses issues in Clone (see comment)

Ultimately the fix should be using an interface for session access/modification, as we need to protect several fields within (access rights, oauthkeys, apply policies, metadata, tags) due to the concurrency. Several functions also need pointer receivers to avoid copying the mutex and causing new issues.

How about avoiding concurrency altogether by having request scoped allocations?

titpetric: Opened https://github.com/TykTechnologies/tyk/pull/6154 ; this should prevent concurrency issues altogether, as the returned value from the session cache is a new allocation (not shared between goroutines, hopefully).

PR #6224, opened 2024-04-11
try

ermirizio:
Try regresion test exclusions over different release branches based on SYSE-354/add-test-controller-exclusions


ermirizio:
Closed , this was a succesfull tests, templates needs to be propagated now

PR #6247, opened 2024-04-24
[SYSE-358 release-5-lts] April template application

ermirizio:
Skip signing when building a snapshot to allow dependabot PRs to build.
Use larger runners for build and test.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,7 @@ env:
 jobs:
   goreleaser:
     name: '${{ matrix.golang_cross }}'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-m
     permissions:
       id-token: write # AWS OIDC JWT
       contents: read # actions/checkout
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -386,8 +431,6 @@ jobs:
           - amazonlinux:2023
           - registry.access.redhat.com/ubi8/ubi
           - registry.access.redhat.com/ubi9/ubi
-          - amazonlinux:2
-          - registry.access.redhat.com/ubi7/ubi
```
alephnull, line 390: If we are shipping packages, we should test them. Perhaps the template conditional is incorrect?

alephnull (changes requested):

.github/workflows/release.yml
ermirizio: added again

.github/workflows/release.yml
```diff
@@ -88,20 +82,20 @@ jobs:
           echo '#!/bin/sh
           ci/bin/unlock-agent.sh
           mkdir -p /go/src
-          go mod tidy
-          go mod vendor
+          GO111MODULE=on go mod tidy
+          GO111MODULE=on go mod vendor
```
alephnull, line 86: This seems weird. L98 below sets it to off, which should be the correct value.

asutosh: We need this, because this is run inside the container which runs with `GO111MODULE=off` as set in L98.
If it isn't set to to `on` for  these two statements, the vendored dependencies are not downloaded(because it's running it with GO111MODULE set to off)

alephnull: Please write up this logic in https://tyktech.atlassian.net/wiki/spaces/EN/pages/449708061/Release+Engineering#Gateway-specific-workflows

alephnull: Or in a template comment.

alephnull (approved):

PR #6297, opened 2024-05-22
Use canonical golang:1.22 image to test otel breaking

titpetric:
- try to build with golang:1.22
- eject arm/other builds for now


alephnull:
I guess this is not needed anymore?

PR #6329, opened 2024-06-07
[TT-12317][PoC] allow validating CA during mTLS

jeffy-mathew:
## Types of changes

## Checklist

### DIFF
--- a/internal/crypto/helpers.go
+++ b/internal/crypto/helpers.go
@@ -97,6 +97,11 @@ func ValidateRequestCerts(r *http.Request, certs []*tls.Certificate) error {
 		return errors.New("Client TLS certificate is required")
 	}
 
+	// no certs to verify
+	if len(certs) == 0 {
+		return nil
+	}
+
 	peerCertificate := r.TLS.PeerCertificates[0]
 
 	certID := HexSHA256(peerCertificate.Raw)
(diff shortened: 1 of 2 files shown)


internal/crypto/helpers.go
```diff
@@ -116,11 +116,30 @@ func ValidateRequestCerts(r *http.Request, certs []*tls.Certificate) error {
...
+			return nil
+		}
 	}
 
 	return errors.New("Certificate with SHA256 " + certID + " not allowed")
 }
 
+func verifyCertAgainstCA(caCert *tls.Certificate, peerCert *x509.Certificate) bool {
+	// Create a certificate pool and add the CA certificate to it
+	roots := x509.NewCertPool()
```
titpetric, line 130: Would we want to keep opts/roots around in memory so they can be reused between requests?

jeffy-mathew: Upon thinking about it again, the client certificates attached in the API would be used during TLS handshake and this additional check against certificate authority isn't required when the uploaded certificate is a CA certificate. 
Also, we could do an improvement by fetching the client certificates during API reload and caching it in middleware struct itself

sarobakhtiyari (approved):

jeffy-mathew:
feature implemented in https://github.com/TykTechnologies/tyk/pull/6405

PR #6338, opened 2024-06-11
LLM reporting Middleware that tracks token counts for AI APIs

lonelycode:
Adds a new middleware that uses tiktoken to count AI message tokens (support OpenAI and Anthropic)

## Description

- Adds `mw_llm_reporter.go` middleware 
- Hooks after transforms and auth
- The middleware is only loaded if the API is tagged with `llm`
- `LLMReporter` middleware will 
  - decode outbound body to a base API message format used by both Anthropic and OpenAI (Gemini uses a different content structure)
  - attempt to build a content blob from the message payload 
  - attempt to detect the model being used in the request
  - use the tiktoken library to estimate the number of tokens, if it can't get a clear read on the model being used it will tag the count as estiamted
  - Set three values into the context: `ctx.LLMReport_Model`, `ctx.LLMReport_NumTokens`, and `ctx.LLMReport_Estimate`.
 - These context values can later be used in the `RecordHit` function to store the data in an an alytics record (as this is part of Tyk Pump I did not extend the pump record)

## Motivation and Context

We've had multiple requests for this kind of reporting and it is relatively easy to add to our analytics, it is also something our competitors already do and will allow us to have a matching feature.

## How This Has Been Tested
- Manual testing directly with dummy requests from OpenAI and Anthropic docs

## Types of changes

## Checklist

### DIFF
--- a/apps/quickstart.json
+++ b/apps/quickstart.json
@@ -6,6 +6,7 @@
         "location": "",
         "key": ""
     },
+    "tags": ["llm"],
     "use_keyless": true,
     "auth": {
         "auth_header_name": ""
@@ -30,5 +31,5 @@
         "target_url": "http://httpbin.org/",
         "strip_listen_path": true
     },
-    "do_not_track": true
+    "do_not_track": false
 }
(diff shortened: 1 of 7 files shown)


buger:
@lonelycode all LLMs this days return amount of spent tokens for request and response, so maybe instead of calculating it before request, calculate it after response? And after update the rates?

PR #6388, opened 2024-07-04
[TT-12561] chore: move some root packages to internal

titpetric:
### **User description**

This moves several packages away from root scope to internal:

https://tyktech.atlassian.net/browse/TT-12561

- rpc
- tcp
- goplugin
- trace
- request (to httputil, single function)
- dnscache
- [...]

These are all in gateway internal scope because:

1) plugins don't reasonably depend on them which means the change is likely non-breaking,
2) should be covered by coding standards under internal scope (and test constraints).

May block/trigger:

If any breakage occurs due to unforseen couplings (e.g. dashboard using any of these gateway packages...) then we need to investigate the coupling and decouple. Gateway is big, changes often, and isn't meant to expose internal API details by default, as breaking changes are considered heavily, the sooner we can move this out of scope, the sooner we can discover and prevent further coupling on these.

Some package like `regexp` remain in gateway scope, as they have known use in plugins. This is documented in 'Code structure guidelines' (confluence).


sarobakhtiyari (approved):

titpetric:
Dashboard uses:

      1 /TykTechnologies/tyk/apidef/adapter"
      1 /TykTechnologies/tyk/apidef/importer"
      1 /TykTechnologies/tyk/config"
      1 /TykTechnologies/tyk/log"
      1 /TykTechnologies/tyk/storage"
      4 /TykTechnologies/tyk/header"
      5 /TykTechnologies/tyk/user"
      8 /TykTechnologies/tyk/certs"
     11 /TykTechnologies/tyk/gateway"
     14 /TykTechnologies/tyk/apidef/oas"
     21 /TykTechnologies/tyk/apidef"
     36 /TykTechnologies/tyk/test"

buger (approved):
Overall looks safe! Does not feel like those packages used as public interfaces.

sredxny (approved):
Looks really good! I do not see any breaking change either

jeffy-mathew (approved):

PR #6393, opened 2024-07-08
Fix: correctly sync multi-value response headers for coprocess-based middlewares

sebkehr:
## Description

When synchronizing single- and multi-valued header representations (of coprocess-based middleware responses) the list of values for any multi-valued header is currently replaced by a list containing only the value given by its single-value representation effectively dropping all but the first value. Instead synchronization should affect/replace only the first value and retain possibly remaining values.

## Related Issue

I could not find any related issue.

## Motivation and Context

We like to employ Tyk Gateway with a coprocess-based response middleware attached to an upstream responding with multiple _Set-Cookie_ headers. We also require our middleware to modify other headers like _Location_. As is due to header synchronization only the first _Set-Cookie_ header passes our middleware.

## How This Has Been Tested

The PR contains a test.

## Types of changes

## Checklist


sebkehr:
will reissue shortly

PR #6251, opened 2024-04-25
[DX-1084] Add documentation to protobuf source code files for review

dcs3spp:
[DX-1084](https://tyktech.atlassian.net/browse/DX-1084)

## Description

- Added documentation for protobuf messages and service for review
- Task file has been updated to use a Python virtual environment.
- The PR includes the bindings regenerated
- The PR is large and has 22 files that includes the comments in the regenerated bindings. If required I can include just the protobuf files and then once reviewed include a separate PR for the bindings?

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Refactoring or add test (improvements in base code or adds test coverage to functionality)

## Checklist

- [x] I ensured that the documentation is up to date
- [ ] I explained why this PR updates go.mod in detail with reasoning why it's required
- [ ] I would like a code coverage CI quality gate exception and have explained why

[DX-1084]: https://tyktech.atlassian.net/browse/DX-1084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6399, opened 2024-07-09
Synchronize master state to 5.3 CI

titpetric:
This PR copies the following locations from master to LTS, updating CI state:

```
.gitignore
.github/workflows/ci-tests.yml
.taskfiles/
ci/tests/tracing/
coprocess/
config/Taskfile.yml
apidef/oas/Taskfile.yml
docker/
Dockerfile
docker-compose.yml
Taskfile.yml
```

coprocess/ is a documentation change, an update of the bindings was made:
https://github.com/TykTechnologies/tyk/pull/6251

Expecting all tests to pass


.gitignore
```diff
@@ -53,4 +54,5 @@ tyk_linux_*
 /dist/
 
 .terraform**
-.task/
\ No newline at end of file
+.task/
+*.test
```
jeffy-mathew, line 58: can we add `coverage/` to gitignore

buger:
My main concern here is bindings, won't it be a breaking change?

jeffy-mathew (approved):

titpetric:
> My main concern here is bindings, won't it be a breaking change?

It was a documentation change. There we're no API changes on the DX ticket reported so we should be safe.

titpetric:
@buger verified again in coprocess/ against release 5.2; to repeat verification:

```
cd coprocess
schema-gen extract
schema-gen list | sort   # This just lists all fields as <Struct>.Field <fieldType>, enough to verify a diff
```

no api changes.

PR #6448, opened 2024-08-02
[TT-12452] Try to clear up quota with a distributed redis lock

titpetric:
- adds a -time field into redis next to quotas, to handle the real expiry
- adds distributed locking to fix races in a cluster of gateways
- cleanups for readability/understanding of the codebase

### DIFF
--- a/gateway/session_manager.go
+++ b/gateway/session_manager.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/TykTechnologies/tyk/config"
 	"github.com/TykTechnologies/tyk/internal/rate"
+	"github.com/TykTechnologies/tyk/internal/rate/limiter"
 	"github.com/TykTechnologies/tyk/internal/redis"
 	"github.com/TykTechnologies/tyk/storage"
 	"github.com/TykTechnologies/tyk/user"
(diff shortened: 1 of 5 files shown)


gateway/session_manager.go
```diff
@@ -334,68 +342,104 @@ func (l *SessionLimiter) RedisQuotaExceeded(r *http.Request, session *user.Sessi
...
-	if qInt == 1 {
-		quotaRenews = time.Now().Unix() + quotaRenewalRate
+	// set up decision fields, time, current quota...
+	now := time.Now()
+	quotaExpired := now.After(expireAt)
+	canRenew := quotaRenewalRate > 0
+
+	count, err := redis.Incr(ctx, rawKey).Result()
+	if err != nil {
+		log.WithError(err).Error("can't update quota")
```
jeffy-mathew, line 376: do we need to return here?

gateway/session_manager.go
```diff
@@ -309,13 +310,20 @@ func (l *SessionLimiter) ForwardMessage(r *http.Request, session *user.SessionSt
 
 }
 
+// RedisQuotaExceeded returns true if the request should be blocked as over quota.
 func (l *SessionLimiter) RedisQuotaExceeded(r *http.Request, session *user.SessionState, quotaKey, scope string, limit *user.APILimit, store storage.Handler, hashKeys bool) bool {
-	// Unlimited?
+	var (
```
jeffy-mathew, line 315: shall we make these values constants?

gateway/session_manager.go
```diff
@@ -334,68 +342,104 @@ func (l *SessionLimiter) RedisQuotaExceeded(r *http.Request, session *user.Sessi
...
-			// The renewal date is in the past, we should update the quota!
-			// Also, this fixes legacy issues where there is no TTL on quota buckets
-			log.Debug("Incorrect key expiry setting detected, correcting")
-			go store.DeleteRawKey(rawKey)
-			qInt = 1
-		} else {
-			// Renewal date is in the future and the quota is exceeded
-			return true
+	// First, ensure a distributed lock
+	redis := l.limiterStorage
```
jeffy-mathew, line 349: `redis` variable name collides with imported package name

gateway/session_manager.go
titpetric: no, local semantics only +0.01% on readability, may as well be unlimited

gateway/session_manager.go
titpetric: linter wasn't configured to report it, updated it

jeffy-mathew (approved):

PR #6519, opened 2024-09-16
[TT-13036] A struct is needed to resemble the info section inside of …

buraksezer:
PR for https://tyktech.atlassian.net/browse/TT-13036

Adding a version for streams config
It is expected that the configuration will evolve heavily in the feature when gathering feedback from users. Therefore a version will be important to allow breaking changes if needed while keeping old configurations possible.

For MVP the version can live inside an info block in the streams config section.

```json
{
  "x-tyk-streaming": {
    "info": {
      "version": "1.0"
    },
    "streams": {
    }
  },
}
```

### DIFF
--- a/gateway/mw_streaming.go
+++ b/gateway/mw_streaming.go
@@ -22,6 +22,13 @@ const (
 // Used for testing
 var globalStreamCounter atomic.Int64
 
+type StreamsConfig struct {
+	Info struct {
+		Version string `json:"version"`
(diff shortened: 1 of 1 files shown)


buger:
I believe for CI/CD use case, this field can be optional, right?

pvormste (approved):

buraksezer:
This PR should be reworked due to conflicts.

PR #6868, opened 2025-02-12
[TT-13836] Key Rotation for MDCB Data Planes

mativm02:
## Description
We're now handling events emitted by MDCB where a user API Key has been reset. When we receive this event, we update the APIKey value in memory and reconnect to MDCB using that new API Key. If the user wants to use Vault or Consul to store their API Key, now it's possible. After this event is processed, the new API Key will be stored in the external KV system.
If they're using env vars or a .conf file, users will still need to update it in case of a restart manually.

## Related Issue
https://tyktech.atlassian.net/jira/software/c/projects/TT/boards/42?selectedIssue=TT-13936

## Motivation and Context
https://tyktech.atlassian.net/jira/software/c/projects/TT/boards/42?selectedIssue=TT-13936

## Types of changes

## Checklist

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -396,6 +396,10 @@ type SlaveOptionsConfig struct {
 
 	// SynchroniserEnabled enable this config if MDCB has enabled the synchoniser. If disabled then it will ignore signals to synchonise recources
 	SynchroniserEnabled bool `json:"synchroniser_enabled"`
+
+	// OriginalAPIKeyPath is the original path to the API key in the configuration file. Only used to modify the external KV store in case of API Key Reset.
+	// This is set automatically in afterConfSetup()
+	OriginalAPIKeyPath string `json:"-"`
 }
 
 type LocalSessionCacheConf struct {
--- a/gateway/redis_signals.go
+++ b/gateway/redis_signals.go
@@ -41,6 +41,7 @@ const (
 	OAuthPurgeLapsedTokens       NotificationCommand = "OAuthPurgeLapsedTokens"
 	// NoticeDeleteAPICache is the command with which event is emitted from dashboard to invalidate cache for an API.
 	NoticeDeleteAPICache NotificationCommand = "DeleteAPICache"
+	NoticeUserKeyReset   NotificationCommand = "UserKeyReset"
 )
 
 // Notification is a type that encodes a message published to a pub sub channel (shared between implementations)
--- a/gateway/redis_signals.go
+++ b/gateway/redis_signals.go
@@ -158,6 +158,8 @@ func (gw *Gateway) handleRedisEvent(v interface{}, handled func(NotificationComm
 		if ok := gw.invalidateAPICache(notif.Payload); !ok {
 			log.WithError(err).Errorf("cache invalidation failed for: %s", notif.Payload)
 		}
+	case NoticeUserKeyReset:
+		gw.handleUserKeyReset(notif.Payload)
 	default:
 		pubSubLog.Warnf("Unknown notification command: %q", notif.Command)
 		return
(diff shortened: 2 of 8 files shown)


buger:

gateway/redis_signals.go
```diff
@@ -322,3 +325,55 @@ func (gw *Gateway) handleDashboardZeroConfMessage(payload string) {
...
+	newKey := keys[1]
+
+	config := gw.GetConfig()
+
+	if oldKey == config.SlaveOptions.APIKey {
+		config.SlaveOptions.APIKey = newKey
+		gw.SetConfig(config)
+
+		// If we're using a KV store, update the API key there as well
+		if config.SlaveOptions.OriginalAPIKeyPath != "" {
```
sredxny, line 353: here you could assign `config.SlaveOptions.OriginalAPIKeyPath` to a var so it will look clearner below:

```
if keyPath!= ""{
    if strings.HasPrefix("vault://", keyPath)
}
.
.
.
```

gateway/redis_signals.go
```diff
@@ -322,3 +325,55 @@ func (gw *Gateway) handleDashboardZeroConfMessage(payload string) {
...
+				actualPath := strings.TrimPrefix(config.SlaveOptions.OriginalAPIKeyPath, "consul://")
+				if gw.consulKVStore != nil {
+					if err := gw.consulKVStore.Put(actualPath, newKey); err != nil {
+						log.WithError(err).Error("Failed to update API key in Consul")
+					} else {
+						log.Info("Successfully updated API key in Consul")
+					}
+				}
+			}
+		}
```
sredxny, line 373: lets reduce here the if{}else{} nesting, am pretty sure we can avoid some of those conditionals or move logic to a different method

gateway/rpc_storage_handler.go
```diff
@@ -1057,6 +1058,21 @@ func (r *RPCStorageHandler) ProcessKeySpaceChanges(keys []string, orgId string)
...
+				}
+				userKeysToReset[userKeys[0]] = userKeys[1]
+				ok := r.Gw.MainNotifier.Notify(Notification{
+					Command: NoticeUserKeyReset,
+					Payload: key,
+					Gw:      r.Gw,
+				})
+				if !ok {
+					log.Error("Failed to notify other gateways about user key reset")
+				}
```
sredxny, line 1075: lets move this logic to a different loop, check the conditionals above, here you might want to do something like:

`
case NoticeUserKeyReset.String():
    userKeysToReset = append(userKeysToReset, splitKeys[0])
    notRegularKeys[key] = true --> with this you just skip this key later when looping over keys
`

Then you can loop over `userKeysToReset` and execute all the other logic that is parsing data and notifying gateways. As in the fuure we need to split this whole function into something more understandable and short function

gateway/server.go
```diff
@@ -1519,14 +1519,22 @@ func (gw *Gateway) afterConfSetup() {
 	if conf.UseDBAppConfigs {
 		conf.DBAppConfOptions.ConnectionString, err = gw.kvStore(conf.DBAppConfOptions.ConnectionString)
 		if err != nil {
-			log.Fatalf("Could not fetch dashboard connection string.. %v", err)
+			log.Fatalf("Could not fetch dashboard connection string.")
```
sredxny, line 1522: why did you remove the err?

config/config.go
```diff
@@ -395,6 +395,10 @@ type SlaveOptionsConfig struct {
 
 	// SynchroniserEnabled enable this config if MDCB has enabled the synchoniser. If disabled then it will ignore signals to synchonise recources
 	SynchroniserEnabled bool `json:"synchroniser_enabled"`
+
+	// OriginalAPIKeyPath is the original path to the API key in the configuration file. Only used to modify the external KV store in case of API Key Reset.
+	// This is set automatically in afterConfSetup()
+	OriginalAPIKeyPath string `json:"-"`
```
sredxny, line 401: why do we want to have `OriginalAPIKeyPath`? this is an historic value and should keep using the new one no matter how many times we rotate the key.
Also, wouldnt this be problematic when generting the docs for config options?

sredxny (changes requested):

config/private.go
```diff
@@ -10,6 +10,9 @@ type Private struct {
 	// none was found, it's the path to the default config file that
 	// was written.
 	OriginalPath string `json:"-"`
+	// OriginalAPIKeyPath is the original path to the API key in the configuration file. Only used to modify the external KV store in case of API Key Reset.
+	// This is set automatically in afterConfSetup()
+	OriginalAPIKeyPath string `json:"-"`
```
sredxny, line 15: lets be clear that this is only used when the gateway is running in slave mode. We can do here 2 things:
- State in the documentation that this is for mdcb envs when the gw is edge
- Or change the name of the var to be explicit

sredxny (approved):

PR #6870, opened 2025-02-13
[TT-13659] add support for custom event handlers in OAS definitions

pvormste:
This PR adds support for custom event handlers in OAS definitions.

## Types of changes


buger:
<p>
    Let's make that PR title a 💯 shall we? 💪
    </p>
    <p>
    Your <em>PR title</em> and <em>story title</em> look <strong>slightly different</strong>. Just checking in to know if it was intentional!
    </p>
    <table>
      <tr>
        <th>Story Title</th>
        <td>[OAS] Event handling: custom handler (MVP)</td>
      </tr>
      <tr>
          <th>PR Title</th>
          <td>[TT-13659] add support for custom event handlers in OAS definitions</td>
        </tr>
    </table>
    <p>
      Check out this <a href="https://www.atlassian.com/blog/git/written-unwritten-guide-pull-requests">guide</a> to learn more about PR best-practices.
    </p>

apidef/oas/event.go
```diff
@@ -116,10 +140,28 @@ func (e *EventHandler) GetWebhookConf() apidef.WebHookHandlerConf {
 	}
 }
 
+// JSVMEvent represents a JavaScript VM event configuration for event handlers.
+type JSVMEvent struct {
```
titpetric, line 144: Can you sanitize the order a bit please, type definitions should be in front/before functions, close to the top of the file next to other (grouping). Move this block there

apidef/api_definitions.go
```diff
@@ -1653,3 +1653,26 @@ func (w *WebHookHandlerConf) Scan(in any) error {
...
+
+// JSVMEventHandlerConf represents the configuration for a JavaScript VM event handler in the API definition.
+type JSVMEventHandlerConf struct {
+	// Disabled indicates whether the event handler is inactive.
+	Disabled bool `bson:"disabled" json:"disabled"`
+	// ID is the optional unique identifier for the event handler.
+	ID string `bson:"id" json:"id"`
+	// MethodName specifies the JavaScript function name to be executed.
+	MethodName string `bson:"name" json:"name"`
+	// Path refers to the file path of the JavaScript source code for the handler.
```
titpetric, line 1665: Is this enough? Source code is an usual input without the file indirection. 
Tests indicate a failing `body` assertion somewhere

apidef/oas/event_test.go
```diff
@@ -255,86 +361,167 @@ func TestEventHandlers(t *testing.T) {
...
+			},
+			expected: map[string]any{
+				"id":           "random-id",
+				"enabled":      true,
+				"trigger":      "QuotaExceeded",
+				"type":         "custom",
+				"name":         "test-custom",
+				"functionName": "myCustomEventHandler",
+				"path":         "event_handlers/session_editor.js",
+				"body":         "console.log('hello world');",
```
titpetric, line 426: May need body next to path to be usable (mentioned on data model).

apidef/api_definitions.go
```diff
@@ -1653,3 +1653,26 @@ func (w *WebHookHandlerConf) Scan(in any) error {
...
+	Path string `bson:"path" json:"path"`
+}
+
+// Scan populates the JSVMEventHandlerConf struct by casting and copying data from the provided input.
+func (j *JSVMEventHandlerConf) Scan(in any) error {
+	conf, err := reflect.Cast[JSVMEventHandlerConf](in)
```
titpetric, line 1671: One problem with reflect.Cast usage is that we don't map the types 1-1; In this case, the issue is mapping `enabled` to `disabled` and need to add handling. On the other hand, the added api definition type could change the Disabled field to Enabled. Other comments raised up on parity of the fields btw classic/oas.

cc @jeffy-mathew

apidef/oas/event.go
pvormste: I would normally agree, but in this case the context is relevant, so it makes sense to break it up. It is also consistent to exisiting implementation. It's sometimes helpful from developer perspective to keep the context grouped instead of the kind of the code

apidef/oas/event_test.go
pvormste: we dropped the body part for this ticket

apidef/api_definitions.go
pvormste: I'm not sure if I understand correctly what you mean here

titpetric (approved):

PR #7063, opened 2025-05-20
adds client ip from XFF by depth

sedkis:
## Types of changes

## Checklist

### DIFF
--- a/request/real_ip.go
+++ b/request/real_ip.go
@@ -5,8 +5,11 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/TykTechnologies/tyk/config"
 	"github.com/TykTechnologies/tyk/header"
 )
+
+var Global func() config.Config
 
 // RealIP takes a request object, and returns the real Client IP address.
 func RealIP(r *http.Request) string {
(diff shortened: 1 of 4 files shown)


cli/linter/schema.json
```diff
@@ -1336,9 +1339,6 @@
     "secrets": {
       "type": ["array", "null"]
     },
-    "enable_http_profiler": {
-      "type": "boolean"
-    },
```
mativm02, line 1341: why are we removing this?

request/real_ip.go
```diff
@@ -22,14 +25,38 @@ func RealIP(r *http.Request) string {
...
+
+		// If depth exceeds available IPs, return empty
+		if depth > len(xffs) {
+			return ""
+		}
+
+		// Choose the appropriate IP based on depth
+		// depth=0 means first IP (leftmost), depth=1 means last IP, depth=2 means second to last, etc.
+		var ip string
+		if depth == 0 {
```
mativm02, line 48: I think this should be `if depth <= 0` to avoid `index out of bounds` panics in the else case

request/real_ip.go
```diff
@@ -22,14 +25,38 @@ func RealIP(r *http.Request) string {
 	}
 
 	if fw := r.Header.Get(header.XForwardFor); fw != "" {
-		// X-Forwarded-For has no port
-		if i := strings.IndexByte(fw, ','); i >= 0 {
-			fw = fw[:i]
+		xffs := strings.Split(fw, ",")
+
+		// If no IPs, return the first IP in the chain
```
mativm02: We are not returning the first IP in the chain, but an empty string

cli/linter/schema.json
sedkis: whoops. will revert.

sedkis: Duplicate key. already defined on line 746 in the same file

request/real_ip.go
sedkis: Thanks Mati - I made this change and wrote a new test for testing negative depth.  I hadn't wanted to include this as it would be a failure on user error but that's fine I guess?  I am wondering if we need to log that this is an invalid config on startup? do we do that?

request/real_ip.go
sedkis: yes! good catch mati.  have fixed + added the test

PR #7117, opened 2025-06-13
[TT-9234] graceful shutdown of gateway improvments and bug fix for mdcb scenario

andrei-tyk:
## Types of changes

## Checklist


buger:

buger:
The use-case which needs be tested, and can be even unit tested in this pr:

Imagine we installing it in fresk k8s cluster. We have 0 APIs and 0 policies, and we use /ready probe.
Gateway will never be included into the load balancer, because there are 0 APIs, Dashboard will not be able connect to it as well. So the actual check is not if we have 0 apis or not, but more like a boolean flag, if api reload happened succesfully

buger:
Also note that we already have dashboard mock here in tests, so we can emulate various error scenarios. Including the one when instead of dashboard we have some load balancer returning 404, or if it returns code 500, or some network timeout. 

It should test the state of initial load, and then validate what will happen if we drop connectivity, and then resumed again.

gateway/server.go
```diff
@@ -1841,46 +1841,12 @@ func Start() {
...
-	}
-
-	// write pprof profiles
-	writeProfiles()
-
-	if gwConfig.UseDBAppConfigs {
-		mainLog.Info("Stopping heartbeat...")
-		gw.DashService.StopBeating()
-		time.Sleep(2 * time.Second)
-		err := gw.DashService.DeRegister()
```
buger, line 1860: @andrei-tyk You no longer call dashboard deregister method anywhere in new code, is that expected?

andrei-tyk: The code block was indeed missed when moving the block inside the new gracefulShutdown function

andrei-tyk: Added via a new commit.

PR #7306, opened 2025-08-13
[TT-2378] Implementing OAS OR auth logic

mativm02:
## Types of changes

## Checklist

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -772,6 +772,10 @@ type APIDefinition struct {
 
 	// UpstreamAuth stores information about authenticating against upstream.
 	UpstreamAuth UpstreamAuth `bson:"upstream_auth" json:"upstream_auth"`
+
+	// SecurityRequirements stores all OAS security requirements (auto-populated from OAS import)
+	// When len(SecurityRequirements) > 1, OR logic is automatically applied
+	SecurityRequirements [][]string `json:"security_requirements,omitempty" bson:"security_requirements,omitempty"`
 }
 
 type JWK struct {
--- a/gateway/mw_auth_or_wrapper.go
+++ b/gateway/mw_auth_or_wrapper.go
@@ -14,9 +14,14 @@ type AuthORWrapper struct {
 
 // ProcessRequest handles the OR logic for authentication
 func (a *AuthORWrapper) ProcessRequest(w http.ResponseWriter, r *http.Request, _ interface{}) (error, int) {
+	logger := a.Logger()
+	logger.Debugf("OR wrapper processing with %d middlewares, %d security requirements", 
+		len(a.authMiddlewares), len(a.Spec.SecurityRequirements))
+	
 	// Auto-detect OR logic from security requirements
 	if len(a.Spec.SecurityRequirements) <= 1 {
 		// Single requirement or empty = AND logic (default behavior)
+		logger.Debug("Using AND logic (single or no security requirement)")
 		for _, mw := range a.authMiddlewares {
 			if err, code := mw.ProcessRequest(w, r, nil); err != nil {
 				return err, code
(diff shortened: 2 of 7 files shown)


buger:
<p>This PR is too huge for one to review :broken_heart: </p>
  
    <table>
      <tr>
          <th>Additions</th>
          <td>1098 :no_good_woman: </td>
      </tr>
      <tr>
          <th>Expected</th>
          <td>:arrow_down: 800</td>
        </tr>
    </table>
    <p>
    Consider breaking it down into multiple small PRs.
    </p>
    <p>
      Check out this <a href="https://www.atlassian.com/blog/git/written-unwritten-guide-pull-requests">guide</a> to learn more about PR best-practices.
    </p>

buger:
<p>
    Let's make that PR title a 💯 shall we? 💪
    </p>
    <p>
    Your <em>PR title</em> and <em>story title</em> look <strong>slightly different</strong>. Just checking in to know if it was intentional!
    </p>
    <table>
      <tr>
        <th>Story Title</th>
        <td>Allow 'OR' logic for multiple authentication modes for an API definition</td>
      </tr>
      <tr>
          <th>PR Title</th>
          <td>[TT-2378] Implementing OAS OR auth logic</td>
        </tr>
    </table>
    <p>
      Check out this <a href="https://www.atlassian.com/blog/git/written-unwritten-guide-pull-requests">guide</a> to learn more about PR best-practices.
    </p>
    

[TT-2378]: https://tyktech.atlassian.net/browse/TT-2378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

apidef/oas/default_test.go
```diff
@@ -1715,10 +1727,20 @@ func TestOAS_importAuthentication(t *testing.T) {
...
+							Name:    "Authorization",
+						},
+					},
+				},
 			}
 
 			assert.Equal(t, expectedSecuritySchemes, authentication.SecuritySchemes)
-			assert.Equal(t, apidef.AuthTypeNone, authentication.BaseIdentityProvider)
+			// Base identity provider should be determined from the security schemes
+			assert.NotNil(t, authentication.BaseIdentityProvider)
```
mativm02, line 1733: Could this be considered a breaking change? Should we add a flag to avoid it?

andrei-tyk: Given it's an OR I think it might be safe to assume that the only working one should be present. If my logic is correct, currently we just use a wildcard one which only is used in logs.###

gateway/mw_auth_or_wrapper.go
```diff
@@ -0,0 +1,121 @@
...
+			return nil, http.StatusOK
+		}
+
+		logger.Debugf("Auth middleware %d failed: %v (code: %d)", i, err, code)
+
+		lastError = err
+		lastCode = code
+	}
+
+	return lastError, lastCode
```
andrei-tyk, line 63: I think here we should just return a generic failing auth error and just log on debug all the tried methods.

gateway/mw_auth_or_wrapper.go
```diff
@@ -0,0 +1,121 @@
...
+	var lastCode int
+
+	for i, mw := range a.authMiddlewares {
+		logger.Debugf("Trying auth middleware %d: %T", i, mw)
+		err, code := mw.ProcessRequest(w, r, nil)
+		if err == nil {
+			logger.Debugf("Auth middleware %d succeeded", i)
+
+			if session := ctxGetSession(r); session != nil {
+				ctxSetSession(r, session, false, a.Gw.GetConfig().HashKeys)
```
andrei-tyk, line 48: should we clear the previous session before setting this one? I'm thinking that maybe any of the middlewares might modify it even if it does not succeed

gateway/mw_auth_or_wrapper.go
mativm02: @andrei-tyk I didn't want to modify the error response so we keep backward compatibility. Therefore, we are returning the same error response that the authentication method returns when there's no "or" condition.

gateway/mw_auth_or_wrapper.go
mativm02: It makes sense. It doesn't seem that any middleware is changing the session on failure, but I added a check just in case

apidef/oas/default_test.go
```diff
@@ -1662,7 +1672,7 @@ func TestOAS_importAuthentication(t *testing.T) {
 		assert.Nil(t, authentication)
 	})
 
-	t.Run("add first authentication in case of OR condition", func(t *testing.T) {
+	t.Run("import all authentication methods for OR condition", func(t *testing.T) {
```
pvormste, line 1675: seems that the test case was changed - I would rather add a complete new test and keep the old one instead, so that we can make sure that we keep the confidence that the backwards-compatible behvaior is not broken

pvormste (approved):

andrei-tyk (approved):

apidef/api_definitions.go
```diff
@@ -772,6 +772,10 @@ type APIDefinition struct {
 
 	// UpstreamAuth stores information about authenticating against upstream.
 	UpstreamAuth UpstreamAuth `bson:"upstream_auth" json:"upstream_auth"`
+
+	// SecurityRequirements stores all OAS security requirements (auto-populated from OAS import)
```
andyo-tyk:
```suggestion
	// SecurityRequirements stores all OAS security requirements (auto-populated from OpenAPI description import)
```

PR #7339, opened 2025-09-03
[TT-15380] Enhance JWKS caching

buraksezer:
PR for https://tyktech.atlassian.net/browse/TT-15380


buger:

apidef/oas/schema/x-tyk-api-gateway.json
```diff
@@ -1618,6 +1618,26 @@
...
+            "null"
+          ],
+          "items": {
+            "type": "object",
+            "properties": {
+              "url": {
+                "type": "string",
+                "format": "uri"
+              },
+              "cache_timeout": {
```
edsonmichaque: shouldn't `cache_timeout` property be `cacheTimeout` for OAS?

edsonmichaque (approved):
LGTM

apidef/oas/schema/x-tyk-api-gateway.json
buraksezer: @edsonmichaque `JwksURIs` in apidef/oas/security.go file has defined like this: 
```go
	// JwksURIs contains a list of JSON Web Key Sets (JWKS) endpoints from which Tyk will retrieve JWKS to validate JSON Web Tokens (JWTs).
	JwksURIs []apidef.JWK `bson:"jwksURIs,omitempty" json:"jwksURIs,omitempty"`
```

So it directly gets `JWK` struct from `apidef` package. Here is the JWK struct:

```go
type JWK struct {
	// URL is the jwk endpoint
	URL string `json:"url"`
	// CacheTimeout defines how long the JWKS is kept in the cache before forcing a refresh.
	CacheTimeout int64 `bson:"cache_timeout" json:"cache_timeout"`
}
```

With this implementation, json tag should be `cache_timeout`. If we want to follow the naming rule, we can add `cacheTimeout` tag to the existing tag but it means we leak the OAS business logic to the `apidef` package. Another option is copying the struct to OAS package and using the proper tags. What do you think?

pvormste: I would suggest having a proper struct for OAS is the way to go, because so far the naming in OAS is very consistent overall and the oas package would have ownership over this struct.

lghiur:
/probe is there a way we can add integration tests in our python test suite, by mocking the auth server?

lghiur:
/probe create integration tests scenarios to write in our python framework

PR #7427, opened 2025-10-09
(TT-16280) Fix #7426: Support underscores in Vault and Consul secret reference paths

### LINKED ISSUE #7426: Support underscores in Vault secret reference paths
We are using Tyk's Vault integration to dynamically retrieve secrets for upstream authentication. Our Vault secret keys follow a naming convention that uses underscores to separate parts (e.g., `API_KEY`, `DB_PASSWORD`). However, Tyk's current regex pattern for Vault references does not support underscores in the secret path, causing the reference to be truncated at the first underscore character.

For example, when we configure:
```yaml
$secret_vault.kv-v2/layer1/layer2/{{ .Values.environment }}.API_KEY
```

Tyk only processes:
```yaml
$secret_vault.kv-v2/layer1/layer2/{{ .Values.environment }}.API
```

This forces us to duplicate all secrets with underscore-free names in Vault, which is:
- Operationally inefficient and error-prone
- Creates unnecessary maintenance overhead
- Not feasible to change original keys as they're used by many other systems

The root cause appears to be in the regex pattern definition (from `mw_url_rewrite.go`):
```go
var vaultMatch = regexp.MustCompile(`\$secret_vault.([A-Za-z0-9\/\-\.]+)`)
```

The pattern lacks the underscore character (`_`), while other secret handlers like `envValueMatch` correctly include it:
```go
var envValueMatch = regexp.MustCompile(`\$secret_env.([A-Za-z0-9_\-\.]+)`)
```

Update the `vaultMatch` regex pattern to include the underscore character, making it consistent with other secret reference patterns:

```go
var vaultMatch = regexp.MustCompile(`\$secret_vault.([A-Za-z0-9_\/\-\.]+)`)
```

This would allow Vault references to properly handle secret keys containing underscores, aligning with common naming conventions used in production environments.

Similarly, the `consulMatch` pattern appears to have the same limitation and should be updated:
```go
var consulMatch = regexp.MustCompile(`\$secret_consul.([A-Za-z0-9_\/\-\.]+)`)
```

1. **Current workaround (unacceptable)**: Duplicating all secrets in Vault with underscore-free names. This creates maintenance burden and potential security risks due to secret sprawl.
2. **Renaming existing secrets**: Not feasible as multiple systems depend on the current naming convention.
3. **Custom middleware**: Would require maintaining custom code and lose official support benefits. we have implemented a plugin for this and it works but it would be much better to support this by default by Tyk.

**Additional context**
- This affects the Vault integration across all Tyk versions/branches (verified in master branch)
- The inconsistency between `envValueMatch` (which supports underscores) and `vaultMatch` (which doesn't) suggests this is an oversight rather than intentional design (maybe other secret engines prohibit this character but surely not Vault)
- Standard secret naming conventions in a lot of organizations use underscores (e.g., `DATABASE_URL`, `API_KEY`, `AWS_SECRET_KEY`)
- This is a breaking limitation for enterprises migrating to Tyk with existing Vault infrastructure

**Code references:**
- File: `mw_url_rewrite.go`
- Regex definitions (lines with pattern declarations)
- Variable replacement logic where `vaultMatch.FindAllString()` is called

bodis:
Fixes #7426

## Description

Added underscore support to `vaultMatch` and `consulMatch` regex patterns to allow secret keys like `API_KEY` and `DB_PASSWORD`.

**Before:** `$secret_vault.([A-Za-z0-9\/\-\.]+)`  
**After:** `$secret_vault.([A-Za-z0-9_\/\-\.]+)`

## Motivation and Context

Vault references are currently truncated at underscores, forcing users to duplicate secrets with different names. This fixes the regex to match standard naming conventions used in production environments.

**Example:**
- Config: `$secret_vault.kv-v2/path.API_KEY`
- Current behavior: Only matches up to `.API` ❌
- Fixed behavior: Matches full path including `.API_KEY` ✅

This aligns vault/consul patterns with `envValueMatch` which already supports underscores.

## Breaking Change Note

**Potential edge case:** If anyone is currently using the underscore as an intentional path terminator (e.g., `$secret_vault.path.API_KEY` expecting only `.API` to be processed with `_KEY` as a literal suffix), this will now include the full path. However, this use case is highly unlikely compared to the standard use of underscores in secret names.

## How This Has Been Tested

- Built successfully with `go build`
- Verified regex patterns compile and match underscore paths correctly
- Tested backward compatibility with existing non-underscore references

## Types of changes
New feature (low probability small breaking change but adds functionality) - 
```

### DIFF
--- a/ci/tests/tracing/Taskfile.yml
+++ b/ci/tests/tracing/Taskfile.yml
@@ -37,6 +37,7 @@ tasks:
       - tracetest run test -f ./scenarios/tyk_test_500.yml -o pretty
       - tracetest run test -f ./scenarios/tyk_testauth_401.yml -o pretty
       - tracetest run test -f ./scenarios/tyk_jwt_200.yml -o pretty
+      - tracetest run test -f ./scenarios/tyk_multiauth_jwt_200.yml -o pretty
       - tracetest run test -f ./scenarios/tyk_tykprotocol_200.yml -o pretty
       - tracetest run test -f ./scenarios/tyk_tykprotocol-auth_401.yml -o pretty
       - tracetest run test -f ./scenarios/tyk_grpcapi_200.yml -o pretty
--- a/ci/tests/tracing/apps/test-multiauth-oas.json
+++ b/ci/tests/tracing/apps/test-multiauth-oas.json
@@ -1,0 +1,90 @@
+{
+    "openapi": "3.0.3",
+    "info": {
+        "title": "MultiAuthTestAPI",
+        "version": "1.0.0"
+    },
+    "paths": {
+        "/anything": {
+            "get": {
+                "operationId": "getAnything",
+                "responses": {
+                    "200": {
+                        "description": "Success"
+                    }
+                }
+            }
+        }
+    },
+    "components": {
+        "securitySchemes": {
+            "jwt": {
+                "type": "http",
+                "scheme": "bearer",
+                "bearerFormat": "JWT"
+            },
+            "authToken": {
+                "type": "apiKey",
+                "in": "header",
+                "name": "X-API-Key"
+            }
+        }
+    },
+    "security": [
+        {"jwt": []},
+        {"authToken": []}
+    ],
+    "x-tyk-api-gateway": {
+        "info": {
+            "id": "multiauth-test-api",
+            "name": "MultiAuthTestAPI",
+            "orgId": "default",
+            "state": {
+                "active": true
+            }
+        },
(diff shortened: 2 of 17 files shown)


andyo-tyk:
Hi @bodis,

Thank you for the contribution - I'm closing this PR because we have implemented a solution via https://github.com/TykTechnologies/tyk/pull/7794. This will be included in the next release: Tyk 5.13.0.

Thanks for supporting Tyk.

PR #7485, opened 2025-10-27
[TT-16032] Add DNS monitoring feature for worker gateway

mativm02:
## Description
- Introduced a new DNSMonitorConfig struct to manage DNS monitoring settings, including enabling/disabling the monitor and setting the check interval.
- Updated the configuration schema to include DNS monitoring options.
- Implemented the DNS monitor functionality, which checks for DNS changes and triggers reconnections as needed.
- Added unit tests to validate the DNS monitor's lifecycle, behavior under various conditions, and its interaction with the system.
- Ensured that the DNS monitor can be started and stopped gracefully, and that it respects emergency mode settings.

This feature enhances the system's ability to detect and respond to DNS changes dynamically, improving reliability in network operations.

## Related Issue
https://tyktech.atlassian.net/browse/TT-16032?atlOrigin=eyJpIjoiMmFhYjY3OTRjMmFlNGNiNDhmYTZkMTQ2ZTBlNmQzN2UiLCJwIjoiaiJ9

## Motivation and Context
https://tyktech.atlassian.net/browse/TT-16032?atlOrigin=eyJpIjoiMmFhYjY3OTRjMmFlNGNiNDhmYTZkMTQ2ZTBlNmQzN2UiLCJwIjoiaiJ9

## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-16032" title="TT-16032" target="_blank">TT-16032</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | Implement background DNS monitor that checks DNS every 30 seconds |

Generated at: 2025-10-31 12:03:53

</details>

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -47,15 +47,6 @@ jobs:
           go-version-file: go.mod
           cache-dependency-path: go.sum
 
-      - name: Cache
-        uses: actions/cache@v4
-        with:
-          path: |
-            ~/.cache/golangci-lint
-          key: 'golangci-lint-${{ runner.os }}-${{ hashFiles(''**/go.sum'') }}'
-          restore-keys: |
-            golangci-lint-${{ runner.os }}-
-
       - name: Setup CI Tooling
         uses: shrink/actions-docker-extract@v3
         with:
@@ -63,25 +54,26 @@ jobs:
           path: /usr/local/bin/.
           destination: /usr/local/bin
 
-      - run: |
+      - name: Run codegen and tidy
+        if: github.event_name == 'pull_request'
+        env:
+          BASE_BRANCH: ${{ github.base_ref || 'master' }}
+        run: |
           # Run linting without auto-fixing
           task tidy
           task --exit-code lint:check-git-state MESSAGE="task tidy made git state dirty, please run task lint locally and update PR"
 
-      - run: golangci-lint version && golangci-lint cache status
-
       - name: golangci-lint
-        if: ${{ github.event_name == 'pull_request' }}
-        run: |  
-          # Generate report for SonarCloud
-          golangci-lint run --out-format colored-line-number,checkstyle:golangci-lint-report.json \
-                            --new-from-rev=origin/${{ env.BRANCH_NAME }} -v ./...
-
-      - name: golangci-lint-on-push
-        if: ${{ github.event_name == 'push' }}
-        # Using the default config, reports stats to sonarcloud. No requirements on passing other than KPIs tracked in dev.
-        run: |
-          golangci-lint run --out-format checkstyle:golangci-lint-report.json --issues-exit-code=0 -v ./...
+        uses: golangci/golangci-lint-action@v8
+        timeout-minutes: 20
+        with:
+          version: v2.5.0
+          only-new-issues: ${{ github.event_name == 'pull_request' }}
+          # Output formats configured in .golangci.yml (v2 approach)
+          # Generates: text to stdout + checkstyle JSON for SonarQube
+          args: --issues-exit-code=0 -v ./...
+          skip-cache: false
+          skip-save-cache: false
 
       - uses: actions/upload-artifact@v4
         if: ${{ always() }}
--- a/.gitignore
+++ b/.gitignore
@@ -81,3 +81,11 @@ ci/tests/specs/gateway.collection.postman.json
 ci/tests/specs/.env
 ci/tests/specs/apps
 ci/tests/specs/policies
+
+# Task cache files
+.task-build-cache
+.task-tidy-cache
+.task-check-imports-cache
+.task-check-config-cache
+.task-check-x-tyk-gateway-cache
+.task-codegen-cache
--- a/.golangci.dev.yml
+++ b/.golangci.dev.yml
@@ -1,12 +1,7 @@
+version: "2"
+
 # Options for analysis running.
 run:
-  timeout: 20m
-  tests: true
-
-  # Be as complete as possible when reporting issues
-  max-issues-per-linter: 0
-  max-same-issues: 0
-
   # If set we pass it to "go list -mod={option}". From "go help modules":
   # If invoked with -mod=readonly, the go command is disallowed from the implicit
   # automatic updating of go.mod described above. Instead, it fails when any changes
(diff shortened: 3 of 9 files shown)


rpc/dns_resolver.go
```diff
@@ -71,9 +73,20 @@ func checkDNSAndReconnect(connectionString string, suppressRegister bool) bool {
...
+		// Check if reconnection is already in progress (from proactive monitor or another reactive check)
+		// Use the same atomic flag as the DNS monitor to prevent concurrent reconnections
+		if !reconnectionInProgress.CompareAndSwap(false, true) {
+			Log.Warning("Reactive DNS check: reconnection already in progress, skipping duplicate reconnection")
+			return false
+		}
+
+		// Perform reconnection synchronously (we're already in error handling path)
+		defer reconnectionInProgress.Store(false)
+
```
mativm02, line 101: The proactive monitor prevents DNS flapping by rate limiting reconnections. The reactive path handles actual connection failures and needs immediate recovery. Both paths share the `reconnectionInProgress` flag and `lastResolvedIPs` state, preventing concurrent reconnections. If the proactive monitor recently reconnected, the reactive check will see no DNS change and skip reconnection anyway. Additionally, the reactive path has its own throttling via the `DNSCheckedAfterError` flag. Applying rate limiting to error recovery would delay service restoration without meaningful benefit.

rpc/dns_monitor.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+}
+
+// checkDNS performs the DNS check and reconnects if needed
+func (m *DNSMonitor) checkDNS() {
+	// Don't check if we're in emergency mode
+	if values.GetEmergencyMode() {
+		Log.Debug("DNS monitor: skipping check (emergency mode active)")
+		return
+	}
```
sredxny: why dont check if we are in emergency? am wondering the situation in which we fell into emergency because dns has changed then we do not check?

rpc/dns_monitor.go
```diff
@@ -0,0 +1,203 @@
...
+
+// DNSMonitor handles background DNS monitoring for worker gateways
+type DNSMonitor struct {
+	enabled       bool
+	checkInterval time.Duration
+	connectionStr string
+	ctx           context.Context
+	cancel        context.CancelFunc
+	stopComplete  chan struct{}
+}
```
sredxny, line 23: what about moving all this code into its own pkg? in the end the functionality grew and is not highly attached to rpc, maybe a new pkg inside `/internal` and rpc just instantiates a new dnsMonitor in case is required and pass configs

rpc/dns_monitor.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	dnsMonitor = nil
+}
+
+// monitorLoop is the main loop that periodically checks DNS
+func (m *DNSMonitor) monitorLoop() {
+	defer close(m.stopComplete)
+
+	ticker := time.NewTicker(m.checkInterval)
```
sredxny: am thinking that maybe the optimal here is to use a different algorithm, something like exponential backoff with floor + max cap and reset inmediatly after a change is detected. In summary, for example:

T0    = 30s     (minimum safe interval)
Tmax  = 10 min  (never go above this even if nothing changes)
Backoff style:  simple doubling (30s → 1m → 2m → 4m → 8m → cap)
Reset to T0 every time a change is detected

However this is something that can be discussed among the team, to define a expected behavior for this. What am trying to avoid is to get blocked by the DNS server because many attempts

rpc/dns_monitor.go
```diff
@@ -0,0 +1,203 @@
...
+	// Validate configuration
+	if connectionString == "" {
+		Log.Warning("DNS monitor enabled but connection string is empty, skipping monitor start")
+		return
+	}
+
+	// Set default interval if not specified or invalid
+	if checkInterval <= 0 {
+		checkInterval = 30 // Default to 30 seconds
+	}
```
sredxny, line 80: not only check that is greater than zero, but also set a minimum accepted value, for exmple if someone sets 1 second is currently valid, but it will lead with issues, because it will check every second and therefore we will get blocked in the DNS server

rpc/dns_monitor.go
```diff
@@ -0,0 +1,254 @@
...
+	// Validate configuration
+	if connectionString == "" {
+		Log.Warning("DNS monitor enabled but connection string is empty, skipping monitor start")
+		return
+	}
+
+	// Validate and set interval with minimum threshold
+	if checkInterval <= 0 {
+		checkInterval = 30 // Default to 30 seconds
+	} else if checkInterval < minCheckInterval {
```
sredxny, line 69: am thinking that this could be reduced to:

```
if checkInterval < minCheckInterval {
    log.WithFields(logrus.Fields{
        "requested_interval": checkInterval,
        "minimum_interval":   minCheckInterval,
    }).Warn("DNS monitor: check interval too low, using minimum")
    checkInterval = minCheckInterval
}
```

And set as minimum 30seconds, I did a quick research and found that the RFC 1035 3.3.13, the SOA Retry timer specifies how long a secondary DNS server waits before retrying after a failed check. Operational defaults (e.g. BIND, AWS Route53) typically set this between 30 and 300 seconds, with 30 seconds being a common conservative baseline.

rpc/dns_monitor.go
```diff
@@ -0,0 +1,254 @@
...
+				// Double the interval (exponential backoff), but cap at max
+				newInterval := m.currentInterval * 2
+				if newInterval > m.maxCheckInterval {
+					newInterval = m.maxCheckInterval
+				}
+				if newInterval != m.currentInterval {
+					m.currentInterval = newInterval
+					Log.WithField("new_interval", m.currentInterval).Debug("DNS monitor: no change, increased check interval")
+				}
+			}
```
sredxny: you dont need to handle the pace by your own, you can just use the library that we are using for exponential backoff but providing extra params, something like this should work:

```
func (m *DNSMonitor) monitorLoop() {
	defer close(m.stopComplete)
	
	// Configure exponential backoff
	b := backoff.NewExponentialBackOff()
	b.InitialInterval = m.baseCheckInterval    // T0 (e.g., 30s)
	b.MaxInterval = m.maxCheckInterval         // Tmax (e.g., 10 min)
	b.MaxElapsedTime = 0                       // Never stop (no total timeout)
	b.Multiplier = 2.0                         // Simple doubling
	b.RandomizationFactor = 0                  // No jitter (set to 0.1 if you want some)
	b.Reset()
	
	// Get initial interval
	currentInterval := b.NextBackOff()
	ticker := time.NewTicker(currentInterval)
	defer ticker.Stop()
	
	Log.WithField("initial_interval", currentInterval).Debug("DNS monitor loop started with exponential backoff")
	
	for {
		select {
		case <-m.ctx.Done():
			Log.Debug("DNS monitor loop received shutdown signal")
			return
			
		case <-ticker.C:
			changed := m.checkDNS()
			
			if changed {
				// Reset to base interval when change detected
				b.Reset()
				currentInterval = b.NextBackOff()
				Log.WithField("new_interval", currentInterval).Info("DNS monitor: DNS changed, reset interval to base")
			} else {
				// Get next interval (exponential backoff with cap)
				currentInterval = b.NextBackOff()
				Log.WithField("new_interval", currentInterval).Debug("DNS monitor: no change, increased check interval")
			}
			
			// Reset ticker with new interval
			ticker.Reset(currentInterval)
		}
	}
}
```

sredxny (changes requested):

PR #7505, opened 2025-11-04
[TT-15606] fix certificate chain mTLS handshake

pvormste:
# Summary

Fixes mTLS authentication when certificate chains are used and improves certificate matching to support hosts with port numbers.

# Problem

## Issue 1 - Certificate Chains Broken:
When a certificate chain (containing both leaf certificate and CA certificate) was configured for mTLS, the system incorrectly added the entire chain, including the leaf certificate, to the CA pool. This caused TLS handshake failures because leaf certificates should be verified against the CA pool, not added to it.

## Issue 2 - Port Handling:
Certificate matching failed when the target host included a port number (e.g., api.example.com:8443). Configuratio entries like "api.example.com": "cert-id" would not match requests to api.example.com:8443.

# Solution

## 1. Certificate Chain Handling

Created a new helper function `AddCACertificatesFromChainToPool()` that intelligently extracts only CA certificates from chains:
  - For certificate chains (length > 1): Skips index 0 (leaf cert) and adds only certificates where IsCA=true
  - For single certificates (length = 1): Maintains backward compatibility by adding the certificate regardless of IsCA flag
  (supports certificate pinning)

Updated two critical locations to use this function:
  - `certs/manager.go:601` - Certificate pool management
  - `gateway/cert.go:466` - Client CA configuration for mTLS APIs

## 2. Port Handling

Refactored certificate matching logic into a testable function getCertificateIDForHost() that:
  - Uses Go's `net.SplitHostPort()` to properly handle hosts with ports
  - Implements priority-based matching (most specific match wins)
  - Supports IPv4, IPv6, and hostnames with or without ports

Matching Priority (low to high):
  1. `*` (wildcard)
  2. `*.example.com:8443` (wildcard with port)
  3. `*.example.com` (wildcard without port)
  4. `example.com` (exact without port)
  5. `example.com:8443` (exact with port - highest priority)

# Changes

  - `internal/crypto/helpers.go` - New AddCACertificatesFromChainToPool() function
  - `internal/crypto/helpers_test.go` - 8 unit tests for chain handling
  - `gateway/cert.go` - Refactored certificate matching with port support
  - `gateway/cert_test.go` - 20 new unit tests + end-to-end mTLS test with chains
  - `certs/manager.go` - Updated to use new chain handling function

## Backward Compatibility

  ✅ Fully backward compatible:
  - Existing certificate configurations (with or without ports) continue to work
  - Single certificate configurations unchanged
  - Certificate chains now work correctly (previously broken)
  - No configuration migration required

# Reviewer Focus Areas

## Critical:
  1. internal/crypto/helpers.go:156-212 - Core certificate chain extraction logic
  2. End-to-end test in gateway/cert_test.go:2203-2309 - Validates the fix works

## Important:
  3. gateway/cert.go:72-138 - Certificate matching with priority handling
  4. Port handling logic using net.SplitHostPort()

## Security Considerations

  - Follows TLS RFC 5246 (leaf certificate at index 0 in chains)
  - Only certificates with IsCA=true are added to CA pools
  - More specific certificate matches override general wildcards
  - Comprehensive negative testing ensures invalid clients are rejected

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-15606" title="TT-15606" target="_blank">TT-15606</a>
</summary>

|         |    |
|---------|----|
| Status  | In Dev |
| Summary | Tyk mTLS Advertises Leaf Subject DNs Instead of CA DNs |

Generated at: 2025-11-06 09:42:00

</details>

### DIFF
--- a/.github/workflows/release-bot.yaml
+++ b/.github/workflows/release-bot.yaml
@@ -1,0 +1,9 @@
+name: Release bot
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  release_bot:
+    uses: TykTechnologies/github-actions/.github/workflows/release-bot.yaml@main
--- a/ctx/ctx.go
+++ b/ctx/ctx.go
@@ -53,6 +53,8 @@ const (
 	CacheOptions
 	OASDefinition
 	SelfLooping
+	// RequestStartTime holds the time when the request entered the middleware chain
+	RequestStartTime
 )
 
 func ctxSetSession(r *http.Request, s *user.SessionState, scheduleUpdate bool, hashKey bool) {
--- a/gateway/api.go
+++ b/gateway/api.go
@@ -3097,6 +3097,19 @@ func ctxSetDoNotTrack(r *http.Request, b bool) {
 	setCtxValue(r, ctx.DoNotTrackThisEndpoint, b)
 }
 
+func ctxSetRequestStartTime(r *http.Request, t time.Time) {
+	setCtxValue(r, ctx.RequestStartTime, t)
+}
+
+func ctxGetRequestStartTime(r *http.Request) time.Time {
+	if v := r.Context().Value(ctx.RequestStartTime); v != nil {
+		if t, ok := v.(time.Time); ok {
+			return t
+		}
+	}
+	return time.Time{}
+}
+
 func ctxGetVersionInfo(r *http.Request) *apidef.VersionInfo {
 	if v := r.Context().Value(ctx.VersionData); v != nil {
 		return v.(*apidef.VersionInfo)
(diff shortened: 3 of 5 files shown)


gateway/cert_test.go
```diff
@@ -996,6 +996,297 @@ func TestUpstreamMutualTLS(t *testing.T) {
...
 }
 
+func TestUpstreamCertificateWithPort(t *testing.T) {
+	ts := StartTest(nil)
+	defer ts.Close()
+
+	ts.Gw.dialCtxFn = test.LocalDialer()
+
+	_, _, combinedClientPEM, clientCert := crypto.GenCertificate(&x509.Certificate{}, false)
+	clientCert.Leaf, _ = x509.ParseCertificate(clientCert.Certificate[0])
```
andrei-tyk: I think you can avoid these kind of comments from linter easily by adding err and assert.Nill(err)

pvormste: Thanks, I've adressed it

PR #7529, opened 2025-11-10
[TT-15942]: Integrate Sentinel One CNS scanner workflow

asutosh:
## Description

This adds a github action workflow that will scan the code base for vulnerabilities, secrets, misconfigurations etc. on every pull request. 
These can be manged based on centrally set up rules on the S1 console.

## Related Issue
TT-15942

## Motivation and Context
This is to integrate S1 with our repositories so that everything can be managed centrally in the S1 console.

## How This Has Been Tested

This is not very testable, as it requires the workflow to be added and then run to actually see how it behaves.
But some scans have been done independently locally for testing.

NA

## Types of changes

## Checklist

### DIFF
--- a/.github/workflows/s1-cns-scans.yml
+++ b/.github/workflows/s1-cns-scans.yml
@@ -1,9 +1,10 @@
+---
 name: SentinelOne CNS Scan
 
-on:
+on:  # yamllint disable-line rule:truthy
   pull_request:
-    types: [ opened, reopened, synchronize ]
-    branches: [ master ]
+    types: [opened, reopened, synchronize, ready_for_review]
+    branches: [master]
 
 jobs:
   s1_scanner:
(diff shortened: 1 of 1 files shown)


.github/workflows/s1-cns-scans.yml
```diff
@@ -0,0 +1,22 @@
+name: SentinelOne CNS Scan
+
+on:
+  pull_request:
+    types: [ opened, reopened, synchronize ]
```
jeffy-mathew: shall we also add `ready_for_review` else it wouldn't run when a PR is marked as ready for review from draft state
```suggestion
    types: [ opened, reopened, synchronize, ready_for_review ]
```

jeffy-mathew (approved):

PR #7560, opened 2025-11-24
[TT-16176] respect gateway tags disabled state in server URL generation

lghiur:
TT-16176

When gateway tags are disabled (server.gatewayTags.enabled: false), the OAS server URLs should only include relative paths. Previously, edge endpoint URLs were generated even when tags were disabled.

The fix adds an early check for apiData.TagsDisabled in determineHostsWithEdgeSupport() before any edge endpoint or tag matching logic, ensuring relative path URLs are returned when gateway tags are disabled.

  Changes:
  - Add TagsDisabled check at start of determineHostsWithEdgeSupport()
  - Add unit tests in TestDetermineHosts for various TagsDisabled
  scenarios

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-16176" title="TT-16176" target="_blank">TT-16176</a>
</summary>

|         |    |
|---------|----|
| Status  | Open |
| Summary | Server URLs not changed when gateway tags are disabled |

Generated at: 2025-11-24 20:49:14

</details>

### DIFF
--- a/gateway/api_test.go
+++ b/gateway/api_test.go
@@ -3278,7 +3278,7 @@ func TestOAS(t *testing.T) {
 		})
 
 		t.Run("retain old OAS servers", func(t *testing.T) {
-			t.Run("should retain first entry in existing API", func(t *testing.T) {
+			t.Run("should merge Tyk servers with user-provided servers", func(t *testing.T) {
 				apiInOAS := copyOAS(oasAPI)
 				fillPaths(&apiInOAS)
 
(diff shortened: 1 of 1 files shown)


lghiur:
> The test case `should deduplicate user servers matching Tyk servers` was changed to expect 3 servers instead of 4. The original test and its comment indicated an expectation of 4 servers (Tyk absolute, Tyk relative, and 2 unique user servers after deduplication). The new assertion for 3 servers suggests that the relative path URL might be missing from the final list, which would be a regression. This change makes the test assert an incorrect behavior and could mask a bug in the server regeneration logic.

This is normal since tags are disabled insteead of 2 URLs using edge endpoints just one relative is generated

PR #7579, opened 2025-11-28
[TT-16013]Fix/oas security scheme race condition 7573

andrei-tyk:
## Types of changes

## Checklist

### DIFF
--- a/ee/middleware/streams/middleware.go
+++ b/ee/middleware/streams/middleware.go
@@ -177,24 +177,11 @@ func (s *Middleware) getStreamsConfig(r *http.Request) *StreamsConfig {
 		return config
 	}
 
-	extension, ok := s.Spec.OAS.T.Extensions[ExtensionTykStreaming]
-	if !ok {
-		return config
-	}
-
-	// Handle both map[string]any (raw) and *XTykStreaming (typed) formats.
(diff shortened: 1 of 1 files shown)


apidef/oas/oas.go
```diff
@@ -248,11 +311,13 @@ func (s *OAS) getTykJWTAuth(name string) (jwt *JWT) {
 	if jwtVal, ok := securityScheme.(*JWT); ok {
 		jwt = jwtVal
 	} else {
-		toStructIfMap(securityScheme, jwt)
+		// Security scheme is stored as map[string]interface{}, convert it to JWT struct
+		if toStructIfMap(securityScheme, jwt) {
+			// Cache the converted struct for future use
+			s.getTykSecuritySchemes()[name] = jwt
```
radkrawczyk, line 317: Hi, I think that by this change we strongly decreased probability of crash, but it's still not zero. In case when two requests will be sent in the same time (it will be hard to reproduce) the call to `toStructMap` can return two times true, so concurrent write will happen. It looks like, the `toStructMap` operation is not locked anyhow, but I may be wrong...

I'm just a bit worried that from the issue which have 'always' frequency, we'll do the `once a week/ once a request flood from many clients`

PR #7597, opened 2025-12-05
Replace GitHub Action runners with WarpBuild runners

bsten-tyk:
This commit syncs latest changes from gromit, introduced via https://github.com/TykTechnologies/gromit/pull/428, which introduces new WarpBuild runners to replace the Github's own runners.

## Types of changes

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -286,7 +286,7 @@ jobs:
   resolve-dashboard-image:
     if: github.event.pull_request.draft == false
     needs: goreleaser
-    runs-on: ${{ vars.DEFAULT_RUNNER }}
+    runs-on: warp-ubuntu-latest-x64-8x
     permissions:
       id-token: write
       contents: read


.github/workflows/release.yml
```diff
@@ -286,7 +286,7 @@ jobs:
   resolve-dashboard-image:
     if: github.event.pull_request.draft == false
     needs: goreleaser
-    runs-on: ubuntu-latest
+    runs-on: ${{ vars.DEFAULT_RUNNER }}
```
jeffy-mathew: so, this runner is different from `ubuntu-latest-m`, I believe it's a 4 core runner. shall we have another GH var corresponding to that wherever `ubuntu-latest` is used?

bsten-tyk: Makes sense! I overwrote the value in 43079c8116851a158e800bca2e6ef20d6faa898f for a while.

once we have again proper environment and understanding on ubuntu-latest-m spec, i can replace it with proper env to prevent Gromit to overwrite in its next sync.

For now, that should work hopefully.

.github/workflows/release.yml
```diff
@@ -80,7 +80,19 @@ jobs:
           registry: docker.tyk.io
           username: ${{ secrets.CLOUDSMITH_USERNAME }}
           password: ${{ secrets.CLOUDSMITH_API_KEY }}
-      - uses: actions/cache@v4
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::754489498669:role/ecr_rw_tyk
```
sredxny, line 85: am wondering if this role-to-assume should be an env var 🤔

sredxny (approved):

.github/workflows/release.yml
bsten-tyk: That makes sense. Let me create a ticket for it, as it needs further changes across repositories

PR #7623, opened 2025-12-10
[TT-5545] fix bug "issue when retrieving a list of keys attached to an api"

NurayAhmadova:
## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-5545" title="TT-5545" target="_blank">TT-5545</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | Dashboard API endpoint does not return API-specific keys |

Generated at: 2026-01-21 07:45:34

</details>

### DIFF
--- a/gateway/middleware_test.go
+++ b/gateway/middleware_test.go
@@ -371,6 +371,118 @@ func TestSessionLimiter_RedisQuotaExceeded_PerAPI(t *testing.T) {
 	})
 }
 
+func TestSessionLimiter_RedisQuotaExceeded_ExpiredAtReset(t *testing.T) {
+	g := StartTest(nil)
+	defer g.Close()
+
+	g.Gw.GlobalSessionManager.Store().DeleteAllKeys()
+	defer g.Gw.GlobalSessionManager.Store().DeleteAllKeys()
+
+	t.Run("expiredAt is set to now.Add(quotaRenewalRate) when lock succeeds", func(t *testing.T) {
+		quotaKey := "test-quota-key-expired-at"
+		quotaRenewalRate := int64(3600)
+		quotaMax := int64(100)
+		expectedTTL := time.Duration(quotaRenewalRate) * time.Second
+
+		limiter := g.Gw.SessionLimiter
+
+		session := &user.SessionState{
+			KeyID: "test-key-expired-at",
+			AccessRights: map[string]user.AccessDefinition{
+				"api1": {
+					Limit: user.APILimit{
+						QuotaMax:         quotaMax,
+						QuotaRenewalRate: quotaRenewalRate,
+					},
+					AllowanceScope: "",
+				},
+			},
+		}
+
+		req := &http.Request{}
+
+		rawKey := QuotaKeyPrefix + quotaKey
+		g.Gw.GlobalSessionManager.Store().DeleteRawKey(rawKey)
+
+		limit := &user.APILimit{
+			QuotaMax:         quotaMax,
+			QuotaRenewalRate: quotaRenewalRate,
+		}
+
+		beforeTime := time.Now()
+		blocked := limiter.RedisQuotaExceeded(req, session, quotaKey, "", limit, g.Gw.GlobalSessionManager.Store(), false)
+		afterTime := time.Now()
+
+		assert.Equal(t, quotaMax-1, session.QuotaRemaining, "Quota remaining should be quotaMax - 1 after increment")
+		assert.Greater(t, session.QuotaRenews, int64(0), "QuotaRenews should be set to a future timestamp")
+
+		expectedRenewTimeMin := beforeTime.Add(expectedTTL).Unix()
+		expectedRenewTimeMax := afterTime.Add(expectedTTL).Unix()
+		assert.GreaterOrEqual(t, session.QuotaRenews, expectedRenewTimeMin,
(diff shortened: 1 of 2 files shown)


gateway/api.go
```diff
@@ -702,44 +710,134 @@ type apiAllKeys struct {
...
+		for _, k := range keys {
+			if !strings.HasPrefix(k, QuotaKeyPrefix) && !strings.HasPrefix(k, RateLimitKeyPrefix) {
+				validSessions = append(validSessions, k)
+			}
 		}
+		return apiAllKeys{validSessions}, http.StatusOK
 	}
 
-	sessionsObj := apiAllKeys{fixedSessions}
+	jobs := make(chan string, KeyListingBufferSize)
```
shults: So if it's hard to fetch a couple of sessions in one query, extract this fetching block into a separate method please. Currently, it's hard to read.

NurayAhmadova: done, please re-review

shults (approved):

gateway/api.go
```diff
@@ -702,44 +710,145 @@ type apiAllKeys struct {
...
+
+// filterKeysByAPIID handles the concurrent processing of key permissions
+func (gw *Gateway) filterKeysByAPIID(c context.Context, keys []string, filter, apiID string, hashed bool) ([]string, error) {
+	jobs := make(chan string, KeyListingBufferSize)
+	validSessions := make([]string, 0)
+	var wg sync.WaitGroup
+	var mu sync.Mutex
+
+	for w := 0; w < KeyListingWorkerCount; w++ {
+		wg.Add(1)
```
shults: `wg.Add(KeyListingWorkerCount)` you can put this line before loop once instead of adding 1 each time :/

gateway/api.go
```diff
@@ -75,6 +77,12 @@ import (
 
 const (
 	oAuthClientTokensKeyPattern = "oauth-data.*oauth-client-tokens.*"
+
+	// KeyListingWorkerCount defines how many parallel goroutines we use to fetch key details.
+	KeyListingWorkerCount = 20
```
shults: Btw, probe has sense. It would be nice to have a mechanism that allows rewriting hardcoded values in the production env. 

P.S.
This mechanism already exists in the config. 
It uses this package https://github.com/kelseyhightower/envconfig. But it uses it wrong, because the default value overrides values from the config. IMO, it is worth discussing this matter in a wider group.

shults (approved):
In general, it looks good to me, but we have to rethink the mechanism that allows us to avoid hardcoded values. This mechanism would make ops easier.

NurayAhmadova:
> In general, it looks good to me, but we have to rethink the mechanism that allows us to avoid hardcoded values. This mechanism would make ops easier.

Based on our conversation with Patric (I asked him if this can be moved to configs) he said it was not necessary for now, we just need to iterate over it to find the minimum values that give us the highest performance. Since Radek has his benchmark tests we will collaborate on finding those numbers.
@shults

NurayAhmadova:
Regarding visor comments:

"N+1 Query Problem" comment:
The option with GetMultiKey, as well as the solution with redis pipelines were reviewed and rejected , as discussed with Tyk team.

"Disallow Empty Filter" comment:
Blocking the empty filter would be a breaking change, as `GET /tyk/keys/` is a legacy endpoint used for system audits (confirmed by `TestKeyHandler/List_keys/filter=""`). We enforced the strict check only for `api_id` because that operation triggers an exponentially expensive N+1 deep fetch (JSON unmarshalling), whereas the legacy list is a significantly lighter shallow key scan.

"Contect Propagation" comment:
Updating SessionDetail to accept context.Context necessitates a breaking change to the SessionHandler interface and refactoring all storage backends (Redis, RPC, Memory), which exceeds the scope of this fix. Instead, we enforce cancellation at the Worker Pool orchestration level by monitoring ctx.Done(), ensuring the HTTP handler unblocks immediately upon timeout even if a specific storage call lingers.

gateway/api.go
```diff
@@ -702,44 +705,154 @@ type apiAllKeys struct {
...
 
-	return sessionsObj, http.StatusOK
+	return apiAllKeys{validSessions}, http.StatusOK
+}
+
+// filterKeysByAPIID handles the concurrent processing of key permissions
+func (gw *Gateway) filterKeysByAPIID(c context.Context, keys []string, filter, apiID string, hashed bool) ([]string, error) {
+	numKeys := len(keys)
+	// the values for keyListingWorkerCount and keyListingBufferSize were averaged through profiling and benchmarking
+	keyListingWorkerCount := int(4 + math.Sqrt(float64(numKeys)))
```
shults: `min(int(4 + math.Sqrt(float64(numKeys))), 350)`

gateway/api.go
```diff
@@ -702,44 +705,154 @@ type apiAllKeys struct {
...
+	if keyListingWorkerCount > 350 {
+		// the number 350 is based on the average of 10000 keys with minor contingency
+		keyListingWorkerCount = 350
+	}
+	keyListingBufferSize := keyListingWorkerCount * 100
```
shults: Extract 100 (magick number) to some sort of const entriesPerKey or some sort of that. The code would be more readable.

gateway/api.go
shults: It's nice to extract 350 to some sort of const.

shults (approved):
lgtm

PR #7656, opened 2025-12-29
[TT-16266] Add option to compress API definitions stored in Redis

imogenkraak:
## Description

This change adds a new storage feature flag `compress_api_definitions` that, when enabled, compresses API definitions stored in Redis RPC backups using Zstd compression.

## Related Issue

[TT-16266
](https://tyktech.atlassian.net/browse/TT-16266)

## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-16266" title="TT-16266" target="_blank">TT-16266</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | Implement compression of API definitions to reduce Redis usage |

Generated at: 2026-01-22 12:17:05

</details>

### DIFF
--- a/gateway/rpc_backup_handlers.go
+++ b/gateway/rpc_backup_handlers.go
@@ -93,7 +93,11 @@ func (gw *Gateway) compressAPIBackup(list string) string {
 		return list
 	}
 
-	compressed := compression.CompressZstd([]byte(list))
+	compressed, err := compression.CompressZstd([]byte(list))
+	if err != nil {
+		log.WithError(err).Error("[RPC] --> Failed to compress API definitions, storing uncompressed")
+		return list
+	}
 	log.Debug("[RPC] --> API definitions compressed with Zstd")
 	return string(compressed)
 }
--- a/internal/compression/zstd.go
+++ b/internal/compression/zstd.go
@@ -2,6 +2,7 @@ package compression
 
 import (
 	"bytes"
+	"errors"
 	"sync"
 
 	"github.com/klauspost/compress/zstd"
@@ -21,7 +22,8 @@ var (
 		New: func() interface{} {
 			encoder, err := zstd.NewWriter(nil)
 			if err != nil {
-				log.WithError(err).Panic("Failed to create Zstd encoder")
+				log.WithError(err).Error("Failed to create Zstd encoder")
+				return nil
 			}
 			return encoder
 		},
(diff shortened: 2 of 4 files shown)


internal/compression/zstd.go
```diff
@@ -0,0 +1,51 @@
...
+		"compressed_size":   len(compressed),
+		"compression_ratio": compressionRatio,
+	}).Debug("Data compressed with Zstd")
+
+	return compressed, nil
+}
+
+// DecompressZstd decompresses Zstd-compressed data
+func DecompressZstd(data []byte) ([]byte, error) {
+	const maxDecompressedSize = 100 * 1024 * 1024 // 100MB
```
imogenkraak: Need to set a limit, but not sure what it should be

imogenkraak:
/review does it make sense to use a `sync.Pool` to cache and reuse `zstd.Encoder` and `zstd.Decoder` objects for this change? Or does this just add complexity?

gateway/rpc_backup_handlers.go
```diff
@@ -71,15 +77,59 @@ func (gw *Gateway) saveRPCDefinitionsBackup(list string) error {
...
+
+	compressed, err := compression.CompressZstd([]byte(list))
+	if err != nil {
+		log.WithError(err).Warning("[RPC] --> Failed to compress API definitions, falling back to uncompressed")
+		return list
+	}
+
+	encoded := base64.StdEncoding.EncodeToString(compressed)
+	log.Debug("[RPC] --> API definitions compressed and base64-encoded")
+	return "zstd::" + encoded
```
buraksezer: [ADR-001](https://github.com/TykTechnologies/adrs/pull/6) says:

> Format Detection:
> 
> Uses Zstd magic bytes (0x28, 0xB5, 0x2F, 0xFD) for reliable identification
> JSON validation as first check ensures backward compatibility
> Graceful fallback for unknown formats

The implementation relies on a string prefix instead of the described method and assumes that any body not starting with `zstd::` is valid. Checking the Zstd magic bytes should be sufficient here; we could validate the frame format directly or reuse an existing validator from the zstd package: https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#zstandard-frames

imogenkraak: Ohh good spot, I can't recall why I changed this implementation, but I agree magic bytes is more reliable.

PR #7665, opened 2026-01-13
[Security] Fix dynamic mTLS bypass vulnerability (TT-16050, TT-13208)

buger:
## Summary

This PR fixes a critical security vulnerability in the Tyk Gateway's dynamic mTLS feature that allowed clients to bypass the certificate requirement.

### Vulnerability Description

When `UseCertificate` is enabled for API authentication, an attacker could:
- **Bypass mTLS by providing only an API key** without presenting a TLS client certificate
- **Use a different valid certificate** to access another user's session

### Root Cause

The previous implementation checked for the presence of a TLS certificate only when generating a token from the certificate hash, but allowed requests with API keys to proceed without verifying that a certificate was presented. Additionally, when a session already had a certificate bound to it, the code did not verify that the presented certificate matched.

### Fix

1. When `UseCertificate` is enabled, a valid TLS client certificate is now **ALWAYS required** at the start of request processing
2. If the session has a certificate already bound to it, the presented certificate must match exactly
3. New error types added: `ErrAuthCertRequired` and `ErrAuthCertMismatch`

## Test plan

- [x] Added `TestDynamicMTLSBypassPrevention` with specific test cases:
  - Bypass attempt with API key but no certificate should fail
  - Attacker with different valid cert trying to use victim's session should fail
  - Legitimate client with correct certificate should succeed
  - Attacker with unregistered certificate should fail
- [x] Existing `TestDynamicMTLS` tests pass (valid cert works, expired cert fails)
- [ ] Manual testing with mTLS-protected API

## Related Issues

- Fixes TT-16050: [Security] It is possible to bypass dynamic mTLS requirements with the orgid and the sha256 hash of the certificate
- Fixes TT-13208: [Security] Dynamic mTLS returns successful response to clients that don't hold the certificate

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-16050" title="TT-16050" target="_blank">TT-16050</a>
</summary>

|         |    |
|---------|----|
| Status  | In Refinement |
| Summary | [Security]It is possible to bypass dynamic mtls requirements with the orgid and the sha256 hash of the certificate |

Generated at: 2026-01-15 11:15:05

</details>


andyo-tyk:
Duplicate - fixed in TT-16050

PR #7668, opened 2026-01-15
TT-15604: Selective Certificate Synchronization for MDCB Data Planes

edsonmichaque:
## Description

This PR implements **usage-based certificate synchronization** for MDCB data planes, allowing data planes to only sync and access certificates that are actually used by loaded APIs.

### Key Features

1. **Certificate Registry** - Tracks which certificates are used by which APIs
2. **Usage-Based Sync** - Only sync certificates referenced by loaded APIs
3. **Access Control** - Block access to unused certificates at application layer
4. **Certificate Cleanup** - Remove orphaned certificates on API reload
5. **Security** - Certificate IDs masked in all logs
6. **Performance** - O(1) lookups using map-based sets with zero-size struct{} values
7. **Backward Compatible** - Features disabled by default

## Related Issue

Relates to TT-15604 - POC for usage-based certificate sync feature.

## Motivation

**Problem:**
In MDCB deployments, data planes sync ALL certificates from the control plane regardless of whether they're used by loaded APIs. In segmented deployments where each data plane serves a subset of APIs, this results in 97-98% wasted storage and log noise from irrelevant expired certificates.

**Solution:**
Track certificate usage from API specs and only sync/access certificates that are actually required.

### Core Components

**Certificate Registry** (`gateway/cert_registry.go`)
- Tracks certificate-to-API mappings using bidirectional indices
- O(1) lookup: `Required(certID) bool`
- Automatic deduplication when APIs reference same certificate
- Thread-safe with RWMutex

**Certificate Manager** (`certs/manager.go`)
- Application-layer access control
- Blocks GetRaw() calls for non-required certificates
- Logs blocked attempts with masked certificate IDs

**RPC Storage Handler** (`gateway/rpc_storage_handler.go`)
- Filters keyspace events by certificate registry
- Skips syncing unused certificates

**MDCB Storage** (`storage/mdcb_storage.go`)
- Filters certificate pulls from control plane
- Blocks before RPC call

**Certificate Cleanup** (`gateway/cert.go`)
- Removes orphaned certificates on API reload
- Cleans up previously synced certificates that are no longer referenced

**Expiry Monitoring** (`internal/certcheck/batcher.go`)
- Skips expiry checks for unused certificates

### Configuration

```json
{
  "slave_options": {
    "use_rpc": true,
    "sync_used_certs_only": false,
    "cleanup_certs": false
  }
}
```

Both flags default to `false` for backward compatibility.

## Performance Optimizations

1. **Zero-size struct{}** - Using `map[string]struct{}` instead of `map[string]bool` saves memory
2. **O(1) lookups** - Map-based sets replace linear slice scans
3. **Automatic deduplication** - Certificates referenced multiple times stored once
4. **Minimal overhead** - Registry only populated during API reload

## Security Enhancements

**Certificate ID Masking:**
All logs mask certificate IDs to prevent information leakage:
```go
func maskCertID(certID string) string {
    if len(certID) <= 8 {
        return certID
    }
    return certID[:8] + "***[len=" + strconv.Itoa(len(certID)) + "]"
}
// Output: cert_id=a1b2c3d4***[len=72]
```

Applied to:
- Certificate sync logs
- Certificate cleanup logs
- Access control block logs
- Storage layer logs

### Application Layer Enforcement

Access control implemented in CertificateManager (application layer) rather than storage layer:
- ✅ Proper separation of concerns
- ✅ Storage layer remains generic
- ✅ Easy to test via interface injection
- ✅ Blocks access before storage/RPC calls

### Backward Compatibility

- Features disabled by default
- No breaking changes to existing deployments
- Registry only created when `use_rpc` is enabled
- Graceful fallback if registry is nil

## Files Changed

- `gateway/cert_registry.go` - NEW - Certificate registry
- `gateway/cert.go` - Certificate cleanup
- `certs/manager.go` - Access control
- `storage/mdcb_storage.go` - Storage filtering
- `gateway/rpc_storage_handler.go` - Sync filtering
- `internal/certcheck/batcher.go` - Expiry optimization
- `config/config.go` - Configuration flags
- `cli/linter/schema.json` - Schema validation
- `internal/httpclient/factory_test.go` - Test mock updates
- Multiple test files - Updated mock signatures

## Types of changes

## Checklist

- [x] All tests pass
- [x] Full binary builds successfully
- [x] Backward compatible (features disabled by default)
- [x] Security considerations addressed (cert ID masking)

### DIFF
--- a/gateway/server.go
+++ b/gateway/server.go
@@ -276,7 +276,7 @@ func NewGateway(config config.Config, ctx context.Context) *Gateway {
 	if config.SlaveOptions.UseRPC {
 		gw.certUsageTracker = newUsageTracker()
 	}
-  
+
 	gw.jwkCache = buildJWKSCache(config)
 
 	return gw


storage/mdcb_storage.go
```diff
@@ -21,24 +30,47 @@
...
+		m.certUsage != nil {
+
+		certID := extractCertID(key)
+		if certID == "" {
+			m.logger.Debugf("Unable to extract cert ID from key: %v", key)
+			return m.getFromRPCAndCache(key)
+		}
+
+		if !m.certUsage.Required(certID) {
+			m.logger.WithField("cert_id", certID).
```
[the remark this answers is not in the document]
edsonmichaque: there's no problem about logging ids

edsonmichaque:
> Is it possible to move the certUsageTracker to a separate module, and define the interface there for importing. There is a lot of repeated UsageTracker interfaces accross the code, could cause a headache to update

That makes sense, let me fix it. Thank you @kofoworola

edsonmichaque:
@kofoworola would you have another look?

certs/manager.go
```diff
@@ -39,6 +42,12 @@ var (
 )
 
 //go:generate mockgen -destination=./mock/mock.go -package=mock . CertificateManager
+
+// UsageTracker is an alias for the consolidated certificate usage tracking interface.
+// This maintains backward compatibility while using the shared interface definition
```
kofoworola: recommendation: the UsageTracker is a new feature so this comment can be removed

edsonmichaque: Thank you! Let me fix that

PR #6910, opened 2025-02-27
TT-14163 when start the rpc set as default in emergency mode

sredxny:
### **User description**
<details open>
  <summary><a href="https://tyktech.atlassian.net/browse/TT-14163" title="TT-14163" target="_blank">TT-14163</a></summary>
  <br />
  <table>
    <tr>
      <th>Summary</th>
      <td>Gateway enters in crashloop in emergency mode in k8n's</td>
    </tr>
    <tr>
      <th>Type</th>
      <td>
        Bug
        Bug
      </td>
    </tr>
    <tr>
      <th>Status</th>
      <td>In Dev</td>
    </tr>
    <tr>
      <th>Points</th>
      <td>N/A</td>
    </tr>
    <tr>
      <th>Labels</th>
      <td>-</td>
    </tr>
  </table>
</details>

---

when start the rpc set as default in emergency mode.Some split in the connect function to improve readability

## Description

With this PR now the dataplane will start in emergency mode until the RPC connection is successful. This will improve the start time when MDCB is down and the gateway is restarted or a new pod is created in a K8s environment.  As a side work, some refactor of the ´Connect´ function to improve readability.

## Related Issue

TT-14163

## How This Has Been Tested

- Run an MDCB environment in k8s
- Shut down MDCB, Replicas=0
- Start a new dataplane gw pod -> it should be initialized and not fall in crashback loop

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Refactoring or add test (improvements in base code or adds test coverage to functionality)

## Checklist

- [ ] I ensured that the documentation is up to date
- [ ] I explained why this PR updates go.mod in detail with reasoning why it's required
- [ ] I would like a code coverage CI quality gate exception and have explained why

___

### **PR Type**
- Enhancement

___

### **Description**
- Extracted configuration setup into a helper function.

- Added initializeClient to manage RPC client initialization.

- Introduced setupDialFunction for custom dialer configuration.

- Improved readability by splitting Connect function parameters.

___

### **Changes walkthrough** 📝
<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>rpc_client.go</strong><dd><code>Refactored Connect and added supporting helper functions.</code></dd></summary>
[...]

PR #7705, opened 2026-01-28
[TT-14618] Fix SSL certificate loading from MDCB at startup

edsonmichaque:
## Description

Data plane gateways fail to load SSL certificates from MDCB during startup in v5.8.0+. This PR fixes the regression by adding exponential backoff retry logic to certificate loading.

## Related Issue

https://tyktech.atlassian.net/browse/TT-14618

## Motivation and Context

Critical regression from TT-14163 (PR #6910). Gateway starts in emergency mode immediately, blocking certificate fetch before RPC ready. Affects v5.8.0+.

## How This Has Been Tested

Unit test in certs/manager_tt14618_test.go demonstrates 6 retry attempts over 0.8s, successfully loading certificate after RPC becomes ready.

## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-14618" title="TT-14618" target="_blank">TT-14618</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | [regression]Data plane gateways cannot start if they're configured with a TYK_GW_HTTPSERVEROPTIONS_SSLCERTIFICATES that is in the certificate store |

Generated at: 2026-02-12 05:45:11

</details>

### DIFF
--- a/gateway/api_loader.go
+++ b/gateway/api_loader.go
@@ -915,6 +915,12 @@ func (gw *Gateway) loadHTTPService(spec *APISpec, apisByListen map[string]int, g
 	}
 
 	// Register routes for each prefix
+	gw.generateRoutesForPrefixes(spec, prefixes, gwConfig.HttpServerOptions.EnableStrictRoutes, router, chainObj)
+
+	return chainObj, nil
+}
+
+func (gw *Gateway) generateRoutesForPrefixes(spec *APISpec, prefixes []string, enabledStrictRoutes bool, router *mux.Router, chainObj *ChainObject) {
 	for _, prefix := range prefixes {
 		subrouter := router.PathPrefix(prefix).Subrouter()
 
@@ -924,13 +930,11 @@ func (gw *Gateway) loadHTTPService(spec *APISpec, apisByListen map[string]int, g
 			subrouter.Handle(rateLimitEndpoint, chainObj.RateLimitChain)
 		}
 
-		httpHandler := explicitRouteSubpaths(prefix, chainObj.ThisHandler, gwConfig.HttpServerOptions.EnableStrictRoutes)
+		httpHandler := explicitRouteSubpaths(prefix, chainObj.ThisHandler, enabledStrictRoutes)
 
 		// Attach handlers
 		subrouter.NewRoute().Handler(httpHandler)
 	}
-
-	return chainObj, nil
 }
 
 func (gw *Gateway) loadTCPService(spec *APISpec, gs *generalStores, muxer *proxyMux) {
@@ -1033,14 +1037,17 @@ func (gw *Gateway) loadGraphQLPlayground(spec *APISpec, subrouter *mux.Router) {
 	})
 }
 
-func sortSpecsByListenPath(specs []*APISpec) {
+func sortSpecsByListenPath(specs []*APISpec, enabledCustomDomain bool) {
 	// sort by listen path from longer to shorter, so that /foo
 	// doesn't break /foo-bar
 	sort.Slice(specs, func(i, j int) bool {
-		// we sort by the following rules:
+		// when custom domains are disabled in config we sort by the following rules:
+		// - decreasing order of listen path length
+
+		// when custom domains are enabled in config we sort by the following rules:
 		// - decreasing order of listen path length
 		// - if a domain is empty it should be at the end
-		if (specs[i].Domain == "") != (specs[j].Domain == "") {
+		if enabledCustomDomain && (specs[i].Domain == "") != (specs[j].Domain == "") {
 			return specs[i].Domain != ""
 		}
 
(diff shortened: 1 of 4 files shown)


gateway/server.go
```diff
@@ -1583,6 +1594,33 @@ func (gw *Gateway) afterConfSetup() {
...
+		if conf.SlaveOptions.RPCCertFetchMaxElapsedTime <= 0 {
+			conf.SlaveOptions.RPCCertFetchMaxElapsedTime = 30
+		}
+
+		if conf.SlaveOptions.RPCCertFetchInitialInterval <= 0 {
+			conf.SlaveOptions.RPCCertFetchInitialInterval = 0.1
+		}
+
+		if conf.SlaveOptions.RPCCertFetchMaxInterval <= 0 {
+			conf.SlaveOptions.RPCCertFetchMaxInterval = 2
```
edsonmichaque: use consts, not magic values

gateway/server.go
```diff
@@ -1583,6 +1594,33 @@ func (gw *Gateway) afterConfSetup() {
 		if conf.SlaveOptions.RPCGlobalCacheExpiration == 0 {
 			conf.SlaveOptions.RPCGlobalCacheExpiration = 30
 		}
+
+		if conf.SlaveOptions.RPCCertFetchMaxElapsedTime <= 0 {
+			conf.SlaveOptions.RPCCertFetchMaxElapsedTime = 30
+		}
+
+		if conf.SlaveOptions.RPCCertFetchInitialInterval <= 0 {
+			conf.SlaveOptions.RPCCertFetchInitialInterval = 0.1
```
edsonmichaque: use consts, not magic values

gateway/server.go
```diff
@@ -1583,6 +1594,33 @@ func (gw *Gateway) afterConfSetup() {
 		if conf.SlaveOptions.RPCGlobalCacheExpiration == 0 {
 			conf.SlaveOptions.RPCGlobalCacheExpiration = 30
 		}
+
+		if conf.SlaveOptions.RPCCertFetchMaxElapsedTime <= 0 {
+			conf.SlaveOptions.RPCCertFetchMaxElapsedTime = 30
```
edsonmichaque: use consts, not magic values

gateway/server.go
```diff
@@ -1583,6 +1594,33 @@ func (gw *Gateway) afterConfSetup() {
...
+		}
+
+		// Default RPCCertFetchRetryEnabled if not explicitly set
+		if conf.SlaveOptions.RPCCertFetchRetryEnabled == nil {
+			enabled := certs.DefaultRPCCertFetchRetryEnabled
+			conf.SlaveOptions.RPCCertFetchRetryEnabled = &enabled
+		}
+
+		// Default RPCCertFetchMaxRetries if not explicitly set (0 = unlimited, negative values default to constant)
+		if conf.SlaveOptions.RPCCertFetchMaxRetries == nil {
```
edsonmichaque, line 1698: use same pattern for initialization?

MFCaballero (approved):

certs/manager.go
```diff
@@ -413,7 +534,57 @@ func (c *certificateManager) List(certIDs []string, mode CertificateType) (out [
 			continue
 		}
 
-		val, err := c.storage.GetKey("raw-" + id)
+		// Retry with exponential backoff when MDCB connection is not ready.
```
pvormste: Can we move this entire new code block into a separate function?
I think there is good potential to make it easier to understand as you could also work with an early return instead of heaving those huge if-else-blocks

edsonmichaque:
## Addressing Visor Performance Concern

Visor flagged a performance issue at `certs/manager.go:408-463` claiming that `backoff.Retry` is called within the loop for each certificate, which would reset the backoff timer.

**This is incorrect.** The implementation already uses a shared backoff strategy correctly:

```go
// Create backoff strategy ONCE for all certificates (line 566)
var sharedBackoff backoff.BackOff
if c.certFetchRetryEnabled {
    expBackoff := backoff.NewExponentialBackOff()
    // ... configure once ...
    sharedBackoff = expBackoff  // Created OUTSIDE the loop
}

for _, id := range certIDs {
    // Pass the SAME sharedBackoff to each certificate fetch (line 591)
    val, err := c.fetchCertificateWithRetry(id, sharedBackoff)
}
```

**Why this works correctly:**

1. The `sharedBackoff` is created **once** before the loop (line 566)
2. The **same instance** is passed to every certificate fetch (line 591)
3. The backoff state is shared across all certificates in the batch
4. First certificate that hits MDCB connection error will retry with exponential backoff
5. By the time it succeeds, subsequent certificates fetch immediately (MDCB is connected)

**Emergency mode context:**

This retry logic **only activates during gateway startup** when in "emergency mode" (MDCB not yet connected). This is a rare, transient state lasting typically 1-5 seconds. Once MDCB connects:
- Emergency mode exits
- All subsequent certificate fetches succeed immediately (no retries)
- Normal operation resumes

The performance impact is minimal and only occurs during the critical startup race condition this PR fixes.

Test coverage validates this behavior in `TestSharedBackoffPreventsMultiplicativeDelay` which confirms that loading 5 certificates takes ~200ms (not 5×30s).

certs/manager.go
edsonmichaque: Thank you @pvormste! I've addressed these issues, would you have another look?

pvormste: Love it!

PR #7719, opened 2026-01-31
[TT-16492] json rpc mcp request handling

lghiur:
## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-16492" title="TT-16492" target="_blank">TT-16492</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | MCP Request Handling & JSON-RPC Routing |

Generated at: 2026-02-05 11:42:50

</details>

### DIFF
--- a/gateway/api_loader_test.go
+++ b/gateway/api_loader_test.go
@@ -9739,7 +9739,9 @@ func TestMCPRequestSizeLimit_Security(t *testing.T) {
 		// Create upstream server
 		upstream := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
 			w.WriteHeader(http.StatusOK)
-			_, _ = w.Write([]byte(`{"jsonrpc":"2.0","result":"success","id":1}`))
+			if _, err := w.Write([]byte(`{"jsonrpc":"2.0","result":"success","id":1}`)); err != nil {
+				t.Errorf("failed to write response: %v", err)
+			}
 		}))
 		defer upstream.Close()
 
--- a/gateway/model_apispec.go
+++ b/gateway/model_apispec.go
@@ -184,7 +184,7 @@ func (a *APISpec) getMatchPathAndMethod(r *http.Request, mode URLStatus) (string
 		}
 	}
 
-	if a.IsMCP() && a.Proxy.ListenPath != "/" && !isJSONRPCVEMPath(r, matchPath) {
+	if a.Proxy.ListenPath != "/" && !isJSONRPCVEMPath(r, matchPath) {
 		matchPath = a.StripListenPath(matchPath)
 	}
 


buger:
/visor I have a question about this PR, change inside the getExtendedPathSpecs function. Is it the right place to put mcp.RegisterVEMPrefixes() and etc, or there is another better suited API initialization logic? Do we have similar examples from different functionality which needs to apply such custom API modifying logic?

buger:
/visor followup on last comment, can we then move 

```
if apiSpec.IsMCP() {
		mcp.RegisterVEMPrefixes()
		a.populateMCPPrimitivesMap(apiSpec)
		a.calculateMCPAllowlistFlags(apiSpec)

		if apiSpec.JsonRpcVersion == apidef.JsonRPC20 {
			apiSpec.JSONRPCRouter = mcp.NewRouter()
		}
		```
		
Inside the makeSpec instead? Or there are nunances on how calculateMCPAllowlistFlags populateMCPPrimitivesMap and etc work, and it needs be actually called inside the getExtendedPathSpecs?

**Comment by buger** [comment]
/visor now question about this line: 

whiteListEnabled := len(whiteListPaths) > 0 || apiSpec.OperationsAllowListEnabled
	if apiSpec.IsMCP() {
		whiteListEnabled = false
	}
	
My assumption is that for JSON RPC we have own logic for handling whitelisting, is that correct? Or we completelly disable whitelist functionality here?

File: gateway/api_definition.go [hunk 1]
```diff
@@ -1513,9 +1549,18 @@ func (a APIDefinitionLoader) getExtendedPathSpecs(apiVersionDef apidef.VersionIn
...
 	combinedPath = append(combinedPath, oasValidateRequestPaths...)
 	combinedPath = append(combinedPath, oasMockResponsePaths...)
-	combinedPath = append(combinedPath, mcpVEMs...)
 
-	return combinedPath, len(whiteListPaths) > 0
+	// Enable whitelist mode if there are whitelist paths or operation-level allows.
+	// For MCP APIs, don't enable global whitelist mode at all because:
+	// 1. It would block the listen path before JSON-RPC routing happens
+	// 2. Allowlist enforcement happens at the VEM level during sequential routing
+	// 3. VEM WhiteList entries are checked correctly in URLAllowedAndIgnored
```
buger: We need to extend the comment to say that it not actually disable it, but instead it has a more complex logic inside the VersionCheck middleware to do that.

buger:
/visor - what if we remove the `if apiSpec.IsMCP() {
whiteListEnabled = false
}` part, and instead inside the whitelist middleware (or I guess it is version check), we will got the same if condition if whitelisting is enabled, but inside it we will have isMCP() flag condition which change the logic. E.g. we don ot disable it inside the make spec or similar, we just change its behavior. Can you draft how this change can look like?

buger:
/visor are there any other checks for whitelisting anywhere in our codebase, except VersionCheck which we need to handle with custom MCP logic?

gateway/mw_version_check.go
```diff
@@ -99,6 +103,20 @@ outside:
 	requestValid, stat := v.Spec.RequestValid(r)
 	if !requestValid {
 		if stat == MCPPrimitiveNotFound {
```
buger: /visor Any idea about this line? It sounds a bit weird. If MCPPrimitiveNotFound actually run specific JSON RPC logic. 

Also I prefer to have here some duplication of spec.IsMCP() feature flag, just in case, to ensure this logic won't run for other apis.

buger:
/visor Any idea about this line? It sounds a bit weird. If MCPPrimitiveNotFound actually run specific JSON RPC logic.

if stat == MCPPrimitiveNotFound {

Can you explain why this logic works right now in versioncheck middleware?

gateway/api_definition.go
```diff
@@ -1605,13 +1650,14 @@ func (a *APISpec) URLAllowedAndIgnored(r *http.Request, rxPaths []URLSpec, white
 		}
 
 		if r.Method == rxPaths[i].Internal.Method && rxPaths[i].Status == Internal {
-			// MCP primitive VEMs require explicit MCP routing context.
+			// MCP primitive VEMs return 404 to avoid exposing internal-only endpoints.
+			// They can only be accessed via JSON-RPC routing (MCP, A2A, etc.), not via generic looping.
 			if mcp.IsPrimitiveVEMPath(rxPaths[i].Internal.Path) {
```
buger: Needs be wrapped to spec.IsMCP()

gateway/api_definition.go
```diff
@@ -1677,23 +1723,31 @@ func (a *APISpec) URLAllowedAndIgnored(r *http.Request, rxPaths []URLSpec, white
 			}
 		}
 
+		// MCP primitive VEMs: continue checking for BlackList even outside whitelist mode.
+		// This allows explicit BlackList entries to block primitives without Allow.
+		if rxPaths[i].Status == Internal && r.Method == rxPaths[i].Internal.Method {
```
buger: Better wrap it ot spec.IsMCP() too

buger:
/visor do we really need RegisterVEMPrefix abstraction, if we already have the protocol field inside the API?

gateway/mw_version_check.go
buger: MCPPrimitiveNotFound can be expected response, if upstream has a resource or tool, but our API definition does not know about it. And if we have allow list configured, it should end up to be error, and if not we should just proxy to upstream - lets add this clarification to comment. And also lets wrap additionally to spec.IsMCP() flag just in case.

gateway/mw_version_check.go
```diff
@@ -117,6 +135,28 @@ outside:
 
 	versionInfo, _ := v.Spec.Version(r)
 	versionPaths := v.Spec.RxPaths[versionInfo.Name]
+
+	// For MCP primitives with allowlist enabled: check if VEM has WhiteList entry
+	// If allowlist is on but VEM doesn't have a WhiteList entry → block
+	if httpctx.IsJsonRPCRouting(r) && mcp.IsPrimitiveVEMPath(r.URL.Path) {
```
buger: Wrap to spec.IsMCP()

gateway/mw_version_check.go
```diff
@@ -117,6 +135,28 @@ outside:
...
+		if allowListEnabled {
+			if err := v.checkVEMWhiteListEntry(r.URL.Path, versionPaths, "access to this resource has been disallowed"); err != nil {
+				return err, http.StatusForbidden
+			}
+		}
+	}
+
+	// For JSON-RPC operations with allowlist enabled: check if operation VEM has WhiteList entry
+	// If OperationsAllowListEnabled but operation doesn't have a WhiteList entry → block
+	if httpctx.IsJsonRPCRouting(r) && strings.HasPrefix(r.URL.Path, jsonrpc.MethodVEMPrefix) {
```
buger: Wrap to spec.IsMCP()

gateway/mw_version_check.go
```diff
@@ -117,6 +135,28 @@ outside:
...
+		if v.Spec.OperationsAllowListEnabled {
```
buger, line 144: Do we really need a separate field for VEM allow list or we can just share WhiteList with Paths for that.

gateway/mw_version_check.go
```diff
@@ -117,6 +135,28 @@ outside:
...
+			if err := v.checkVEMWhiteListEntry(r.URL.Path, versionPaths, "Access to this operation has been disallowed"); err != nil {
```
buger, line 145: Lets add a comment that at this point in time r.URL.Path now actually a VEM endpoint, e.g. like /mcp/jsonoperation:tool/call or similar

gateway/mw_version_check.go
```diff
@@ -147,3 +187,26 @@ outside:
 
 	return nil, http.StatusOK
 }
+
+// getPrimitiveAllowListFlag returns the allowlist flag for a given MCP primitive VEM path.
+func (v *VersionCheck) getPrimitiveAllowListFlag(path string) bool {
+	if strings.HasPrefix(path, mcp.ToolPrefix) {
```
buger, line 232: What if our API has a listen path like this and URL request path becomes smth like that: /myapi2/v2/mcp/jsonrpc:tool1

Does the path argument, which we pass to this function already gets cleaned up from the listen path or not.

buger:
/visor exaplain me about getPrimitiveAllowListFlag function. 

What if our API has a listen path like this and URL request path becomes smth like that: /myapi2/v2/mcp/jsonrpc:tool1

Does the path argument, which we pass to this function already gets cleaned up from the listen path or not.

Wehre do we override r.URL.Path for Vem requests? When we override it - does it include listen path or not?

buger:
/visor question about last comment. Can it introduce some issues to our existing routers, when we strip the listen path and override request like this? ALso we have things like strict matching config, strict prefix and suffix etc!

buger:
/visor what about plugins and etc? What will be current behavior? Pls evaluate differnet plugin types we have, and how they will behave with this VEM endpoints? Will it call plugin 3 times, first for real url, then tool VEM url, and tool name VEM url? 

Also how it compare to the old tyk looping middleware mechanism, does it have the same issue? Does it re-run the full mw chain each time?

buger:
/visor what if we disable running plugins for "internal" and "vem" requests, only for the real request?

buger:
/visor I meant custom Go and gRPC plugins

gateway/mw_version_check.go
lghiur: For MCP is better to use another flag as this WhiteList is used in other parts of the code base and triggers other flows if truthy

PR #7755, opened 2026-02-09
TT-17896: add ability to turn off transfer encoding to specific API targets

### LINKED ISSUE #7754: Add per-API option to disable upstream chunked transfer encoding by buffering in Tyk (with safe defaults)
Some upstream services do not reliably accept requests sent with chunked transfer encoding. When this happens through Tyk, requests can be rejected or handled incorrectly.

The workaround we use at BYU is to run an additional proxy between Tyk and the upstream just to normalize request bodies. That adds extra infrastructure, latency, and operational overhead.

Please add a way to configure this behavior at the API level so Tyk can send upstream requests in a non-chunked form when needed.

At a high level, the feature should:
- Be opt-in per API.
- Preserve normal behavior for APIs that do not enable it.
- Include safe body-size controls to avoid unbounded buffering/memory risk.
- Work consistently across Tyk API definition formats.

- Keeping the current behavior and requiring upstream-side fixes.
- Requiring clients to always send fixed-length request bodies.
- Using a separate intermediary proxy to rewrite/normalize requests before upstream.

All of these alternatives either increase operational complexity or are not always feasible.

Note that WSO2 has this feature: https://apim.docs.wso2.com/en/latest/api-gateway/policies/disabling-message-chunking/

jvisker:
## Description

Add an opt-in, per-API capability to avoid chunked transfer encoding to upstreams by buffering eligible request bodies in Tyk and forwarding with `Content-Length`.

The change also introduces stronger safeguards for this buffering path, including bounded body-size handling and a default safety limit when no explicit limit is configured.

## Related Issue

Closes #7754

## Motivation and Context

Some upstreams/proxies/WAFs do not reliably handle chunked request bodies. The current workaround is often to add another proxy tier between Tyk and upstream purely for request normalization, which increases complexity and latency.

This PR allows that compatibility behavior to be configured directly in Tyk at API level while preserving default behavior for APIs that do not opt in.

## How This Has Been Tested

- Unit tests (targeted):
  - `go test ./gateway -run "TestDisableChunkedEncodingForUpstream|TestDeepCopyBody"`
  - `go test ./apidef/oas -run "TestDisableChunkedEncoding|TestPreserveHostHeader"`
- Lint/tooling:
  - `task lint`

All commands above completed successfully in this branch.

N/A

## Types of changes

## Checklist

Additional notes:
- This PR does not intentionally introduce persistent `go.mod`/`go.sum` changes.
- No code coverage gate exception is requested.

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -1026,6 +1026,8 @@ type RequestSigningMeta struct {
 
 type ProxyConfig struct {
 	PreserveHostHeader          bool                          `bson:"preserve_host_header" json:"preserve_host_header"`
+	DisableChunkedEncoding      bool                          `bson:"disable_chunked_encoding" json:"disable_chunked_encoding"`
+	ChunkedEncodingMaxBodySize  int64                         `bson:"chunked_encoding_max_body_size" json:"chunked_encoding_max_body_size"`
 	ListenPath                  string                        `bson:"listen_path" json:"listen_path"`
 	TargetURL                   string                        `bson:"target_url" json:"target_url"`
 	DisableStripSlash           bool                          `bson:"disable_strip_slash" json:"disable_strip_slash"`
--- a/apidef/oas/upstream.go
+++ b/apidef/oas/upstream.go
@@ -53,6 +53,11 @@ type Upstream struct {
 	// Tyk classic API definition: `proxy.disable_strip_slash`.
 	PreserveTrailingSlash *PreserveTrailingSlash `bson:"preserveTrailingSlash,omitempty" json:"preserveTrailingSlash,omitempty"`
 
+	// DisableChunkedEncoding controls whether Tyk buffers request bodies and forwards them upstream
+	// with a Content-Length header instead of chunked transfer encoding.
+	// Tyk classic API definition: `proxy.disable_chunked_encoding`.
+	DisableChunkedEncoding *DisableChunkedEncoding `bson:"disableChunkedEncoding,omitempty" json:"disableChunkedEncoding,omitempty"`
+
 	// TLSTransport contains the configuration for TLS transport settings.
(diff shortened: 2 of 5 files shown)


nerdydread:
Thanks for contribution @jvisker.  Looks good, there are just a couple of items that we will need to address.

1. **OOM Protection** - In `gateway/reverse_proxy.go`, the core proxy function `WrappedServeHTTP` calls `deepCopyBody(req, outreq)` at line 1103, which is before `disableChunkedEncodingForUpstream` is called at line 1171.
`deepCopyBody` uses `io.ReadAll(source.Body`) to read the entire request body into memory. If a user sends a massive chunked request and the gateway-level `MaxRequestBodySize` is 0 (the default), `deepCopyBody` will read the entire unbounded body into memory and cause an OOM crash before `disableChunkedEncodingForUpstream` is even reached. The PR's size control (`ChunkedEncodingMaxBodySize`) is bypassed because the unbounded buffering already happened.

2. **Double Memory Allocation** - Because `deepCopyBody` already reads the body and replaces req.Body with a `bytes.Buffer`, when `disableChunkedEncodingForUpstream` calls `readBodyForChunkedEncoding`, it uses `io.ReadAll` again on that buffer. This allocates a second byte slice of the same size, doubling the memory required for the request body.

**Suggestion**
To fix this, the chunked encoding disablement logic needs to be integrated earlier in the request lifecycle (e.g., inside or before `deepCopyBody`, or within `copyRequest/nopCloseRequestBodyErr`) so that the body is read from the network with the `io.LimitReader` first, preventing both the OOM risk and the double memory allocation.

jvisker:
If you do get to a place where you want me to look at addressing those let me know. For now (from other channels of communication) it sounds like you will run with it.

PR #7782, opened 2026-02-16
[TT-16609] be create edit internal well known url prm

andrei-tyk:
## Description

 Converts the PRM (Protected Resource Metadata) well-known endpoint from a subrouter HandlerFunc (loadPRMWellKnownEndpoint) into a proper TykMiddleware (PRMMiddleware). The middleware is placed in the chain 
  after MiddlewareContextVars and TrackEndpointMiddleware but before the authentication block. It intercepts GET requests matching the well-known path (including the listen path prefix), serves the PRM JSON
  document, and terminates the chain via middleware.StatusRespond. Non-matching requests pass through to the next middleware. The old loadPRMWellKnownEndpoint function and its subrouter registration in       
  generateSubRoutes have been removed.  

## Motivation and Context
The PRM endpoint was previously registered directly on the subrouter, which meant it executed before the entire middleware chain. As a result, MiddlewareContextVars never ran, so $tyk_context.* variables
  (headers, path, cookies, remote_addr, etc.) were never populated — only $secret_env.* and other secret-based variables could be resolved in the PRM resource field. By moving PRM into the middleware chain
  after MiddlewareContextVars, $tyk_context.* variables are now available for substitution. The endpoint remains accessible without authentication since it runs before the auth middlewares, and $tyk_meta.*
  still won't work as expected since it requires an authenticated session.

## How This Has Been Tested

 Go unit tests (go test -run "TestPRM" ./gateway/ -v -count=1) — all existing tests pass unchanged (TestPRMWellKnownEndpoint, TestPRMWWWAuthenticateHeader, TestPRMHappyPath_FullDiscoveryFlow,TestJWTMiddleware_PRMWWWAuthenticate, TestAuthKeyMiddleware_PRMWWWAuthenticate). Two new test cases were added in TestPRMMiddleware_ContextVariables: one verifying that $tyk_context.headers_X_Custom_Header is resolved in the resource field when EnableContextVars is true, and another verifying that requests to non-well-known paths pass through without interception. Python integration tests (pytest tests/mcp/mcp_prm_test.py) — all 4 tests pass against a running gateway and dashboard, including the updated test_context_variables_in_resource which now uses $tyk_context.headers_X_Test_Header instead of $secret_env.HOSTNAME for a deterministic assertion.

## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-16609" title="TT-16609" target="_blank">TT-16609</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | (BE) Create/Edit Internal Well Known URL PRM |

Generated at: 2026-02-19 13:18:37

</details>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -741,7 +741,7 @@ jobs:
           org_gh_token: ${{ secrets.ORG_GH_TOKEN }}
       - name: Run API tests
         uses: TykTechnologies/github-actions/.github/actions/tests/api-tests@main
-        timeout-minutes: 30
+        timeout-minutes: 45
         id: test_execution
         with:
           user_api_secret: ${{ steps.env_up.outputs.USER_API_SECRET }}


gateway/mw_protected_resource.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (m *PRMMiddleware) ProcessRequest(w http.ResponseWriter, r *http.Request, _ interface{}) (error, int) {
+	prm := m.Spec.GetPRMConfig()
+	if prm == nil {
+		return nil, http.StatusOK // pass through
+	}
+
+	// Only intercept GET requests to the well-known path (include listen path prefix)
+	wellKnownPath := path.Join(m.Spec.Proxy.ListenPath, prm.GetWellKnownPath())
```
lghiur: I woudl split a bit into. compose well known path, build resource, write response small functions

andrei-tyk:
/visor  I thnk the r.Host issue raised is an infrastructure-level mitigation (reverse proxy should set/validate Host), and that this follows the same r.Host pattern already used across the codebase. If host header injection protection is needed, it should be a gateway-wide feature, not a per-middleware fix.

PR #7806, opened 2026-02-24
[TT-12238] Javascript regex issue on OAS API

MaciekMis:
## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-12238" title="TT-12238" target="_blank">TT-12238</a>
</summary>

|         |    |
|---------|----|
| Status  | In Test |
| Summary | Javascript regex issue on OAS API |

Generated at: 2026-03-11 12:25:33

</details>


lib/apidef/utils.go
```diff
@@ -0,0 +1,59 @@
...
+	re2Regex     = regexp.MustCompile(`\\x\{([0-9a-fA-F]{4})}`)
+)
+
+// TransformUnicodeEscapesToRE2 takes a byte array containing a regular expression
+// pattern and transforms ECMA-262 compliant Unicode escape sequences (`\uXXXX`)
+// into a format that is compatible with Go's RE2 regex engine (`\x{XXXX}`).
+// This is necessary because RE2 does not support the `\u` escape sequence but
+// does support hexadecimal escapes, which can represent any Unicode code point.
+// The function returns a new byte array with the transformed pattern.
+func TransformUnicodeEscapesToRE2(data []byte) []byte {
```
pvormste: I can think of 2 ways to improve this solution a tiny bit and make it more future proof:

1. A general bytes modifier func that takes a splat argument:
```go
type DataBytesModifier func(data []byte) []byte

func ModifyDataBytes(data []byte, modifiers ...DataBytesModifier) {
 // iterate over the modifiers
}

func TransformUnicodeEscapesToRE2(data []byte) []byte {
 // ...
}

func RestoreUnicodeEscapesFromRE2(data []byte) []byte {
  // ...
}
```

2. Use builder pattern ( = requires more modifications to PR)
```go
type DataBytesModifier struct {
	data []byte
}

func NewDataBytesModifier(data []byte) *DataBytesModifier {
	return &DataBytesModifier{data: data}
}

func (d *DataBytesModifier) TransformUnicodeEscapesToRE2() *DataBytesModifier {
	// ...
	return d
}

func (d *DataBytesModifier) RestoreUnicodeEscapesFromRE2() *DataBytesModifier {
	// ...
	return d
}

func (d *DataBytesModifier) Execute() []byte {
	return d.data
}
```

Feel free to adjust names

pvormste (changes requested):

gateway/api.go
```diff
@@ -3592,6 +3606,8 @@ func extractOASObjFromReq(reqBody io.Reader) ([]byte, *oas.OAS, error) {
 		return nil, nil, ErrRequestMalformed
 	}
 
+	reqBodyInBytes = lib.TransformUnicodeEscapesToRE2(reqBodyInBytes)
```
shults: You are replacing unicode with re2 in the whole body.
The sequence of "\uXXXX" may occur not only in regex fields.

Maybe it's worth parsing the body to OAS object, and after that remap fields with regex?

lib/apidef/utils.go
MaciekMis: Introduced builder pattern for bytes modifier

gateway/api.go
MaciekMis: The solution would be much more complex. The easiest way would be to iterate over OAS document and verify if predefined fields (containing regexp) have to be changed. However this solution would be prone to error when any new Tyk exclusive field (containing regexp) is implemented.

PR #7811, opened 2026-02-25
fix(mdcb): proactively sync required certs after API reload

edsonmichaque:
## Description

Fixes upstream certs not syncing to DC Redis when `sync_used_certs_only: true` is enabled (option A: proactive sync on reload).

Adds `syncRequiredCertificates()` called after each `loadApps()` tracker update. It iterates all cert IDs in the usage tracker and calls `CertificateManager.GetRaw()` for each. If the cert is already in local DC Redis the call is a cheap no-op; if missing it pulls from MDCB and caches it locally.

## Related Issue

Fixes: Tyk MDCB upstream certs not syncing to DC Redis on API reload (`sync_used_certs_only: true`)

## Motivation and Context

When `sync_used_certs_only: true`, `CertificateAdded` keyspace events arrive before the API reload that populates the cert usage tracker. `ProcessKeySpaceChanges` checks `Required()` — which returns false at that moment — and silently skips the cert. After the reload the tracker is correct (`cert_count=N`) but the certs were never fetched from MDCB.

This fix ensures that after every reload, all certs currently required by loaded APIs are present in DC Redis, regardless of whether the keyspace event arrived before or after the reload.

## How This Has Been Tested

- Docker Compose MDCB environment with 3 DC gateways, `sync_used_certs_only: true`, `group_id` set per DC
- Uploaded certs and created APIs with `upstream_certificates` referencing each cert
- Verified DC Redis contains all referenced certs after reload
- Unit tests: `TestLoadApps_CertificateTracking`, `TestCollectCertUsageMap`, `TestUsageTracker_ReplaceAll` all pass

## Types of changes

## Checklist

### DIFF
--- a/gateway/api_loader.go
+++ b/gateway/api_loader.go
@@ -1291,27 +1291,36 @@ func (gw *Gateway) syncRequiredCertificates() {
 		return
 	}
 
-	fetched := 0
-	cached := 0
+	var (
+		mu      sync.Mutex
+		fetched int
+		failed  int
+		wg      sync.WaitGroup
+	)
 
 	for _, certID := range certIDs {
-		_, err := gw.CertificateManager.GetRaw(certID)
-		if err != nil {
-			mainLog.WithFields(logrus.Fields{
-				"cert_id": certID,
-				"err":     err,
-			}).Debug("syncRequiredCertificates: could not fetch cert")
-			continue
-		}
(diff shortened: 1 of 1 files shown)


edsonmichaque:
Closing in favour of option-b (pending queue approach, PR #7812).

PR #7812, opened 2026-02-25
[TT-15604] sync certs skipped during API reload when sync_used_certs_only is enabled

edsonmichaque:
## Description

Fixes upstream certs not syncing to DC Redis when `sync_used_certs_only: true` is enabled (option B: pending queue + drain on reload).

Adds a `pendingCerts sync.Map` to the Gateway struct. When `ProcessKeySpaceChanges` would skip a cert because `Required()` is false, it parks the cert ID in `pendingCerts` instead of discarding it. After each `loadApps()` tracker update, the pending set is drained and `GetRaw()` is called for any cert that is now required.

## Related Issue

Fixes: Tyk MDCB upstream certs not syncing to DC Redis on API reload (`sync_used_certs_only: true`)

## Motivation and Context

When `sync_used_certs_only: true`, `CertificateAdded` keyspace events arrive before the API reload that populates the cert usage tracker. `ProcessKeySpaceChanges` checks `Required()` — which returns false at that moment — and silently drops the cert. After the reload the tracker is correct (`cert_count=N`) but the certs were never fetched.

This fix is surgical: only certs that were explicitly skipped due to the timing race are re-fetched, and only if they become required after the subsequent reload.

## How This Has Been Tested

- Docker Compose MDCB environment with 3 DC gateways, `sync_used_certs_only: true`, `group_id` set per DC
- Observed `"skipping certificate - not used by loaded APIs, queuing for re-check after reload"` debug log followed by `"synced pending certificate after reload"` info log after API creation
- Verified DC Redis contains all referenced certs after reload
- Unit tests: `TestLoadApps_CertificateTracking`, `TestCollectCertUsageMap`, `TestUsageTracker_ReplaceAll` all pass

## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-15604" title="TT-15604" target="_blank">TT-15604</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | [Feature] Data plane should pull only the required certificates when synchroniser is in use |

Generated at: 2026-02-27 13:19:22

</details>

### DIFF
--- a/gateway/api.go
+++ b/gateway/api.go
@@ -513,6 +513,11 @@ func (gw *Gateway) handleAddOrUpdate(keyName string, r *http.Request, isHashed b
 		return apiError(err.Error()), http.StatusBadRequest
 	}
 
+	if err := gw.validateNonMCPFieldsOnMCPProxy(newSession.AccessRights); err != nil {
+		log.Error(err)
+		return apiError(err.Error()), http.StatusBadRequest
+	}
+
 	mw := &BaseMiddleware{Gw: gw}
 	mw.ApplyPolicies(newSession)
 
@@ -1116,6 +1121,11 @@ func (gw *Gateway) handleAddOrUpdatePolicy(polID string, r *http.Request) (inter
 	}
 
 	if err := gw.validateMCPFieldsInAccessRights(newPol.AccessRights); err != nil {
+		log.Error(err)
+		return apiError(err.Error()), http.StatusBadRequest
+	}
+
+	if err := gw.validateNonMCPFieldsOnMCPProxy(newPol.AccessRights); err != nil {
 		log.Error(err)
 		return apiError(err.Error()), http.StatusBadRequest
 	}
--- a/gateway/policy.go
+++ b/gateway/policy.go
@@ -105,6 +105,35 @@ func (gw *Gateway) validateMCPFieldsInAccessRights(accessRights map[string]user.
 	return nil
 }
 
+// hasNonMCPFields returns true if the access definition contains any HTTP/REST or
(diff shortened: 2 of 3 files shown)


edsonmichaque:
/visor The unbounded goroutine concern here is not a practical issue. The `pendingCerts` map only accumulates cert IDs that arrived via Redis keyspace events (CertificateAdded) in the narrow race window between the event and the next API reload. In normal operation this is a handful of entries — one event per cert uploaded, not a batch. Each goroutine does a single RPC call (GetRaw) which is network-bound, not CPU-bound, and goroutines are cheap. On initial startup there are no pending certs yet (the map starts empty); the pending queue only fills during subsequent reloads. A large configuration change would push N cert events, but N is bounded by the number of certs changed, not the total cert count. Introducing a semaphore adds complexity and a magic number with no meaningful benefit for this use case.

edsonmichaque:
/visor This is intentional. If a cert is not available in the local store it will be fetched lazily when it is actually needed, so discarding it from the pending queue on failure is safe.

edsonmichaque:
/visor Certs are fetched lazily during request processing — if a cert is not found in the local store, MdcbStorage transparently pulls it from MDCB at that point. So a failed prefetch here has no impact on correctness.

edsonmichaque:
/visor Not an issue. Even if the prefetch fails, the cert will be fetched lazily during request processing — MdcbStorage pulls from MDCB transparently when a cert is not found in the local store. The pending queue is a best-effort optimisation to warm the local cache ahead of time, not the authoritative sync path.

pvormste (approved):

PR #7903, opened 2026-03-18
[TT-16812]  add MCPStats to handler_success and error handler response records

lghiur:
Add recordMCPDetails() which reads JSONRPCRoutingState from request context and populates MCPStats (method, primitive type/name) on the AnalyticsRecord.
Call it from both handler_success.go (RecordHit) and handler_error.go (HandleError), mirroring the existing recordGraphDetails pattern.
Tag MCP records with tyk-mcp-analytics so pumps can filter them.
Normalise the recorded path to the original MCP listen path (not the routed sub-path).
Add unit tests for all three behaviours: no routing state, full state, path normalisation.

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-16812" title="TT-16812" target="_blank">TT-16812</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | [Gateway] Capture MCP Identity in Analytics Records |

Generated at: 2026-03-19 17:18:09

</details>


tbuchaillot (changes requested):
We need to update Tyk pump dependency so it picks up the structure changes https://github.com/TykTechnologies/tyk/blob/master/go.mod#L33

tbuchaillot:
Also - maybe for inspiration - https://github.com/TykTechnologies/tyk/blob/master/gateway/mw_redis_cache_test.go#L374 here's some test around "super pseudo kinda" e2e of the analytics mechanism for unit test - it would be amazing if we have some of these on this

tbuchaillot (approved):

PR #7941, opened 2026-03-31
TT-6075: Update rate limit implementation

vladzabolotnyi:
## Description

Update logic with handling rate limit headers data. Before we always set the data from quota. Now we're adding configuration to select the source for headers value.

## Motivation and Context
-  Updated config with new fields to support context variables and RL headers source
- Added new context variables for RL and Quotas
- Added HeaderSender interface and quota, RL implementations to support polymorphism based on headers source config

## How This Has Been Tested
1. Unit tests

## Demos

_Given API with 3/15s RL and 5 Quotas && "ratelimit_headers_source": "quotas",  "enable_redis_rolling_limiter": true_

https://github.com/user-attachments/assets/3ca49358-76f2-4d65-b881-9097a6f08483

_When client sends 3 request, Headers are present and reflect Quota state(3/5). After 3 requests the RL is hit and 429 code returned. Wait 15 secs, send 2 more requests to return 403 and headers are omitted. We don't include headers on error response to support backward compatibility. When rate limit is hit we can see 429 status code and for quotas limit we could see 403, that's why we have different status codes - rate limit is recovering but quotas has much longer delay to recover_

https://github.com/user-attachments/assets/24538cbc-6109-4226-8093-460065fae154

_Given API with 3/5s RL and 5 Quotas && "ratelimit_headers_source": "rate_limits",  "enable_redis_rolling_limiter": true. Send 3 requests, hit 429 code, headers must be present reflecting current state. Wait till RL refresh, send 2 more requests to hit the quotas limit and 403 status code. Headers are present reflecting rate limit state._

https://github.com/user-attachments/assets/428e378b-a75f-4144-ba00-4061165e0006

_Context variables attached to response headers_

https://github.com/user-attachments/assets/743fde4c-7da2-4cfa-9eb4-65ae11e0913d

## Types of changes

## Checklist


vladzabolotnyi:
/visor 
>The `limitSentinel` function spawns a 'fire-and-forget' goroutine for every request to update the Redis rolling window counter. Under high traffic, this can lead to unbounded goroutine creation, causing excessive memory consumption, CPU scheduler contention, and potentially overwhelming the Redis server. This creates a denial-of-service vulnerability where the gateway can be crashed by a flood of requests.

This change is out of scope of current PR and requires business logic change

vladzabolotnyi:
/visor 
> The key for the in-memory rate-limiting bucket (`bucketKey`) is constructed using `session.LastUpdated`. Since `session.LastUpdated` can change on every request for certain configurations, a new in-memory bucket may be created for each request when using the default Distributed Rate Limiter (DRL) in single-node mode. This leads to unbounded memory growth as old buckets are never reused or garbage-collected, causing a memory leak that can result in a Denial of Service.

Its legacy logic which I'm not sure we're ready to change right now.

vladzabolotnyi:
All sonarqube issues are related to constant initialization at test file with avoiding value duplication and cognitive complexity in legacy logic. This changes can be fixed but not in the scope of current ticket.

cli/linter/schema.json
```diff
@@ -1610,6 +1610,12 @@
         }
       }
     },
+    "rate_limit_response_headers": {
+      "description": "defines the source for the rate limit headers",
```
andyo-tyk: Does this end up in documentation? If so should could this be:

```suggestion
      "description": "Determines the type of data that will be returned in the rate limit headers",
```

andyo-tyk (changes requested):
One suggested tweak

andyo-tyk (approved):
Go docs LGTM. This does not cover developer/engineer code review.

pvormste (approved):

Issue #802, opened 2017-06-13
It is recommended to replace the command line resolution package.


penhauer-xiao:
In the command line parameter analysis; suggested using spf13/cobra packet replacement docopt/docopt.go

mvdan:
@penhauer-xiao any specific reason why? Also note that these two libraries are vastly different. `docopt` just handles usage printing and options, while `cobra` is a CLI app library and has many more features, adding complexity.

penhauer-xiao:
@mvdan   I agree with your opinion. I think it's complicated.

mvdan:
Note that I agree that `docopt` is not particularly idiomatic in Go, but at the moment we're trying to focus on 2.4 features and fixes. We can't be refactoring the codebase forever. We might replace it at some point in the future.

penhauer-xiao:
I see, thanks. @mvdan

Issue #801, opened 2017-06-13
How do I restrict the parameters in the API like this ?


artemus717:
![image](https://user-images.githubusercontent.com/16465173/27070119-21843330-504a-11e7-8475-d5f91320feca.png)

buger:
Hello! You probably mean not restrict, but mark some of the as required?

artemus717:
Yes, i want to define every parameter in the request, then Illegal request will not be forwarded to backend servers .
For example, i define an id param as a long data type, when a request with ?id=test will be forbidden.
I want to build a waf with modsecurity in front of api gateway to filter illegal request first, and api gateway as a second protection

lonelycode:
We don't have an input validation middleware at the moment, though it's been asked for enough times - maybe we should generalise this to include XML schemas and JSON schema?

artemus717:
Glad to hear that. Is there anything i can help ?

buger:
You can help with planning phase if you want: define the scope of the issue, check what is currently missing in Tyk to make it work and etc. Is there anything we can re-use to implement this task faster, for example, jsonschema? Swagger definition by itself support this kind of attributes, but internal Tyk api definition not. 

In addition to Gateway we will need to update dashboard as well, so we need to think about the UI.

But as mentioned above

artemus717:
We want to define request input parameter information and restrict them.

1. param order

Issue #797, opened 2017-06-12
Deprecate the old rate limiter


mvdan:
Like https://github.com/TykTechnologies/tyk/issues/784, but for the old rate limiter. It's slow and it's known to be a memory hog too.

https://tyk.io/docs/deploy-tyk-premise-production/ has more information on the new rate limiter.

mvdan:
@buger remind me what the old rate limiter is again - was it `RedisRolling` or `Sentinel`? I understand that `NonTransactional` is the new and recommended one.

mvdan:
And `SessionLimiter.ForwardMessage` also has a fallback in-memory rate limiter. I'm a bit confused.

PR #7975, opened 2026-04-08
TT-16948: Upgrade JSVM engine from Otto to Goja

sedkis:
## Context

Why `Goja` instead of `Otto`?
  - [Otto](https://github.com/robertkrimen/otto) community dead.  [Goja](https://github.com/dop251/goja/commits/master/) is alive and well.
  - ES2015+ instead of ES5
  - 8-19x performance improvement
  - Virtual endpoints + all middleware hooks on goja
  - Zero breaking changes for existing JS plugin users

## What we did
- adds goja as a drop-in replacement for Otto.  It's feature-flagged on the api level.  instead of "otto", users would say "javascript" for plugin driver. no other changes needed.
- added support for response plugins and post-auth plugins to make jsvm (goja) feature parity with other plugin types
- added inline code delivery on the api definition for plugin types
- completely backwards compatible.  the same interface continues to exist.

# Integration Tests & Backwards compat
Running existing `otto` test suite against `goja` to test for backwards compatibility.  All tests passed.
`goja` is designed to be a drop-in replacement for `otto`.  no other changes needed.

| Test | Otto | Goja | What it proves |
|------|------|------|----------------|
| PreHookHeaderInjection | PASS | PASS | Pre hooks via file-based middleware |
| BodyManipulation | PASS | PASS | b64 round-trip through DoProcessRequest |
| SessionMetadataUpdate | PASS | PASS | Post-hook session metadata |
| Timeout | PASS | PASS | Infinite loop killed by timeout |
| ConfigData | PASS | PASS | spec.config_data from JS |
| Auth | PASS | PASS | Custom auth plugin (401/200) |
| PrePostStages | PASS | PASS | Pre + Post hooks combined |
| Base64 | PASS | PASS | b64enc/dec/raw directly on GojaJSVM |
| Bundle | PASS | PASS | Bundle-based middleware loading |
| PostKeyAuthHook | PASS | PASS | Post-auth JS middleware with auth flow (401/200) |
| ResponseHeaderInjection | PASS | PASS | Response hook adds header |
| ResponseBodyModification | PASS | PASS | Response hook replaces body |
| ResponseStatusCode | PASS | PASS | Response hook changes status code |
| ResponseHeaderDeletion | PASS | PASS | Response hook removes header |

# New test coverage (Plugin Studio)

| Test suite | Tests | What it covers |
|------------|-------|----------------|
| TestCollectAllMiddleware | 4 | Gathers middleware from all hook types for inline loading |
| TestLoadCustomMiddleware | 6 | Code-vs-Path routing, disabled MW skip, mixed mode, driver passthrough |
| TestCustomPlugins_InlineCodeRoundtrip | 5 | OAS Fill/ExtractTo preserves Code, PluginID, PluginHash |
| TestGoja_LoadJSPaths | 4 | Unsupported extensions, missing files, compile errors, valid loading |

# Benchmarks

## Sustained load
Goja significantly faster over sustained peak workloads. 1k rps vs 8k rps.

**Date:** 2026-03-26 | **Tool:** hey | **Mode:** Sustained 60s, concurrency 2

| Metric | Otto | Goja |
|--------|------|------|
| Requests/sec | 1,025 | **8,275** |
| Avg latency | 2.0ms | **0.2ms** |
| Total responses | 61,515 | **496,478** |

## Bursty test
Goja significantly faster in bursts. 1k rps vs 19k rps.

**Date:** 2026-03-26 | **Tool:** hey | **Mode:** 10000 requests, concurrency 10

| Metric | Otto | Goja |
|--------|------|------|
| Requests/sec | 1,083 | **19,341** |
| Total time | 9.23s | **0.52s** |

## Memory stability (30 minutes)

| Metric | Goja | Otto |
|--------|------|------|
| Throughput | **11,614 ops/s** | 3,601 ops/s |
| Heap avg | **6.4 MiB** | 20.6 MiB |
| Heap max | **17.1 MiB** | 52.2 MiB |
| GC avg pause | 203 us | 198 us |
| **Verdict** | **PASS — no leak** | **PASS — no leak** |

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-16948" title="TT-16948" target="_blank">TT-16948</a>
</summary>

|         |    |
|---------|----|
| Status  | Merge |
| Summary | [Innersource] Improve JS plugin experience and add multi-bundle support |

Generated at: 2026-07-20 15:33:22

</details>


sedkis:
/review 

> The JSVM API function `TykSetKeyData` allows a JS plugin to write session data for any API key, without restriction to the API the plugin is running for. A plugin can call `TykSetKeyData("some-other-apis-key", ...)` to overwrite session data, leading to privilege escalation, session hijacking, or denial of service against other users. The function does not validate that the target `apiKey` is associated with the current API (`h.Spec.APIID`). A similar issue exists in `TykGetKeyData` on lines 179-187, allowing unauthorized reads.

> A new `http.Client` and `http.Transport` are created for every call to `TykMakeHttpRequest` from within a JS plugin. This is a known performance anti-pattern in Go, as it prevents the reuse of TCP connections (via HTTP keep-alives). This leads to increased latency due to repeated TCP and TLS handshakes for each request and can cause file descriptor exhaustion on the gateway under high load.

> The new Goja JSVM implementation does not set a memory limit on the JavaScript runtime. A malicious or poorly-written JS plugin could allocate large amounts of memory, leading to excessive memory consumption by the Tyk Gateway process. This can cause a crash (OOM kill), resulting in a denial of service for the entire gateway. While a timeout is implemented, there is no protection against large memory allocations.

These three issues you highlighted are pre-existing issues, in-use by today's "otto" engine.  This PR upgrades the engine in place, which re-uses these existing bindings.  These limitations are noted and will be addressed in future work.  These are not new issues which were introduced.

> The new `/tyk/plugins/test` endpoint accepts `org_id` and `api_id` from the request body to construct the JavaScript execution context. It fails to validate whether the authenticated user (via `x-tyk-authorization` token) has permissions for the provided `org_id` and `api_id`. This allows a user with permissions for one API to execute code in the context of any other API on the gateway, including those in other organizations. This could lead to sensitive information disclosure, as the executed script can access the target API's `config_data`.

These fields are "assembled by dashboard, org_id and api_id always injected server-side".  they're  server-controlled now but structurally present for when the dashboard starts passing real values.  The gateway endpoint is a general-purpose contract — it accepts org_id and api_id because it needs them to build the JS execution context. The dashboard is what enforces policy about which values are allowed. Today that policy is "always sandbox." Tomorrow it could be "the API you're editing, if you have permission to it." The overwrite pattern means the gateway doesn't need to change when that policy evolves — only the dashboard handler does.

sedkis:
## Response to outstanding bot findings

### CodeQL path-traversal alerts (#797, #798, #799, #800) — won't fix, inherited trust boundary
The four alerts on `gateway/coprocess_bundle.go` (652/665/692) and `gateway/mw_js_plugin_goja.go:174` flag flows where a path string from the API definition reaches `Open` / `ReadFile`. The bundle/middleware source paths are operator-controlled at API-definition time — the same trust boundary as the existing otto code path in `gateway/mw_js_plugin.go` (`LoadJSPaths`). Anyone able to set the bundle URL or middleware path is already authorised to run arbitrary JS through that bundle; reading a file is strictly less powerful than executing one. No new exposure is introduced by this PR. Dismissing as "Won't fix — used in safe context."

### CodeQL weak-hash alerts (#801, #802) and the two Sonar Security Hotspots — fixed
- `coprocess_bundle.go:778` (`bundleSubdirName` fallback) and `mw_js_plugin_goja.go:322` (`gojaHandlerAlias`) used MD5 and SHA-1 purely for filesystem-safe / JS-identifier-safe naming — no secret, no integrity claim, no auth. Both swapped to truncated SHA-256 in the latest push. Behavioural impact: zero (the hash is only used as a short stable id).

### Visor: `mw_virtual_endpoint.go:45` (`preLoadVirtualMetaCodeGoja`) — inherited from otto
The goja branch reads `meta.FunctionSourceURI` exactly the same way the existing otto branch (`preLoadVirtualMetaCode`, same file, line 62) has done for years. `FunctionSourceURI` is part of the API definition and therefore administrator-controlled. This PR mirrors the existing behaviour for the new engine and does not change the trust boundary. Path hardening would need to apply to both engines and is tracked separately.

### Sonar new-code coverage (78.3% → ≥80%)
Added validation tests covering the `pluginTestHandler` 400-path branches (malformed body, missing code, invalid hook_type, missing org_id, response hook without response payload) to lift new-code coverage over the gate.

gateway/server.go
sedkis, line 1146: The claim's stated impact is wrong: the goja loader never reads mwPaths (it re-derives the  Code-vs-Path decision itself in api_loader.go), so the extra Response paths are dead data with no runtime cost or behavioral bug. At most it's a cosmetic inconsistency affecting only the legacy Otto branch.

sedkis:
/review The claim's stated impact is wrong: the goja loader never reads mwPaths (it re-derives the Code-vs-Path decision itself in api_loader.go), so the extra Response paths are dead data with no runtime cost or behavioral bug. At most it's a cosmetic inconsistency affecting only the legacy Otto branch.

kofoworola (approved):

PR #7995, opened 2026-04-13
[TT-16966][TT-16971] Fix ci/tests/specs Error: Cannot find module 'tslib' and policies schema

MFCaballero:
Addresses https://tyktech.atlassian.net/browse/TT-16966 and https://tyktech.atlassian.net/browse/TT-16971

Durable guard (follow-up)
To prevent any future transitive drift from silently breaking this job:

Commit a package-lock.json in ci/tests/specs/.

Change ci/tests/specs/Taskfile.yml to use npm ci instead of npm install for the tests task. npm ci installs exactly what's in the lockfile and fails loudly on drift.

Optional: file an upstream issue on apideck-libraries/portman asking them to move tslib from devDependencies (or wherever it sits) into dependencies, since the compiled dist/index.js requires it at runtime.


lghiur (approved):

MFCaballero:
LGTM

PR #8047, opened 2026-04-16
[TT-16342] revert: undo incorrect OTel scenario update (PR #8041)

buger:
## Summary
Reverts PR #8041 which incorrectly updated OTel test scenarios to use NEW attribute names on release-5.12.

release-5.12 uses otelhttp v0.49.0 (pinned via go.mod replace), which emits OLD names:
- `http.method` (not `http.request.method`)
- `http.status_code` (not `http.response.status_code`)
- `http.user_agent` (not `user_agent.original`)
- `http.wrote_bytes` (not `http.response.body.size`)

The scenario update was only correct for master (otelhttp v0.65.0).

### DIFF
--- a/ci/tests/tracing/scenarios/tyk_body_size_200.yaml
+++ b/ci/tests/tracing/scenarios/tyk_body_size_200.yaml
@@ -12,7 +12,7 @@ spec:
         - key: Content-Type
           value: application/json
   specs:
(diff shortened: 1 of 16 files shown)


buger:
Closing: Go 1.25 updated the OTel HTTP instrumentation library on release-5.12.1, so scenarios need the NEW-style attribute names (e.g. `http.request.method` instead of `http.method`). This revert was incorrect. Correct fix applied in PR #8059.

PR #8048, opened 2026-04-16
[TT-16342] revert: undo incorrect OTel scenario update (PR #8042)

buger:
## Summary
Reverts PR #8042 which incorrectly updated OTel test scenarios to use NEW attribute names on release-5.12.1.

release-5.12.1 uses otelhttp v0.60.0 which defaults to OLD attribute names unless OTEL_SEMCONV_STABILITY_OPT_IN is set.


buger:
Closing: Go 1.25 updated the OTel HTTP instrumentation library on release-5.12.1, so scenarios need the NEW-style attribute names (e.g. `http.request.method` instead of `http.method`). This revert was incorrect. Correct fix applied in PR #8059.

PR #8248, opened 2026-05-26
[TT-7099] [Customer FR] Send info and debug messages to stdout not stderr

shults:
## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-7099" title="TT-7099" target="_blank">TT-7099</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | Configurable sinks for application logs [Gateway] |

Generated at: 2026-08-12 11:10:16

</details>


shults:
/visor ask 

`rpc.Log` uses the same global reference as tyk logger. I suppose that your theory regarding removing `rpc.Log = log` is wrong.

log/logger.go
```diff
@@ -0,0 +1,265 @@
...
+		lgr.setupOnce.Do(func(executed bool) {
+			if executed {
+				tmpLogs.Forward(inner)
+			} else {
+				// send logs to emergency logger in case if Fatal was called
+				tmpLogs.Forward(emergencyLogger)
+			}
+		})
+
+		lgr.ExitFunc(code)
```
MFCaballero: I think this would lead to a recursive stack overflow, as `lgr.ExitFunc(code)` resolves to this exact closure. `Logger` embeds `*logrus.Logger` anonymously and doesn't declare its own `ExitFunc` field, so `lgr.ExitFunc` is just promoted through to `lgr.innerLogger.ExitFunc`

log/logger.go
```diff
@@ -0,0 +1,265 @@
...
+		panic("invokeOnce.Must has to be executed only once")
+	}
+
+	s.value = true
+
+	fn()
+}
+
+func (s *invokeOnce) Do(fn func(executed bool)) {
+	s.mu.Lock()
```
MFCaballero: I think this would deadlock, as `s.mu` here is the same mutex `MustOnce` (line 189) holds for the entire duration of `Setup()`'s callback. 
If `f(&builder)` (line 113, i.e. `gw.setupLogger`) calls `Fatal`, it triggers `ExitFunc` → `lgr.setupOnce.Do(...)`, which tries to lock `s.mu` again from the same goroutine that already holds it.

log/builder.go
```diff
@@ -0,0 +1,139 @@
...
+	b.discardLogger(dest.innerLogger)
+	b.applyHooksAndSinks(dest.innerLogger)
+
+	if b.propagate {
+		b.discardLogger(logrusStdLog)
+		b.applyHooksAndSinks(logrusStdLog)
+	}
+}
+
+func (b *Builder) discardLogger(target *logrus.Logger) {
```
MFCaballero, line 89: I think this would default a stock gateway to trace-level logging, as `level` falls back to `target.Level` (line 93) whenever `b.level` is nil — and it stays at whatever `New()` set it to (`TraceLevel`), since nothing calls `WithLevel` in that case.
`tyk.conf.example` has no `log_level` key, and the schema allows `""` with no default, so on a completely stock config with no env overrides, `gateway/server.go:1553`'s `logLevel != ""` check is false and `builder.WithLevel(...)` is never called

log/logger.go
shults: It's a typo. I had to call `lgr.OsExit`

log/logger.go
shults: I'll write a test. to simmulate situation you describe. And let you know sooner.

log/builder.go
shults: Thank you for your response. 90 and 93 lines can be removed. 
`discardLogger`  modifies the root loggers, Root logger is not responsible for filtering by severity. Therefore the root logger has to bypass all the logs to it's children (sinks). And each sink filters logs by it's own.


If log_level in a config file is empty "", then info log level is used as defafult value. 
- "" -> Info
- "brbrbr" -> Fatal error and exit(1)
- "warn" -> parses properly

MFCaballero (approved):

config/log_format.go
```diff
@@ -0,0 +1,84 @@
...
+		lf.sinks = sinks
+		return nil
+	} else {
+		errs = append(errs, err)
+	}
+
+	return fmt.Errorf("invalid log_format: must be a string, null, or an array of tyklog.SinkConfig: %w", errors.Join(errs...))
+}
+
+func (lf LogFormat) MarshalJSON() ([]byte, error) {
```
MaciekMis, line 73: this should be a pointer receiver to keep LogFormat consistent

config/log_format.go
```diff
@@ -0,0 +1,84 @@
...
+	switch lf.formatType {
+	case LogFormatString:
+		return json.Marshal(lf.format)
+	case LogFormatSinks:
+		return json.Marshal(lf.sinks)
+	case LogFormatUndefined:
+		fallthrough
```
MaciekMis: this is counterintuitive: when the format is undefined, we should return the default behavior (a null byte array), so this fallthrough into default is redundant. It also creates a potential maintenance problem - if another case is added here in the future, the existing fallthrough could cause unexpected behavior

config/log_format.go
```diff
@@ -0,0 +1,84 @@
...
+	default:
+		return []byte("null"), nil
+	}
+}
+
+type LogFormatType int
```
MaciekMis: this type, along with the consts, should be at the top of the file

gateway/server.go
```diff
@@ -1544,8 +1545,100 @@ func (gw *Gateway) reloadURLStructure(done func()) {
...
+		}
+	} else {
+		level = logrus.InfoLevel
+	}
+
+	return level
+}
+
+func (gw *Gateway) buildSinkWithDest(
+	case_ string,
```
MaciekMis: why is there an underscore in this var name?

log/field_map.go
```diff
@@ -27,3 +31,24 @@ func (f FieldMap) Resolve(field string) string {
...
+	var m map[string]string
+	if err := json.Unmarshal(data, &m); err != nil {
+		return err
+	}
+
+	f.fields = m
+	return nil
+}
+
+func (f FieldMap) MarshalJSON() ([]byte, error) {
```
MaciekMis, line 49: this should be a pointer receiver to keep FieldMap consistent

log/log.go
```diff
@@ -134,20 +142,35 @@ func newFormatterLegacy() logrus.Formatter {
 	}
 }
 
-func IsLegacyFormatter(formatter logrus.Formatter) bool {
-	textFormatter, ok := formatter.(*logrus.TextFormatter)
+type Format string
```
MaciekMis: type and consts should be at the top of the file

log/output.go
```diff
@@ -0,0 +1,40 @@
...
+
+const (
+	OutputStdout Output = "stdout"
+	OutputStderr Output = "stderr"
+)
+
+func MakeOutput(output Output, opts json.RawMessage) (io.Writer, error) {
+	if fn, ok := outputRegistry[output]; !ok {
+		return nil, fmt.Errorf("unknown output %q for logger", output)
+	} else {
```
MaciekMis: redundant else

gateway/server.go
```diff
@@ -1544,8 +1545,100 @@ func (gw *Gateway) reloadURLStructure(done func()) {
...
+	format tyklog.Format,
+	cfg *config.Config,
+) ([]tyklog.SinkerExtended, error) {
+
+	level := gw.extractLogLevel(cfg)
+	sink := tyklog.NewSink(
+		os.Stderr,
+		tyklog.NewFormatter(format),
+		tyklog.NewAcceptorRange(level, logrus.FatalLevel),
```
MaciekMis: shouldn't we use the lowest possible log level `PanicLevel` here?

log/builder.go
```diff
@@ -0,0 +1,139 @@
...
+	var logrusStdLog = logrus.StandardLogger()
+
+	if b.stdLog != nil {
+		logrusStdLog = b.stdLog
+	}
+	b.applyHooksToRawLog()
+
+	if b.level != nil {
+		for _, sinker := range b.sinkers {
+			sinker.SetAcceptor(NewAcceptorRange(*b.level, logrus.FatalLevel))
```
MaciekMis: shouldn't we use the lowest possible log level `PanicLevel` here?

MaciekMis (changes requested):

config/log_format.go
shults: why?

config/log_format.go
shults: empty struct or underfined -> null
string -> "some string value"
slice of sinks -> [{sinks}]
---
if another case is added justadd case before undefined

gateway/server.go
shults: why?

gateway/server.go
shults: case is a keword; it cant be used as a argument name

log/log.go
shults: I prefer this approach. Are there any conventions in tyk regarding file organization?
---
Type1 (exported)
const/vars of Type1
PublicMethods of Type1
privateMethods of Type1
...
TypeN (exported)
const/vars of TypeN
PublicMethods of TypeN
privateMethods of TypeN
---
typePrivate1 (exported)
const/vars of typePrivate1
methods of typePrivate1
---

log/output.go
shults: why?

go.mod
```diff
@@ -630,3 +630,6 @@ require (
 replace go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.54.0 => go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.49.0
 
 replace github.com/getkin/kin-openapi => github.com/TykTechnologies/kin-openapi v0.92.1-0.20260211103127-0004a8367058
+
+// todo: remove temporary pump replace after merging pump
+replace github.com/TykTechnologies/tyk-pump => github.com/TykTechnologies/tyk-pump v0.0.0-20260728202309-958a99ec60f1
```
shults: todo: resolve

config/log_format.go
shults: nope. its intentionally.

MaciekMis (approved):

PR #8254, opened 2026-05-27
[TT-16259] [Security] Unable to delete Session that uses a quota and is under heavy load [Plan A]

shults:
## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-16259" title="TT-16259" target="_blank">TT-16259</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | [Security] Unable to delete Session that uses a quota and is under heavy load |

Generated at: 2026-06-05 11:28:49

</details>


shults:
/visor ask  "SetRawKeyEx" is implemented the same way as "SetRawKey" for MDCBStorage; If someone uses this MDCBStorage as session storage, it will fail anyway. So it is not a problem at all from my point of view.

shults:
/visor ask `APISpec.Init` session store was not used; there was no reason of precence that parameter

go.mod
```diff
@@ -630,3 +630,5 @@ require (
 replace go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.54.0 => go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.49.0
 
 replace github.com/getkin/kin-openapi => github.com/TykTechnologies/kin-openapi v0.92.1-0.20260211103127-0004a8367058
+
+replace github.com/TykTechnologies/storage => github.com/TykTechnologies/storage v0.0.0-20260601101308-a5fb7c875ace // todo: remove before merge to master
```
radkrawczyk: this changes are of course temporary changes, to make pipeline running. It will be removed when new release is generated on storage.

gateway/session_manager.go
```diff
@@ -496,6 +496,7 @@ func (l *SessionLimiter) RedisQuotaExceeded(
 
 	increment := func() bool {
 		var res *redis.IntCmd
+		// todo: change to lua script, it does not works like author suppose
```
MaciekMis, line 499: please fix the todo

storage/storage.go
kofoworola, line 1: Why was this refactored?

user/session.go
```diff
@@ -348,6 +348,11 @@ type SessionState struct {
 	// modified holds the hint if a session has been modified for update.
 	// use Touch() to set it, and IsModified() to get it.
 	modified bool
+
+	// isRestored
+	// The flag holds information if session was restored or not.
+	// Use MarkAsRestored() to mark session as restored, and IsRestored() to get it.
+	isRestored bool
```
kofoworola, line 355: If the session is serialized and transmitted over the wire, e.g MDCB, this value will be lost

kofoworola (changes requested):

storage/storage.go
shults: storage.Handler violates ISP

user/session.go
shults: Show me the place where the session is being sent over the wire.
I don't need this value to be saved.

shults: Lifecycle looks like this: fetch -> modify -> save.
It could be a problem if the process looks like the next way:  fetch -> modify -> serialize -> send to other process -> deserialize -> save.
So it would not be a problem.
It works the same way as a `modified` field.

storage/storage.go
kofoworola: It is beyond the scope of the ticket

shults: If I extend `storage.Handler` interface with an additional method. It would provide more coupling.

tyk-identity-broker depends on tyk, and I should implement that method all over the tyk-ecosystem, which is inefficient.

Here is proof of my words. 
- https://github.com/TykTechnologies/tyk-identity-broker/pull/481

user/session.go
kofoworola: Example, the RPCHandler is used by MDCB workers
https://github.com/TykTechnologies/tyk/blob/d72a70e43686e0caf1da1f6aa154c90f6e977043/gateway/rpc_storage_handler.go#L365

storage/storage.go
shults: https://github.com/TykTechnologies/tyk/pull/8254/changes#diff-78e768b2719ac9f70038499f847de2843db20d8ca21a963ea63b82010d711039R138-R163

I need those single-method interfaces to reduce coupling.
tyk-identity-broker -> uses github.com/TykTechnologies/tyk/certs

user/session.go
shults: @kofoworola, here is only one place where this flag is being used.
If under this `b.store ` lays `RPCStorageHandler`. It will call RPC server part on tyk-sink (MDCB). 
MDCB receives a serialized session and just saves it into the storage (probably redis-cluster) like a list of bytes.

```
// UpdateSession updates the session state in the storage engine
func (b *DefaultSessionManager) UpdateSession(
	keyName string,
	session *user.SessionState,
	resetTTLTo int64,
	hashed bool,
) error {

	defer b.clearCacheForKey(keyName, hashed)

	v, err := json.Marshal(session)
	if err != nil {
		log.Error("Error marshalling session for sync update")
		return err
	}

	if !session.IsRestored() {
		if hashed {
			keyName = b.store.GetKeyPrefix() + keyName
			err = b.store.SetRawKey(keyName, string(v), resetTTLTo)
		} else {
			err = b.store.SetKey(keyName, string(v), resetTTLTo)
		}
	} else {
		if hashed {
			keyName = b.store.GetKeyPrefix() + keyName
			err = b.store.SetRawKeyEx(keyName, string(v), resetTTLTo)
		} else {
			err = b.store.SetKeyEx(keyName, string(v), resetTTLTo)
		}
	}

	return err
}
```

MDCB - RPC server side. I don't need that flag.
```
func SetKeyFunc(clientID string, input interface{}) error {
	ibd, err := extractInboundDataForSetOperation(clientID, input)
	if err == nil {
		storage.Store.SetKey(ibd.KeyName, ibd.SessionState, ibd.Timeout)
	}
	return err
}
```

kofoworola (approved):

MaciekMis (approved):

PR #5756, opened 2023-11-08
Syse 288/final phase deprecate old pipelines

ermirizio:
This PR looks to address the following jira issues:

[SYSE-289](https://tyktech.atlassian.net/browse/SYSE-289): Final Phase: Configure X-Ray to Accept test matrix results

[SYSE-290](https://tyktech.atlassian.net/browse/SYSE-290): Final Phase: Configure Slack notifications to tag the PR author

[SYSE-291](https://tyktech.atlassian.net/browse/SYSE-291): Final Phase: Configure PR comments to be updated in place to avoid spam

[SYSE-289]: https://tyktech.atlassian.net/browse/SYSE-289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SYSE-290]: https://tyktech.atlassian.net/browse/SYSE-290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SYSE-291]: https://tyktech.atlassian.net/browse/SYSE-291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #8268, opened 2026-06-02
[TT-17312] Gateway: client-IdP registry for JWT key & scope resolution

mativm02:
## Summary

Adds an in-memory **`IdPRegistry`** (`gateway/idp_registry.go`) that joins an API to its client IdPs **at request time**, as a sibling dataset of the API definition travelling the same transports. The JWT middleware consults it **only as a last fallback** after the existing manual API-def branches, so DCR IdPs resolve without ever entering the API def, manual config stays authoritative, and a registry change refreshes just the registry — no router rebuild, no fleet-wide reload.

Pairs with the dashboard side (`TykTechnologies/tyk-analytics#5756`, the `/system/clientidps` feed + `NoticeClientIdPChanged` signal) and the MDCB `GetClientIdPs` RPC handler (separate ticket).

## What changed

- **`gateway/idp_registry.go`** (new) — registry with debounced (~100ms) `Refresh`, mode-branching `fetch` (direct-Dashboard HTTP `/system/clientidps`, RPC `GetClientIdPs`, file-deferred), build-then-swap `rebuild` with a **segment-aware backstop** keyed on `apisByID`, and `scopeToPolicyMapForRequest` (manual wins on collision, binding fills gaps).
- **`gateway/mw_jwt.go`** — `secretFromIdPRegistry` last-fallback probe in `getSecretToVerifySignature`; URL-keyed `fetchJWKSKey` (avoids `getSecretFromURL`'s `JWTSource`-gated cache miss); manual-first scope resolution in `processCentralisedJWT`; `hasRegistryBinding` routing gate in `ProcessRequest`.
- **`ctx/ctx.go` + `gateway/api.go`** — `MatchedIdPBinding` context key + helpers. Per-request binding lives on the **request context, never the shared middleware struct** (one `JWTMiddleware` per API, `ProcessRequest` runs concurrently).
- **`gateway/rpc_storage_handler.go` + `internal/policy/{rpc.go,store_mock.gen.go}`** — `GetClientIdPs` on `RPCDataLoader` + impl + regenerated/handwritten mocks.
- **`gateway/redis_signals.go`** — `NoticeClientIdPChanged` (registry-only refresh).
- **`gateway/server.go`** — `idpRegistry` field, init in `NewGateway`, `doRefresh` after `loadGlobalApps` in `DoReloadWithError`.

## Key properties

- **Manual config is authoritative** — appended as the last branch; an admin's `JWTSource`/`JWTJwksURIs`/scope map always wins. For manual APIs the probe never runs (cost zero, byte-identical behaviour).
- **`iss` is an unverified hint** — only selects which JWKS to try; signature verification + `validateIssuer` still run afterward.
- **Per-binding `scope_to_policy`** makes scope-name collisions between IdPs structurally impossible.
- Works for **both classic and OAS** APIs.
- HTTP fetch goes through `executeDashboardRequestWithRecovery`, inheriting nonce-failure re-registration and client config like `FromDashboardService`.

## Testing

- New unit/integration tests (all green, incl. `-race`): decoder envelope/bare-array tolerance, scope-merge collision/fill, segment backstop, build-then-swap, RPC + Dashboard fetch (incl. nonce capture), ctx roundtrip, shutdown no-op, and the three acceptance criteria — AC1 registry resolves key+scope (200), AC2 manual `JWTSource` wins, AC3 `NoticeClientIdPChanged` refreshes registry without an API reload.
- `go build ./...` + `go vet` clean; existing `TestJWT*`/`TestJWKS*`/`TestJWTScopeToPolicyMapping`/`TestManagementNodeRedisEvents` pass (no regressions).
- E2E framework available (dashboard-direct, portal DCR, MDCB) covering the full round-trip.

## Acceptance criteria

- [x] **AC1** — JWT API with no `JWTSource`/`JWTJwksURIs` + a registry binding returns 200; token scope maps to the bound policy.
- [x] **AC2** — Setting `JWTSource` bypasses the registry; manual wins on collision, binding fills absent scopes.
- [x] **AC3** — A registry change refreshes the registry with no "Reloading endpoints"/`loadGlobalApps`.

## Notes / follow-ups

- The `GetClientIdPs` RPC call on edge nodes requires the MDCB-side handler (parallel ticket); they ship coordinated. Against an older sink the call errors and is swallowed (previous snapshot kept).
- Code review flagged JWKS fan-out for tokens with absent/non-matching `iss` (walk-all per spec). Bounded by binding count (typically 1) and on par with existing manual-JWKS refetch behaviour; negative-caching is a possible future hardening.

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-17312" title="TT-17312" target="_blank">TT-17312</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | Gateway: client IdP registry + JWT integration |

Generated at: 2026-06-05 14:28:38

</details>

### DIFF
--- a/gateway/api.go
+++ b/gateway/api.go
@@ -3299,9 +3299,9 @@ func ctxGetCacheOptions(r *http.Request) *cacheOptions {
 }
 
 // ctxSetMatchedBinding stores the client-IdP registry binding matched for this
-// request. JWTMiddleware is a single shared instance per API (api_loader.go:403)
-// and its ProcessRequest runs concurrently, so per-request state must live on
-// the context, not the middleware struct.
+// request. JWTMiddleware is a single shared instance per API and its
+// ProcessRequest runs concurrently, so per-request state must live on the
+// context, not the middleware struct.
 func ctxSetMatchedBinding(r *http.Request, b *Binding) {
 	setCtxValue(r, ctx.MatchedIdPBinding, b)
 }
--- a/gateway/api.go
+++ b/gateway/api.go
@@ -3298,6 +3298,23 @@ func ctxGetCacheOptions(r *http.Request) *cacheOptions {
 	return key
 }
 
+// ctxSetMatchedBinding stores the client-IdP registry binding matched for this
+// request. JWTMiddleware is a single shared instance per API (api_loader.go:403)
+// and its ProcessRequest runs concurrently, so per-request state must live on
+// the context, not the middleware struct.
+func ctxSetMatchedBinding(r *http.Request, b *Binding) {
+	setCtxValue(r, ctx.MatchedIdPBinding, b)
+}
+
+// ctxGetMatchedBinding returns the matched binding for this request, or nil.
+func ctxGetMatchedBinding(r *http.Request) *Binding {
+	b, ok := r.Context().Value(ctx.MatchedIdPBinding).(*Binding)
+	if !ok {
+		return nil
+	}
+	return b
+}
+
 func ctxGetSession(r *http.Request) *user.SessionState {
 	return ctx.GetSession(r)
 }
--- a/gateway/idp_registry.go
+++ b/gateway/idp_registry.go
@@ -196,7 +196,7 @@ func (r *IdPRegistry) fetchFromDashboard() ([]IdP, error) {
 	endpoint := r.gw.buildDashboardConnStr("/system/clientidps")
 
 	buildReq := func() (*http.Request, error) {
-		req, err := http.NewRequest(http.MethodGet, endpoint, nil)
+		req, err := http.NewRequestWithContext(r.gw.ctx, http.MethodGet, endpoint, nil)
 		if err != nil {
 			return nil, fmt.Errorf("build client-idps request: %w", err)
 		}
--- a/gateway/idp_registry.go
+++ b/gateway/idp_registry.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/TykTechnologies/tyk/header"
+	"github.com/TykTechnologies/tyk/rpc"
 )
 
 const idpRefreshDebounce = 100 * time.Millisecond
@@ -242,8 +243,14 @@ func (r *IdPRegistry) fetchFromDashboard() ([]IdP, error) {
 }
 
 // fetchFromRPC pulls the registry over RPC (MDCB/edge). The payload is a bare
-// JSON array; an empty string is a no-op.
+// JSON array; an empty string is a no-op. In emergency mode (MDCB unreachable)
+// it restores the registry from the Redis backup instead — mirroring how APIs
+// and policies recover — and on a successful sync it refreshes that backup.
 func (r *IdPRegistry) fetchFromRPC() ([]IdP, error) {
+	if rpc.IsEmergencyMode() {
+		return r.gw.LoadIdPsFromRPCBackup()
+	}
+
 	conf := r.gw.GetConfig()
 	var tags []string
 	if conf.DBAppConfOptions.NodeIsSegmented {
@@ -252,6 +259,9 @@ func (r *IdPRegistry) fetchFromRPC() ([]IdP, error) {
 	payload := r.rpcLoaderFn().GetClientIdPs(conf.SlaveOptions.RPCKey, tags)
 	if payload == "" {
 		return nil, nil
+	}
+	if err := r.gw.saveRPCIdPsBackup(payload); err != nil {
+		log.WithError(err).Warning("Failed to back up client-IdP registry to Redis")
 	}
 	return unmarshalIdPs([]byte(payload))
 }
--- a/gateway/idp_registry.go
+++ b/gateway/idp_registry.go
@@ -1,0 +1,301 @@
+package gateway
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/TykTechnologies/tyk/header"
+)
+
+const idpRefreshDebounce = 100 * time.Millisecond
+
+// IdP is a client auth IdP stored beside the API definition, joined to APIs by
+// api_id. The wire shape is identical across the direct-Dashboard HTTP feed
+// (GET /system/clientidps) and the MDCB/edge RPC feed (GetClientIdPs); the JSON
+// tags mirror the dashboard's model.ClientIdP.
+type IdP struct {
+	ID          string                  `json:"client_idp_id"`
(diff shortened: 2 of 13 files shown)


mativm02:
### Validated end-to-end against the e2e framework (T1)

Built the gateway from this branch + the dashboard from tyk-analytics #5756 (`TT-17310`) and ran the **dashboard Client IdP** e2e (real Keycloak DCR, no portal). Both classic **and** OAS JWT APIs — each with empty `jwt_source` and empty `jwt_scope_to_policy_mapping` — resolved their signing key from the registry's Keycloak JWKS and mapped token scopes to the bound policies:

```
classic R1 (quota 1/4, rate 1/2) → 200
classic R2 (quota 2/4, rate 2/2) → 200
classic R3 (rate exceeded)       → 429
OAS     R1 (quota 3/4, rate 1/2) → 200
OAS     R2 (quota 4/4, rate 2/2) → 200
OAS     R3 (rate exceeded)       → 429
classic R4 (shared quota 4/4)    → 403
=> DASHBOARD CLIENT IDP TEST PASSED
```

This exercises the full chain (registry refresh via `/system/clientidps` → `secretFromIdPRegistry` → JWKS fetch → scope→policy via the binding → product ACL + plan rate/quota partitions) for **both API formats**.

### Bug caught & fixed by the e2e

The OAS run initially crashed the gateway with a nil-pointer panic in `getScopeClaimNameOAS` (`mw_jwt.go`): a registry-only OAS API has no JWT block in its OAS definition, so `OAS.GetJWTConfiguration()` (and its `.Scopes`) is `nil`. My unit `AC1` used a classic API and missed it. Fixed with a nil guard (falls back to the default `scope` claim) and added a regression test `TestGetScopeClaimNameOAS_NilJWTConfig`. Amended into the branch.

mativm02:
### Addressed Visor findings (all Visor checks now green)

Folded into the branch:

- **Security (SSRF, `fetchJWKSKey`)** — `jwks_uri` is admin-controlled config (Dashboard/MDCB), the same trust domain + `ExternalHTTPClientFactory` as the existing `JWTSource`/`JWTJwksURIs` fetches, so it inherits their SSRF posture rather than adding a new one. Added an explicit `http(s)`-scheme guard (rejects `file://`/`gopher://`/… vectors) and documented the boundary. Host/IP egress blocking is intentionally **not** applied — DCR IdPs (e.g. internal Keycloak) legitimately live on private networks.
- **Security (log hygiene, `fetchFromDashboard`)** — no longer logs the raw 403 response body; returns a static `403 Forbidden` error.
- **Performance (thundering herd, `fetchJWKSKey`)** — concurrent JWKS fetches for the same `(API, URL)` are now coalesced via `singleflight` (`golang.org/x/sync` — already a direct dep), so a burst on a cold/expired cache triggers one network fetch.
- **Performance (memory, `fetchFromDashboard`)** — stream-decode the feed with `json.NewDecoder` instead of buffering the whole body via `io.ReadAll`.

New tests: `TestFetchJWKSKey_RejectsNonHTTPScheme`; strengthened the 403 test to assert the body isn't leaked. Existing JWT/JWKS suite + the T1 e2e key-resolution path (AC1/AC2) still pass.

Note: the `s1_scanner` (SentinelOne CNS) failure is the repo-wide dependency-vulnerability scan — this PR changes no `go.mod`/`go.sum`/vendor, so it's pre-existing on `master` and not addressable here.

gateway/idp_registry.go
```diff
@@ -0,0 +1,287 @@
...
+		// detail. The node recovers via executeDashboardRequestWithRecovery.
+		return nil, errors.New("client-idps login failure (403 Forbidden)")
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("client-idps dashboard error: status %d", resp.StatusCode)
+	}
+
+	// Stream-decode straight off the response body to avoid buffering the whole
+	// payload in memory (the registry can be large in big deployments).
+	var env struct {
```
tbuchaillot: Why we aren't using `unmarshalIdPs` func here?

mativm02: `unmarshalIdPs` only gives back the IdPs, but here we also need the `Nonce` from the envelope to keep the node's nonce in sync — and we decode straight off the body (no `io.ReadAll`) for the perf check. So it can't reuse it directly. I did pull the shared shape into an `idpFeedEnvelope` type so both use it now.

tbuchaillot (approved):

MFCaballero (approved):
🚀

gateway/api.go
```diff
@@ -3298,6 +3298,23 @@ func ctxGetCacheOptions(r *http.Request) *cacheOptions {
 	return key
 }
 
+// ctxSetMatchedBinding stores the client-IdP registry binding matched for this
+// request. JWTMiddleware is a single shared instance per API (api_loader.go:403)
```
sredxny: I would remove this reference

gateway/idp_registry.go
```diff
@@ -0,0 +1,311 @@
...
+// nonce-failure re-registration and client config as FromDashboardService — the
+// registry is a first-class /system/* citizen, not a fragile passive reader.
+// The response is a NodeResponseOK envelope {"Status","Message":[...],"Nonce"};
+// the Nonce is captured into gw.ServiceNonce so the node stays in sync — the
+// dashboard rotates the node's stored nonce on every /system/* call.
+func (r *IdPRegistry) fetchFromDashboard() ([]IdP, error) {
+	endpoint := r.gw.buildDashboardConnStr("/system/clientidps")
+
+	buildReq := func() (*http.Request, error) {
+		req, err := http.NewRequest(http.MethodGet, endpoint, nil)
```
sredxny: I think here we can take advantage of the gw context:

`http.NewRequestWithContext(r.gw.ctx, endpoint...)`

gateway/rpc_backup_handlers.go
```diff
@@ -152,6 +154,80 @@ func (gw *Gateway) decompressPolicyBackup(decrypted string) (string, error) {
 	return gw.decompressBackup(decrypted, backupKindPolicies)
 }
 
+// The client-IdP registry payload is small (one record per org IdP), so it is
+// stored uncompressed; decompressBackup still auto-detects on load.
+func (gw *Gateway) compressClientIdPBackup(list string) string {
+	return gw.compressBackup(list, false, backupKindClientIdPs)
+}
```
sredxny: am wondering why to have a compress function that doesnt compress, if we are not going to compress this maybe the best is to fully remove it?

sredxny (changes requested):
left some comments, we can discuss it

sredxny (approved):

PR #8276, opened 2026-06-03
[TT-17356] Join the MCP agent's trace from params._meta and propagate upstream

lghiur:
MCP-native agents carry the W3C trace context inside the JSON-RPC body (params._meta, SEP-414), not the traceparent header. Tyk only read the header, so for those requests it started a fresh, disconnected trace — its spans, access logs, and analytics couldn't be correlated to the agent's journey.

This makes Tyk a first-class participant in an MCP request's trace:

Read/join the trace context from configurable sources; when the header carried none, install it so every span/log/audit later in the request inherits the agent's trace_id. Reconcile (never fork) when both channels are present.
Write the current trace context into the outbound body so the MCP server joins too — the body-channel parallel of the traceparent header Tyk already injects for ordinary upstreams. Runs automatically when tracing is on (no per-API flag).
Span labels: mcp.method.name, mcp.{tool,resource,prompt}.name (keyed by primitive type), mcp.trace_source.
Config: opentelemetry.mcp_trace_context.read_sources (ordered, first-match-wins; default [{header}, {body, params._meta}]). Applies to all MCP primitive types; no-op when tracing is off or the body is non-MCP/malformed. Covered by unit tests (internal/mcp, internal/otel) and gateway integration tests.
Screenshot 2026-06-03 at 20 03 31

## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-17356" title="TT-17356" target="_blank">TT-17356</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | MCP trace-context bridge, join the agent's trace via params._meta (read/write) |

Generated at: 2026-06-04 12:35:36

</details>


internal/otel/config.go
```diff
@@ -76,6 +76,11 @@ type OpenTelemetry struct {
 
 	// Metrics holds the OpenTelemetry metrics configuration.
 	Metrics MetricsConfig `json:"metrics"`
+
+	// MCPTraceContext configures where Tyk reads an MCP request's W3C trace
+	// context from (ordered, first-match-wins). Omitted ⇒ the canonical
+	// [{header}, {body, path: params._meta}] default.
+	MCPTraceContext MCPTraceContextConfig `json:"mcp_trace_context"`
```
tbuchaillot: This should be part of `traceConfig` structure. In 5.13 we introduced this division in configuration for metrics and traces so new tracing related config configs should live inside `opentelemetry.traces` instead of root level `opentelemetry`

internal/otel/mcp/tracecontext.go
```diff
@@ -0,0 +1,183 @@
+package mcp
```
tbuchaillot, line 1: nitpick - shouldn't this live in otel package or otel/mcp since it's purely related to opentelemetry?

internal/otel/mcp_trace_context.go
```diff
@@ -0,0 +1,196 @@
...
+		ctx := extractTraceContext(r.Context(), res.Context)
+		*r = *r.WithContext(ctx)
+	}
+	return res.Source
+}
+
+// extractTraceContext returns ctx with tc installed as the active (remote) span
+// context via the W3C propagator — the same propagator Tyk uses for inbound
+// header extraction, so the joined trace is byte-identical to a header-carried one.
+func extractTraceContext(ctx context.Context, tc mcp.TraceContext) context.Context {
```
tbuchaillot: I'm wondering if this functions shouldn't be part of https://github.com/TykTechnologies/opentelemetry/tree/main/trace package so we avoid importing/using `go.opentelemetry.io/otel` directly here

tbuchaillot (changes requested):
Please have a look at the config comment and add some tracing test case for this

tbuchaillot (approved):

edsonmichaque (approved):
LGTM

PR #8286, opened 2026-06-08
[releng master] gromit: sync templates

olamilekan000:
Auto-generated from gromit templates by policy sync.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -174,7 +174,7 @@ jobs:
           platforms: ${{ github.event_name == 'pull_request' && 'linux/amd64' || 'linux/amd64,linux/arm64' }}
           file: ci/Dockerfile.distroless
           provenance: mode=max
-          push: true
+          outputs: type=image,push=true,compression=gzip,force-compression=true,oci-mediatypes=true
           cache-from: type=gha
           cache-to: type=gha,mode=max
           tags: ${{ steps.ci_metadata_ee.outputs.tags }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -183,6 +183,20 @@ jobs:
...
             BASE_IMAGE=tykio/dhi-busybox:1.37-fips
             NONROOT_CHOWN=true
+      - name: Verify compression format for ee CI image
+        if: ${{ matrix.golang_cross == '1.25-bullseye' }}
+        run: |
+          # Resolve the first tag for verification
+          TAG=$(echo "${{ steps.ci_metadata_ee.outputs.tags }}" | tr ',' '\n' | head -n1)
+          echo "Inspecting layer compression for tag: $TAG"
+
+          docker buildx imagetools inspect --raw -- "$TAG" > manifest.json
```
rafalgolarz, line 193: The new verification steps may not actually catch zstd-compressed layers for multi-platform images. `docker buildx imagetools inspect --raw "$TAG"` can return the top-level image index/manifest list for a multi-arch tag, while the layer mediaType values live in the per-platform child manifests. In that case, `grep -q "tar+zstd" manifest.json` can pass even if one of the platform manifests contains zstd layers.

I think this should inspect each platform manifest digest from the index, for example `"$TAG@$digest"`, and validate the `.layers[].mediaType` values there. It would also be stronger to assert the expected gzip media types explicitly, rather than only checking that tar+zstd is absent, since the current success message says all layers are gzip-compressed.

PR #8278, opened 2026-06-04
[TT-12339] Rename OAS enforceTimeout field `timeout` to `duration`

jay-deshmukh:
The sub-second enforced-timeout field added for TT-12339 was serialized as `timeout`. Rename it to `duration` to match the intended API contract:

    x-tyk-api-gateway:
      middleware:
        operations:
          <operationId>:
            enforceTimeout:
              enabled: true
              duration: 100ms

Changes:
- apidef/oas/middleware.go: rename EnforceTimeout.Timeout -> Duration and its json/bson tags `timeout` -> `duration`; update the Value deprecation note and the Fill/ExtractTo mappings.
- apidef/oas/schema/x-tyk-api-gateway.json and .strict.json: rename the X-Tyk-EnforceTimeout `timeout` property to `duration`.
- Update references in middleware_test.go, linter_test.go and operation_test.go.

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-12339" title="TT-12339" target="_blank">TT-12339</a>
</summary>

|         |    |
|---------|----|
| Status  | Ready for Testing |
| Summary | [BE] Implement sub-second timeout for the Enforced Timeout middleware |

Generated at: 2026-06-05 10:07:56

</details>

PR #8335, opened 2026-06-19
[TT-17515] Rename Tyk Classic hard_timeouts duration wire field

mativm02:
## Summary

Renames the granular enforced-timeout field on the **Tyk Classic** API definition from `timeout_duration` to `duration`, so it matches the Tyk OAS field name (`enforceTimeout.duration`). Closes the Classic half of the `duration` rename agreed during TT-12339 (OAS was done in #8278).

Ticket: **TT-17515** (parent TT-12339).

## What changed

Minimal **wire-tag rename only** — the Go field name `HardTimeoutMeta.TimeoutDuration` is unchanged, so the enforcement path (`gateway/reverse_proxy.go`) and the negative-value validator (`apidef/validator.go`) are untouched.

- `apidef/api_definitions.go` — `HardTimeoutMeta.TimeoutDuration` json/bson tag `timeout_duration` → `duration`
- `apidef/schema.json` — `hard_timeouts[]` item property renamed (format pattern preserved)
- `apidef/oas/middleware.go` — doc-comment reference updated
- `apidef/api_definitions_test.go` — new `TestHardTimeoutMeta_DurationWireTag` (marshal/unmarshal/omitempty)

## Out of scope (by product decision)

- **No round-up for Classic.** Setting only `duration` is *not* folded into the legacy whole-second `timeout` field. This was a deliberate omission (agreed with product) — a Classic API created with only `duration` has no enforced timeout on an older Gateway unless `timeout` is also set. This is an accepted, documented limitation; users wanting sub-second precision are steered to Tyk OAS.
- **No new backend validation.** Negatives still rejected (existing behaviour); empty/`0`/`0ms` remain valid (no timeout).

## Compatibility

- **Backwards compatible:** the legacy whole-second `timeout` field is untouched; existing APIs keep working on upgrade and downgrade.
- **No JSON alias for the old key.** The pre-release `timeout_duration` name was **never released**, so there is nothing to migrate — any definition still carrying that key would deserialize the granular value as zero. This rename is safe on that premise.

## Testing

- `go test ./apidef/ -run 'TestSchema$|HardTimeoutMeta_DurationWireTag' -count=1` ✅ (incl. struct-vs-embedded-schema validation)
- `go test ./gateway/ -run TestTimeoutPrioritization -count=1` ✅ (sub-second + legacy enforcement)

## Downstream

Dashboard support is in tyk-analytics-ui (Classic designer note) + a `tyk` dependency bump in tyk-analytics that lands after this merges.

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -231,7 +231,7 @@ type HardTimeoutMeta struct {
 	Method   string `bson:"method" json:"method"`
 	// Deprecated: Use TimeoutDuration instead.
 	TimeOut         int                      `bson:"timeout" json:"timeout"`
-	TimeoutDuration tyktime.ReadableDuration `bson:"timeout_duration,omitempty" json:"timeout_duration,omitempty"`
+	TimeoutDuration tyktime.ReadableDuration `bson:"duration,omitempty" json:"duration,omitempty"`
 }
 
 type TrackEndpointMeta struct {
--- a/apidef/api_definitions_test.go
+++ b/apidef/api_definitions_test.go
@@ -54,6 +54,42 @@ func TestAPIDefinition_JsonRpcVersion(t *testing.T) {
 	})
 }
 
+// TestHardTimeoutMeta_DurationWireTag locks in the TT-17515 rename of the
+// granular enforced-timeout field from `timeout_duration` to `duration`.
+// The Go field name (TimeoutDuration) is unchanged; only the JSON/BSON tags moved.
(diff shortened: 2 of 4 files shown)


buger:
Isn't it backward incompatible change which breaks legacy API definitions which had this field before?

mativm02:
> Isn't it backward incompatible change which breaks legacy API definitions which had this field before?

@buger not really, as the old naming wasn't released yet, it's only present in `master`

buger (approved):

PR #8401, opened 2026-06-30
[TT-5893] Error Log: "Couldn't find a matching policy for scope item" should be optional

shults:
## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-5893" title="TT-5893" target="_blank">TT-5893</a>
</summary>

|         |    |
|---------|----|
| Status  | In Test |
| Summary | Error Log: "Couldn't find a matching policy for scope item" should be optional |

Generated at: 2026-07-10 11:22:33

</details>

### DIFF
--- a/gateway/mw_jwt.go
+++ b/gateway/mw_jwt.go
@@ -10,7 +10,9 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"maps"
 	"net/http"
+	"slices"
 	"strings"
 	"time"
 
@@ -834,23 +836,27 @@ func getScopeFromClaim(claims jwt.MapClaims, scopeClaimName string) []string {
 }
 
 func mapScopeToPolicies(mapping map[string]string, scope []string) []string {
-	polIDs := []string{}
-
 	// add all policies matched from scope-policy mapping
-	policiesToApply := map[string]bool{}
+	policiesToApplySet := make(map[string]struct{}, len(mapping))
+	unmatchedScopes := make(map[string]struct{}, len(mapping))
+
 	for _, scopeItem := range scope {
 		if policyID, ok := mapping[scopeItem]; ok {
-			policiesToApply[policyID] = true
+			policiesToApplySet[policyID] = struct{}{}
 			log.Debugf("Found a matching policy for scope item: %s", scopeItem)
 		} else {
-			log.Errorf("Couldn't find a matching policy for scope item: %s", scopeItem)
+			unmatchedScopes[scopeItem] = struct{}{}
 		}
 	}
-	for id := range policiesToApply {
-		polIDs = append(polIDs, id)
+
+	// https://tyktech.atlassian.net/browse/TT-5893
+	if len(policiesToApplySet) > 0 {
+		for scopeItem := range unmatchedScopes {
+			log.Debugf("Couldn't find a matching policy for scope item: %s", scopeItem)
+		}
 	}
 
-	return polIDs
+	return slices.Collect(maps.Keys(policiesToApplySet))
 }
 
 func (k *JWTMiddleware) getOAuthClientIDFromClaim(claims jwt.MapClaims) string {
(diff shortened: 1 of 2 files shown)


gateway/mw_jwt.go
```diff
@@ -834,23 +836,27 @@ func getScopeFromClaim(claims jwt.MapClaims, scopeClaimName string) []string {
 }
 
 func mapScopeToPolicies(mapping map[string]string, scope []string) []string {
-	polIDs := []string{}
-
 	// add all policies matched from scope-policy mapping
-	policiesToApply := map[string]bool{}
+	policiesToApplySet := make(map[string]struct{}, len(mapping))
+	unmatchedScopes := make(map[string]struct{}, len(mapping))
```
MFCaballero: I think it would be better to set capacity bounded at len(scope).. if admin configured mappings is too large extra allocation could slow down requests in the hot path

gateway/mw_jwt_test.go
```diff
@@ -5365,3 +5366,74 @@ func TestJWTPostExpiry(t *testing.T) {
...
+
+			res := mapScopeToPolicies(map[string]string{
+				"scope1": "policy1",
+				"scope2": "policy2",
+			}, []string{"scope1", "scope2"})
+
+			assert.True(t, lo.EveryBy(hook.AllEntries(), func(item *logrus.Entry) bool {
+				return item.Level == logrus.DebugLevel && strings.HasPrefix(item.Message, "Found a matching policy for scope item")
+			}))
+			assert.Equal(t, []string{"policy1", "policy2"}, res)
```
MFCaballero: I think this assertion would cause flakyness as Go map iteration order is deliberately randomized by the runtime

MFCaballero (changes requested):

gateway/mw_jwt.go
shults: oK.

shults: I can be implemented in three wrong cases:
1. make(map[string]struct{}, len(scopes))  -> scopes can be duplicated, and the system will allocate more memory than needed
2. make(map[string]struct{}, len(mapping)) -> configured mappings can be huge, and the system will allocate more memory than needed; leaves a window for a DOS attack
3. make(map[string]struct{}) -> default allocation; size can be to low and realocation needed;
4. omit preallocation, and allocate only when the first insert occurs;
Which one of the offered solutions seems the best?

MaciekMis (approved):

gateway/mw_jwt.go
MFCaballero: I'd say 3 as it's the simplest.. benchmarks to back it up:
  | Scenario | Strategy | ns/op | B/op | allocs/op |
  |---|---|---|---|---|
  | Small mapping (20), no dup | SizeByMapping (current) | 1248 | 984 | 4 |
  | | SizeByScope | 328 | 256 | 2 |
  | | DefaultCap | 301 | 256 | 2 |
  | | NoPrealloc | 280 | 256 | 2 |
  | Small mapping, heavy dup (50 items→5 distinct) | SizeByMapping (current) | 2566 | 984 | 4 |
  | | SizeByScope | 2566 | 1880 | 4 |
  | | DefaultCap | 1878 | 256 | 2 |
  | | NoPrealloc | 1757 | 256 | 2 |
  | Huge mapping (5000), no dup | SizeByMapping (current) | 30451 | 218480 | 18 |
  | | SizeByScope | 257 | 256 | 2 |
  | | DefaultCap | 243 | 256 | 2 |
  | | NoPrealloc | 272 | 256 | 2 |
  | Huge mapping, heavy dup | SizeByMapping (current) | 34636 | 218480 | 18 |
  | | SizeByScope | 2590 | 1880 | 4 |
  | | DefaultCap | 1868 | 256 | 2 |
  | | NoPrealloc | 1931 | 256 | 2 |

MFCaballero: 4 is ok too 👍

MFCaballero (approved):

PR #8456, opened 2026-07-06
[TT-16259] [Security] Unable to delete Session that uses a quota and is under heavy load

shults:
## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-16259" title="TT-16259" target="_blank">TT-16259</a>
</summary>

|         |    |
|---------|----|
| Status  | In Test |
| Summary | [Security] Unable to delete Session that uses a quota and is under heavy load |

Generated at: 2026-07-15 05:19:14

</details>


gateway/auth_manager.go
```diff
@@ -212,6 +225,7 @@ func (b *DefaultSessionManager) SessionDetail(orgID string, keyName string, hash
 		return user.SessionState{}, false
 	}
 	session.KeyID = keyId
+	session.MarkAsRestored()
```
MFCaballero, line 228: Just one finding here: this unconditionally marks every loaded session restored, including reads from the RPC/master store, which breaks local caching on MDCB edges. On an edge, CheckSessionAndIdentityForValidKey falls back to AuthManager.SessionDetail (the RPC store) for a key not in local Redis; that read lands here and sets isRestored=true. The deferred write goes to local Redis (GlobalSessionManager), so the IsRestored() branch above (:140) picks SET … XX which no-ops because the key isn't local yet. The key is served (HTTP 200) but never cached, so every request re-pulls it over RPC. 
**Suggested Fix:** only mark restored when the read came from the local write-target store; or, at the middleware, mark the recreation as a create in CheckSessionAndIdentityForValidKey before session.Touch(). I'd lean to the second as it names the intent (a create) right where it happens.

shults: @MFCaballero
Thank you for your response.  
I've checked the code. Sessions are stored in KV storage with the `apikey-` prefix.
Getting a key with `apikey-` is impossible because setting this kind of key is impossible (take a look at SetKeyFunc function from tyk-sink repository). 
Therefore, the fallback scenario is not reproducible. 

```
func SetKeyFunc(clientID string, input interface{}) error {
        /// ...
	if strings.Contains(ibd.KeyName, "apikey-") {
		if len(ibd.SessionState) > 0 {
			log.Warning("Unauthorized upstream write detected!")
			return errors.New("write dissallowed for API Tokens")
		}
	}

	if strings.Contains(ibd.KeyName, "cert-") {
		log.Warning("Unauthorized upstream write detected!")
		return errors.New("write dissallowed for API Certificates")
	}
        // ...
}
```

2. And the second argument, MdcbStorage, always writes to the local (Redis) storage.
```
func (m MdcbStorage) SetKey(key string, content string, TTL int64) error {
	// only set the value locally as rpc writtes is not allowed
	errLocal := m.local.SetKey(key, content, TTL)

	if errLocal != nil {
		return errors.New("cannot save key in local")
	}

	return nil
}
```

MFCaballero: Hi @shults thanks for taking a look. However SetKeyFunc is the master-side RPC SetKey handler — it blocks an edge pushing a session upstream over RPC. My finding doesn't involve any upstream apikey- write. The key is created normally on the master (/keys), the edge reads it via the RPC fallback in CheckSessionAndIdentityForValidKey, and the local cache write-back (GlobalSessionManager → local edge Redis, no RPC) is what no-ops.
MarkAsRestored() at auth_manager.go:228 is set from the RPC/master read, but it drives SET…XX on the local write target at :140. On a cold key the local SET…XX no-ops, so the pulled key is served (200) but never persisted — every request re-pulls over RPC. apikey reads also aren't cached at read time (processResourceByType only caches OAuth/certs), so this write-back was the only thing persisting a cold-pulled key.

I'm attaching a testing framework for different scenarios PTAL in particular session-delete-race/scripts/test-edge-cold-pull-persist.sh which reproduces this

[session-delete-race.zip](https://github.com/user-attachments/files/29797450/session-delete-race.zip)

shults: @MFCaballero 
Take a look, please, at the last two commits.
It seems that they resolve the issue you described. 
If the session is not able to fetch from the local redis (GlobalSessionManager) then it can be extracted from Remote (RPC, LDAP, whatever)
After fetching it from the remote service, it can be saved into the local Redis.

Thank you for the set of scripts.

shults:
/visor ask 
1. pipeline improvement (INCR) is out of scope for this bug. I'll create the other task to solve this issue. 
2. ErrNotImplement is ok, session (api-key) can't be changed via RPC.
3. RaceCondition existed long before my changes, and I think it's worth adding a new task to solve that race condition.

shults:
/visor ask Even if I replace SetKey with SetkeyNx, the race condition will stay. The session can be removed, for example, between fetching from one system and putting into the other. A distributed transaction in this case is not possible; I don't know how to implement it in the actual codebase;

gateway/middleware.go
```diff
@@ -697,12 +700,23 @@ func (t *BaseMiddleware) CheckSessionAndIdentityForValidKey(originalKey string,
...
 
+		{
+			// insert new session
+			clonedSession := session.Clone()
+			clonedSession.MarkAsNew()
+
+			lifetime := clonedSession.Lifetime(t.Spec.GetSessionLifetimeRespectsKeyExpiration(), t.Spec.SessionLifetime, t.Gw.GetConfig().ForceGlobalSessionLifetime, t.Gw.GetConfig().GlobalSessionLifetime)
+			if err := t.Gw.GlobalSessionManager.UpdateSession(key, &clonedSession, lifetime, false); err != nil {
+				t.Logger().WithError(err).Error("Cant copy session from AuthManager into GlobalSessionManager")
+				return user.SessionState{}, false
```
MFCaballero, line 716: raised issue by visor is actually true here and reproduced by attached script test-delete-race-multigw.sh 
[test-delete-race-multigw.sh](https://github.com/user-attachments/files/29886908/test-delete-race-multigw.sh)
Suggested fix: Only seed the local copy when AuthManager reads from a genuinely different store than GlobalSessionManager (a real cross-store cold pull, e.g. MDCB). If they're the same store, AuthManager just found a key GlobalSessionManager already has — nothing to seed, and doing it anyway is an unconditional write eachable during ordinary delete+recreate churn (see harness test-delete-race-multigw.sh: 3/60 cycles reproduced this reliably).
PTAL 🙏

gateway/api_definition.go
```diff
@@ -1999,7 +1991,12 @@ func (a APIDefinitionLoader) getExtendedPathSpecs(apiVersionDef apidef.VersionIn
 	return combinedPath, whiteListEnabled
 }
 
-func (a *APISpec) Init(authStore, sessionStore, healthStore, orgStore storage.Handler) {
+func (a *APISpec) Init(
```
MaciekMis: group parameters with the same type

gateway/session_manager.go
```diff
@@ -496,6 +496,7 @@ func (l *SessionLimiter) RedisQuotaExceeded(
 
 	increment := func() bool {
 		var res *redis.IntCmd
+		// todo: change to lua script, it does not works like author suppose
```
MaciekMis, line 499: left todo in the code

log/hook.go
MaciekMis, line 1: Why this log implementation exists in this PR?

gateway/middleware.go
```diff
@@ -717,13 +731,12 @@ func (t *BaseMiddleware) CheckSessionAndIdentityForValidKey(originalKey string,
 
 		t.Logger().Debug("Lifetime is: ", session.Lifetime(t.Spec.GetSessionLifetimeRespectsKeyExpiration(), t.Spec.SessionLifetime, t.Gw.GetConfig().ForceGlobalSessionLifetime, t.Gw.GetConfig().GlobalSessionLifetime))
 
-		session.Touch()
```
MaciekMis, line 720: Removal of Touch() breaks session TTL refresh. 
From oel:
`Previously, when a session was loaded from the AuthManager (e.g., after local cache expiration), session.Touch() marked it as modified. This caused UpdateRequestSession to update the session in Redis at the end of the request, effectively refreshing its TTL (acting as an idle timeout).

By removing session.Touch(), sessions without quotas are no longer marked as modified, so they are never updated in Redis. This means their TTL is never refreshed, causing them to expire absolutely rather than acting as an idle timeout.`

gateway/session_manager.go
shults: ? 
IMO it's out of scope. A separate task has to be created.

gateway/middleware.go
shults: The session is inserted a couple of blocks before. Therefore, Touch can be removed.

shults: Hovever, I'll revert it.

log/hook.go
shults: Leftovers from the first reverted version.

MaciekMis (approved):

MFCaballero (approved):

PR #8492, opened 2026-07-16
Disambiguate the /tyk/keys/create OpenAPI summary from POST /tyk/keys

sharadregoti:
## Summary

tyk-docs added a per-spec Mintlify directory split to stop cross-spec collisions (different specs sharing a generated docs page). That surfaced a within-spec duplicate here: `POST /tyk/keys` and `POST /tyk/keys/create` both had the summary "Create a key.", producing an identical generated docs page URL for two different operations - one silently overwrites the other.

## Investigation

Traced both handlers to confirm this isn't a copy-paste bug or a legacy-vs-current situation before touching anything:

- `POST /tyk/keys` -> `keyHandler` -> `handleAddOrUpdate`, which calls `gw.generateToken(...)` to produce the key.
- `POST /tyk/keys/create` -> `createKeyHandler`, which calls `gw.keyGen.GenerateAuthKey(...)`.

Both auto-generate the key the same way, and neither is marked `deprecated`. They're genuinely two long-standing, equivalent routes for the same operation - not something to unilaterally declare one "canonical" over the other.

## Fix

Updated `/tyk/keys/create`'s summary and description to say it's an alternative, explicitly-named endpoint equivalent to `POST /tyk/keys`, rather than asserting a preference between them. `POST /tyk/keys` is unchanged.

## Test plan

- [x] `yaml.safe_load` confirms the file is still valid YAML
- [x] Confirmed only one `summary: Create a key.` remains; the other now reads "(alternative endpoint)"
- [ ] Swagger/redocly CI lint on this PR

### DIFF
--- a/swagger.yml
+++ b/swagger.yml
@@ -2738,7 +2738,6 @@ paths:
   /tyk/keys/create:
     post:
       description: Creates a new API key. Automatically generates a key ID (unless overridden by the 'certificate' field) and an 'hmac_secret' (if 'hmac_enabled' is true). For Basic Auth hashing, MCP validation, or update parameters (suppress_reset, hashed), use POST /tyk/keys instead.
-        as an alternative, explicitly-named endpoint.
       operationId: createKey
       requestBody:
         content:
--- a/swagger.yml
+++ b/swagger.yml
@@ -2827,7 +2827,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/ApiStatusMessage'
           description: Malformed body.
-      summary: Create a key (alternative endpoint).
+      summary: Create a key (auto-generates Key ID and HMAC secret)
       tags:
       - Keys
   /tyk/keys/policy/{keyID}:
--- a/swagger.yml
+++ b/swagger.yml
@@ -2737,7 +2737,7 @@ paths:
       - Keys
   /tyk/keys/create:
     post:
-      description: Create a key. Functionally equivalent to `POST /tyk/keys`, provided
+      description: Creates a new API key. Automatically generates a key ID (unless overridden by the 'certificate' field) and an 'hmac_secret' (if 'hmac_enabled' is true). For Basic Auth hashing, MCP validation, or update parameters (suppress_reset, hashed), use POST /tyk/keys instead.
         as an alternative, explicitly-named endpoint.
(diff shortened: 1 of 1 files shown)


swagger.yml
```diff
@@ -2737,7 +2737,8 @@ paths:
       - Keys
   /tyk/keys/create:
     post:
-      description: Create a key.
+      description: Create a key. Functionally equivalent to `POST /tyk/keys`, provided
```
andyo-tyk:
```suggestion
      description: Creates a new API key. Automatically generates a key ID (unless overridden by the 'certificate' field) and an 'hmac_secret' (if 'hmac_enabled' is true). For Basic Auth hashing, MCP validation, or update parameters (suppress_reset, hashed), use POST /tyk/keys instead.
```

swagger.yml
```diff
@@ -2826,7 +2827,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/ApiStatusMessage'
           description: Malformed body.
-      summary: Create a key.
+      summary: Create a key (alternative endpoint).
```
andyo-tyk:
```suggestion
      summary: Create a key (auto-generates Key ID and HMAC secret)
```

andyo-tyk (changes requested):
My review with Oel suggests there's a difference, this isn't just a duplicate.

swagger.yml
```diff
@@ -2737,7 +2737,8 @@ paths:
       - Keys
   /tyk/keys/create:
     post:
-      description: Create a key.
+      description: Creates a new API key. Automatically generates a key ID (unless overridden by the 'certificate' field) and an 'hmac_secret' (if 'hmac_enabled' is true). For Basic Auth hashing, MCP validation, or update parameters (suppress_reset, hashed), use POST /tyk/keys instead.
+        as an alternative, explicitly-named endpoint.
```
sharadregoti:
```suggestion
```

sharadregoti:
Closing in favor of a different DX-2380 fix: Mintlify's per-operation `x-mint.href` override lets us give colliding operations an explicit unique URL without renaming any summary text or restructuring directories. Redoing this collision as an `x-mint.href` addition instead - see the DX-2380 tracking discussion for details.

PR #8499, opened 2026-07-20
[TT-16039] Add configurable client authentication method for upstream OAuth

mativm02:
## Description

Upstream OAuth currently relies on the oauth2 library's auto-detect for client authentication at the IdP token endpoint: it always tries HTTP Basic Auth first and falls back to sending the credentials in the request body. Against body-only IdPs this produces a guaranteed failing 400 before every successful token fetch — noisy logs, false alerts, and an extra round-trip (and because the OAuth config is rebuilt per fetch, the probe never "sticks").

This PR adds an optional `method` field on the shared `ClientAuthData` (so it applies to both the Client Credentials and Password/ROPC grants, in both the classic and OAS formats):

| Value | Behaviour |
|---|---|
| empty / unset | **Unchanged default** — auto-detect (header first, body fallback) |
| `client_secret_basic` | Credentials in the Authorization header only, single request, no fallback |
| `client_secret_post` | Credentials in the request body only, single request, no fallback |

The values and constants are reused from the OAuth2 Token Exchange feature (`OAuth2ClientAuthBasic`/`OAuth2ClientAuthPost`, relocated from `apidef/oas` to `apidef` so both formats can share them; `apidef/oas` keeps aliases so existing consumers are untouched).

**Deliberate divergence from Token Exchange:** empty does NOT default to `client_secret_basic`. Existing upstream OAuth users pointed at body-only IdPs depend on the fallback, so empty must preserve auto-detect. This is called out in the field docs.

### Changes

- `apidef`: `Method` on `ClientAuthData` (classic `method`), constants relocation, new `RuleUpstreamAuth` validation (`ErrInvalidUpstreamOAuthClientAuthMethod`).
- `apidef/oas`: `method` on OAS `ClientAuthData` with Fill/ExtractTo wiring for both grant blocks; schema enum for `upstream.authentication.oauth.{clientCredentials,password}.method` (strict schema regenerated via `build.sh`).
- `ee/middleware/upstreamoauth`: `authStyle()` maps the field to `oauth2.AuthStyle` (`AuthStyleInHeader` / `AuthStyleInParams` / `AuthStyleAutoDetect`); unrecognised values (possible for file-loaded definitions that bypass validation) log a warning and fall back to auto-detect.
- **Drive-by fix (call-out):** `UpstreamAuth.requestSigningExtractTo` reset `u.BasicAuth` instead of `u.RequestSigning` in its cleanup defer, so `ExtractTo` permanently mutated the OAS struct. Covered by the new round-trip test.

### Tests

- Validator table tests (invalid values on both grants, valid values pass).
- OAS round-trip test for `method` on both grant blocks.
- Integration tests (`-tags ee`, real gateway + httptest IdPs with request counting):
  - `client_secret_post` vs body-only IdP → exactly 1 token request, no Basic probe (both grants);
  - `client_secret_basic` vs header-only IdP → exactly 1 header request;
  - empty method vs body-only IdP → pins today's probe (400 then 200) for back-compat;
  - `client_secret_post` vs header-only IdP → single rejected request, no silent fallback, error surfaced.
- Also manually verified end-to-end against a local Dashboard + mock IdP (single body-creds request with the field set; legacy 400→200 probe with it unset).

### Follow-ups (separate PRs)

- Dashboard: designer dropdown for both grants (webclient PR) + `go.mod` tyk pin bump after this merges.
- tyk-docs update.
- MDCB needs no change (field travels inside the standard API definition).

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-16039" title="TT-16039" target="_blank">TT-16039</a>
</summary>

|         |    |
|---------|----|
| Status  | Merge |
| Summary | Upstream OAuth: avoid duplicate token requests and log noise |

Generated at: 2026-08-04 11:05:01

</details>

### DIFF
--- a/apidef/api_definitions.go
+++ b/apidef/api_definitions.go
@@ -137,6 +137,13 @@ const (
 	// OAuthAuthorizationTypePassword is the authorization type for password flow.
 	OAuthAuthorizationTypePassword = "password"
 
+	// OAuth2ClientAuthBasic and OAuth2ClientAuthPost are the client authentication
+	// methods defined by RFC 6749 Section 2.3.1 (the `token_endpoint_auth_method`
+	// values IdPs publish). Shared by upstream OAuth (ClientAuthData.Method) and
+	// the OAS OAuth2 token-exchange feature (oas.OAuth2ClientAuth.Method).
+	OAuth2ClientAuthBasic = "client_secret_basic"
+	OAuth2ClientAuthPost  = "client_secret_post"
+
 	// JSON-RPC protocol versions
 	JsonRPC20 = "2.0"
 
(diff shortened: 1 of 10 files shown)


vladzabolotnyi (approved):
Well done!

PR #8532, opened 2026-07-31
TT-17848: return 401 (not 400) on missing auth for PRM-enabled APIs

edsonmichaque:
## TT-17848

https://tyktech.atlassian.net/browse/TT-17848

### Problem

A request to a JWT- or external-OAuth-protected API with **no `Authorization` header** returns **HTTP 400 Bad Request**. Per RFC 6750 a missing credential is **401 Unauthorized** (400 is reserved for a *malformed* request), and the MCP authorization spec makes 401 a hard requirement: MCP clients begin their OAuth / Protected-Resource-Metadata (PRM) discovery flow **only on a 401 (or 403) challenge**.

So a PRM-enabled MCP API behind Tyk can never be authenticated — the client hits it, gets 400, and gives up, even though the gateway already attaches a correct `WWW-Authenticate: Bearer realm="tyk", resource_metadata=...` header. This was verified against source across the MCP client ecosystem (VS Code, Cursor, the embedded Tyk MCP Inspector, and the official SDKs) — none act on a 400.

Both middlewares are also internally inconsistent: they already return **401** for invalid/expired tokens, yet **400** when the credential is simply *missing*. Auth-key already returns 401.

### Fix

Return **401** instead of 400 on missing auth, **gated on PRM being enabled** (`Spec.GetPRMConfig() != nil`). Only PRM/MCP APIs change; every other JWT / external-OAuth API keeps its long-standing 400, avoiding a global breaking change. No `error=` parameter is added, matching RFC 6750 §3.1.

The same defect existed in both auth middlewares, so the fix is applied to both:
- `gateway/mw_jwt.go`
- `gateway/mw_external_oauth.go`

### Tests

Updated the PRM-enabled missing-auth assertions in `gateway/mw_protected_resource_test.go` to expect 401 (the JWT `WWW-Authenticate` case and step 1 of the full MCP discovery flow). The no-PRM cases continue to assert 400.

### Scope / risk

Blast radius is bounded to PRM-enabled specs. Non-PRM APIs are unchanged (they still return 400 on missing auth).

### Spec basis

- RFC 6750 §3.1 (Bearer Token Usage)
- RFC 9728 §5.1 (Protected Resource Metadata)
- MCP Authorization spec (2025-06-18)

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-17848" title="TT-17848" target="_blank">TT-17848</a>
</summary>

|         |    |
|---------|----|
| Status  | Merge |
| Summary | JWT middleware returns 400 (not 401) on missing Authorization, breaking the MCP/PRM OAuth flow |

Generated at: 2026-08-06 13:00:31

</details>


lghiur (approved):

edsonmichaque:
Addressed the PRM-location precedence feedback (test both old and new config).

The missing-auth 401 gate previously checked only the deprecated top-level `protectedResourceMetadata` (`GetPRMConfig`). Since PRM moved under the `oauth2` security scheme — and `GetOAuth2PRMConfig` **wins** over the top-level block — an API configured PRM only the new way would still return 400.

Fix (`52ec7fff`):
- Added `APISpec.IsPRMEnabled()` — checks the oauth2-scheme block first (priority), falls back to the top-level block. This matches the precedence `PRMMiddleware` already uses; `EnabledForSpec()` now delegates to it.
- Both `mw_jwt` and `mw_external_oauth` gate on `IsPRMEnabled()`.

Tests cover **both** locations:
- 401 challenge for PRM under the oauth2 scheme (new location).
- 401 challenge for PRM under the top-level block (existing).
- `TestIsPRMEnabled` unit test: no auth block, top-level PRM, oauth2-scheme PRM, both-disabled.

PR #8562, opened 2026-08-10
[TT-17576] JWT sessions on MDCB data plane gateways retain stale rate limit values after policy update via Dashboard UI

shults:
## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-17576" title="TT-17576" target="_blank">TT-17576</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | JWT sessions on MDCB data plane gateways retain stale rate limit values after policy update via Dashboard UI |

Generated at: 2026-09-16 17:02:35

</details>


pvormste:
From Visor:
```
When applying non-ACL policy partitions (like rate limiting), the code incorrectly overwrites the session's existing endpoint permissions with the new policy's endpoint list. If the new policy has ACL disabled, its endpoint list will be empty, which erases the user's access to all specific API paths they previously had, causing a denial of service for those endpoints.
💡 Suggestion
Remove the line that overwrites the endpoints. Endpoint permissions should only be managed when a policy's ACL partition is being applied, which is handled elsewhere. This block should only update the limits (rate, quota, etc.) as intended.
```
and
```
The `Endpoints` field, which defines endpoint-specific access control lists, is being updated based on the `didRateLimit` flag. Endpoint permissions are part of the ACL and should not be modified when only rate limits are being applied, especially within a block that executes when the main ACL partition (`!applyState.didAcl[k]`) is disabled. This could lead to unintended and incorrect permission changes.
💡 Suggestion
The line `existingAR.Endpoints = v.Endpoints` should be removed from the `if applyState.didRateLimit[k]` block. Endpoint permissions should only be handled when the ACL partition is explicitly being applied.
```

If visor's claims are correct, this would introduce a new bug - can you please double-check?

shults:
@pvormste Checked Visor's claim about `existingAR.Endpoints = v.Endpoints` and it is a false positive. `AccessDefinition.Endpoints` is not ACL. In `user/session.go` it is documented as "configuration for endpoint rate limiting" (path plus per-method rate limits). ACL lives in `AllowedURLs`, `Versions` and the GraphQL/MCP restriction fields, none of which this block touches. The existing code a few lines below already treats `Endpoints` as rate-limit data: it sets `v.Endpoints = nil` when the rate-limit partition was *not* applied. The new line is the mirror image for the case where it *was* applied, so it is consistent with what was there.

To be transparent about what else changes: the `len(rights) == 0` check is replaced by a counter of policies actually found. Before, a key whose policies are all non-ACL partitions was rejected with "key has no valid policies to be applied" (that check dates from TT-12710). After, such a key is accepted and its own access list is kept with refreshed limits. I will call this out in the PR description as an intentional behaviour change.

Separately, I found `Apply` hard to review in its current shape (four parallel `did*` maps keyed by an unnamed string, a 60-line loop with two unrelated branches). I pushed a pure, behaviour-preserving refactor stacked on this branch as #8708, and I would like to keep going in that direction, splitting `Apply` into named phases until each one reads on its own, with the existing tests as the guard. Happy to do that as follow-up PRs so this fix can go first.

PR #8640, opened 2026-08-21
[TT-17873] proxy_default_timeout overrides enforced timeout middleware but should only be applied when enforced timeout is unset

shults:
## Problem / Task
Added unit tests for the `internal/roundtrippers` package and an integration test in `gateway/reverse_proxy_test.go` to verify correct timeout inheritance.

## Changes
- Added `internal/roundtrippers/timeout_test.go`
- Added `internal/roundtrippers/header_timeout_test.go`
- Added `internal/roundtrippers/roundtrippers_test.go`
- Added `TestTimeoutCachingBugFix` in `gateway/reverse_proxy_test.go`

## Testing
- Ran unit tests for `internal/roundtrippers`
- Verified integration test logic for `TestTimeoutCachingBugFix`

---
## Technical Solution (What was done)

1. Decoupled Timeouts from Shared Transport: 
   We stopped setting ResponseHeaderTimeout directly on the shared http.Transport. This eliminates the cache poisoning issue and ensures that the transport can be safely shared across endpoints with different timeout requirements.

2. Introduced RoundTripper Middleware Pattern (Splitting the Base RoundTripper): 
   We refactored the monolithic TykRoundTripper pipeline into a clean chain of responsibility (middlewares) using a decorator pattern. We split the timeout logic into separate middlewares:
   • Timeout Middleware: Manages the total request lifecycle timeout using context.WithTimeout.
   • HeadersTimeout Middleware: Manages the Time-To-First-Byte (TTFB) / header timeout.

   Justification (Why we did it this way): Without splitting the base RoundTripper into separate middlewares, it would be impossible to achieve local, per-request disabling of the HeaderTimeout. The standard Go http.Transport shares the ResponseHeaderTimeout across the entire connection pool. By extracting this logic into a dynamic middleware, we can now evaluate conditions on the fly without affecting other requests sharing the same transport.

3. Dynamic Middleware Disabling for Internal Loops:
   We introduced conditional execution for middlewares (e.g., WithHeadersTimeoutDisabledFn(isInternalLoop)). This dynamically disables the HeadersTimeout for Tyk's internal loopback requests (tyk:// and tyk-internal://), preventing false 500 errors caused by premature context cancellation during in-memory GraphQL queries.

4. Thread-Safe Header Timeout Tracking: 
   Implemented a headersTimeoutTracker using time.AfterFunc (instead of goroutines with select) and atomic flags. It hooks into httptrace.ClientTrace (WroteRequest, GotFirstResponseByte) to accurately measure header timeouts without leaking goroutines.

5. Safe Resource Cleanup & Streaming Support: 
   • Added proper panic recovery (invokeRtWithCancel) to prevent resource leaks.
   • Fixed premature context cancellation that broke streaming (SSE, gRPC, GraphQL) by wrapping the response body in a custom ReadWriteCloser. The context is now safely canceled only when the body is explicitly closed.

6. Final Transport as a Simple Selector:
   The original TykRoundTripper at the end of the pipeline was stripped down to its core responsibility: simply routing the request to either the standard http.Transport or h2ctransport (used for HTTP/2 Cleartext / gRPC).

7. Testing & Documentation: 
   • Added comprehensive unit tests for the new RoundTripper middlewares.
   • Added an integration test (TestTimeoutCachingBugFix) to verify that timeout inheritance works correctly and the caching bug is resolved.
   • Added inline code comments explaining why ResponseHeaderTimeout must not be set on the shared transport to prevent future regressions.

---
Requested by: U08QWRUM11Q
Slack thread: https://tyktech.slack.com/archives/D0AF7G82CSV/p1787221640049799

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-17873" title="TT-17873" target="_blank">TT-17873</a>
</summary>

|         |    |
|---------|----|
| Status  | Ready for Testing |
| Summary | [regression] Enforced Timeout endpoint middleware not working as expected  |

Generated at: 2026-08-28 08:38:02

</details>

### DIFF
--- a/gateway/reverse_proxy.go
+++ b/gateway/reverse_proxy.go
@@ -1275,6 +1275,12 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 
 	requestTimeout, isRequestTimeoutEnforced := p.GetEnforcedTimeoutSettings(p.TykAPISpec, outreq)
 
+	if isRequestTimeoutEnforced {
+		timeoutContext, cancel := context.WithCancel(outreq.Context())
+		defer cancel()
+		outreq = outreq.WithContext(timeoutContext)
+	}
+
 	// create HTTP transport
 	createTransport := p.TykAPISpec.HTTPTransport == nil
 
--- a/gateway/reverse_proxy.go
+++ b/gateway/reverse_proxy.go
@@ -846,10 +846,10 @@ func (p *ReverseProxy) httpTransport(dialTimeout float64, req *http.Request, out
 			},
 			AllowHTTP: true,
 		}
-		return &TykRoundTripper{transport, h2t, p.logger, p.Gw}
+		return &TykRoundTripper{transport: transport, h2ctransport: h2t}
 	}
 
-	return &TykRoundTripper{transport, nil, p.logger, p.Gw}
+	return &TykRoundTripper{transport: transport, h2ctransport: nil}
 }
 
 func (p *ReverseProxy) setCommonNameVerifyPeerCertificate(tlsConfig *tls.Config, hostName string) {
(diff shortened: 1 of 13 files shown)


gateway/reverse_proxy.go
```diff
@@ -897,42 +901,45 @@ func (p *ReverseProxy) setCommonNameVerifyPeerCertificate(tlsConfig *tls.Config,
...
-		handler, _, found := rt.Gw.findInternalHTTPHandlerForLoop(r.Host, nil, r)
-		if !found {
-			rt.logger.WithField("looping_url", "tyk://"+r.Host).Error("Couldn't detect target")
-			return nil, errors.New("handler could")
-		}
+			handler, _, found := gw.findInternalHTTPHandlerForLoop(r.Host, nil, r)
+
+			if !found {
+				logger.WithField("looping_url", "tyk://"+r.Host).Error("Couldn't detect target")
+				return nil, errors.New("handler could")
```
MaciekMis, line 927: missing partial error message

internal/roundtrippers/header_timeout.go
```diff
@@ -0,0 +1,106 @@
...
+func (t *headersTimeoutTracker) OnWroteRequest(_ httptrace.WroteRequestInfo) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	// If the request was already canceled (e.g., lower-level error), ignore the timer.
+	if t.ctx.Err() != nil {
+		return
+	}
+
+	t.timer = time.AfterFunc(t.timeout, func() {
```
MaciekMis, line 86: timer could be already initialized and running (ex due to retries this func could be invoked multiple times) - we should stop timer if it is present to prevent any leaks

gateway/reverse_proxy.go
```diff
@@ -1351,6 +1342,17 @@ func (p *ReverseProxy) WrappedServeHTTP(rw http.ResponseWriter, req *http.Reques
 		err             error
 	)
 
+	var decoratedRoundTripper = roundtrippers.Combine(
+		roundTripper,
+		internalLoopRoundTripperMiddleware(p.Gw, p.logger),
```
MaciekMis, line 1347: `internalLoopRoundTripperMiddleware` stops executing other RTs for internal loops so `roundtrippers.Timeout` will not be invoked - is it intentional?

gateway/reverse_proxy.go
shults: It's 1:1 port from an old TyjkRoundTripper

gateway/reverse_proxy.go
shults: Yep it's intentional. Take a look at the old TykRoundTripper. I have to put Timeout middleware and HeadersTimeout (time-to-first-byte) before the real transport.

internal/roundtrippers/header_timeout.go
shults: Good point, but IMO we should set the timer only once if it's not already set. 
Let's imagine that we have... retry MW in the chain after a timeout. 
Each retry would extend timeout...

MaciekMis: I agree, we should keep the first initialized timer

MaciekMis (approved):

pvormste (approved):

PR #8651, opened 2026-08-26
[TT-17981] docs: add a testing section to the README

kofoworola:
## What

The README currently covers running tests in two lines, tucked inside the compiling section:

> To run tests locally use the following command: `go test ./...`
> Note that tests require Redis to be running on the same machine (default port).

That is not enough to get a clean run. Following it produces failures that look like real test failures but are actually setup gaps. This PR moves testing into its own `## Testing` section and fills in what was missing.

## Why

Running the suite from the README as written, the following bite you and none of them are documented:

- **Go's test cache hides most of the run.** Without `-count=1`, the majority of packages report cached results and never execute. It looks like a full pass.
- **`-p 1` matters.** Each package starts its own gateway against the same Redis. Run them concurrently and you can blow the gateway's two-second connect timeout, which surfaces as `panic: can't connect to redis, timeout` rather than anything test-related.
- **`dlpython` looks for Python 3.5.** On any modern Python it fails with `Couldn't find Python 3.5`. `PYTHON_VERSION` fixes it, but that is not mentioned anywhere.
- **httpbin is needed too**, not just Redis.
- **`task test:integration` already does all of this** and is only mentioned in CONTRIBUTING.md.

## Notes

- Uses Docker Compose throughout rather than raw `docker run`, so the instructions work the same on any platform and Compose handles the network.
- The AOF workaround is written as a troubleshooting step rather than blanket advice. The committed `redis.yml` uses `--appendonly yes` and CI passes with it, so this looks specific to local disks under load.
- Kept the existing link to TESTING.md for writing tests.

## Out of scope

The compiling section still says Go 1.22 is required. `go.mod` is on 1.26.5. Left alone here to keep this diff to the testing content, but worth a follow-up.

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-17981" title="TT-17981" target="_blank">TT-17981</a>
</summary>

|         |    |
|---------|----|
| Status  | Closed |
| Summary | Update README.md / docs about testing our Open Source Core Components |

Generated at: 2026-09-02 06:14:50

</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ Compile from Source
 git clone https://github.com/TykTechnologies/tyk
 go build
 ```
-Go version 1.22 is required to build `master`, the current development version. Tyk is officially supported on `Linux/amd64`, `Linux/i386` and `Linux/arm64`.
+Go version 1.26 is required to build `master`, the current development version. Tyk is officially supported on `Linux/amd64`, `Linux/i386` and `Linux/arm64`.
 
 ## Testing
 
--- a/README.md
+++ b/README.md
@@ -183,13 +183,85 @@ go build
 ```
 Go version 1.22 is required to build `master`, the current development version. Tyk is officially supported on `Linux/amd64`, `Linux/i386` and `Linux/arm64`.
 
-To run tests locally use the following command:
-```console
-go test ./...
-```
-Note that tests require Redis to be running on the same machine (default port).
+## Testing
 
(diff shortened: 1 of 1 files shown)


rafalgolarz (approved):
nice, thank you. Maybe it would be worth to update Go version in docs 1.22->1.26. Thanks.

PR #8661, opened 2026-09-01
[TT-17765] Prevent Dashboard recovery reload deadlock

andrei-tyk:
## Summary

- make Dashboard registration responsible only for renewing the node ID and nonce
- keep reload ownership with startup, request recovery, and explicit reload callers
- add a regression covering nonce recovery during an active reload with a queued notification burst
- verify classic and MCP definitions are installed by the same completed reload cycle

## Root cause

`DoReloadWithError` holds `reloadMu` for the full reload. A nonce or network error during Dashboard API synchronization re-registers the node. Registration then called `DoReloadWithRetry`, causing a nested reload to wait forever on the same non-reentrant mutex. Reload notifications continued to queue, but the reload loop could no longer complete.

## Testing

- `go test ./gateway -run "^(TestRegister_DoesNotReload|TestReloadLoop_DashboardNonceRecoveryDoesNotDeadlock)$" -count=1`
- `go test -race ./gateway -run "^TestReloadLoop_DashboardNonceRecoveryDoesNotDeadlock$" -count=10`
- focused Dashboard registration, policy nonce recovery, API nonce recovery, and reload-loop tests

Both pre-fix and fixed Gateway Docker images were built successfully. The full Gateway package was also exercised; remaining local failures were unrelated Python coprocess and timing-sensitive Sentinel tests. Live Dashboard comparison was blocked before fault injection because the local Dashboard license was expired and `/register/node` returned `No nodes available`.

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-17765" title="TT-17765" target="_blank">TT-17765</a>
</summary>

|         |    |
|---------|----|
| Status  | Merge |
| Summary | Gateway reload wedges after container recreate - queued reloads never apply, APIs serve 404 until restart |

Generated at: 2026-09-11 13:07:31

</details>


andyo-tyk:
Relinked to the correct Jira ticket TT-17765

pvormste (approved):

PR #8675, opened 2026-09-04
[TT-1678] Gateway silently continues request when body transformation fails instead of returning 500

shults:
### WHAT WAS DONE

1. Proper Error Handling for Body Transformations (Fixing TT-1678)
• Template Compilation Status (result.Result[T]):
  Introduce a generic Result[T] container in internal/result to capture template compilation results (success or error) during API definition loading (compileTransformPathSpec) instead of silently skipping broken templates.
• Route Matching Fix:
  Update URLAllowedAndIgnored in gateway/api_definition.go to match endpoints based on Status == Transformed rather than checking for a non-nil template pointer. This ensures broken transformation specs are routed to the transformation middleware rather than bypassed to upstream.
• Request Body Transformation (TransformMiddleware):
  When template compilation fails (meta.Template.Get() returns an error) or execution fails, stop request processing and immediately return HTTP 500 Internal Server Error.
• Response Body Transformation (ResponseTransformMiddleware):
  When response template compilation fails, abort the response chain and invoke handleError with HTTP 500 Internal Server Error (with descriptive error message) instead of silently returning upstream's 200 OK with empty body.

2. Legacy JQ Transformation Middleware stays as is. It is currently non-compilable.

3. Testing & Verification
• Added comprehensive unit and end-to-end tests for result.Result[T] (internal/result/result_test.go).
• Added end-to-end tests in gateway/mw_transform_test.go covering:
  • Unmatched paths (passthrough with 200 OK)
  • Valid request body transformation (upstream verification)
  • Invalid template syntax (returns 500 Internal Server Error)
  • Invalid template mode (returns 500 Internal Server Error)
  • Malformed request body (returns 500 Internal Server Error)
•  Added unit/integration tests for ResponseTransformMiddleware confirming HTTP 500 on invalid response templates.

## Types of changes

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-1678" title="TT-1678" target="_blank">TT-1678</a>
</summary>

|         |    |
|---------|----|
| Status  | In Code Review |
| Summary | Gateway silently continues request when body transformation fails instead of returning 500 |

Generated at: 2026-09-08 14:08:56

</details>


pvormste (approved):

MaciekMis (approved):

Santoshkumarpuppala:
When a response template parses but fails at execute, `res_handler_transform.go:203-204` (`e26a836`) only logs the error, then `:209-217` make the partial buffer the body, with a matching Content-Length, and return nil. The request path returns a 500 (`mw_transform.go:126-127`, `:58-60`). This predates the PR (master `bc2308a:193-197`), and the description scopes the response fix to compile errors, so it may be out of scope, like the `//todo return error` unmarshal branches at `:151-154`/`:165-168`.

Run with Go's text/template: the new test's `hello {{.name` fails at Parse, so it only covers compile. With a `["a","b"]` body (mapped to `.array` at `:171-173`), `{"first":"{{index .array 0}}","x":"{{index .array 5}}"}` parses, then Execute fails with "index out of range: 5" and leaves `{"first":"a","x":"`. The control, `index .array 1`, gives `{"first":"a","x":"b"}`. Missing keys render `<no value>` without error, so this needs a real execute error.

Not run end to end: sprig's `{{fail "boom"}}` also errors at execute, so a template using it to reject a response would still send the upstream status with a truncated body. Was log-only intentional? If not, a generic 500 via `handleError` would also avoid the `err.Error()` leak at `:196` that probelabs flagged. Happy to add an execute-failure test.

PR #8696, opened 2026-09-11
[TT-18228] Migrating a Tyk Classic API to OAS can silently drop overlapping regex-based endpoints

shults:
## Description

Two separate things, in order.

**Migration — first 11 commits (TT-18228).**
Two Classic endpoints that differ only by a regex, for example `/users/[0-9]+`
and `/users/[a-zA-Z]+`, used to become one OAS path and one of them was lost
with no error. Each now gets its own path and keeps its own regex.

**Routing — last commit `4136007b1` (TT-16292).**
After migration both endpoints were in the document, but the gateway still
matched both with `([^/]+)`, so every request went to the same one. The matcher
is now built from the regex instead of the placeholder. This commit is
self-contained and can be reverted on its own.

### Please read before merging

The last commit changes behaviour. It makes a path parameter behave differently
depending on its **name**:

| parameter | regex applied when matching? |
| --- | --- |
| `{customRegex1}` - generated by migration | yes |
| `{myParam}` - written by a user | no |

This matches what Tyk Classic did, so it is deliberate. But you cannot see it in
the document: both parameters have a `pattern`, and nothing says only one of
them is enforced.

Today every path parameter matches any single segment. After this commit,
migrated APIs match only what their regex allows. **Requests that work today can
start returning 404.**

That is a change to the model, not a bug fix. I would prefer to put it behind a
feature flag, off by default:

- flag on - new migrations route correctly
- flag off - existing APIs behave exactly as they do now

Happy to do that if you agree.

## Related Issue

- https://tyktech.atlassian.net/browse/TT-18228 - migration (first 11 commits)
- https://tyktech.atlassian.net/browse/TT-16292 - routing, item 3 (last commit)

## Motivation and Context

A Classic API can have two endpoints at the same position told apart only by a
regex. Classic routes them correctly. Migrating such an API to OAS lost one of
them, and the one that survived answered requests meant for the other.

## How This Has Been Tested

- Unit tests in `apidef/oas` and `internal/pathnormalizer`, including a fuzz
  target for the path parser.
- An end-to-end gateway test that loads a migrated API and sends the two
  requests above (`gateway/mw_mock_response_regex_test.go`).
- Verified by hand against a locally migrated API through the Dashboard.

Note: the full `./gateway/` suite has one pre-existing failure,
`TestSyncAPISpecsRPCFailure_CheckGlobals`, which fails on the base commit too.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

### Ticket Details

<details>
<summary>
<a href="https://tyktech.atlassian.net/browse/TT-18228" title="TT-18228" target="_blank">TT-18228</a>
</summary>

|         |    |
|---------|----|
| Status  | Blocked |
| Summary | Migrating a Tyk Classic API to OAS can silently drop overlapping regex-based endpoints |

Generated at: 2026-09-15 11:53:35

</details>


shults:
Classic → OAS migration itself is correct: endpoints that differ only by a regex
are now preserved as separate OAS paths, each keeping its own pattern in
`parameters[].schema.pattern`.

Routing, however, breaks on the first save of a migrated API. `OAS.ExtractTo`
rebuilds the Classic view from the OAS path keys alone and does not carry the
parameter pattern back into the path, so `/users/[a-z]+` and `/users/[0-9]+`
both return as `/users/{customRegexN}` and compile to the same matcher
`^/users/([^/]+)$`. An anonymous regex therefore degrades to the default
`[^/]+` and the endpoints stop being distinguishable.

The regression is latent: migration leaves the original Classic paths
untouched, so a freshly migrated API still routes correctly and only breaks
after an unrelated edit in the dashboard. Confirmed by reproduction — after the
save the Classic side holds `white_list: /users/{customRegex1},
/users/{customRegex2}` with an empty `mock_response`.

Fixing this requires `ExtractTo` to reconstruct the original regex from the
path parameter; scope TBD.

shults:
# Classic → OAS: endpoints told apart by a regex

Tyk Classic lets two endpoints sit at the same position in the URL and be told
apart only by a regex:

```
/users/[0-9]+      numbers only
/users/[a-zA-Z]+   letters only
```

Classic routes these correctly today. Migrating them to OAS did not.

---

## It was three problems stacked, not one

**1. Migration lost an endpoint.** Both became `/users/{customRegex1}`. The
second overwrote the first. No error, no warning, migration reported success.

**2. Middleware landed on the wrong endpoint.** The one that survived carried
the *other* one's configuration — mock body, cache, rate limit, all of it. The
operation was named after endpoint A and configured as endpoint B.

**3. Routing could not tell them apart.** Even with both endpoints in the
document, the gateway compiled both to `^/users/([^/]+)$`, so every request went
to whichever sorted first.

Each problem was invisible while the one above it was still broken. Fixing the
first is what made the second visible, and so on.

---

## What is fixed

| what | ticket |
| --- | --- |
| Migration keeps both endpoints, each with its own regex | TT-18228 |
| Middleware stays on the endpoint it belongs to | TT-18228 |
| OAS → Classic keeps the regex in the path | TT-18228 |
| Gateway matches each endpoint by its own regex | TT-16292 |

PR #8696, 12 commits. The routing commit is last and self-contained.

---

## Round trip: what comes back

A regex can reach a Classic path in three shapes. What survives a
Classic → OAS → Classic round trip:

| Classic path | comes back as | matcher before | matcher after |
| --- | --- | --- | --- |
| `/users/[0-9]+` | `/users/[0-9]+` | `^/users/[0-9]+$` | `^/users/[0-9]+$` |
| `/users/{id:[0-9]+}` | `/users/{id}` | `^/users/([^/]+)$` | `^/users/([^/]+)$` |
| `/users/{id}` | `/users/{id}` | `^/users/([^/]+)$` | `^/users/([^/]+)$` |

**Behavioural symmetry is restored.** Every shape routes after the round trip
exactly as it routed before. Note the middle row: Classic never applied a regex
written inside a named parameter, and it still does not.

**Visual symmetry is not, and will not be without a further change.** A named
parameter loses its `:regex` from the path text — the regex survives on the OAS
parameter instead, but the Classic path reads `/users/{id}`. Restoring that text
would mean rewriting the Classic path of every OAS API carrying a pattern,
including APIs that were never migrated.

The matcher is the contract. The path text is not, and we are not treating it
as one.

---

## Decision needed: this changes behaviour

The routing fix makes a path parameter behave differently depending on its
**name**:

| parameter | regex enforced when matching? |
| --- | --- |
| `{customRegex1}` — generated by migration | yes |
| `{myParam}` — written by a user | no |

This mirrors what Classic does, so it is deliberate. But it cannot be seen in
the document: both parameters carry a `pattern`, and nothing marks which one is
enforced.

Today every path parameter matches any single segment. After this change,
migrated APIs match only what their regex allows.

**Requests that work today can start returning 404.**

One option is a feature flag, off by default — new migrations route correctly
with it on, existing APIs unchanged with it off. The commit reverts cleanly
either way.

---

## Still open

**Candidate disambiguation does not see migrated APIs.**
`pathParamsMatchOperation` and `pathParamRestrictiveness` read
`operation.Parameters`. Migration writes path parameters on `pathItem`. Measured
on both shapes:

```
parameters on the operation   restrictiveness 2 / 2, each request resolves correctly
parameters on the path item   restrictiveness 0 / 0, every candidate matches everything
```

Every existing test for this builds parameters on the operation, so the shape
migration produces is not covered. The routing fix sidesteps it rather than
repairing it.

**`compileOASValidateRequestPathSpec`** has the same problem as the mock
response path and takes the same one-line change.

**kin-openapi's duplicate-path rule is dead upstream** — it stores under the raw
path and looks up by the normalized one, so two paths differing only in
placeholder name pass validation. If that is ever fixed, every migrated API
fails to load, both in the gateway and on the Dashboard's migrate button.

---

## The question underneath all of this

Today a path parameter's `schema` has **no effect on routing** in Tyk OAS. Not
`pattern`, not `type`, not `enum`, not `minLength`. Every `{anything}` compiles
to `([^/]+)`. Request Validation is the only thing that reads the schema, and it
runs *after* the endpoint has been chosen — so it can reject a request, never
send it somewhere else.

Nobody has decided whether that is the intended design.

> **Should a path parameter's `schema` affect routing in Tyk OAS, or stay
> validation-only, as it is today?**

Worth saying plainly: **the last commit in this PR already answers it** — "yes,
but only for parameters that migration generated". That answer fell out of the
implementation rather than being chosen. That, more than the 404 risk, is why it
belongs behind a flag.

What follows from each answer:

- **Validation-only** — migrating regex endpoints is lossy by design. Document
  the limitation and drop the routing commit.
- **Affects routing** — the model needs a way to express it. Classic has no form
  for "named parameter with an enforced regex", which is why a generated name
  (`customRegexN`) is currently standing in for one. That is a feature with a
  flag and a migration path, not a bug fix.

---

## Status

- **PR #8696** — open, tests and lint green
- **TT-18228** — Blocked
- The three red checks are the Jira linter refusing a Blocked ticket, not code

PR #8712, opened 2026-09-17
TT-18236 Gateway register information

tbuchaillot:
## Types of changes

## Checklist


gateway/node_metadata.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+// setHeaders writes the Dashboard node metadata headers onto h. It is safe to
+// call repeatedly on a reused request: values are replaced and the tags header
+// is removed when there are no tags.
+func (m nodeMetadata) setHeaders(h http.Header) {
+	h.Set(header.XTykNodeVersion, m.Version)
+	h.Set(header.XTykNodeSegmented, strconv.FormatBool(m.IsSegmented))
+	if tags := strings.Join(m.Tags, ","); tags != "" {
```
sredxny, line 45: what would happen if a tag contains a comma? Eg: "team,eu"? is it valid? or is it an edge case and should be covered by documentation asking users to not set any comma inside the tags?

gateway/node_metadata_test.go
```diff
@@ -0,0 +1,152 @@
...
+		t.Cleanup(func() { getIpAddress = orig })
+		getIpAddress = func() ([]string, error) { return []string{"192.0.2.10", "192.0.2.11"}, nil }
+
+		gw.SetConfig(config.Config{})
+		gw.getHostDetails()
+		assert.Equal(t, "192.0.2.10", gw.nodeMetadata().HostDetails.Address)
+	})
+}
+
+func TestNodeMetadata_TagsHeaderBoundedSize(t *testing.T) {
```
sredxny, line 108: The test name implies setHeaders bounds the tags header size, but setHeaders enforces no bound at all. The test just hardcodes one input (50 tags × 32 bytes ≈ 1649 bytes) and asserts it happens to be < 2048. It verifies nothing about a boundary.
Here we must either add a real bound in setHeaders  and test the behavior at the boundary, or rename the test to reflect that it's a fixed-input smoke check, not a bound guarantee.

sredxny (changes requested):

gateway/node_metadata.go
tbuchaillot: IMO it's an edge case

gateway/node_metadata_test.go
tbuchaillot: fixed

sredxny (approved):

